PR #55, opened 2021-12-23
Add simple block template

kachkaev:
## In scope

- Create `packages/block-template` based on the template we used internally.
  - Excludes ESLint for simplicity
  - Excludes (partially out-of-date) README content in favor of https://blockprotocol.org/docs/developing-blocks
  - Is completely autonomous, i.e. does not depend on other parts of the monorepo
  - Includes `src/block-protocol-types.ts` (to be replaced with an npm package import in the future)
  - Contains updated `package.json` deps via `npx npm-check-updates --upgrade --target=minor`. We had a mix of pinned and pinned deps, I used `^` everywhere for now to reduce duplicates in the monorepo. We can reconsider this later if needed.
  - Uses `yarn dev` instead of `yarn start` for dev mode (e.g. like in Next.js)
  - Excludes `serve`. This is only useful for local embeds, which is outside the current docs narrative.
  
- Create `yarn create-block-app` and `packages/create-block-app` based on the `new:block` script used internally
   - Includes a new required arg for folder path
   - Name is closer to `npx create block-app` which we might want to use in the future

## Out of scope

The script and the template can be optimised, which can reduce boilerplate and cover more edge cases. Feel free to contribute! 🙌
   
## How to test

Follow `/docs/developing-blocks`. E.g. to create a new block use `yarn create-block-app hello /tmp/blocks/hello`.


packages/block-template/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "name": "@hashintel/block-template",
+  "version": "0.0.0",
+  "private": true,
+  "description": "Block template",
+  "license": "MIT",
+  "author": "HASH",
+  "main": "src/index.ts",
+  "browser": "dist/main.js",
+  "scripts": {
+    "build": "yarn clean; cross-env NODE_ENV=production webpack --mode production && yarn schema",
```
benwerner01, line 11: Do you think we should add `**/dist/**` to the `.gitignore` for devs creating a block in this repo?

packages/create-block-app/scripts/create-block-app.js
```diff
@@ -0,0 +1,56 @@
+const { promisify } = require("util");
+const child_process = require("child_process");
+const fs = require("fs");
+const path = require("path");
+
+const exec = promisify(child_process.exec);
+
+(async () => {
+  const blockName = process.argv[2];
+  const blockPath = process.argv[3];
```
benwerner01, line 10: I think it's worth making the path optional, and just defaulting to using a directory with the same name as the `blockName` (similar to how `create-react-app` automatically creates a directory with your react app's name). Happy for this to be done in follow-up though.

benwerner01 (approved):
Thank you @kachkaev , this looks good to me and has worked for me locally 🎉

So far I haven't actually written any blocks myself, so it's probably a good idea to get @teenoh 's thoughts on this PR as well.

packages/block-template/package.json
kachkaev: Good idea! I added `dist` to `.gitignore` in https://github.com/blockprotocol/blockprotocol/pull/55/commits/37ccbfb0a121726cb769796ad4dc35804295e235

packages/create-block-app/scripts/create-block-app.js
kachkaev: Yeah it make sense to have block path equal to block name by default. The problem with doing so now is that blocks will be created _inside_ the repo, which may lead to misinterpretation. We want third-party blocks to be stored in external repos rather than ours, but may end up seeing PRs with new blocks.

When we can run `npx create block-app myblock`, omitting path will make more sense because the command will be probably called not from the repo folder.

packages/block-template/src/block-protocol-types.ts
```diff
@@ -0,0 +1,258 @@
...
+  description?: string;
+  displayName?: string;
+  externals?: Record<string, string>;
+  license?: string;
+  icon?: string;
+  name?: string;
+  schema?: string;
+  source?: string;
+  variants?: BlockVariant[];
+  version?: string;
```
teenoh, line 30: This should also include image `image?: string`

packages/create-block-app/scripts/create-block-app.js
```diff
@@ -0,0 +1,56 @@
...
+
+  try {
+    const { stdout } = await exec("git config --get user.name");
+    packageJson.author = stdout.trim();
+  } catch {
+    delete packageJson.author;
+  }
+
+  fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, undefined, 2));
+
```
teenoh, line 52: I think it might be better DX if we also run `yarn install` after updating the package.json file. I think people are used to that flow.

kachkaev: Yeah I did not include `yarn install` for that reason. We should probably fallback to `npm` if `yarn` is not around. That's something to think about when we work on the npm package.

packages/block-template/src/block-protocol-types.ts
kachkaev: Thanks! I’ll add `image?` before merging. I’ll add image field here and upstream before merging.

PR #57, opened 2021-12-23
Sync block implementation with the updated spec

kachkaev:
Follows https://github.com/hashintel/dev/pull/432 (private link)


teenoh (approved):
Looks good to me. Thanks!

PR #61, opened 2021-12-24
Improve meta tags

kachkaev:
Follows #60 

Refs:

- https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started
- https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary-card-with-large-image
- https://developers.facebook.com/docs/sharing/webmasters/

### DIFF
--- a/site/src/pages/_document.page.tsx
+++ b/site/src/pages/_document.page.tsx
@@ -26,29 +26,30 @@ class MyDocument extends Document {
             content="A standardized way to create blocks whose contents are mapped to schemas, which are both
             human and machine-readable."
           />
-          <meta name="twitter:card" content="summary" />
+          <meta name="twitter:card" content="summary_large_image" />
           <meta name="twitter:title" content="Block Protocol" />
           <meta
             name="twitter:description"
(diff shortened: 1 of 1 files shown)


site/src/pages/_document.page.tsx
```diff
@@ -26,29 +26,30 @@ class MyDocument extends Document {
...
           <meta
             name="og:description"
             content="An open standard for data-driven blocks"
           />
           <meta
-            name="og:image"
+            property="og:image"
             content="https://blockprotocol.org/assets/bp_og_cover.png"
           />
-          <meta name="og:url" content="https://blockprotocol.org/" />
```
kachkaev, line 49: `og:url` is meant to be a normalized URL of a given page. Always referring to the home page URL can confuse social media platforms.

We can bring this tag back when we convert hardcoded meta tags into some kind of `<Meta title="..." />` component. It will need to be pulled out of `_document`. I can copy-paste this solution from a side project when it’s time to add custom previews for blocks etc.

nonparibus (approved):

PR #62, opened 2021-12-24
Fix HTTP 500 for non-existing blocks

kachkaev:
**Before**  
https://blockprotocol-f6y3xupq0-hashintel.vercel.app/@hello/world

**After**  
https://blockprotocol-bcjr91k7x-hashintel.vercel.app/@hello/world

I noticed this bug when working on #60. Opening https://blockprotocol.org/assets/bp_twitter_cover.png before merging returned 500.

### DIFF
--- a/site/src/components/BlockCard.tsx
+++ b/site/src/components/BlockCard.tsx
@@ -79,7 +79,7 @@ export const BlockCard: VFC<BlockCardProps> = ({ loading, data }) => {
   } = data;
 
   return (
-    <Link href={`${packagePath}`}>
+    <Link href={`/${packagePath}`}>
       <Box
         sx={{
           ...blockWidthStyles,
--- a/site/src/pages/[shortname]/[blockSlug].page.tsx
+++ b/site/src/pages/[shortname]/[blockSlug].page.tsx
@@ -52,7 +52,7 @@ const Bullet: VoidFunctionComponent = () => {
 };
 
 type BlockPageProps = {
-  metadata: BlockMetadata;
+  blockMetadata: BlockMetadata;
   schema: BlockSchema;
   blockStringifiedSource: string;
   catalog: BlockMetadata[];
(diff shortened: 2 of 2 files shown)


site/src/components/BlockCard.tsx
```diff
@@ -79,7 +79,7 @@ export const BlockCard: VFC<BlockCardProps> = ({ loading, data }) => {
   } = data;
 
   return (
-    <Link href={`${packagePath}`}>
+    <Link href={`/${packagePath}`}>
```
kachkaev, line 82: Follows #44

Blocks used relative path (i.e. `@hash/abc`) and were classified as external because of that. Adding `/` ensures that they are not opened in the new tab (which is annoying).

site/src/pages/[shortname]/[blockSlug].page.tsx
```diff
@@ -110,28 +110,43 @@ export const getStaticProps: GetStaticProps<
...
     `${BASE_URL}/blocks/${shortname}/${blockSlug}/metadata.json`,
-  ).then((res) => res.json());
+  );
+
+  if (blockMetadataResponse.status === 404) {
+    // TODO: Render custom 404 page for blocks
+    return { notFound: true };
+  } else if (!blockMetadataResponse.ok) {
+    // TODO: Render details of upstream error
+    throw new Error("Something went wrong");
```
Comment (teenoh) [inline site/src/pages/[shortname]/[blockSlug].page.tsx:128]: Thanks. Wouldn't this cause the page to crash? Perhaps we still return `{notFound: true}`, then add a comment to log the error

Comment (kachkaev) [reply site/src/pages/[shortname]/[blockSlug].page.tsx:128]: Yeah that will cause the page to crash but that's deliberate. We don't want to mask network or parse errors with a 404. Ideally, we'll need to convert the error into some props and render a nice error page, but just throwing 500 looks more honest than showing 404 for a block that does exist.

Comment (kachkaev) [reply site/src/pages/[shortname]/[blockSlug].page.tsx:128]: My current goal is to fix errors in paths like https://blockprotocol.org/assets/bp_twitter_cover.png_oops

teenoh (approved):

PR #63, opened 2021-12-28
[SITE] Add homepage header animation

teenoh:
This PR
- adds the homepage header animation
- improves homepage header styling on large screens and small screens
- Right-Aligns "Supported by HASH" link in the footer

[Deploy link](https://blockprotocol-jek440h09-hashintel.vercel.app/)

---
  - https://app.asana.com/0/0/1201562956971378
  - https://app.asana.com/0/0/1201562956971377
  - https://app.asana.com/0/0/1201594305573237
  - https://app.asana.com/0/0/1201588774378635
  - https://app.asana.com/0/0/1201588774378626


nonparibus:
Hero notes:
- **animation bug:** the background blur elements disappear for me after a second or so (they should both remain present and float about the screen)
- **animation inconsistency:** concentric circles that fade to 100% transparency should emanate outwards from the person node (see Figma)
- **animation inconsistency:** lines should be drawn, not 'fade in' (see Figma)
- **improving the graphic's clarity:** we should probably add `Full Name` alongside `Job Title` and `Image` for completeness of the labels shown on the person entity. It's strange we only show 2 of 3. If it were 2 of 30 I'd understand being selective for data viz simplicity, but here it looks like name is omitted for some other reason (which isn't the case).

nonparibus:
Given Ben's function-critical work for Jan 14th I suggested he hold reviewing this at least until the above changes requested are made. Opportunity cost considered, it doesn't make sense to look at this beforehand.

site/src/components/pages/home/Header/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+  useTheme,
+  useMediaQuery,
+} from "@mui/material";
+import { gsap } from "gsap";
+import { Button } from "../../../Button";
+import { BoltIcon } from "../../../SvgIcon/BoltIcon";
+import { Link } from "../../../Link";
+import { HeaderCard } from "./HeaderCard";
+
+export const Header = () => {
```
benwerner01, line 15: During page-load the animation is displayed in its final sate, ideally it shouldn't do this:

https://user-images.githubusercontent.com/42802102/148939573-28e93eda-767b-4b30-8bf5-c6cb050a261c.mov

This is likely going to be more noticeable with slower internet connections. If this is an easy fix I'd suggest it's worth doing as part of this PR, if not let's create a follow-up task.

benwerner01: Also noticed for wider screensizes the header card looses its fade on the right:

https://user-images.githubusercontent.com/42802102/148940990-ef9389c9-c2ab-4c25-b165-326e8f02d0dd.mp4

I think this is worth addressing, as anyone with a 4k+ monitor is going to see that screen size.

teenoh: Thanks. Noticed the first issue much earlier, completely forgot to fix that. 
The header card losing its fade is intentional ([context](https://hashintel.slack.com/archives/C02NW18QGUR/p1641822258008500)). Although the card still being cutoff at those screen sizes is a bug and I'll fix that.

nonparibus: @teenoh It's fine for the card to lose it's fade, but I think the top-right and bottom-right corners should be rounded like their left-side counterparts when this is the case.

CiaranMn:
@teenoh once you've updated the visuals please tag me for a quick review of the code - Ben will be focusing on the signup/signin work

nonparibus:
Let's aim to have this merged no later than EOD tomorrow.

CiaranMn (changes requested):
Thanks @teenoh - I spotted a few things:
~1. **To action:** - there's a random png file committed, what's that about?~
2. **To action:** - the background blur now appears only intermittently, I can't quite work out on what basis. We should fix this as it's weird.

https://user-images.githubusercontent.com/37743469/149327079-a559f451-8ede-4e99-a625-27b6e2eec280.mp4


3. **For followup** - the final state of the animation is shown at the beginning. I'm assuming this is tricky and therefore should be deferred, unless you can fix it relatively quickly. It is lower priority.

site/src/components/pages/home/Header/HeaderCardSvg.tsx
```diff
@@ -0,0 +1,720 @@
...
+          >
+            <stop stopColor="#2A2B37" />
+            <stop offset={1} stopColor="#F7FAFC" stopOpacity={0} />
+          </linearGradient>
+        )}
+        <image
+          id="image0_2132_16425"
+          width={200}
+          height={200}
+          href="/assets/header-card-svg-photo.png"
```
teenoh, line 715: @CiaranMn the committed image is used here in svg

CiaranMn: Oops, thank you

CiaranMn (approved):
Thanks @teenoh - I've mentioned that the background issue disappeared when I updated Chrome.

Please create a follow-up task for improving initial state on load - good to merge

site/src/components/pages/home/Header/index.tsx
teenoh: Follow up [task](https://app.asana.com/0/1201519755227635/1201656851219256)

PR #64, opened 2021-12-28
Fix broken links in spec

nonparibus:
In preparation for launch we updated the slug of Appendix A and this PR corrects old broken links which were resulting in 404s.


benwerner01 (approved):
Thanks, looks good 👍

PR #65, opened 2021-12-29
[SITE] Change selected link in `PageSidebar` based on scroll position on the current page

benwerner01:
This PR introduces the feature where the selected link in the `PageSidebar` (or the Navbar breadcrumbs) is updated based on the scroll position of the user for the specification/documentation pages.

https://user-images.githubusercontent.com/42802102/147665286-41e1e204-6c5f-4a2a-8c95-be90e64b63ed.mp4

---
  - https://app.asana.com/0/0/1201503372043151

### DIFF
--- a/site/src/components/MDXPageContent.tsx
+++ b/site/src/components/MDXPageContent.tsx
@@ -123,7 +123,7 @@ export const MDXPageContent: VFC<MDXPageContentProps> = ({
         {...boxProps}
         sx={{
           width: "100%",
-          overflow: "scroll",
+          overflow: "auto",
           "& > :not(.info-card-wrapper), > a:not(.info-card-wrapper) > *": {
             maxWidth: {
               xs: "100%",
--- a/site/src/components/MDXPageContent.tsx
+++ b/site/src/components/MDXPageContent.tsx
@@ -1,7 +1,10 @@
-import { VFC } from "react";
+import { useEffect, useState, VFC } from "react";
+import { unstable_batchedUpdates } from "react-dom";
+import { useRouter } from "next/router";
 import { MDXRemote, MDXRemoteSerializeResult } from "next-mdx-remote";
 import Box, { BoxProps } from "@mui/material/Box";
 import { mdxComponents } from "../util/mdxComponents";
+import MDXPageContext, { Heading } from "./context/MDXPageContext";
 
 export const MDX_TEXT_CONTENT_MAX_WIDTH = 680;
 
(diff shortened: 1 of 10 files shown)


nonparibus:
Functional UAT LGTM.

I think we want to rename some of our section headers to fit them into the sidebar better. They were written before the current site existed and while appropriate at the time perhaps don't work so well in the live context (specifically they're a bit long/wordy and things like "putting it all together - the data gets sent to blocks" are perhaps not the best in a navbar/title... better as body copy). cc @CiaranMn for appropriate updates. Captured as https://app.asana.com/0/1200211978612931/1201562956971367

site/src/util/mdxComponents.tsx
```diff
@@ -1,10 +1,37 @@
+/* eslint-disable react-hooks/rules-of-hooks */
```
kachkaev, line 1: It might be safer to replace this comment with `// eslint-disable-next-line react-hooks/rules-of-hooks` called three times. Otherwise, we can miss real bugs in the future such as conditional hook calls after early exits.

I can also imagine us splitting the lookup as it has grew too big. If we do that, we won't have to silence the rule.

site/src/components/MDXPageContent.tsx
```diff
@@ -1,30 +1,138 @@
...
-  </Box>
-);
+}) => {
+  const router = useRouter();
+
+  const [headings, setHeadings] = useState<Heading[]>([]);
+
+  useEffect(() => {
+    setHeadings([]);
+  }, [serializedPage]);
```
kachkaev, line 26: I wonder if we can avoid this if `useMDXHeading` can both register and unregister a heading. But this is probably not worth a change at this point.

site/src/components/context/MDXPageContext.ts
```diff
@@ -0,0 +1,18 @@
...
+  headings: Heading[];
+  setHeadings: Dispatch<SetStateAction<Heading[]>>;
+};
+
+const MDXPageContext = createContext<MDXPageContextProps>({
+  headings: [],
+  setHeadings: () => undefined,
+});
+
+export default MDXPageContext;
```
kachkaev, line 18: How about co-locating React context with the hook and the provider? This will help us see the relationship between various parts of the same feature.

Here’s how I imagine this at the exterior:

```
import { ScrollSyncProvider, useScrollSyncAnchor } from "../../../shared/scroll-sync"`
```

```ts
const SomePage = () => {
  return (
   <HeadingScrollSyncProvider>
      <SomeContent />
   </HeadingScrollSyncProvider>
   );
}
```

```tsx
const SomeContent = () => {
  const a1Ref = useScrollSyncAnchor();
  const a2Ref = useScrollSyncAnchor();
  
  return <div>
  <h1><a href="#hello" ref={a1Ref}>⚓️</a> Hello</h1>
  <p>lorem ipsum</p>
  <h1><a href="#world" ref={a2Ref}>⚓️</a> World</h1>
  </div>
}
```

site/src/util/mdxComponents.tsx
```diff
@@ -1,10 +1,37 @@
...
+  useContext,
+  useEffect,
+  useRef,
+} from "react";
 import { TypographyProps, Typography, Box, Paper } from "@mui/material";
 import { Link } from "../components/Link";
 import { InfoCardWrapper } from "../components/InfoCard/InfoCardWrapper";
 import { InfoCard } from "../components/InfoCard/InfoCard";
 import { Snippet } from "../components/Snippet";
+import MDXPageContext from "../components/context/MDXPageContext";
```
kachkaev, line 16: I generally prefer `StrictCamelCase`, so it’d be `MdxPageContext`. I guess you’ve used all caps in `MDX` here because we have `import { MDXRemoteSerializeResult } from "next-mdx-remote"`.

The new functionality is agnostic of markdown so I guess we can avoid `MDX` / `Mdx` at all.

site/src/components/context/MDXPageContext.ts
kachkaev: Or even:

```ts
const ScrollSyncAnchor = /* useScrollSyncAnchor + <a>*/;
```

```ts
import { ScrollSyncAnchor } from "../../../shared/scroll-sync";

const StyledScollSyncAnchor = styled(ScrollSyncAnchor)`
  ...
`;

StyledScollSyncAnchor.defaultProps = {
  children: "⚓️"
}

const SomeContent = () => {
  return <div>
    <h1><ScrollSyncAnchor anchor="hello" /> Hello</h1>
    <p>lorem ipsum</p>
    <h2><ScrollSyncAnchor anchor="world" /> World</h2>
  </div>
}
```

Note that the anchor is separate from the header, like in 

- https://beta.reactjs.org
- https://www.typescriptlang.org/docs/handbook/typescript-from-scratch.html
- https://prettier.io/docs/en/options.html

site/src/util/mdxComponents.tsx
benwerner01: Fixed in https://github.com/blockprotocol/blockprotocol/pull/65/commits/64ff9926ff1067cdcb2abe3705a955fbe2ffab61

site/src/util/mdxComponents.tsx
benwerner01: Agreed, addressed this in https://github.com/blockprotocol/blockprotocol/pull/65/commits/9100cac4862dabd0bd2c3cf1b6ebae7cb8377696

kachkaev (approved):
Thanks Ben, neat stuff!

Going forward, we might benefit from moving some `useState` and `useEffect` calls from `MdxPageContent` into `<PageHeadingsProvider />` (a wrapper around `<PageHeadingsContext.Provider />`). This can help us keep all parts of the feature under one roof and thus ease future code maintenance. That’s not a blocker though, let’s get the PR in! 🚀

PR #66, opened 2021-12-29
[SITE] Implement mobile designs individual block page

teenoh:
This PR implements 
- mobile design for the individual block page.
- fixes the issue of Explore blocks not showing content and block pages not being built
- fixes styling issue with the main icon in the block page

[Deploy Link](https://blockprotocol-qftthq7ks-hashintel.vercel.app/)

TODO:
[ ] Replace Variant tabs and Data Source/Block Schema tabs with a select dropdown on mobile

---
  - https://app.asana.com/0/0/1201589261424264
  - https://app.asana.com/0/0/1201601100049648
  - https://app.asana.com/0/1201414765539664/1201547408490492


nonparibus:
Teenoh advises that what’s left in this draft PR could either be added in before merge, or added when this [Block Variants task](https://app.asana.com/0/1201414765539664/1201540186206729) is being worked on.

site/src/pages/[shortname]/[blockSlug].page.tsx
```diff
@@ -63,19 +67,13 @@ type BlockPageQueryParams = {
   blockSlug?: string[];
 };
 
-export const getStaticPaths: GetStaticPaths<
-  BlockPageQueryParams
-> = async () => {
+export const getStaticPaths: GetStaticPaths<BlockPageQueryParams> = () => {
   return {
-    paths: [],
+    paths: readBlocksFromDisk().map((metadata) => `/${metadata.packagePath}`),
```
Comment (benwerner01) [inline site/src/pages/[shortname]/[blockSlug].page.tsx:72]: **Blocking:** Right now we can afford to read all the blocks from disk at build time, but the idea behind `fallback: "blocking"` is that we _don't_ determine the paths at build time. Therefore I suggest leaving this as:

```suggestion
    paths: [],
```

site/src/pages/[shortname]/[blockSlug].page.tsx
```diff
@@ -114,35 +112,20 @@ export const getStaticProps: GetStaticProps<
...
   }
 
+  const packagePath = `${shortname}/${blockSlug}`;
   const catalog = readBlocksFromDisk();
 
-  const blockMetadataResponse = await fetch(
-    `${BASE_URL}/blocks/${shortname}/${blockSlug}/block-metadata.json`,
+  const blockMetadata = catalog.find(
+    (metadata) => metadata.packagePath === packagePath,
   );
```
Comment (benwerner01) [inline site/src/pages/[shortname]/[blockSlug].page.tsx:120]: This is more expensive than it should be, as we are reading all the blocks from the disk just to find a specific one. Ideally we'd have a `readBlockFromDisk` function to use instead. I appreciate this is likely out of scope for this PR, so if you could leave a `@todo` here that would be great.

site/src/components/pages/hub/BlockTabsModal.tsx
```diff
@@ -26,7 +26,15 @@ export const BlockTabsModal: VoidFunctionComponent<BlockTabsModalProps> = ({
   setText,
 }) => {
   return (
-    <Modal open={open} onClose={() => setOpen((oldValue) => !oldValue)}>
+    <Modal
+      open={open}
+      onClose={() => setOpen((oldValue) => !oldValue)}
+      sx={{
+        "& .MuiBackdrop-root": {
+          backgroundColor: "rgba(77, 92, 108, 0.6)",
```
benwerner01, line 34: Instead of using `rgba()`, you could use one of the theme palette colours and set the backdrop's opacity to `0.6`, or whatever else you'd like.

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -58,119 +61,178 @@ export const BlockDataContainer: VoidFunctionComponent<
...
+          value={activeMobileTab}
+          onChange={(_event, newValue: number) => setActiveMobileTab(newValue)}
+          sx={{
+            "&.MuiTabs-root": {
+              backgroundColor: theme.palette.gray[10],
+              border: `1px solid ${theme.palette.gray[20]}`,
+              borderRadius: "6px",
+              display: "flex",
+              position: "relative",
+            },
```
benwerner01, line 76: This is equivalent to:
```suggestion
          sx={{
              backgroundColor: theme.palette.gray[10],
              border: `1px solid ${theme.palette.gray[20]}`,
              borderRadius: "6px",
              display: "flex",
              position: "relative",
```
because this will always have the `MuiTabss-root` CSS class

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -58,119 +61,178 @@ export const BlockDataContainer: VoidFunctionComponent<
...
+          }}
+        >
+          <Tab label="Preview" />
+          <Tab label="Source Code" />
+        </Tabs>
+      )}
+
+      <Box
+        mt={5}
```
benwerner01, line 103: In the figma designs there's less margin on mobile

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -58,119 +61,178 @@ export const BlockDataContainer: VoidFunctionComponent<
...
-            borderBottomLeftRadius: 6,
-            borderBottomRightRadius: 6,
+            ...(isMobile && {
+              // @todo use an enum value instead
+              display: activeMobileTab === 1 ? "block" : "none",
+            }),
           }}
         >
-          <BlockDataTabPanels
+          <BlockDataTabs
```
benwerner01, line 163: In the designs this is a dropdown on mobile, not a tab navigator. If you choose not to add this in this PR please create a follow up task so that it's captured.

benwerner01 (changes requested):
Thanks Teeno, this is looking great, and fixes the carousel issue for me 👍

Spotted one blocking change and made a couple other minor suggestion.

site/src/pages/[shortname]/[blockSlug].page.tsx
Comment (teenoh) [reply site/src/pages/[shortname]/[blockSlug].page.tsx:120]: I agree. Will update

site/src/components/pages/hub/BlockDataContainer.tsx
teenoh: True. I planned to implement that in a follow-up task.

site/src/pages/[shortname]/[blockSlug].page.tsx
Comment (teenoh) [reply site/src/pages/[shortname]/[blockSlug].page.tsx:72]: We can determine paths to be built while using `fallback: blocking`, only difference is that instead of immediately returning a 404 for paths that don't exist there, it tries to generate the page.

The reason we set the paths is that we can't access the `blocks` directory at runtime. As a result, the explore blocks section is always empty because readBlocksFromApi (which tries to access the blocks directory) returns an empty array. The block page was rendering correct information previously because we were making [requests](https://github.com/blockprotocol/blockprotocol/blob/main/site/src/pages/%5Bshortname%5D/%5BblockSlug%5D.page.tsx#L138-L145) to fetch the respective info from the public folder instead of accessing them from the filesystem. Setting the paths ensures each block page specified is built at build time.


I understand building all the block pages at build time is not a great idea when we have a huge list of blocks. But I think the ideal situation will be to build common block pages (specified in the path list) at build time, then build and cache the remaining blocks as the user requests for them. It will be much easier to access the file system at runtime once we upgrade to [Next 12]() though.

For now, I think we can work with building all the block pages at build time. In an example Nextjs gives about Incremental Static Regeneration, you can build up to 1k product pages and generate the rest on demand. I doubt we'll notice any serious difference in build times for the time being. If we do, we can always try to improve it then

Comment (CiaranMn) [reply site/src/pages/[shortname]/[blockSlug].page.tsx:72]: I don't think it affects your discussion about building _pages_, but worth noting that the 'metadata and block package files in GitHub repo' approach is a temporary implementation. 

We'll store metadata in a database and the package files in some external object store.

Comment (benwerner01) [reply site/src/pages/[shortname]/[blockSlug].page.tsx:72]: Thanks both, after more thought I agree @teenoh we can leave this as is for now. Perhaps a `@todo` about generating the paths for a limited subset of blocks would be good, but agree we should address this when we have a more long-term implementation of how we're storing the medatadata.

benwerner01 (approved):
Thanks Teeno, happy for this to be merged 👍

site/src/pages/[shortname]/[blockSlug].page.tsx
Comment (teenoh) [reply site/src/pages/[shortname]/[blockSlug].page.tsx:72]: Ahh I see, I didn't consider that @CiaranMn.  Thanks all, will add a `@todo`

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -58,119 +61,178 @@ export const BlockDataContainer: VoidFunctionComponent<
...
+                  textTransform: "none",
+                  color: "#64778C",
+                },
+                "& .MuiTab-root.Mui-selected": {
+                  backgroundColor: theme.palette.gray[10],
+                  color: theme.palette.purple[700],
+                  borderTopLeftRadius: 6,
+                  borderTopRightRadius: 6,
+                },
+              }}
```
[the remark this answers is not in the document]
teenoh: Fixed here acef6d4

site/src/components/pages/hub/BlockDataContainer.tsx
teenoh: Fixed here acef6d4

site/src/components/pages/hub/BlockDataContainer.tsx
teenoh: Fixed here acef6d4

site/src/components/pages/hub/BlockTabsModal.tsx
teenoh: Fixed here acef6d4. Couldn't use opacity since mui modal uses that internally for transitions. Realized mui exports an alpha method which can be used instead. So we can do

```
{ backgroundColor: ({ palette }) => alpha(palette.gray[70], 0.6) }
```

site/src/pages/[shortname]/[blockSlug].page.tsx
Comment (teenoh) [reply site/src/pages/[shortname]/[blockSlug].page.tsx:120]: Added a @todo for this here acef6d4

site/src/components/pages/hub/BlockDataContainer.tsx
teenoh: https://app.asana.com/0/1201519755227635/1201664899405923

teenoh: [Task](https://app.asana.com/0/1201519755227635/1201664899405923)

PR #68, opened 2021-12-29
[SITE] add fallback fonts

benwerner01:
Previously there were no fallback fonts, so before the fonts are loaded by the browser the user could briefly see a sarif font:
image
This PR introduces fallback fonts (`"Helvetica"`, `"Arial"` and `sans-serif`) which are now displayed instead whilst the browser is loading the correct fonts:
image

---
  - https://app.asana.com/0/0/1201543651334385


nonparibus (approved):
This is great, thanks. Let's use a similar approach in HASH but with separate `serif` and `sans-serif` fallback consts (where required).

PR #69, opened 2021-12-29
[SITE] stop truncating sidebar navigation items, make sidebar scrollable

benwerner01:
This PR stops truncating the link text in the `PageSidebar` on the specification/documentation pages:
image

In addition this PR makes the `PageSidebar` scrollable when its expanded contents exceed the height of the viewport, and it is in the "sticky" state.

---
  - https://app.asana.com/0/0/1201588774378622

### DIFF
--- a/site/src/components/PageSidebar.tsx
+++ b/site/src/components/PageSidebar.tsx
@@ -21,9 +21,7 @@ import { Link } from "./Link";
 
 const SidebarLink = styled(Link)(({ theme }) => ({
   display: "block",
-  whiteSpace: "nowrap",
-  overflow: "hidden",
-  textOverflow: "ellipsis",
+  lineHeight: "1.25em",
   transition: theme.transitions.create(["color"]),
   color: theme.palette.gray[70],
   ":hover": {
@@ -73,7 +71,7 @@ const SidebarPageSection: VFC<SidebarPageSectionProps> = ({
 
   return (
     <>
-      <Box mb={1} display="flex" alignItems="center">
+      <Box mb={1.5} display="flex" alignItems="flex-start">
         <SidebarLink
           ref={(ref) => {
             if (ref && isSectionSelected) {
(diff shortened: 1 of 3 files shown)


MaggieAppleton:
Now that we're wrapping the section titles, it would help readability a lot if we extended the width of the sidebar. Text wrapped on multiple lines gets hard to read and distinguish from the sections above and below it.

It's currently at 220px wide with 48px right-padding. If we bump that up to 300px it makes a big difference:

220px
Screenshot - 2022-01-03 11 32 06


300px
Screenshot - 2022-01-03 11 31 55

nonparibus:
> Now that we're wrapping the section titles, it would help readability a lot if we extended the width of the sidebar. Text wrapped on multiple lines gets hard to read and distinguish from the sections above and below it.
> 
> It's currently at 220px wide with 48px right-padding. If we bump that up to 300px it makes a big difference:
> 
> 220px Screenshot - 2022-01-03 11 32 06
> 
> 300px Screenshot - 2022-01-03 11 31 55
> 
> Moves most of the titles that were wrapping onto 3 lines down to 2

I'd created a separate ticket to shorten the titles of all the pages (as they're unnecessarily verbose in many cases). I think this will solve the same issue. Let's see if it does, and then we can always increase left-sidebar width if not.

kachkaev (approved):
Brilliant. Thanks Ben!

kachkaev:
One minor observation:

Screenshot 2022-01-04 at 20 08 42

The content has moved left but the two bottom links are still centred. This creates an offset between the text edge and the _previous_ link. This does not block the merge and can be addressed separately.

benwerner01:
> The content has moved left but the two bottom links are still centred. This creates an offset between the text edge and the previous link. This does not block the merge and can be addressed separately.

@kachkaev thanks for spotting, fixed this in https://github.com/blockprotocol/blockprotocol/pull/69/commits/8b20f213d09b9c147bf2620779c5f666d90e8c8c

> I'd created a separate ticket to shorten the titles of all the pages (as they're unnecessarily verbose in many cases). I think this will solve the same issue. Let's see if it does, and then we can always increase left-sidebar width if not.

@nonparibus @MaggieAppleton thanks both for the feedback, I've created an [asana task](https://app.asana.com/0/1201095311341924/1201618515575267/f) for this, but am also happy to change the width of the `PageSidebar`.

In my most recent commit (https://github.com/blockprotocol/blockprotocol/pull/69/commits/8b20f213d09b9c147bf2620779c5f666d90e8c8c) I've re-defined the width in a `SIDEBAR_WIDTH` constant, so this can be easily tweaked (as this should now only be a one-line change).

PR #73, opened 2021-12-31
[SITE] added block definition to the docs

MaggieAppleton:
Took a swing at defining a 'block' in the introduction of the docs and added some extra context about them.

https://blockprotocol-6t1rj7dem-hashintel.vercel.app/docs

I think it could certainly be improved, but gives us a starting point.

We should definitely add images and diagrams to this in the future.


site/src/_pages/docs/0_index.mdx
```diff
@@ -1,12 +1,26 @@
...
+- [develop blocks](https://blockprotocol.org/docs/developing-blocks) and publish them to the block registry
+- [embed blocks](https://blockprotocol.org/docs/embedding-blocks) in their own application
 
 For the full technical details, read [the Block Protocol specification](https://blockprotocol.org/spec).
 
+### What's a block?
+
+Blocks are the basic units we use to build pages on the modern web. They are rectangular, full-width sections stacked on top of one another – also known as "block-level" elements.
+
+Blocks can be created using plain HTML, or by using a block editor interface like Notion or Wordpress' Gutenberg. With block editors you can usually rearrange the order of blocks on a page using drag and drop.
```
nonparibus, line 16:
```suggestion
Blocks can be inserted using plain HTML/JavaScript, or through visual interfaces provided by modern content management systems and applications like all-in-one workspaces. With these block-based editors you can usually rearrange the order of blocks on a page using drag and drop.
```

site/src/_pages/docs/0_index.mdx
```diff
@@ -1,12 +1,26 @@
...
+
+- [develop blocks](https://blockprotocol.org/docs/developing-blocks) and publish them to the block registry
+- [embed blocks](https://blockprotocol.org/docs/embedding-blocks) in their own application
 
 For the full technical details, read [the Block Protocol specification](https://blockprotocol.org/spec).
 
+### What's a block?
+
+Blocks are the basic units we use to build pages on the modern web. They are rectangular, full-width sections stacked on top of one another – also known as "block-level" elements.
+
```
nonparibus, line 15:
```suggestion

A block is any valid HTML published to the [Block Hub](https://blockprotocol.org/hub).

```

cc @CiaranMn @MaggieAppleton per our 'quickest start guide' @ https://blockprotocol.org/docs/developing-blocks

site/src/_pages/docs/0_index.mdx
```diff
@@ -1,12 +1,26 @@
...
+Blocks can be created using plain HTML, or by using a block editor interface like Notion or Wordpress' Gutenberg. With block editors you can usually rearrange the order of blocks on a page using drag and drop.
+
+Each block has a type that determines how it displays data. You're probably familiar with the most common block types such as _paragraph_, _header_, _image_, _list_, and _table_. Some block types have more advanced interactive features such as filtering, search, and sorting.
```
nonparibus, line 18: cc @CiaranMn for feedback

site/src/_pages/docs/0_index.mdx
```diff
@@ -1,12 +1,26 @@
...
+
+There are an inifinite number of possible block types. One goal of the Block Protocol is to encourage developers to build and share brand new block types we've never seen before.
```
nonparibus, line 20:
```suggestion
There are an infinite number of possible block types. One goal of the Block Protocol is to encourage developers to build and share brand new types of blocks that haven't been seen before. We don't believe application's users should be limited to a dozen or so blocks (the bounded imaginations or expected maintenance bandwidth of their developers).
```

Should be rephrased to make simpler, but adds a bit of missing context re: the why. Can be improved now or later.

kachkaev:
Sorry for being to to pedantic, but it’d be cool if we could use typography symbols in our public-facing texts 😅 This mainly includes quotation marks (‘’ “”) and dashes (- – —).

Context:

- https://www.typewolf.com/cheatsheet
- https://smartquotesforsmartpeople.com

Texts with typography symbols look more professional and trustworthy, similar to how books compare to facebook posts. Paired quotes and long dashes are not necessary in every piece of text we write, but can be worth the investment when we expect a high _read/write ratio_ 🙂

I personally access typography symbols via https://ilyabirman.net/projects/typography-layout/, but that’s an old habit and just one of possible ways. macOS has built-in symbol shortcuts, which are listed in the first link. It’s also possible to write text in Google docs and then paste it into markdown. Word processing apps replace typewriter symbols with the typography ones automatically and do this quite well in most cases. Slack replaces '' "" with ‘’ “” too, which is quite pleasing to me personally! 🙂

site/src/_pages/docs/0_index.mdx
```diff
@@ -1,12 +1,26 @@
...
+
+From a developer perspective, blocks are similar to [web components](https://en.wikipedia.org/wiki/Web_Components). They are reusable sets of HTML elements that can accept any data that matches the block's schema. For example, an image block can only accept an image URL or file, but not a text string.
```
CiaranMn, line 22: I'm assuming this is talking about generic blocks, rather than BP blocks - we could add the sentence I suggest above after this ("A block which complies..."). 

This could be rephrased but I'm trying to get across that:
- web components are an implementation method rather than an alternative to blocks
- "schema" is something we use rather than being universally in use for different versions of "blocks"
- pedantry: tweaked the example to make it starker, as technically an image URL is a text string
 
```suggestion
From a developer perspective, blocks offer encapsulation of specific functionality ([web components](https://en.wikipedia.org/wiki/Web_Components) are one possible way of implementing a block). They are reusable elements that expect certain data to be passed to them (what we refer to as a block's "schema"). For example, an image block might only accept an image URL in its `src` property, not a number.
```

CiaranMn (commented):
I think this does a good job of explaining web blocks. I've suggested a few tweaks, which could no doubt be improved upon.

site/src/_pages/docs/0_index.mdx
CiaranMn: I'd also move the "goals of the BP" to the bottom as I think it'd flow better with a new para on BP blocks. So I'm suggesting the final paras would be something like this. Maybe with the paras swapped?

> There are an infinite number of possible block types. One goal of the Block Protocol is to encourage developers to build and share brand new types of blocks that haven't been seen before, and for applications to offer users far more blocks than any one team could imagine or build.
> 
> A block which complies with the Block Protocol has the added advantage of specifying the data it receives - and modifying it - in a standardized way. This makes it easier to re-use Block Protocol blocks across applications - and if they are published to the Block Hub, easily discoverable by users.

site/src/_pages/docs/0_index.mdx
```diff
@@ -1,12 +1,26 @@
...
+The following pages provide a practical introduction for developers who want to:
+
+- [develop blocks](https://blockprotocol.org/docs/developing-blocks) and publish them to the block registry
+- [embed blocks](https://blockprotocol.org/docs/embedding-blocks) in their own application
 
 For the full technical details, read [the Block Protocol specification](https://blockprotocol.org/spec).
 
+### What's a block?
+
+Blocks are the basic units we use to build pages on the modern web. They are rectangular, full-width sections stacked on top of one another – also known as "block-level" elements.
```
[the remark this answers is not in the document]
MaggieAppleton: Updated and removed this language 👍

site/src/_pages/docs/0_index.mdx
MaggieAppleton: Updated this to read:

> Blocks can be added to webpages using plain HTML and/or JavaScript. Some modern content management systems (eg. Wordpress, Squarespace) and applications (eg. Notion, Coda) allow you to add blocks to a page using a visual interface. With these block-based editors you can usually rearrange the order of blocks on a page using drag and drop. Blocks can be stacked on top of one another, arranged in columns, or embedded inside other blocks.

I know pointing to specific apps isn't ideal, so could remove the examples here. It would help paint a more concrete picture in people's minds of what we're talking about though.

site/src/_pages/docs/0_index.mdx
MaggieAppleton: Rephrased this but tried to keep the same gist in the "What is a Block Protocol block" section

site/src/_pages/docs/0_index.mdx
MaggieAppleton: Updated with suggestions 👍

site/src/_pages/docs/0_index.mdx
```diff
@@ -1,12 +1,36 @@
...
+
+Blocks can be added to webpages using plain HTML and/or JavaScript. Some modern content management systems (eg. Wordpress, Squarespace) and applications (eg. Notion, Coda) allow you to add blocks to a page using a visual interface. With these block-based editors you can usually rearrange the order of blocks on a page using drag and drop. Blocks can be stacked on top of one another, arranged in columns, or embedded inside other blocks.
+
+From a developer perspective, blocks encapsulate a set of specific functionality. Building a block is no different to writing a React, Vue, or [Web Component](https://en.wikipedia.org/wiki/Web_Components) – in fact, these are all ways you might implement a block. They are reusable elements that expect certain data to be passed to them (what we refer to as a block's “schema”). For example, an image block might only accept an image URL in its `src` property, not a number.
```
kachkaev, line 22:
```suggestion
From a developer perspective, blocks encapsulate a set of specific functionality. Building a block is no different to writing a React, Vue, or [Web Component](https://en.wikipedia.org/wiki/Web_Components) – in fact, these are all ways you might implement a block. They are reusable elements that expect certain data to be passed to them (what we refer to as a block's ‘schema’). For example, an image block might only accept an image URL in its `src` property, not a number.
```

The most common rule I‘ve seen was: _“” for quotes, ‘’ for emphasis_. We don’t have to stick with it rigorously, but it might be worth keeping the rule in our heads.

nonparibus:
Let's aim to have this merged by EOD tomorrow.

site/src/_pages/docs/0_index.mdx
```diff
@@ -9,6 +9,20 @@ The following pages provide a practical introduction for developers who want to:
 
 For the full technical details, read [the Block Protocol specification](https://blockprotocol.org/spec).
 
+### Why is structured data useful?
+
+Structured data allows computers to understand information. It is easier to create and easier to understand than unstructured data.
```
nonparibus, line 14: "It is easier to create and easier to understand than unstructured data." I take issue with the claim that it's easier to create. The Block Protocol was (partly) born of frustration that this wasn't the case. When in rapid-notetaking mode most people just want to dump ideas onto a page and not worry about structure. But blocks that are both useful to folks in that type of a setting (where speed matters), and which organically add semantic meaning to the information being captured at the same time, lay the groundwork for this to be true.

site/src/_pages/docs/0_index.mdx
```diff
@@ -9,6 +9,20 @@ The following pages provide a practical introduction for developers who want to:
...
+
+Most of the data on the web is unstructured. When we write webpages in HTML, it's easily readable by humans, but not by computers. The computer only knows you've added a bunch of `<p>` tags to a page, but doesn't know anything about what you've written inside them.
```
nonparibus, line 16:
```suggestion
Most of the data on the web is unstructured. When we write webpages in HTML, it's easily readable by humans, but not by computers. The computer only knows you've added a bunch of `<p>` tags to a page, but doesn't understand the contents or shape of what you've written inside (at least without any fancy natural language processing).
```

site/src/_pages/docs/0_index.mdx
```diff
@@ -9,6 +9,20 @@ The following pages provide a practical introduction for developers who want to:
...
+
+Unlike plain HTML, blocks created with the Block Protocol have **types**, such as _Person_, _Company_, and _Date_ that help computers understand what your text is about.
```
nonparibus, line 18:
```suggestion
Unlike plain HTML, blocks created with the Block Protocol have **types**, such as _Person_, _Company_, and _Date_ that help computers understand what they actually contain.
```

site/src/_pages/docs/0_index.mdx
```diff
@@ -9,6 +9,20 @@ The following pages provide a practical introduction for developers who want to:
...
+
+{diagram of HTML vs. structure block data}
```
nonparibus, line 20: @MaggieAppleton We should ready this for close of play Wednesday or remove this placeholder from the docs. Do you think we'll have something for then?

site/src/_pages/docs/0_index.mdx
```diff
@@ -9,6 +9,20 @@ The following pages provide a practical introduction for developers who want to:
...
+
+Imagine you are creating a webpage to announce a book signing event with a well-known author. You could add an _Author_ block, a _Location_ block, and a _Time_ block to this page. Instead of being some mystery HTML, your information is now readable by computers. It would be much easier to automatically pass this information to other websites and applications, such as a calendar application, or the author's personal website.
+
+We are trying to fulfil the vision of the Semantic Web – the original Web 3.0 (before all the NFTs took over). Our goal is to update the web from a pile of randomly formatted, human-readable pages into a structure trove of knowledge.
```
nonparibus, line 24:
```suggestion
We are trying to fulfil the vision of the Semantic Web (aka. the OG Web3, before all the NFTs took over!), transforming the web's randomly formatted, human-readable pages into structured troves of connected knowledge, improving both user and developer experiences as we go.
```

site/src/_pages/docs/0_index.mdx
MaggieAppleton: I was unsure about that claim too - copied off the rough notes from Joel, but figured it was in the context of pre-selecting entities from a known list vs. relying on memory. We can adapt to "Structured data makes it much easier for computers to understand information"

MaggieAppleton: Updated to: "Structured data makes it much easier for computers to understand the information we give them."

nonparibus (approved):

PR #75, opened 2022-01-04
[SITE] Add Github Star Button to Footer

MaggieAppleton:
This PR adds a button to the footer than prompts visitors to star the project on Github

Previews
image
image

Mobile:
image

### DIFF
--- a/site/src/components/Footer.tsx
+++ b/site/src/components/Footer.tsx
@@ -12,6 +12,7 @@ import {
 import { Link } from "./Link";
 import { BlockProtocolLogoIcon } from "./SvgIcon/BlockProtocolLogoIcon";
 import { HASHLogoIcon } from "./SvgIcon/HASHLogoIcon";
+import { Button } from "./Button";
 
 const FOOTER_NAVIGATION_LINKS: { href: string; name: string }[] = [
   {
@@ -73,11 +74,6 @@ const FooterNavigationLinks = FOOTER_NAVIGATION_LINKS.map(({ href, name }) => (
 
 const SOCIALS: { name: string; icon: ReactNode; href: string }[] = [
   {
-    name: "GitHub",
-    icon: <Icon className="fab fa-github" />,
-    href: "https://github.com/blockprotocol",
-  },
-  {
     name: "Twitter",
     icon: <Icon className="fab fa-twitter" />,
     href: "https://twitter.com/blockprotocol",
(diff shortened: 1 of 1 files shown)


nonparibus:
Addresses https://app.asana.com/0/1201414765539664/1201562956971357/f

site/src/components/Footer.tsx
```diff
@@ -87,38 +83,67 @@ const SOCIALS: { name: string; icon: ReactNode; href: string }[] = [
...
+    display="flex"
+    sx={{
+      flexDirection: {
+        xs: "column",
+        md: "row",
+      },
+      alignItems: { xs: "left", md: "center" },
+    }}
+  >
+    <Box marginBottom="1rem">
```
benwerner01, line 105: Just curious, is there a reason for using `1rem` instead of the default theme spacing here?
```suggestion
    <Box marginBottom={2}>
```

site/src/components/Footer.tsx
```diff
@@ -87,38 +83,67 @@ const SOCIALS: { name: string; icon: ReactNode; href: string }[] = [
...
+    mt={3}
+    display="flex"
+    sx={{
+      flexDirection: {
+        xs: "column",
+        md: "row",
+      },
+      alignItems: { xs: "left", md: "center" },
+    }}
+  >
```
benwerner01, line 104: My preference would be for the button to wrap to the next line when only when it no longer fits on the same line, like so:

https://user-images.githubusercontent.com/42802102/148071025-f7616df4-ed7e-4fab-b7eb-55d52ab52578.mov

If you agree, the following would accomplish this (in addition to adding `flexShrink: 0` to both flex children):
```suggestion
<Box
    mt={3}
    display="flex"
    flexDirection="row"
    alignItems="center"
    flexWrap="wrap"
  >
```

site/src/components/Footer.tsx
```diff
@@ -87,38 +83,67 @@ const SOCIALS: { name: string; icon: ReactNode; href: string }[] = [
...
-              theme.transitions.create("color", { duration: 150 }),
-            color: (theme) => theme.palette.gray[50],
-          },
+          }}
+        >
+          {icon}
+        </Link>
+      ))}
+    </Box>
+    <a href="https://github.com/blockprotocol/blockprotocol">
```
benwerner01, line 133: For consistency with the remainder of the repo, and to gain access to the `sx` prop, I think we should use our `Link` component here instead of directly using an `a` tag:
```suggestion
    <Link href="https://github.com/blockprotocol/blockprotocol">
```
(it's currently defined in `src/components/Link`)

site/src/components/Footer.tsx
```diff
@@ -87,38 +83,67 @@ const SOCIALS: { name: string; icon: ReactNode; href: string }[] = [
...
+          backgroundColor: (theme) => theme.palette.gray[70],
+          color: (theme) => theme.palette.gray[20],
         }}
+        startIcon={<Icon className="fa fa-star" />}
       >
-        {icon}
-      </Link>
-    ))}
+        Star us on Github
+      </Button>{" "}
```
benwerner01, line 145: I think this trailing space is unintentional, likely introduced whilst running `prettier` or `lint:fix`:
```suggestion
      </Button>
```

benwerner01 (approved):
Thank you Maggie, looks good to me 👍

Spotted some minor things which might be worth improving, am happy to commit these changes to the branch myself (if you agree with them)

MaggieAppleton:
Those changes all look good - thanks for the fixes!

site/src/components/Footer.tsx
kachkaev, line 133: If I understand correctly, the new button pretty much replicates the GitHub social icon. It points to https://github.com/blockprotocol/blockprotocol instead of https://github.com/blockprotocol, but maybe the social link is wrong and should be pointing directly to the monorepo. There’s not much else in the org.

Not sure if it’s just me, but I’ve been always afraid of pressing custom-looking _fork_ and _star_ buttons. They feel like you’d make an action right away, without any further confirmation. Seeing the repo makes sense first, so I tend to not press them.

Perhaps, we could use widgets that look more GitHub-y? Here’s an example of what I mean: https://prettier.io

Screenshot 2022-01-04 at 14 45 52

Buttons like that still just take you to the repo because GitHub does not allow XSS, but they are subjectively less scary to press. You can also click the number of stargazers, which would take you to https://github.com/blockprotocol/blockprotocol/stargazers.

Here are some unofficial GitHub widgets, but I’m sure there are more:
- https://buttons.github.io
- https://ghbtns.com

All this is optional, especially before the launch. Just sharing some random thoughts 🙂

benwerner01: Have made this change in https://github.com/blockprotocol/blockprotocol/pull/75/commits/ae821f0ba94307366141bb6428de3c7c7abe9030

site/src/components/Footer.tsx
benwerner01: Have made this change in https://github.com/blockprotocol/blockprotocol/pull/75/commits/ae821f0ba94307366141bb6428de3c7c7abe9030

site/src/components/Footer.tsx
benwerner01: Have made this change in https://github.com/blockprotocol/blockprotocol/pull/75/commits/ae821f0ba94307366141bb6428de3c7c7abe9030

site/src/components/Footer.tsx
MaggieAppleton: Appreciate the thoughts! We could consider adding official github buttons later on - agree it might lead to a greater sense of trust / less mystery about what the button does.  For the moment we intentionally decided to leave the star count off until it's 100+

PR #76, opened 2022-01-04
[SITE] Fix a few React warnings

kachkaev:
See commit messages for details

### DIFF
--- a/site/src/components/Footer.tsx
+++ b/site/src/components/Footer.tsx
@@ -42,13 +42,13 @@ const FooterNavigationLinks = FOOTER_NAVIGATION_LINKS.map(({ href, name }) => (
     component="p"
     variant="bpSmallCopy"
     sx={{
-      "&:first-child": {
+      "&:first-of-type": {
         marginTop: {
           xs: 1.5,
           sm: 0,
         },
       },
-      "&:not(:first-child)": {
+      "&:not(:first-of-type)": {
         marginTop: 1.5,
       },
     }}
--- a/site/src/components/Navbar.tsx
+++ b/site/src/components/Navbar.tsx
@@ -440,9 +440,7 @@ export const Navbar: VFC<NavbarProps> = ({ navbarHeight, setNavbarHeight }) => {
 
   const [displayMobileNav, setDisplayMobileNav] = useState<boolean>(false);
   const [idleScrollPosition, setIdleScrollPosition] = useState<boolean>(false);
-  const [scrollY, setScrollY] = useState<number>(
-    typeof window === "undefined" ? 0 : window.scrollY,
-  );
+  const [scrollY, setScrollY] = useState<number>(0);
 
   useEffect(() => {
     let timer: NodeJS.Timeout | undefined = undefined;
--- a/site/src/components/pages/home/Section2/Steps.tsx
+++ b/site/src/components/pages/home/Section2/Steps.tsx
@@ -343,7 +343,7 @@ const SchemaBlock: VFC<SchemaBlockProps> = ({
           borderRadius: "6px",
         }}
       >
-        {content.map(({ key, value }) => (
+        {content.map(({ key, value }, index) => (
           <Box
             sx={{
               display: "flex",
@@ -354,7 +354,8 @@ const SchemaBlock: VFC<SchemaBlockProps> = ({
               typography: "bpSmallCopy",
               fontWeight: 500,
             }}
-            key={key}
+            // eslint-disable-next-line react/no-array-index-key -- ListItem is repeated, so "key" cannot be used
+            key={index}
           >
             <Box
               sx={{
--- a/site/src/components/pages/home/Section2/Steps.tsx
+++ b/site/src/components/pages/home/Section2/Steps.tsx
@@ -445,8 +445,9 @@ const App: FC<AppProps> = ({ name, block }) => {
 const LayoutBg = () => {
   return (
     <svg
-      width={619}
-      height="auto"
+      width={620}
+      height={520}
+      viewBox="-20 -20 640 540"
       fill="none"
       xmlns="http://www.w3.org/2000/svg"
       style={{ width: "80%" }}
--- a/site/src/pages/_document.page.tsx
+++ b/site/src/pages/_document.page.tsx
@@ -54,10 +54,6 @@ class MyDocument extends Document {
             href="https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500;600;700;800;900&display=swap"
             rel="stylesheet"
           />
-          <script
-            src="https://kit.fontawesome.com/31605611a9.js"
-            crossOrigin="anonymous"
-          />
           <link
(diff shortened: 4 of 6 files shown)


site/src/components/Navbar.tsx
```diff
@@ -440,9 +440,7 @@ export const Navbar: VFC<NavbarProps> = ({ navbarHeight, setNavbarHeight }) => {
 
   const [displayMobileNav, setDisplayMobileNav] = useState<boolean>(false);
   const [idleScrollPosition, setIdleScrollPosition] = useState<boolean>(false);
-  const [scrollY, setScrollY] = useState<number>(
-    typeof window === "undefined" ? 0 : window.scrollY,
-  );
+  const [scrollY, setScrollY] = useState<number>(0);
```
benwerner01, line 443: Just for my understanding, what's the thinking behind this? Why not initialise the state with the correct value on the first render when the page is being rendered in the browser?

site/src/components/pages/home/Section2/Steps.tsx
```diff
@@ -354,7 +354,8 @@ const SchemaBlock: VFC<SchemaBlockProps> = ({
               typography: "bpSmallCopy",
               fontWeight: 500,
             }}
-            key={key}
+            // eslint-disable-next-line react/no-array-index-key -- ListItem is repeated, so "key" cannot be used
+            key={index}
```
benwerner01, line 358: Good spot 🙏

site/src/pages/_document.page.tsx
```diff
@@ -92,6 +88,11 @@ class MyDocument extends Document {
         <body>
           <Main />
           <NextScript />
+          <script
+            defer
+            src="https://kit.fontawesome.com/31605611a9.js"
+            crossOrigin="anonymous"
+          />
```
kachkaev, line 95: We should probably change how we use font awesome because the current one changes the DOM outside React. This may lead to crashes because React reconciler expects that it ‘owns’ the tree.

Here’s what the DOM looks like at the moment:

Screenshot 2022-01-04 at 23 20 54

The comment is a Node that was created by React but then removed by a third-party Font-awesome script. My fix is to do with delaying this dodgy behaviour till the tree is re-hydrated, but the approach is still fundamentally wrong.

site/src/components/pages/home/Section2/Steps.tsx
```diff
@@ -444,8 +445,9 @@ const App: FC<AppProps> = ({ name, block }) => {
 const LayoutBg = () => {
   return (
     <svg
-      width={619}
-      height="auto"
+      width={620}
+      height={520}
+      viewBox="-20 -20 640 540"
```
kachkaev, line 450: Adding viewbox with negative values helps avoid gradient trimming.

site/src/components/Footer.tsx
```diff
@@ -42,13 +42,13 @@ const FooterNavigationLinks = FOOTER_NAVIGATION_LINKS.map(({ href, name }) => (
     component="p"
     variant="bpSmallCopy"
     sx={{
-      "&:first-child": {
+      "&:first-of-type": {
```
kachkaev, line 45:
```
The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type".
```

Background: https://infinum.com/handbook/frontend/react/common-ssr-errors

Not sure if this change can have any negative side-effects, so worth a quick visual check.

site/src/pages/_document.page.tsx
```diff
@@ -54,10 +54,6 @@ class MyDocument extends Document {
             href="https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500;600;700;800;900&display=swap"
             rel="stylesheet"
           />
-          <script
-            src="https://kit.fontawesome.com/31605611a9.js"
-            crossOrigin="anonymous"
-          />
```
benwerner01, line 60: What's the thinking behind moving this script into the body? To increase page-load speed?

The font [awesome docs](https://fontawesome.com/v5.15/how-to-use/javascript-api/setup/configuration) seem to indicate to place this into the header instead, so I'd be inclined to keep it here for now.

Either-way this isn't a big deal as I think we're going to change how these icons are loaded soon anyway for two reasons:
1. so the SVGs aren't fetched by the client, but statically added when the page is built by NextJs
2. so that the SVGs are treated as react children (currently react is not aware of them, so we can't do things like update the styling of an SVG element using JS)

benwerner01 (approved):
I'm happy for this to be merged, thanks @kachkaev 😀

site/src/components/Footer.tsx
benwerner01: Did a quick check by looking at the footer navigation links, and seems to have worked 👍

site/src/pages/_document.page.tsx
benwerner01: Understood, happy to do it like this for. I've come across this before and didn't think of that solution. The reason I've left this unaddressed because I think we should change how we load the font awesome icons.

As I understand it, currently the SVG icons are loaded and added to the DOM by the front awesome script on the client side. If I'm correct, this would explain why the icons currently aren't instantly loading (they appear a couple milliseconds after the rest page has loaded).

Ideally we want the SVG icons to be attached to the document when the page is built on the server by Next.js instead, so that they can be provided to the browser directly (instead of relying on the browser to fetch them). I haven't tested this out, but I think we could accomplish this by using the font awesome NPM packages (such as `@fortawesome/fontawesome-svg-core`) to import the svgs directly.

I haven't spent a lot of time researching/testing this further as I've considered it a lower priority item - nonetheless I think we should do this at some point soon, so that the way we are using icons in development of the FE is more future-proof.

kachkaev: Yeah I think we are aligned here. We don’t have to change the approach immediately, but worth being cautious about what the non-react script does.

`@fortawesome/fontawesome-svg-core` and `@fortawesome/react-fontawesome` should be the right way to go. I would not worry too much about adding more icons before the switchover, because we can bulk-replace them with Search later on. Still, it's something we want to do within weeks than months.

site/src/components/Navbar.tsx
kachkaev: When we use SSR, the first CSR render should be identical to the SSR results. So instead of:

```ts
const [answer, setAnswer] = useState(typeof window === "undefined" ? 0 : 42);
```

it should be:

```ts
const [answer, setAnswer] = useState(0);

useEffect(() => {
  setAnswer(42);
}, []);
```

Calls to `useEffect` are ignored on the server. The page will flickr in both cases, because the first version of the rendered page is server-side-rendered and there is a time gap between html rendering and bundle loading. `useEffect` just adds an extra 10ms delay, but makes UI behaviour more stable.

If we want to start with `42` in the case above, we should think how to get that value on the server. This is not always possible, e.g. because the value depends on window size. If that’s the case, we can make the component client-only or think about ways to replace state with CSS media queries etc.

PR #83, opened 2022-01-07
[SITE] Signup API Routes

benwerner01:
This PR adds the following Next API user signup routes:
- `POST /api/signup`: creates a user with an email address, and sends an verification email to that email address
- `POST /api/verifyEmail`: verifies the email address of a created user, and authenticates them with the BP API
- `POST /api/completeSignup`: completes the signup process of the user, by letting them set their `shortname` and `preferredName`
- `GET /api/isShortnameTaken`: returns a boolean whether a shortname has already been taken by an existing user

Future todos:
- [subscribe users to mailchimp list on signup](https://app.asana.com/0/1201414765539664/1201588774378622/f)


site/src/pages/api/verifyEmail.api.ts
```diff
@@ -0,0 +1,81 @@
...
+        formatErrors({
+          msg: "Could not find user with the provided id",
+          value: userId,
+        }),
+      );
+    }
+
+    const { emailVerificationCodeId, code } = body;
+
+    const emailVerificaitonCode = await user.getVerificationCode(db, {
```
CiaranMn, line 47: Typo here and below: `emailVerificaitonCode`

site/src/pages/api/verifyEmail.api.ts
```diff
@@ -0,0 +1,81 @@
...
+      verificationCodeId: emailVerificationCodeId,
+      variant: "email",
+    });
+
+    if (!emailVerificaitonCode) {
+      return res.status(404).json(
+        formatErrors({
+          msg: "Could not find email verification code associated with user",
+          param: "emailVerificaitonCode",
+          value: emailVerificaitonCode,
```
CiaranMn, line 57: Shouldn't this be returning the invalid `emailVerificationCodeId`?

site/src/pages/api/verifyEmail.api.ts
```diff
@@ -0,0 +1,81 @@
...
+    const { db, body } = req;
+    const { userId } = body;
+
+    const user = await User.getById(db, { userId });
+
+    if (!user) {
+      return res.status(404).json(
+        formatErrors({
+          msg: "Could not find user with the provided id",
+          value: userId,
```
CiaranMn, line 40: We don't include `param: "userId"` here like we do in the other error messages - I'm assuming something to do with how the FE will use them?

site/src/lib/model/user.model.ts
```diff
@@ -66,72 +89,131 @@ export class User {
...
+    const { verificationCodeId, variant } = params;
+
+    const possibleVerificationCodes =
+      this[variant === "login" ? "loginCodes" : "emailVerificationCodes"];
+
     if (
-      !this.loginCodes.find(({ oid }) => oid.toString() === params.loginCodeId)
+      !possibleVerificationCodes.find(
+        ({ oid }) => oid.toString() === verificationCodeId,
+      )
```
CiaranMn, line 194: I think storing the `userId` and type on the codes rather than in the user->code direction would be better as it would allow you to get rid of this, in favour of trying to find a code in the db with the right id and type, but on balance I don't think it worth spending the time to update, since this is a temporary implementation.

site/src/pages/api/completeSignup.api.ts
```diff
@@ -0,0 +1,54 @@
...
+      const errors = validationResult(req);
+      if (!errors.isEmpty()) {
+        return res.status(400).json(formatErrors(...errors.array()));
+      }
+
+      const { db, user, body } = req;
+
+      const { shortname, preferredName } = body;
+
+      if (user.shortname) {
```
CiaranMn, line 38: you could only throw if they are trying to set a different one

```suggestion
      if (user.shortname !== shortname) {
```

site/src/pages/api/completeSignup.api.ts
```diff
@@ -0,0 +1,54 @@
...
+        return res.status(400).json(
+          formatErrors({
+            msg: "Cannot update shortname of user",
+            param: "shortname",
+            value: body.shortname,
```
CiaranMn, line 43:
```suggestion
            value: shortname,
```

site/src/pages/api/completeSignup.api.ts
```diff
@@ -0,0 +1,54 @@
...
+          }),
+        );
+      }
+
+      /** @todo: validate shortname for correct characters */
```
CiaranMn, line 48: and length (min 4, max 24, inclusive)

CiaranMn (changes requested):
Thanks Ben - great work powering through this!

A few small things - there's a return that needs fixing which is the only thing required to merge.

site/src/pages/api/verifyEmail.api.ts
benwerner01: Yes, good spot 👍

site/src/pages/api/verifyEmail.api.ts
benwerner01: I simply forgot to add it, will do so. I'm yet sure whether the errors displayed on the frontend should be defined in the backend or not.

site/src/pages/api/completeSignup.api.ts
benwerner01: Sure 👍

site/src/lib/model/user.model.ts
benwerner01: I'm honestly not sure why I didn't think to do this, have made this change in a recent commit becuase this would b bother me every time I see it (https://github.com/blockprotocol/blockprotocol/pull/83/commits/aeae7876d331e986a31acabbe971c6e9778a3064)

site/src/pages/api/completeSignup.api.ts
benwerner01: Have added shortname validation in my most recent commit:  https://github.com/blockprotocol/blockprotocol/pull/83/commits/4c5c1fe586c2d4de085c5cad466a63fac4096f45

site/src/lib/model/user.model.ts
```diff
@@ -220,19 +263,30 @@ export class User {
...
+            variant: "login",
+          },
+          {
+            createdAt: {
+              $gt: new Date(
+                new Date().getTime() - User.LOGIN_CODE_RATE_LIMIT_PERIOD_MS,
+              ),
+            },
+          },
+        ],
```
CiaranMn, line 280: You can express conditions like this (require multiple key-value pairs on a document) as a single object, but in practice it's probably optimized to make no difference except to the code. 

```TypeScript
          {
            user: this.toRef(),
            variant: "login",
            createdAt: {
              $gt: new Date(
                new Date().getTime() - User.LOGIN_CODE_RATE_LIMIT_PERIOD_MS,
              ),
            },
          },
```

CiaranMn (approved):
Thanks Ben, looks good to me

PR #84, opened 2022-01-09
[SITE] BP Site login UI

benwerner01:
This PR introduces the login UI for the BP site (both as a modal, and at a dedicated /login page). Quick demo:

https://user-images.githubusercontent.com/42802102/149024317-ef9f5e1e-5dbd-4d49-b61c-c328288a4efd.mp4

Additional changes include:
- added an `apiClient` object that let's users easily query the API, which uses `axios` under the hood for conveniences such as body/error parsing
- added a custom `TextField` component (as a wrapper of the MUI `TextField` component), so that a collapse animation can be used for the helper text in a reusable way
- added a loading animation for the `Button` component
- added a `AccountDropdown` which displays the account navigation options in the navbar when the user is logged in
- split the `Navbar` underlying component into several files (the file was approaching 1000 lines of code so deemed this as a more maintainable alternative)
- added a global user react context which makes the user object accessible when the user is logged in
- modified the `sendLoginCode` user model method to generated a magic link

Known issues:
- at least in local development, the `me` request hangs indefinitely and a `ERR_INVALID_ARG_TYPE` is thrown by the API on every second request. This causes a bug where the UI reflects the logged in state on every second page load (I need to look into this further as it has to be resolved for Friday)

Future Todos:
- display custom avatar in the Navbar uploaded by user


site/src/components/Navbar/AccountDropdown.tsx
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  const handleLogout = async () => {
+    await apiClient.post("logout");
+
+    refetch();
```
CiaranMn, line 33: It might be better to have a 'clear' method for the user - it saves the network request, and achieves what we know we want to do (clear the user from state) in a more direct way.

CiaranMn (approved):
Thanks @benwerner01 - I've tested this and it's working for me. I don't encounter the `me` query hanging (tried switching pages client-side and refreshing a couple of times).

On environment variables, I think we could usefully provide either (a) an example of the `.env.local` file with local defaults in the README, or (b) an `.env.local` file with the defaults. 

Can we also move the 'local development' section of the readme above the API routes? This could be in a separate PR.

site/src/components/Navbar/AccountDropdown.tsx
benwerner01: I'll do the same for the `handleLogin` function (directly update the user context using a `setUser` dispatch method using the response of the `/api/loginWithLoginCode` request)

site/src/components/Modal/LoginWithLoginCodeScreen.tsx
```diff
@@ -0,0 +1,325 @@
...
+        );
+      } else {
+        throw error;
+      }
+    } else if (data) {
+      setVerificationCode("");
+      setApiLoginErrorMessage(undefined);
+      setTouchedVerificationCodeInput(false);
+      setLoginCodeId(data.loginCodeId);
+      if (loginCodeInputRef.current) {
```
teenoh, line 107: This will lead to multiple re-renders. React (<v18) batches updates in event handlers but that doesn't apply when the handler is an async function.

> However, React wasn’t consistent about when it batches updates. For example, if you need to fetch data, and then update the state in the handleClick above, then React would not batch the updates, and perform two independent updates.
This is because React used to only batch updates during a browser event (like click), but here we’re updating the state after the event has already been handled (in fetch callback)

[Link](https://github.com/reactwg/react-18/discussions/21#discussion-3385721)

It might be better if we handle state management with useReducer. Can also use `unstable_batchedUpdates` to fix the immediate issue if we don't want to go through `useReducer` route

site/src/components/Modal/LoginModal.tsx
```diff
@@ -0,0 +1,172 @@
...
+export const LoginModal: VFC<LoginModalProps> = ({
+  onClose,
+  ...modalProps
+}) => {
+  const { user, refetch } = useContext(UserContext);
+  const [currentPage, setCurrentPage] = useState<LoginModalPage>("Email");
+  const [email, setEmail] = useState<string>("");
+
+  const [userId, setUserId] = useState<string>();
+  const [loginCodeId, setLoginCodeId] = useState<string>();
```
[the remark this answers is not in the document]
benwerner01: Yes that assumption is correct, have made the change here https://github.com/blockprotocol/blockprotocol/pull/84/commits/82348baae69b63f1684883fb73e5063227e79cd8

site/src/lib/model/user.model.ts
```diff
@@ -7,6 +7,9 @@ import {
   VerificationCodeDocument,
   VerificationCodeVariant,
 } from "./verificationCode.model";
+import { ApiLoginWithLoginCodeRequestBody } from "../../pages/api/loginWithLoginCode.api";
+
+const FRONTEND_URL = "http://localhost:3000";
```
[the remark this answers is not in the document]
benwerner01: Added the environment variable here: https://github.com/blockprotocol/blockprotocol/pull/84/commits/3d31f5a3196445a179ee2dcec02c5f268de31b0f

site/src/components/Navbar/AccountDropdown.tsx
benwerner01: Have done this here: https://github.com/blockprotocol/blockprotocol/pull/84/commits/264e2a914e89b89d4882eb71a894fb97075f8f79

site/src/components/Modal/LoginWithLoginCodeScreen.tsx
benwerner01: Good spot, thank you. I'm going to opt for `unstable_batchedUpdates` for now, as I don't want to invest too much time into perfecting how the state is stored here as it will likely change once we come up with a common approach for capturing the `TextField` related state.

benwerner01: But if I'm understanding correctly this will be resolved once we move to React > 18 anyway

benwerner01: Addressed in: https://github.com/blockprotocol/blockprotocol/pull/84/commits/4d73656ae7318724825a036574c4e486f6e5d9e7

site/src/components/Button.tsx
```diff
@@ -12,17 +79,21 @@ type ButtonProps = {
...
+    return (
+      <MuiButton
+        {...props}
+        sx={{
+          borderRadius: squared ? "6px" : undefined,
+          /** @todo: figure out how to properly merge this object with `props.sx` */
+          ...props.sx,
+          lineHeight: "1.5",
+        }}
+        ref={ref}
```
teenoh, line 93: You can do 
```suggestion
  ({ children, squared, loading, sx, ...props }, ref) => {
    return (
      <MuiButton
        {...props}
        sx={{
          borderRadius: squared ? "6px" : undefined,
          ...sx,
          lineHeight: "1.5",
        }}
        ref={ref}
```

benwerner01: Unless I'm missing something, that would be equivalent. What I'm referring to here is the fact that the `sx` prop can be a function, and when that is the case cannot be spreaded using `..props.sx` (or `...sx`). For now this means don't provide the a function to the `sx` prop of the custom `Button` wrapper. A better long-term solution might be to enforce this by modifying the type of the `sx` prop for this component, or preferably figuring out a way to properly merge `props.sx` with the extended styling.

site/src/components/theme.ts
```diff
@@ -320,6 +313,17 @@ export const theme = createTheme({
               scroll-behavior: smooth;
             }
 
+            body {
+              overflow: overlay;
```
teenoh, line 317: overlay isn't a valid value for overflow.

site/src/pages/login.page.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  useEffect(() => {
+    const { query } = router;
+
+    if (tbdIsLoginPageParsedUrlQuery(query)) {
+      const { redirectAsPath, ...requestBody } = query;
+
+      setCurrentPage("VerificationCode");
+
+      /** @todo: set to verification code screen */
```
teenoh, line 76: seems you already handled this todo

site/src/components/Button.tsx
teenoh: Ohh... I assumed the issue was the fact that we were assigning the `sx` prop twice. I get what you mean now.

site/src/components/context/UserContext.ts
```diff
@@ -0,0 +1,16 @@
+import { createContext, Dispatch, SetStateAction } from "react";
+import { SerializedUser } from "../../lib/model/user.model";
+
```
teenoh, line 3: I think we should move the `context` folder up since it isn't exactly a component.
Structure should look like this

```
site
  - context
  - hooks
  - components
    - Button
    - Modal
  - pages
```

instead of 

```
site
  - components
     - context
     - Modal
  - pages     
```

teenoh (approved):
Thanks @benwerner01. This is working for me too. I encountered the issue with `ERR_INVALID_ARG_TYPE`. 
Asides from that and some comments I made, this looks good to go 👍

site/src/components/context/UserContext.ts
benwerner01: I have no objections to that, done this in https://github.com/blockprotocol/blockprotocol/pull/84/commits/eac5c0b3d7f9cb9ac1a85570be221ac2bfc7724a

site/src/pages/login.page.tsx
benwerner01: Good spot 😅

site/src/components/theme.ts
benwerner01: It is for some browsers (see https://caniuse.com/css-overflow-overlay), when it isn't supported the browser would default to `overflow: auto`.

The reason I added this is because this prevents the Navbar from shifting slightly to the right when the login modal is opened.

Before:
https://user-images.githubusercontent.com/42802102/149164575-7d19a6e0-02a0-4619-97a0-48a6cceabb2a.mov


After:
https://user-images.githubusercontent.com/42802102/149164609-1e15b2ec-9918-47e0-acae-2aebe47c7b00.mov

But I'm not sure if this is the best solution, so I'll add a todo here for now.

PR #89, opened 2022-01-12
[tool] build all blocks everytime

espretto:
Vercel persists nextjs' cache between builds. That's why I used it to persist block builds, too.
However, vercel provides fresh cache instances for each branch and seems to clear the target branch's
cache when merging. This PR removes all caching functionality and changes the build command so that all blocks get build everytime.

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -5,7 +5,7 @@
   "repository": "https://github.com/blockprotocol/blockprotocol.git",
   "license": "MIT",
   "scripts": {
-    "build": "./scripts/build-blocks.sh && yarn generate-sitemap && next build",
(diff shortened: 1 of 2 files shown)


teenoh (approved):
Thanks Marius!

CiaranMn (approved):
Thanks Marius - spotted something that might be redundant now - looks good otherwise.

```sh
  # mirror deletions in the cache
  if ! [[ -f "$build_config" ]]; then
    log info "removing cached build for removed ${build_config}"
    [[ -d "$public_dir" ]] && rm -rf "$public_dir"
    return
  fi
``

PR #90, opened 2022-01-12
[Site] v0.2

benwerner01:
This PR implements v0.2 of the BP site, introducing:
- user accounts
- a search endpoint that enables consumers to search for blocks using an API key


kachkaev:
@benwerner01 @CiaranMn let’s enable rebasing when the PR is ready to merge. It’d be nice to keep commits in this case as long as each of them represents a single PR. This will make `git blame` a bit better.

CiaranMn:
> @benwerner01 @CiaranMn let’s enable rebasing when the PR is ready to merge. It’d be nice to keep commits in this case as long as each of them represents a single PR. This will make `git blame` a bit better.

Sure, ping me when that is

PR #91, opened 2022-01-12
Add redirect from `/discord`

kachkaev:
This PR unifies links to Discord our server by creating a new `/discord` redirect endpoint. It is now used both in internal pages and markdowns. Markdown links are absolute, so we need to merge this PR for them to start working.

Screenshot 2022-01-12 at 16 02 36

I was thinking of putting the link to an env variable, but then decided to hardcode it right in `next.config.js`. That’s one less abstraction to think about if a link changes. I don’t think it will happen too often if at all TBH.

[Asana task](https://app.asana.com/0/1201414765539664/1201562956971361/f) (internal link)

### DIFF
--- a/site/src/components/Footer.tsx
+++ b/site/src/components/Footer.tsx
@@ -81,7 +81,7 @@ const SOCIALS: { name: string; icon: ReactNode; href: string }[] = [
   {
     name: "Discord",
     icon: <Icon className="fab fa-discord" />,
-    href: "https://discord.gg/PefPteFe5j",
+    href: "/discord",
   },
   {
     name: "GitHub",
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Thanks for taking the time to contribute! 🎉 HASH has established a set of [community guidelines](https://hash.ai/legal/community) to enable as many people as possible to contribute to and benefit from the Block Protocol. Please follow these when interacting with this repo.
 
-We have a community [Discord server](https://discord.gg/PefPteFe5j) where you can ask questions about contributing to the Block Protocol, and get support on its use. If you'd like to make a significant change or re-architecture, please first discuss the change there (or create an [issue](https://github.com/blockprotocol/blockprotocol/issues)) to get feedback before spending too much time.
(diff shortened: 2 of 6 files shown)


nonparibus (approved):

site/src/components/Footer.tsx
```diff
@@ -81,7 +81,7 @@ const SOCIALS: { name: string; icon: ReactNode; href: string }[] = [
   {
     name: "Discord",
     icon: <Icon className="fab fa-discord" />,
-    href: "https://discord.gg/PefPteFe5j",
+    href: "/discord",
```
CiaranMn, line 84: I expect that this link will not work as the redirects in `next.config.js` are for server-side only. We could either (a) make this absolute or (b) add a `pages/discord.tsx` which handles the redirect client-side.

kachkaev: I checked this locally by clicking on the footer link – all worked.
https://blockprotocol-8h51t8v39-hashintel.vercel.app/contact

When we run `next build`, Next creates a catalog of all routes including the redirects, so the client bundle knows how to distinguish a redirect from another page.

If we want Discord link to open in a new tab, I can force it by adding some attributes.

kachkaev: 82ed24ca8fdeb08eaf94a8396e0113825de2574a

CiaranMn: Oh, weird - it never used to and [their own docs](https://nextjs.org/docs/api-reference/next.config.js/redirects) still say it doesn't. Maybe a Vercel-specific thing? Apologies!

I think opening in a new tab would be better for anything off-site.

kachkaev: > Redirects are only available on the Node.js environment and do not affect client-side routing.

Yeah, that’s a strange statement. Looks like it’s outdated indeed. Just ran `npx create-next-app --example redirects redirects-app` to make sure I’m not crazy – the [example](https://github.com/vercel/next.js/tree/canary/examples/redirects) worked for me (even for internal redirects).

Screenshot 2022-01-12 at 16 29 48

This does not mean I’m not crazy though 😅

PR #93, opened 2022-01-12
Prepare `blockprotocol@0.0.1` release

kachkaev:
This PR moves blockprotocol typings to `packages/blockprotocol`, which is a new package. I will publish it as https://npmjs.com/package/blockprotocol after we merge. I’ve picked a very modest version number for now to support quick iterations early in the game.


site/package.json
```diff
@@ -1,8 +1,7 @@
 {
   "name": "@blockprotocol/site",
-  "version": "0.1.0",
-  "description": "The public-facing blockprotocol.org website",
-  "repository": "https://github.com/blockprotocol/blockprotocol.git",
+  "version": "0.0.0",
```
kachkaev, line 3: V `0.0.0` is a way of saying ‘this won’t be published on NPM. Turborepo examples to that. I used `0.0.0-local` in the past, but `0.0.0` should work too.

site/package.json
```diff
@@ -1,8 +1,7 @@
...
+  "private": true,
```
kachkaev, line 4: This will help us avoid accidental publishing to NPM. I removed description and repository because they are not practically useful for non-punishable packages.

site/package.json
```diff
@@ -48,8 +47,8 @@
     "@babel/runtime": "^7.16.3",
     "@types/lodash": "4.14.171",
     "@types/prismjs": "^1.16.6",
-    "@types/react": "17.0.7",
-    "@types/react-dom": "17.0.8",
+    "@types/react": "17.0.38",
+    "@types/react-dom": "17.0.11",
```
kachkaev, line 51: Ensures we have only one of each in yarn.lock

CiaranMn (approved):
Thanks Alex - we need to agree on how we'll manage bringing over any changes to the BP package in `dev` and then switch to using the published package in the `dev` repo.

As there are a couple of PRs in flight which affect those types, I suggest we get a version published so that's all in place, and then pick a suitable point to move the latest types from `dev` to here, publish a new version, and replace imports in `dev`.

PR #97, opened 2022-01-14
[SITE] BP site signup UI

benwerner01:
This PR implements the signup user flow on the BP site. Additional changes
- replaced `express-session` with `next-session` (resolves session issues locally and in deployment)

Future tasks:
- [suggest usernames based on email](https://app.asana.com/0/1201095311341924/1201652492640268/f)

### DIFF
--- a/site/src/components/Screens/CompleteSignupScreen.tsx
+++ b/site/src/components/Screens/CompleteSignupScreen.tsx
@@ -1,0 +1,152 @@
+import { Box, Icon, Typography } from "@mui/material";
+import { useContext, useRef, useState, VFC } from "react";
+import { TextField } from "../TextField";
+import { Button } from "../Button";
+import { useShortnameTextField } from "../hooks/useShortnameTextField";
+import { apiClient } from "../../lib/apiClient";
+import {
+  ApiCompleteSignupRequestBody,
+  ApiCompleteSignupResponse,
+} from "../../pages/api/completeSignup.api";
+import UserContext from "../../context/UserContext";
+
+type CompleteSignupScreenProps = {
+  email: string;
+};
+
+export const CompleteSignupScreen: VFC<CompleteSignupScreenProps> = ({
+  email,
+}) => {
+  const { setUser } = useContext(UserContext);
+
+  const shortnameInputRef = useRef<HTMLInputElement>(null);
+  const preferredNameInputRef = useRef<HTMLInputElement>(null);
+
+  const [touchedPreferredNameInput, setTouchedPreferredNameInput] =
+    useState<boolean>(false);
+  const [touchedShortnameInput, setTouchedShortnameInput] =
+    useState<boolean>(false);
+
+  const [completingSignup, setCompletingSignup] = useState<boolean>(false);
+
+  const { shortname, setShortname, isShortnameValid, shortnameHelperText } =
+    useShortnameTextField({
+      touched: touchedShortnameInput,
+    });
+
+  const [preferredName, setPreferredName] = useState<string>("");
+
+  const isPreferredNameEmpty = !preferredName;
+
+  const isPreferredNameValid = !isPreferredNameEmpty;
+  const preferredNameHelperText = isPreferredNameEmpty
+    ? "You must choose a preferred name"
+    : undefined;
+
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    if (isShortnameValid && isPreferredNameValid) {
+      setCompletingSignup(true);
+      const { data, error } = await apiClient.post<
+        ApiCompleteSignupRequestBody,
+        ApiCompleteSignupResponse
+      >("completeSignup", { shortname, preferredName });
+      setCompletingSignup(false);
+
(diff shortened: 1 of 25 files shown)


site/src/components/Screens/SendLoginCodeScreen.tsx
```diff
@@ -98,7 +114,7 @@ export const SendLoginCodeScreen: VFC<SendLoginCodeScreenProps> = ({
           inputRef={emailInputRef}
           fullWidth
           label="Email address"
-          placeholder="you@example.com"
+          placeholder="claude@shannon.com"
```
CiaranMn, line 117: This is a real non-exampley website, I don't think we should be using it. maybe `claude@example.com`

site/src/components/Screens/CompleteSignupScreen.tsx
```diff
@@ -0,0 +1,152 @@
...
+      >
+        Add your account details
+      </Typography>
+      <Box
+        width="100%"
+        component="form"
+        onSubmit={(event: React.FormEvent<HTMLFormElement>) => {
+          event.preventDefault();
+          void handleSubmit();
+        }}
```
CiaranMn, line 101: I'd move the `event.preventDefault()` to `handleSubmit` as the function is split at the moment where it doesn't really need to be(?)

site/src/components/hooks/useShortnameTextField.ts
```diff
@@ -0,0 +1,77 @@
...
+    () =>
+      debounce(async (params: { shortname: string }) => {
+        const { data: isShortnameTaken, error } = await apiClient.get<
+          ApiIsShortnameTakenResponse,
+          ApiIsShortnameTakenQueryParams
+        >("isShortnameTaken", params);
+
+        if (error) {
+          throw error;
+        } else if (isShortnameTaken !== undefined) {
```
CiaranMn, line 40: Why compare to `undefined` rather than for truthiness? 

If `null` or `false` ever end up as a value here for `isShortnameTaken`, it'll show as taken. Appreciate that's not how it is but might trip us up later.

site/src/components/Screens/SendLoginCodeScreen.tsx
```diff
@@ -90,6 +105,7 @@ export const SendLoginCodeScreen: VFC<SendLoginCodeScreenProps> = ({
       </Typography>
       <Box width="100%" component="form" onSubmit={handleSubmit} noValidate>
         <TextField
+          sx={{ marginBottom: 2 }}
           required
           type="email"
           onBlur={() => setTouchedEmailInput(true)}
```
CiaranMn, line 111: It's not ideal that users who want to sign up will see an error message on this input when they blur it to click the 'Create an account' link - maybe we can set the condition for showing 'empty email' errors to be clicking on the 'log in' button? Or just rely on the native browser 'required' handling?

CiaranMn (approved):
Excellent work getting all this out so quickly, Ben!

I've tested and works well. A few nits on the code.

The only blocking thing is that I don't think we should use `@shannon.com` as a domain in the placeholder.

site/src/components/Screens/CompleteSignupScreen.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const { shortname, setShortname, isShortnameValid, shortnameHelperText } =
+    useShortnameTextField({
+      touched: touchedShortnameInput,
+    });
+
+  const [preferredName, setPreferredName] = useState<string>("");
+
+  const isPreferredNameEmpty = !preferredName;
+
+  const isPreferredNameValid = !isPreferredNameEmpty;
```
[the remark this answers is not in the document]
benwerner01: Ah this was to aid with the `preferredNameHelperText`, will add it to make use of the `isPreferredNameEmpty` variable

site/src/components/Screens/CompleteSignupScreen.tsx
benwerner01: Yeah it doesn't need to be handled here, just like keeping the event of `handle` functions unless is needed. But will change as I don't have strong feelings on this.

site/src/components/hooks/useShortnameTextField.ts
benwerner01: I'm not understanding, `isShortnameTaken` here is the response from the server (either `true`, `false` or `undefined` if an error occurred). If we were to use truthiness, `isTaken` would never be set to `false`, and we couldn't indicate to the user their shortname is taken.

Not sure if I've misunderstood something though

site/src/components/Screens/SendLoginCodeScreen.tsx
benwerner01: The native `required` handling would introduce popups to the UI which I believe in this context are unwanted. Will change this to be called when clicking the `log in` button.

site/src/components/hooks/useShortnameTextField.ts
CiaranMn: Ahh sorry, I misunderstood `setIsTaken`.

PR #98, opened 2022-01-14
[SITE] Add Google Tag Manager init

CiaranMn:
Adds GTM to the site.

### To test
1. Install the [Google Analytics debugger extension](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna/related) - alternatively, watch the network tab for requests
2. Remove the 'VERCEL_ENV' check
3. Switch pages, check pageview events are fired
4. Will need to test again once merged


site/src/pages/_app.page.tsx
```diff
@@ -30,6 +31,12 @@ const MyApp = ({
 }: MyAppProps) => {
   const router = useRouter();
 
+  useEffect(() => {
+    if (process.env.NEXT_PUBLIC_VERCEL_ENV === "production") {
```
benwerner01, line 35: Just to confirm my understanding, `process.env.NEXT_PUBLIC_VERCEL_ENV` is not equivalent to `process.env.NODE_ENV` as `NODE_ENV === "production"` for all deployed branches, but `NEXT_PUBLIC_VERCEL_ENV === "production"` only on the `main` branch deployment?

benwerner01 (approved):
Thanks Ciaran, this is working locally for me 👍

site/src/pages/_app.page.tsx
CiaranMn: Correct - it's one of `production`, `preview`, or `development`, with `production` being the main deployment in Vercel.

PR #100, opened 2022-01-14
[SITE] send verification emails in production using AWS SES

benwerner01:
This PR configures the `sendLoginCode` and `signup` Next API routes to send emails when not run in developer mode. In developer mode the code will continue to be printed to the console for convenience, and so that AWS credentials aren't required to run the developer environment locally.

Note: this branch merged off `bw/signup-ui` to make use of the latest version of the related API routes. So that the diff isn't confusing for reviewers, I've temporarily set-up the PR to be merged into `bw/signup-ui` - but ideally this branch will be merged directly into `v0.2` once `bw/signup-ui` has been merged, which is why I've temporarily assigned the `DO NOT MERGE` label.

Future todos:
- [add styled email template](https://app.asana.com/0/1201095311341924/1201652492640274/f)


site/src/lib/awsSes.ts
```diff
@@ -0,0 +1,48 @@
+import * as aws from "@aws-sdk/client-ses";
+import nodemailer from "nodemailer";
+import { convert } from "html-to-text";
+import { defaultProvider } from "@aws-sdk/credential-provider-node";
+import SESTransport from "nodemailer/lib/ses-transport";
+import { isProduction } from "./config";
+import { mustGetEnvVar } from "../util/api";
+
+const FROM_ADDRESS = "Support <noreply@blockprotocol.org>";
```
kachkaev, line 9: How about moving this value to an env var too? Alternatively, can we specify the sender on the AWS end?

benwerner01: Can create an environment variable, the reason I didn't was because I can't envision this value changing before we deprecate the Next API, unless I've missed something?

CiaranMn: I think it's fine as-is - we don't expect anyone to host their own version of this site so we don't need to account for as much configuration as we do for HASH.

site/src/lib/config.ts
```diff
@@ -5,4 +5,6 @@ export const FRONTEND_DOMAIN =
 
 export const USE_HTTPS = process.env.NODE_ENV !== "development";
 
-export const FRONTEND_URL = `http${USE_HTTPS ? "" : "s"}//${FRONTEND_DOMAIN}`;
+export const FRONTEND_URL = `http${USE_HTTPS ? "s" : ""}://${FRONTEND_DOMAIN}`;
```
CiaranMn, line 8: I think `USE_HTTPS` is unnecessary - we could just have a `FRONTEND_URL` or `FRONTEND_ORIGIN` which includes the protocol, and turn on secure cookies if it starts with `https`.

The current implementation would also prevent (a) running a production build locally without HTTPS enabled, (b) running a development build locally with HTTPS enabled.

site/README.md
```diff
@@ -44,6 +44,14 @@ The public-facing [blockprotocol.org](https://blockprotocol.org) website serves
     yarn dev
     ```
 
+### AWS configuration
+
+If you want to send verification codes to an email address, the following AWS environment variables have to be additionally configured:
+
+- `BP_AWS_REGION`: The region, eg. `us-east-1`
+- `BP_AWS_ACCESS_KEY_ID`: The AWS access key
+- `BP_AWS_SECRET_ACCESS_KEY`: The AWS secret access key
```
CiaranMn, line 53: Do we need these `BP_` prefixes?

CiaranMn (approved):
Thanks Ben - a couple of comments quibbling over variables, nothing blocking.

site/README.md
benwerner01: Unfortunately for now I think we do. Found out the hard way that Vercel reserves the AWS default environment variable names: https://vercel.com/docs/concepts/projects/environment-variables#reserved-environment-variables

CiaranMn: Ahhh, right. Fair enough - sorry for the pain this must've caused before discovering it!

site/src/lib/config.ts
benwerner01: My only issue with this would is that it would make it harder to determine the `FRONTEND_DOMAIN`, as we now have to parse `FRONTEND_URL` for this. But I suppose the following would work:
```TypeScript
const FRONTEND_DOMAIN = (new URL(FRONTEND_URL)).hostname;
```

benwerner01: Are you suggesting we do the following to prevent breaking deployments?
```TypeScript
export const FRONTEND_URL = process.env.NEXT_PUBLIC_VERCEL_URL
  ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
  : process.env.FRONTEND_URL ?? "localhost:3000";
```

CiaranMn: Yes, assuming  `VERCEL_URL` doesn't include the protocol

benwerner01: It doesn't 👍

PR #103, opened 2022-01-15
[SITE] mailchimp

benwerner01:
This PR introduces functionality that subscribes a user to mailchimp and sets their shortname/preferred name during signup.


site/src/lib/mailchimp.ts
```diff
@@ -0,0 +1,66 @@
...
+      baseURL,
+      headers: {
+        "Content-Type": "application/json",
+      },
+      auth: {
+        username,
+        password,
+      },
+    },
+  );
```
CiaranMn, line 65: Could do with a `catch` here

CiaranMn (approved):
Thanks @benwerner01 - looks good for the initial process.

Something we should address in follow-up: this won't re-subscribe users who have entered their email, abandoned signup, unsubscribed from any mail we send them, and then come back to go through the process again. 

I believe the MailChimp API requires you to `patch` the existing user in that case.

nonparibus:
> Thanks @benwerner01 - looks good for the initial process.
> 
> Something we should address in follow-up: this won't re-subscribe users who have entered their email, abandoned signup, unsubscribed from any mail we send them, and then come back to go through the process again.
> 
> I believe the MailChimp API requires you to `patch` the existing user in that case.

This is a good flag. At the same time we don't want to resubscribe users who've consciously unsubscribed when they simply come back to log in again.

site/src/lib/mailchimp.ts
benwerner01: I've added a `.catch` function which for now just logs the error (so that completing signup doesn't fail if the mailchimp api returns an error). Was there additional functionality you had in mind for this?

CiaranMn: Nope, that's good, thank you

PR #104, opened 2022-01-17
[SITE] Add preview images for blocks 

teenoh:
This PR 
- updates the codebase to pull block preview images from the block's metadata file
- adds some styling improvements to the block card and block slider components.

How to test this
- Change the `branch` property in the registry files (`hub/*.json`)  to `vu/block-preview-images`
- Run `yarn build-blocks` from the `site/` directory (might need to delete the `public/blocks` folder before doing this)
- Run `yarn dev`
- The preview images should show up when you go to the hub page ("/hub")

---
  - https://app.asana.com/0/0/1201555682469130


teenoh:
Needs to be merged in after https://github.com/hashintel/dev/pull/476

site/src/pages/api/blocks.api.ts
```diff
@@ -107,11 +66,10 @@ export const readBlocksFromDisk = (): BlockMetadata[] => {
     }))
     .map((metadata: BlockMetadata) => ({
       ...metadata,
+      image: `/blocks/${metadata.packagePath}/${metadata.image}`,
```
CiaranMn, line 69: This assumes that the image URL is relative - can we also handle absolute URLs please?

e.g. check here if it starts with `/`, and if so do this, and if not just include the URL?

Can we also make the same fix for icons?

CiaranMn (approved):
Thanks @teenoh - small fix requested, good to merge with that in

site/src/pages/api/blocks.api.ts
teenoh: Added in this commit 1146653

PR #106, opened 2022-01-18
[SITE] Blocks Search API

teenoh:
This PR adds a basic search endpoint for Blocks.

How to Test
- Open Postman or any REST client
- Set the url to `https://blockprotocol-n4g31hg5w-hashintel.vercel.app/api/blocks?q=<any_word>`
- Set a `x-api-key` header and give it any value
- Confirm it returns filtered responses based on the query

---
  - https://app.asana.com/0/0/1201637454224598


site/scripts/generate-blocks-data.ts
```diff
@@ -0,0 +1,12 @@
+import { writeFileSync } from "fs";
+import path from "path";
+import { readBlocksFromDisk } from "../src/lib/blocks";
+
```
teenoh, line 4: Had to go with the approach of generating a json file that contains blocks data (similar to what we do for sitemap). This is because of the difficulty I faced in making the `api` access the `hub/` folder at runtime. I feel we can work with this for the main time since it's a temporary solution

site/src/pages/api/blocks.api.ts
```diff
@@ -1,125 +1,47 @@
...
+    return res.status(401).json(
+      formatErrors({
+        msg: "Unauthorized",
+      }),
+    );
+  }
+
+  let data: BlockMetadata[] = [];
+
+  // do we return data when the query string is empty?
```
teenoh, line 30: What do we do when the query string is empty? Do we return an empty list or a default set of data. I'm currently returning an empty list

site/src/pages/[shortname]/[blockSlug].page.tsx
```diff
@@ -114,35 +112,20 @@ export const getStaticProps: GetStaticProps<
...
-  )?.lastUpdated;
-
-  const [schema, blockStringifiedSource] = await Promise.all([
-    fetch(
-      `${BASE_URL}/blocks/${shortname}/${blockSlug}/${blockMetadata.schema}`,
-    ).then((res) => res.json()),
-    fetch(
-      `${BASE_URL}/blocks/${shortname}/${blockSlug}/${blockMetadata.source}`,
-    ).then((res) => res.text()),
-  ]);
```
Comment (teenoh) [inline site/src/pages/[shortname]/[blockSlug].page.tsx:145]: This changes are already in main. Just brought them over since I didn't want to pull all the changes there

site/src/pages/api/blocks.api.ts
CiaranMn: Let's return the full list please

site/scripts/generate-blocks-data.ts
CiaranMn: Yes, sounds good.

I'm assuming the hub folder is not accessible as it's outside of the `site` folder, where this application is built from.

site/package.json
```diff
@@ -4,13 +4,14 @@
...
     "build-block": "./scripts/build-blocks.sh",
     "build-blocks": "find ../hub -type f | xargs ./scripts/build-blocks.sh",
     "clean": "rimraf ./.next/",
-    "dev": "yarn generate-sitemap && next dev",
+    "dev": "yarn generate-sitemap && yarn generate-blocks-data && next dev",
     "dev:db": "docker-compose -f docker-compose.dev.yml --env-file .env.local up",
     "dev:seed-db": "node --max-old-space-size=2048 node_modules/.bin/ts-node --files --project tsconfig.json scripts/seed-db",
+    "generate-blocks-data": "node --max-old-space-size=2048 node_modules/.bin/ts-node --files --project tsconfig.json scripts/generate-blocks-data",
     "generate-sitemap": "node --max-old-space-size=2048 node_modules/.bin/ts-node --files --project tsconfig.json scripts/generate-sitemap",
     "postinstall": "yarn generate-sitemap",
```
CiaranMn, line 16: You need to add `yarn generate-blocks-data` here so that the file is available reliably (e.g. in CI)

site/src/pages/api/blocks.api.ts
```diff
@@ -1,125 +1,49 @@
...
-    }));
-};
+  if (query && typeof query === "string") {
+    data = blocksData.filter(
+      ({ displayName, variants }) =>
+        displayName?.toLowerCase().includes(query) ||
+        variants?.some((variant) =>
+          variant.displayName?.toLowerCase().includes(query),
+        ),
+    );
```
CiaranMn, line 39: Can we include 'author' and 'name' in this too please?

site/src/lib/blocks.ts
```diff
@@ -0,0 +1,135 @@
...
+/** @todo make part of blockprotocol package and publish */
+export type BlockVariant = {
+  description?: string;
+  displayName?: string;
+  icon?: string;
+  properties?: BlockProps;
+};
+
+/** @todo make part of blockprotocol package and publish */
+export type BlockMetadata = {
```
CiaranMn, line 13: These can all be imported from "blockprotocol" now (you'll need to add `packagePath` locally) - `blockprotocol` will need adding as a dependency

CiaranMn (changes requested):
Thanks @teenoh, working well - a few bits of cleanup suggested

site/src/pages/api/blocks.api.ts
```diff
@@ -1,125 +1,49 @@
...
 
-export type BuildConfig = {
-  workspace: string;
-  repository: string;
-  branch: string;
-  distDir: string;
-  timestamp: string;
-};
+const blocks: NextApiHandler<Response> = (_req, res) => {
+  const query = _req.query.q;
```
benwerner01, line 17: Depending on the priority of this PR, this might be a good opportunity to make this API handler consistent with the other API handlers. Things that I'm noticing off the top of my head:
1. `express-validator` could be used to validate the query (ensure that `q` is a string if it is supplied for instance)
2. define `ApiBlocksRequestBody` and `ApiBlocksResponse` types for the FE to make use of

site/src/pages/api/blocks.api.ts
```diff
@@ -1,125 +1,49 @@
...
+type Response = { results: BlockMetadata[] } | { errors: any };
 
-export type BuildConfig = {
-  workspace: string;
-  repository: string;
-  branch: string;
-  distDir: string;
-  timestamp: string;
-};
+const blocks: NextApiHandler<Response> = (_req, res) => {
```
benwerner01, line 16: Since `req` is not longer an unused parameter:
```suggestion
const blocks: NextApiHandler<Response> = (req, res) => {
```

site/scripts/generate-blocks-data.ts
teenoh: Yes. That's correct

site/src/pages/api/blocks.api.ts
teenoh: Thanks will do this

site/src/lib/blocks.ts
teenoh: Added here e41bce7

site/src/pages/api/blocks.api.ts
teenoh: Added here 59570d6

site/src/pages/api/blocks.api.ts
teenoh: Thanks. Fixed here 59570d6

site/package.json
```diff
@@ -28,6 +28,7 @@
     "@mui/material": "^5.2.1",
     "@twind/next": "1.0.8",
     "axios": "^0.24.0",
+    "blockprotocol": "*",
```
teenoh, line 31: @kachkaev please let me know if I am doing this the right way

site/src/lib/blocks.ts
CiaranMn: Small follow up, sorry for not being clear - I think we should add `packagePath` in this file, e.g.

```typescript
type ExpandedBlockMetadata = BlockMetadata & { packagePath: string }
```

because `packagePath` is just something temporary we're putting in, and we generate it here, so we don't want to make it required for other people using the type as a guide for what to include. We should only include things in `blockprotocol` that are consistent with what the spec says.

CiaranMn: There are a couple of other type errors blocking merge I suggest can be fixed in the same way (adding extra fields in `site` rather than to `blockprotocol`, where they aren't something we want to mention in the spec)

teenoh: Ahh I see. Will update

teenoh: a651d1c6ef43762

CiaranMn (approved):
Great, thanks @teenoh !

PR #107, opened 2022-01-18
Update block metadata + build script for new source repo

CiaranMn:
We recently moved source code for our initial blocks from a private repository to [`hash`](https://github.com/hashintel/hash).

As a consequence this PR:
- updates the block config files to point to the new repo
- removes authentication handling from the block build script - auth is no longer required as the source is available publicly.

### DIFF
--- a/site/scripts/build-blocks.sh
+++ b/site/scripts/build-blocks.sh
@@ -112,7 +112,6 @@ function build_block {
 
     log info "downloading ${zip_url}"
 
-    zip_url="${zip_url:0:8}@${zip_url:8}"
 
     res=$(curl -sL -w '%{http_code}' -o "${repo_path}.zip" "$zip_url")
 
--- a/site/scripts/build-blocks.sh
+++ b/site/scripts/build-blocks.sh
@@ -53,18 +53,6 @@ else
     log error "missing commandline tool 'rsync'"
     exit 2
   fi
-
-  if [[ -f "${REPO_ROOT}/site/.env.local" ]]; then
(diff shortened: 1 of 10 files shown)


site/scripts/build-blocks.sh
```diff
@@ -124,9 +112,7 @@ function build_block {
 
     log info "downloading ${zip_url}"
 
-    # insert http basic auth credentials (https://<user:password>@domain.com)
-    # cannot use curl's -u options because that would also require a password
-    zip_url="${zip_url:0:8}${GH_ACCESS_TOKEN}@${zip_url:8}"
+    zip_url="${zip_url:0:8}@${zip_url:8}"
```
espretto, line 115:
```suggestion
```
This line can be removed completely. The `zip_url` is good as is.

espretto (changes requested):
Thank you. This PR is good for merge except for one line which can be removed. Please see down below.

site/scripts/build-blocks.sh
espretto: This would result in urls like `https://@github.com....`

kachkaev (approved):
I made a similar PR last week but forgot about it 😅
https://github.com/blockprotocol/blockprotocol/pull/101

LGTM with or without changes suggested by Marius. I hope we will rewrite the script in TypeScript at some point. Bash is OS-dependent looks more fragile.

espretto (approved):
Ready for merge.

@kachkaev Again, I don't think it's worth rewriting before some major change is required, like moving things over to an s3 instance.

PR #108, opened 2022-01-18
[SITE] User API key generation, listing, validation

CiaranMn:
## What this does
Introduces an `ApiKey` model to generate user API keys.

An API key is in the following format - `b10ck5.[publicId].[privateId]` - where:
- `b10ck5` is a constant prefix 
  - it allow users and repo-scanning tools to identify BP API keys (where it appears as the prefix of a string which otherwise matches the format)
- `publicId` is 16 random bytes encoded in hex to produce 24 characters:
  - it is used to identify API keys when presented by users
  - it is also used to identify API keys when displaying them to users, along with any display name
  - in future it can be used to identify specific API keys users want to revoke, rename, etc
- `privateId` is a uuid (36 characters including dashes)

This format [inspired by Dynatrace's](https://www.dynatrace.com/news/blog/further-enhance-security-by-easily-automating-your-api-token-protection/).

The full API key is only known at the point of creation, with only the `publicId` and `salt` persisted to the database, along with the hash of `[publicId].[privateId].[salt]`.

When a user presents an API key, the key is retrieved by `publicId` and the hashes compared.

For MVP, users can only have a single active key at a time. Generating a new key will revoke any keys associated with the user that have not already been revoked.

We could probably do without a salt for this but I don't think it does any harm, either.

I've added a `searchStub` API endpoint as an example endpoint which requires a valid API key - this will be replaced with the logic in #106 when we pull this together.

## How to test it
Log in to `http://localhost:3000` in the browser, and then either run the following in the console or get your cookie and do it via some other tool.

1. Generate an API key
```JavaScript
fetch("http://localhost:3000/api/me/generateApiKey", { 
    headers: { "content-type": "application/json" }, 
    method: "POST", 
    body: JSON.stringify({ displayName: "test" }) 
 }).then(resp => resp.json()).then(console.log)
```
make a note of the key responded to use in future tests.

2. List keys
```javascript
fetch("http://localhost:3000/api/me/apiKeys").then(resp => resp.json()).then(console.log)
```

3. Try using a key in the dummy endpoint
```javascript
fetch("http://localhost:3000/api/searchStub", { 
  headers: { "x-api-key": "b10ck5.xxxx" } 
}).then(resp => resp.json()).then(console.log)
```
Try invalidly formatted keys, or no key.

4. Generate a new key, then try using the one you generated previously - it should fail.


site/src/lib/middleware/hasValidApiKey.middleware.ts
```diff
@@ -0,0 +1,34 @@
...
+      );
+    }
+
+    const { db } = req;
+
+    await ApiKey.validate(db, { apiKeyString: apiKey });
+
+    next();
+  } catch (err) {
+    return res.status(401).send(
```
kachkaev, line 25: Different code styles (`throw new Error` and `res.status().send()) made me think that the responses would be shaped differently. They are the same, which is good for the end users:

Screenshot 2022-01-18 at 20 07 02

I wonder if we can align the implementation or at least state why we cannot do this. I originally thought that `throw new Error` would give us `5xx` – that’s why I went to check this.

site/src/lib/model/apiKey.model.ts
```diff
@@ -0,0 +1,239 @@
...
+  revokedAt?: Date | null;
+  salt: string;
+  useCount: number;
+  user: DBRef;
+
+  static readonly COLLECTION_NAME = "bp-api-keys";
+
+  private static readonly HASHING_SECRET = process.env.HASHING_SECRET;
+
+  private static readonly PREFIX = "b10ck5";
```
kachkaev, line 41: Does prefix need to look cryptic? It’d might be nice to add a small comment explaining what this prefix is for and why it’s not `"blocks"` for instance.

site/src/lib/model/apiKey.model.ts
```diff
@@ -0,0 +1,239 @@
...
+  lastUsedAt?: Date | null;
+  publicId: string;
+  revokedAt?: Date | null;
+  salt: string;
+  useCount: number;
+  user: DBRef;
+
+  static readonly COLLECTION_NAME = "bp-api-keys";
+
+  private static readonly HASHING_SECRET = process.env.HASHING_SECRET;
```
kachkaev, line 39:
```suggestion
  private static readonly HASHING_SECRET = mustGetEnvVar(process.env.HASHING_SECRET);
```

This will give us an early crash.

kachkaev: We also won’t need these checks below:

```
    if (!this.HASHING_SECRET) {
      throw new Error(
        "Cannot hash API key: no HASHING_SECRET present in environment.",
      );
    }
```

site/src/lib/model/apiKey.model.ts
```diff
@@ -0,0 +1,239 @@
...
+        "Cannot hash API key: no HASHING_SECRET present in environment.",
+      );
+    }
+
+    const stringToHash = `${publicId}.${privateId}.${salt}`;
+
+    return crypto
+      .createHmac("sha256", this.HASHING_SECRET)
+      .update(stringToHash)
+      .digest("hex");
```
kachkaev, line 90: 👍

site/src/lib/model/apiKey.model.ts
```diff
@@ -0,0 +1,239 @@
...
+        { publicId: this.publicId },
+        { $inc: { useCount: 1 }, $set: { lastUsedAt: new Date() } },
+        { returnDocument: "after" },
+      );
+
+    if (value) {
+      this.lastUsedAt = value.lastUsedAt;
+      this.useCount = value.useCount;
+    }
+  }
```
kachkaev, line 238: What do we need `useCount` for? Doing too many writes may become a bottleneck 🤔 If we only update `lastUsedAt`, we can do this with some threshold (like at most once per minute).

kachkaev (commented):
Great stuff! I left a few minor comments / questions.

site/src/lib/model/apiKey.model.ts
```diff
@@ -0,0 +1,239 @@
...
+      .collection<ApiKeyDocument>(this.COLLECTION_NAME)
+      .findOne({
+        publicId: params.publicId,
+      });
+
+    if (!apiKeyRecord) {
+      return undefined;
+    }
+
+    return new ApiKey(apiKeyRecord);
```
benwerner01, line 140: Just for consistency, I've been returning `null` instead when something isn't found in the other model classes:
```suggestion
    return apiKeyRecord ? new ApiKey(apiKeyRecord) : null;
```

site/src/lib/model/user.model.ts
```diff
@@ -361,6 +362,19 @@ export class User {
     return emailVerificationCode;
   }
 
+  async generateApiKey(db: Db, params: { displayName: string }) {
+    const { displayName } = params;
+
+    /* @todo allow users to have multiple API keys - remove this once implemented */
+    await ApiKey.revokeAll(db, { user: this });
+
+    return await ApiKey.create(db, { displayName, user: this });
```
benwerner01, line 371: This is likely best addressed on the frontend, but I don't think it's expected behaviour for a "Create API Key" button to immediately invalidate an existing API key. This might lead a user to accidentally invalidate an API key they are using in some infrastructure. Until we remove this limitation, it might be a good idea to have some warning dialog on the FE to ensure the user is aware of the consequences of generating another API key, or just not let them generate one if they already have one.

site/README.md
```diff
@@ -151,6 +162,20 @@ Retrieves the user object of the currently logged in user.
...
+
+Retrieves metadata on the API keys associated with the authenticated user.
+
+- Request Response:
+  - `apiKeys`: metadata on API keys (the key itself is only visible at the point of generation)
+
+### `POST /api/me/generateApiKey` [authenticated]
+
+Generates a new API key for the authenticated user, and revokes any others.
+
```
benwerner01, line 175: I think this is missing the definition of the `Request Body`, which requires a `displayName` in the request body

benwerner01 (approved):
Thanks Ciaran for getting this done, this looks good to me and is working for me locally 👍

Had two non-blocking suggestion, and a thought about the temporary behaviour of the `generateAPIKey` method (which is likely best addressed in the FE)

site/src/lib/middleware/hasValidApiKey.middleware.ts
CiaranMn: The error is immediately caught by the catch block - I threw it like this to make sure that this error is handled the same as anything thrown by `ApiKey.validate`.

I can instead repeat the `res.status(401).send`, although this makes it more likely they will diverge - is this clearer/better?

```typescript
  try {
    const apiKey = req.headers["x-api-key"];
    if (!apiKey || typeof apiKey !== "string") {
      return res
        .status(401)
        .send(
          formatErrors({
            msg: "A valid API key must be provided in a 'x-api-key' header.",
          }),
        );
    }
    ...
  } catch (err) {
    return res.status(401).send(
      formatErrors({
        msg:
          err instanceof Error
            ? err.message
            : "Unknown error validating API key.",
      }),
    );
  }
```

site/README.md
```diff
@@ -151,6 +162,20 @@ Retrieves the user object of the currently logged in user.
 - Request Response:
   - `user`: the user currently authenticated with the API
 
+### `GET /api/me/apiKeys` [authenticated]
```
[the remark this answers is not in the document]
CiaranMn: I agree but would rather we do this in follow-up since it'll involve renaming every other API route (or having this one oddly inconsistent with the others)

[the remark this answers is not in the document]
CiaranMn: https://app.asana.com/0/1201095311341924/1201685191027589/f

site/src/lib/model/apiKey.model.ts
CiaranMn: Added a comment

site/README.md
CiaranMn: Thanks, added

site/src/lib/middleware/hasValidApiKey.middleware.ts
CiaranMn: I've gone ahead and done this - let me know if it's better

site/src/lib/model/apiKey.model.ts
CiaranMn: Some indication of activity would be helpful - until we have better event tracking generally, and levels of activity that cause concern, I think this is fine.

site/src/lib/model/user.model.ts
CiaranMn: Agreed - the button will be called 'Regenerate API key' and a warning displayed when the user does so.

CiaranMn (commented):
Feedback applied unless otherwise noted - thanks both!

site/src/lib/model/apiKey.model.ts
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  static async create(
+    db: Db,
+    params: { displayName: string; user: User },
+  ): Promise<string> {
+    const { displayName, user } = params;
+
+    const privateId = uuid();
+    const publicId = crypto.randomBytes(16).toString("hex");
```
thehabbos007, line 99: Non blocking - continuation of above comment, the `16` byte count here could be the same const divided by 2. Also perhaps making it clear that 1 byte = 2 hex digits so no one will change it by accident

site/src/lib/model/apiKey.model.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  /**
+   * This prefix helps to identify keys for both humans and automated code-scanners.
+   * It aims to balance human-readability with avoiding false positives (which plain "blocks" might do).
+   */
+  private static readonly PREFIX = "b10ck5";
+
+  private static readonly KEY_FORMAT_REGEXP = new RegExp(
+    // e.g. b10ck5.e84a616cedc8822b02ac96761b068f19.91ea0c90-acee-4215-a01e-85a4438af0ea
+    `^${ApiKey.PREFIX}\\.[a-z0-9]{32}\\.[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}$`,
```
thehabbos007, line 50: Non-blocking - could it be helpful to extract the publicId length `{32}` into a const that the generated bytes can use below?

thehabbos007 (approved):
Looks good to me. Minor thing about `publicId` part of the key, but not blocking a merge.

site/src/lib/model/apiKey.model.ts
CiaranMn: Agreed, thank you - I've done this [here](https://github.com/blockprotocol/blockprotocol/pull/108/commits/1c004c374c9935397a3969731a46a4bd9c9109e7)

site/README.md
```diff
@@ -13,13 +13,24 @@ The public-facing [blockprotocol.org](https://blockprotocol.org) website serves
...
-    - `SESSION_SECRET`: he secret used to sign the session ID cookie
+    - `HASHING_SECRET`: the secret used to hash API keys
+    - `SESSION_SECRET`: the secret used to sign the session ID cookie
     - `MONGODB_URI`: the URL where the mongo developer db instance is hosted (for example at `mongodb://root:password@localhost:27017/`)
     - `MONGODB_DB_NAME`: the name of the database (for example `local`)
     - `MONGODB_USERNAME`: the database username
     - `MONGODB_PASSWORD`: the database password
     - `FRONTEND_DOMAIN` (optional): the domain where the frontend is hosted (defaults to `localhost:3000`)
 
+Example minimal file at `site/.env.local` (with **zero** security) to make local development work when following the instructions below:
```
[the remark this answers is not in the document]
kachkaev: The new offset fixes the list, thanks!

I still think we might be slightly better off by having the vars in `site/.env.development` which is tracked by git. One less step to do to start the app locally. But this is optional.

site/src/lib/middleware/hasValidApiKey.middleware.ts
kachkaev: Yeah that looks a bit more coherent thanks! 👍

kachkaev (approved):
Thanks Ciaran! Just left one optional suggestion about `site/.env.development` (see unresolved comment).

site/README.md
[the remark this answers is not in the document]
CiaranMn: Captured for followup - https://app.asana.com/0/1201095311341924/1201685786367415/f

PR #109, opened 2022-01-19
[SITE] Dashboard to create API Keys

akash-joshi:
## What this Does

The Block Protocol website presently allows users to browse through all available docs, and go through documentation to create their own blocks. Further PRs are going to add functionality like submitting blocks through the website. To make this functionality programmatic, a supporting API is going to be created.

This PR creates a dashboard to manage the API keys, create new ones and regenerate existing ones. The user comes to the [dashboard page](https://github.com/blockprotocol/blockprotocol/pull/109/files#diff-cc355ff8d00f008095e316c85f3ff5a860efe27d2429bc1ae405d81fa15ded05R1), where they can read the guide, try out blocks, or generate keys.

The [DashboardCard](https://github.com/blockprotocol/blockprotocol/pull/109/files#diff-736547c724d34abc7b666e11cb974cefade8586609830fc3aa8cef519812ea2fR1) component is responsible for rendering this data.

The [page for API keys](https://github.com/blockprotocol/blockprotocol/pull/109/files#diff-a28f6424bb04944ec0b75cbf7a9175098d0ae271ea9762aadc299610ba6c4655R1) is responsible for generating and refreshing API keys, with helper components being created for [Modal](https://github.com/blockprotocol/blockprotocol/pull/109/files#diff-a42a6d82a58ce282f5ae980c4cb729880437e6f86b30db1bbab775fed0e94746R1) and [Table](https://github.com/blockprotocol/blockprotocol/pull/109/files#diff-c37885c632d5f871b9c5d7630cae8c3e1ff2745668df75e498fabd927aa5f999R1), which are used in the rendering of the API generation modals, and the actual table on the page.

## Future Work

- Warn users that API keys are only visible once (I'll pick this up as a part of this PR).
- Integrate the components with actual data.
- Update some of the component styles to match the design better.

## How to test it

1. Navigate to `/dashboard`.
2. Follow the demo below and try to create and regenerate API keys.

## Demo

chrome_GAvTXsZoLP

> Note: I've taken some liberty with the API key generation design. Instead of having 2 different closable modals, I've tried to integrate into a single modal, since we would be displaying the API key in a modal for all generated keys.

---
  - https://app.asana.com/0/0/1201668052739249


CiaranMn:
I've hooked this up to the real API - see demo below.

@akash-joshi can you do the following tomorrow as a priority:
1. Sort the mobile styling out
2. Make the whole of dashboard cards clickable (not just the link)

Note that I've removed some hardcoding styles that were applied when we should be using values from the MUI theme. We shouldn't be specifying any font sizes, button styles, colours, etc, anywhere except the MUI theme.

We'll need to revisit this to clear up the other hardcoded values, and sort out some of the button styles in the theme which don't match the design system, but for now please just focus on (1) and (2) above.

https://user-images.githubusercontent.com/37743469/150208868-b423a922-405c-4f86-b02c-bcee4855c438.mp4

akash-joshi:
@CiaranMn commit https://github.com/blockprotocol/blockprotocol/pull/109/commits/659648f3a4058169533f7f34f5a7ba4d910a91b1 implement 1 & 2 from your previous comment.

1. Sort the mobile styling out

I believe the buttons were missing some margin on mobile, which I fixed.

NptjzmiYZD


2. Make the whole of dashboard cards clickable (not just the link)

Have moved the links from the child element to the parent.

CiaranMn (changes requested):
Thanks Akash - there are a few more quick things we can do to make mobile better and more closely aligned with the designs:

1. The "Dashboard" / "API Keys" tabs at the top disappears on mobile - it is present in the designs on mobile. This component also repeats between the dashboard and api-keys page - I'd extract this and re-use it.

2. Various boxes could do with less padding on smaller sizes, e.g. this one (please check others for padding in the designs vs here)
Screenshot 2022-01-20 at 07 47 53

3. We should give the table cells sufficient width to avoid wrapping here
Screenshot 2022-01-20 at 07 49 32

4. Buttons should generally be full width on mobile
Screenshot 2022-01-20 at 07 48 17

akash-joshi:
Have made the above changes.

chrome_jIlEscrPEE

The header won't wrap either now - [link](https://github.com/blockprotocol/blockprotocol/pull/109/commits/e00229f8ef10dc40c320a92ba5e810d7069e153e#diff-c37885c632d5f871b9c5d7630cae8c3e1ff2745668df75e498fabd927aa5f999R28)

CiaranMn (changes requested):
Please extract the tabs into its own component - it's going to be much easier to add to it later rather than updating in two places.

site/src/components/pages/dashboard/TopNavigationTabs.tsx
```diff
@@ -0,0 +1,54 @@
...
+        marginTop: {
+          xs: 2,
+          md: 0,
+        },
+      }}
+    >
+      <Container>
+        <Tabs
+          value={
+            dashboardPages.find((dashboardPage) =>
```
akash-joshi, line 30: This is a heuristic way to determine the current page, so it doesn't get affected by hashes or query parameters. This component in itself could eventually be created into a generic reusable component.

CiaranMn (approved):
Thanks Akash, let's get this in

PR #112, opened 2022-01-20
Make fn signatures consistent between spec and blockprotocol

CiaranMn:
Cleans up the functions described in the spec and typed in the `blockprotocol` package to ensure completeness and consistency: 
- Adds functions described in the spec which were missing from the `blockprotocol` types
- Adds some fields that were missing or inconsistently applied across both, e.g. `accountId`
- Makes optional fields consistently `| null` (since different apps may omit or provide `null`)

**Next steps:**
- Release v0.0.2
- See what breaks in HASH and fix it (this PR should actually apply some of the patches we have there rather than break anything)

### DIFF
--- a/packages/blockprotocol/core.d.ts
+++ b/packages/blockprotocol/core.d.ts
@@ -1,3 +1,4 @@
+// -------------------------- BLOCK METADATA -------------------------- //
 export type BlockVariant = {
   description?: string;
   displayName?: string;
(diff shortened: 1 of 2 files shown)


kachkaev (approved):
👍

Optional suggestion: replace `type` with `interface` given the opportunity.

From [TS docs](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces):

For the most part, you can choose based on personal preference, and TypeScript will tell you if it needs something to be the other kind of declaration. If you would like a heuristic, use interface until you need to use features from type.

CiaranMn:
> Optional suggestion: replace type with interface given the opportunity.

Let's discuss this separately - I'm not convinced of the benefit so I think it requires more thought that we can take outside of this PR.

The key practical difference to my mind is that someone can intentionally or accidentally declare an interface again to add stuff to it ("declaration merging"). Beyond that, everything else is achievable, albeit with slightly different syntax.

PR #114, opened 2022-01-21
[FE] fix account button glitch

espretto:
This PR replaces MUI's scroll-lock mechanism which adds a padding to the body to replace the scrollbar which however breaks the layout our case. The solution is to apply different yet equally effective styles to the html element.

How to reproduce: Fire up the site, log in, then press the account button in the top-right corner. The popover should pop over and the menu should no longer shift to the left due to prior body-padding applied by MUI.


espretto:
replaced by #115

PR #115, opened 2022-01-21
[SITE] Replace mui's troublesome scroll-lock implementation

espretto:
This PR replaces MUI's scroll-lock mechanism which adds a padding to the body to replace the scrollbar which however breaks the layout our case. The solution is to apply different yet equally effective styles to the html element.

How to reproduce: Fire up the site, log in, then press the account button in the top-right corner. The popover should pop over and the menu should no longer shift to the left due to prior body-padding applied by MUI.

### DIFF
--- a/site/src/components/Popover.tsx
+++ b/site/src/components/Popover.tsx
@@ -1,0 +1,15 @@
+/* eslint-disable-next-line -- allow import of original popover to extend it */
+import { Popover as MuiPopover, PopoverProps } from "@mui/material";
+import React from "react";
+import { useScrollLock } from "../util/muiUtils";
+
+/**
+ * Custom Popover re-implementing MUI's troublesome scroll-lock mechanism.
+ */
+export const Popover: React.FC<PopoverProps> = ({
+  disableScrollLock = false,
+  ...props
+}) => {
+  useScrollLock(!disableScrollLock && props.open);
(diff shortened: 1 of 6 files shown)


site/src/components/Popover.tsx
```diff
@@ -0,0 +1,24 @@
...
+export const Popover: React.FC<PopoverProps> = ({
+  disableScrollLock = false,
+  ...props
+}) => {
+  const applyScrollLock = !disableScrollLock && props.open;
+
+  useLayoutEffect(() => {
+    document.documentElement.style.cssText = applyScrollLock
+      ? "position: fixed; overflow-y: scroll; width: 100%"
+      : "";
```
CiaranMn, line 20: I think this effect needs cleaning up - at the moment if you click the account window, and then click 'log out', the fixed position remains.

```suggestion
      : "";
      
      return () => {
        document.documentElement.style.cssText = "";
       };
```

CiaranMn (changes requested):
Thanks Marius - spotted a bit of cleanup required.

Could you also apply this magic to `LoginModal`? Maybe that needs a wrapper around `Modal` also? Could be done in follow up.

CiaranMn (approved):
Works great, thanks Marius!

PR #119, opened 2022-01-24
[API] Filter blocks by JSON schema validity

thehabbos007:
This PR adds a query parameter for filtering blocks by a JSON payload on the query parameter on the `/api/blocks` endpoint.

The added query parameter is `json` and can be used as follows

`https://blockprotocol.org/api/blocks?json={"name":"Bob","email":"bob@test.com"}`
or URL encoded:
`https://blockprotocol.org/api/blocks?json=%7B%20%22name%22%3A%20%22Bob%22,%20%22email%22%3A%20%22bob%40test.com%22%20%7D`

The implementation requires any JSON object to at least supply one value from any block it returns, otherwise blocks that have only optional fields/no fields would always validate and be returned (such as the Divider block.)

### DIFF
--- a/site/src/_pages/docs/3_embedding-blocks.mdx
+++ b/site/src/_pages/docs/3_embedding-blocks.mdx
@@ -16,15 +16,21 @@ You can also [generate an API key](https://blockprotocol.org/settings/api-keys) 
 
 To authenticate with the API, pass your API key in an `x-api-key` header.
 
-### Coming soon
+It is also possible to search blocks by data structure: allowing you to provide a data object, and be told which blocks can display or edit it.
 
-We will soon introduce a feature to search blocks by data structure: allowing you to provide a data object, and be told which blocks can display or edit it.
+For example, if you provide `{ "name": "Bob", "email": "bob@test.com" }` in the `json` query parameter, the API will suggest the
+[person block](https://blockprotocol.org/@hash/person). It's preferable to URL encode the json object.
 
-For example, if you provide `{ "name": "Bob", email: "bob@test.com" }`, the API will suggest the
-[person block](https://blockprotocol.org/@hash/person).
+`https://blockprotocol.org/api/blocks?json={"name":"Bob","email":"bob@test.com"}`
+
+URL encoded:
+
+`https://blockprotocol.org/api/blocks?json=%7B%20%22name%22%3A%20%22Bob%22,%20%22email%22%3A%20%22bob%40test.com%22%20%7D`
 
 Once your app implements the block protocol, your users can search and use blocks to work with a
 wide variety of data structures, without further effort on your part.
+
+Both the `q` and `json` query parameters work together.
 
 ## Building the block's properties
 
--- a/site/src/_pages/docs/3_embedding-blocks.mdx
+++ b/site/src/_pages/docs/3_embedding-blocks.mdx
@@ -30,7 +30,7 @@ URL encoded:
 Once your app implements the block protocol, your users can search and use blocks to work with a
 wide variety of data structures, without further effort on your part.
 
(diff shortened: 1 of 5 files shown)


site/src/_pages/docs/3_embedding-blocks.mdx
```diff
@@ -16,16 +16,22 @@ You can also [generate an API key](https://blockprotocol.org/settings/api-keys)
...
+`https://blockprotocol.org/api/blocks?json={"name":"Bob","email":"bob@test.com"}`
+
+URL encoded:
+
+`https://blockprotocol.org/api/blocks?json=%7B%20%22name%22%3A%20%22Bob%22,%20%22email%22%3A%20%22bob%40test.com%22%20%7D`
 
 Once your app implements the block protocol, your users can search and use blocks to work with a
 wide variety of data structures, without further effort on your part.
 
+Both the `q` and `json` query parameters work together.
```
CiaranMn, line 33:
```suggestion
Both the `q` and `json` query parameters work together. You may provide either, both, or none.
```

site/src/pages/api/blocks.api.ts
```diff
@@ -1,10 +1,16 @@
...
-import { ExpandedBlockMetadata as BlockMetadata } from "../../lib/blocks";
+import {
+  ExpandedBlockMetadata as BlockMetadata,
+  readBlockDataFromDisk,
+} from "../../lib/blocks";
 import { createApiKeyRequiredHandler } from "../../lib/handler/apiKeyRequiredHandler";
 
 export type ApiSearchRequestQuery = {
   q: string;
+  json: string;
```
CiaranMn, line 13: These should be optional, so that users can provide either (or neither, for a full list - later, paginated)

```suggestion
  q?: string;
  json?: string;
```

site/src/pages/api/blocks.api.ts
```diff
@@ -13,8 +19,11 @@ export type ApiSearchResponse = {
 
 export default createApiKeyRequiredHandler<null, ApiSearchResponse>()
   .use(queryValidator("q").isString().toLowerCase())
+  .use(queryValidator("json").isJSON())
   .get(async (req, res) => {
-    const { q: query } = req.query as ApiSearchRequestQuery;
+    const { q: query, json: jsonText } = req.query as ApiSearchRequestQuery;
+
+    const json = JSON.parse(jsonText);
```
CiaranMn, line 26: This will crash if `json` is undefined.

site/src/pages/api/blocks.api.ts
```diff
@@ -30,6 +39,40 @@ export default createApiKeyRequiredHandler<null, ApiSearchResponse>()
...
+          const schema = readBlockDataFromDisk(block).schema;
+          const validate = ajv.compile(schema);
+
+          // withoutAdditional is transformed in validate.
+          // eslint-disable-next-line prefer-const
+          let withoutAdditional = cloneDeep(json);
+          const valid = validate(withoutAdditional);
+
+          // removeAdditional lets us count how many keys are a part of the schema
+          const keyCountWithouTAdditional =
```
CiaranMn, line 59: typo `keyCountWithouTAdditional` -> `keyCountWithoutAdditional`

CiaranMn (changes requested):
Thanks Ahamd - this works well when providing a `json` parameter - just a couple of tweaks required to handle `json` being omitted.

site/src/pages/api/blocks.api.ts
thehabbos007: Thanks for pointing out!

CiaranMn (approved):
Looks good to me!

PR #120, opened 2022-01-24
blockprotocol 0.0.4: refactor fn types, add linkedAggregations

CiaranMn:
Make a few changes to the `blockprotocol` package types, and bump version to `0.0.4` - will publish immediately after merging.

Changes
1. Remove `[function]Loading`/`[function]Error` properties - these aren't mentioned in the spec and not clear they are desirable.
2. add `linkedAggregations` field typing
3. Rename `sourceEntityAccountId` back to `sourceAccountId` - I thought the additional `Entity` was clearer when adding this in #112, but it's inconsistent with the others, which are named `source[NameOfFieldOnEntity]`, e.g. `entityId` becomes `sourceEntityId`. by this logic, `accountId` should become `sourceAccountId`.
4. Better organise functions
5. add `links` argument to `createEntities`
5. `DistributedOmit` to handle `Omit` on unions properly

**Next steps**
1. Publish
2. Upgrade in `hash` and fix breakages


packages/blockprotocol/core.d.ts
```diff
@@ -273,23 +309,45 @@ export type BlockProtocolDeleteEntityTypesFunction = {
...
+  | BlockProtocolAggregateEntityTypesFunction
+  | BlockProtocolCreateEntityTypesFunction
+  | BlockProtocolGetEntityTypesFunction
+  | BlockProtocolDeleteEntityTypesFunction
   | BlockProtocolUpdateEntityTypesFunction
+  | BlockProtocolCreateLinksFunction
+  | BlockProtocolGetLinksFunction
   | BlockProtocolUpdateLinksFunction
+  | BlockProtocolDeleteLinksFunction
   | BlockProtocolUploadFileFunction;
```
kachkaev, line 328: The original list was ordered alphabetically. I prefer sticking with this where I can to simplify visual search. I understand why you’ve regrouped the types and am happy both ways.

yarn.lock
```diff
@@ -3110,6 +3110,13 @@ bl@^1.0.0:
     readable-stream "^2.3.5"
     safe-buffer "^5.1.1"
 
+blockprotocol@0.0.1, blockprotocol@^0.0.1:
+  version "0.0.1"
+  resolved "https://registry.yarnpkg.com/blockprotocol/-/blockprotocol-0.0.1.tgz#2723d6b80f50c3da7cf4984b4e5824fc26570f76"
+  integrity sha512-neITCxiFVxM5S0C0KTPucMCB2Xip5xIYo3LZFXUY/vamUKM5Ov2viv39+og1X6Vfd/IZwuwONK3gjO6dtGDXDA==
+  dependencies:
+    "@types/react" "^17.0.38"
+
```
kachkaev, line 3119: I guess we need to update dependency in `site/package.json`. We should not depend on an external BP version.

packages/blockprotocol/core.d.ts
CiaranMn: I've ordered these alphabetically - I've retained the 'by target' grouping in the type below it.

yarn.lock
CiaranMn: Done

packages/block-template/package.json
```diff
@@ -29,7 +29,7 @@
     "schema": "typescript-json-schema tsconfig.json AppProps --required true --out dist/block-schema.json"
   },
   "dependencies": {
-    "blockprotocol": "0.0.1"
+    "blockprotocol": "*"
```
kachkaev, line 32: Not sure we want to make this requirement so loose. A specific version of a template conceptually depends on a specific version of `blockprotocol`. We can still use `*` in `site/package.json`, but I’d go for 0.0.4 everywhere for ‘symmetry’.

Tools like Lerna automate version bumps like this, so this manual edits are temporary. 

```suggestion
    "blockprotocol": "0.0.4"
```

site/package.json
```diff
@@ -27,7 +27,7 @@
     "@twind/next": "1.0.8",
     "ajv": "^8.9.0",
     "axios": "^0.24.0",
-    "blockprotocol": "^0.0.1",
+    "blockprotocol": "*",
```
kachkaev, line 30:
```suggestion
    "blockprotocol": "0.0.4",
```

packages/blockprotocol/core.d.ts
```diff
@@ -37,6 +45,13 @@ export type BlockProtocolCreateEntitiesAction<T> = {
   entityTypeVersionId?: string | null;
   data: T;
   accountId?: string | null;
+  links?: DistributedOmit<
+    BlockProtocolCreateLinksAction,
+    | "sourceAccountId"
```
akash-joshi, line 50: Does this need renaming to `sourceEntityAccountId`?

akash-joshi: Same for all the other instances of the string.

packages/blockprotocol/core.d.ts
```diff
@@ -130,48 +148,63 @@ export type BlockProtocolAggregateEntitiesFunction = {
...
 export type BlockProtocolFileMediaType = "image" | "video";
 
 export type BlockProtocolUploadFileFunction = {
   (action: {
+    accountId?: string;
     file?: File | null;
     url?: string | null;
     mediaType: BlockProtocolFileMediaType;
   }): Promise<{
+    accountId?: string;
```
akash-joshi, line 161: I can't just reply with an emote to a line, but this is helpful ✅

packages/blockprotocol/core.d.ts
CiaranMn: Nope - see the PR description

packages/block-template/package.json
CiaranMn: This change (to `"*"`) was in response to your previous comment:
> I guess we need to update dependency in site/package.json & packages/block-template/package.json. We should not depend on an external BP version.

This returns it to depending on an external version, which does not exist. Is the intention that we do in fact go back to depending on an external version?

CiaranMn: Or that we should always have this be the same value as the latest published, but pin it so that when people use it elsewhere they get that version?

CiaranMn: Actually, maybe it wasn't picking up 0.0.4 from the workspace before because of #124 - will commit this and see what happens

kachkaev (approved):

PR #121, opened 2022-01-24
Stop using `npm pack` in `create-block-pack` to support npm v6

kachkaev:
@CiaranMn could not run `npx create-block-app` on his laptop because of the following error:

```txt
npm ERR! Could not install from "../../../var/folders/lf/yycqw_q52wzc77xg1b9ghf740000gn/T/tmp-55410-lZLZsIF4dliY" as it does not contain a package.json file.
```

It was caused by `npm pack`, which did not support `--destination` in v6. This PR replaces `npm pack` with [`pacote`](https://www.npmjs.com/package/pacote), which allows us to download and extract a package in a single function call.

Because of `pacote`, we have a few more sub-dependencies than before. This is not ideal but tolerable for the time being. When the number of downloads increases, we can replace `pacote` with lower-level calls to NPM registry. They require more edge case handling and testing, so are probably not worth it now.

To test the change locally before I publish, run:

```sh
cd /path/to/blockprotocol-monorepo

npm i -g npm@6
./node_modules/.bin/create-block-app test1 /tmp/blocks/test1

npm i -g npm
./node_modules/.bin/create-block-app test2 /tmp/blocks/test2
```

### DIFF
--- a/packages/create-block-app/package.json
+++ b/packages/create-block-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-block-app",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "description": "Create new https://blockprotocol.org blocks",
   "keywords": [
     "blockprotocol",
--- a/packages/create-block-app/package.json
+++ b/packages/create-block-app/package.json
@@ -24,7 +24,9 @@
     "lint:tsc": "echo \"Skipping: TypeScript not configured\""
   },
   "dependencies": {
-    "decompress": "4.2.1",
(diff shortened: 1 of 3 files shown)


packages/create-block-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "create-block-app",
-  "version": "0.0.2",
+  "version": "0.0.3",
```
kachkaev, line 3: To be publish after PR merge

CiaranMn (approved):
Tested and working, thanks Alex!

PR #126, opened 2022-01-25
Hide `/settings/api-keys` for guests, unify auth wall

kachkaev:
This reuses guest redirects in `/dashboard` in `/settings/api-keys`. To avoid repetition, I’ve introduced `withAuthWall`, which we can use onwards in the new pages.

To further reduce repetition, I have replaced `useContext(UserContext)` with `useUser()`, which is a common pattern in React hooks.

When looking at the original logic in `/dashboard`, I noticed a timing bug which was to do with the speed of `/auth/me`. If we rendered the dashboard before `fetch` was resolved, we ended up with an unwanted redirect for logged in users. This is fixed by adding `UserState = SerializedUser | "loading" | undefined`, which allows us to represent reality more accurately.

[Asana task](https://app.asana.com/0/1201095311341924/1201694916711161/f) _(internal link)_

### DIFF
--- a/site/src/components/Navbar.tsx
+++ b/site/src/components/Navbar.tsx
@@ -19,7 +19,7 @@ import { BlockProtocolLogoIcon } from "./SvgIcon/BlockProtocolLogoIcon";
 import { BoltIcon } from "./SvgIcon/BoltIcon";
 import { HOME_PAGE_HEADER_HEIGHT } from "../pages/index.page";
 import SiteMapContext from "../context/SiteMapContext";
-import UserContext from "../context/UserContext";
+import { useUser } from "../context/UserContext";
 import { AccountDropdown } from "./Navbar/AccountDropdown";
 import { MobileNavItems } from "./Navbar/MobileNavItems";
 import { itemIsPage, NAVBAR_LINK_ICONS } from "./Navbar/util";
@@ -91,7 +91,7 @@ export const Navbar: VFC<NavbarProps> = ({
   const theme = useTheme();
   const router = useRouter();
   const { pages } = useContext(SiteMapContext);
-  const { user } = useContext(UserContext);
+  const { user } = useUser();
 
   const [displayMobileNav, setDisplayMobileNav] = useState<boolean>(false);
   const [idleScrollPosition, setIdleScrollPosition] = useState<boolean>(false);
@@ -343,7 +343,7 @@ export const Navbar: VFC<NavbarProps> = ({
                       </Typography>
                     </Button>
                   )}
-                  {user?.isSignedUp ? null : (
+                  {user !== "loading" && !user?.isSignedUp ? (
                     <Link href="/docs/developing-blocks">
                       <Button
                         size="small"
@@ -355,7 +355,7 @@ export const Navbar: VFC<NavbarProps> = ({
                           : "Quick Start Guide"}
                       </Button>
                     </Link>
-                  )}
+                  ) : null}
                 </>
               ) : (
                 <IconButton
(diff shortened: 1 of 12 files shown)


site/src/components/Navbar.tsx
```diff
@@ -369,7 +369,7 @@ export const Navbar: VFC<NavbarProps> = ({
                   <Icon className="fas fa-bars" />
                 </IconButton>
               )}
-              {user?.isSignedUp ? <AccountDropdown /> : null}
+              <AccountDropdown />
```
kachkaev, line 372: Account dropdown handles calls `useUser` internally, so returns `null` by iteslf.

site/src/pages/_app.page.tsx
```diff
@@ -70,13 +73,18 @@ const MyApp = ({
...
-    if (user && !user.isSignedUp && route !== "/signup") {
+    if (user && user !== "loading" && !user.isSignedUp && route !== "/signup") {
       void router.push("/signup");
     }
   }, [user, router]);
 
+  const userContextValue = useMemo<UserContextValue>(
+    () => ({ user, setUser, refetch: refetchUser }),
+    [refetchUser, user],
+  );
```
kachkaev, line 84: Potentially prevents rerenders

site/src/components/pages/authWall.tsx
```diff
@@ -0,0 +1,42 @@
...
+};
+
+/**
+ * We don’t know if a user is logged in during server-side-rendering, but we want
+ * some pages to be available behind the auth wall. Once UserContext is populated
+ * with the result of /api/me, we want to redirect guest users from to /.
+ * use `withAuthWall` to enable this behavior.
+ *
+ * You can also get user object from Content props and thus avoid extra checks.
+ */
```
CiaranMn, line 39: We could find out if the user is logged in during SSR, since we have the cookie on the request.

We can do this in follow up since it is only an issue if the user attempts to visit a protected page directly. I've created [a task (internal)](https://app.asana.com/0/1201095311341924/1201715945634534/f).

CiaranMn (commented):
Personally I find the fact that `user` can be either a user or the string `"loading"` less intuitive than the Apollo Client or React Query approach, whereby there's an additional `loading` or `isLoading` prop returns from the hook.

Did you consider this approach?

I appreciate your way means that consumers are forced to deal with the fact that there's a loading state (whereas they can just ignore the `isLoading` property on the hook return). It just feels a bit less familiar.

kachkaev:
I was going to make a comment on why I mixed `"loading"` with `SerializedUser`, but forgot to do it in the end 🤦

I agree that having something like `{ userIsLoading, user } = useUser()` is more intuitive, but I’m not sure if this is worth it at the moment. The shape of `UserContextValue` will get a bit more complex, same for `useState` in `_app`. I thought our user detection logic was temporary anyway, so decided to not over-engineer the solution. Switching from this `UserState` to `userIsLoading` won’t be too hard if needed. If you feel too uncomfortable about mixing `object | string | undefined` in the meantime, I can implement the change as part of this PR.

site/src/components/pages/authWall.tsx
kachkaev: Yeah that might be useful for some pages. We’ll probably want to keep public content statically rendered to avoid SSR everywhere. This means that user loading state will be still there, we’ll just replace `Router.push()` with a redirect inside some `getServerSideProps`. I believe we can live with the current approach for quite long TBH.

CiaranMn (approved):
Thanks Alex - agree it's fine to leave with "loading" on `user` for now. Helpful to get this discussion logged for future travellers!

PR #128, opened 2022-01-25
[SITE] API routes for entity types / schemas, and listing user blocks

CiaranMn:
This introduces API routes for working with entity types / schemas. These are documented in the README - a few things to try below. These assume you log in as `alice@example.com`:

Before doing anything, sign up a user with the shortname `hash` and the name `HASH` - this will create the dummy user to allow searching hash for blocks.

- `POST /api/types/create`, with body `{ schema: { "title": "MySchema" } }`
- `PUT /api/types/[entityTypeId]/update` where `entityTypeId` is that of the one you just created, with some new `schema` in the body
- `GET /api/users/alice/types/MySchema` - get a type by username and title (so we can have friendly URLs that use username/title - these will break if the schema name or username changes, but for references in schemas to other schemas we'll use a different URI to avoid this - see point below)
- `GET /api/types/[entityTypeId]` - get an entity type by its id.

It also adds the ability to get info on users by shortname, and list blocks in a namespace, to support user profile pages:
- `GET /api/users/hash/blocks`: there isn't a 'hash' user but blocks are tied to namespaces manually at the moment, so this works. 
- `GET /api/users/alice`: this will work, but `/api/users/hash` won't, since there isn't a 'hash' user.
- `GET /api/users/alice/types`: try after creating a type
- `GET /api/users/alice/types/MySchema`: try after creating a type

Other points to note:
- current implementation uses a URI/`$id` for schemas of `https://blockprotocol.org/types/[uuid]`, as this won't change and means that references to these schemas from other schemas will be valid forever. Alternatively we could use a combination of `[namespace]` and `[title]` in the URI, and accept they might break if either the user or schema title changes.
- I've added an `author` property to schemas so that when viewing the JSON you can tell who it belongs to 
- schemas are unique by namespace (username) / title

**Follow ups**
- Frontend for creating / updating schemas
- User profile page to list user blocks / schemas

---
  - https://app.asana.com/0/0/1201706915162193


thehabbos007:
Should `POST /api/schemas/create` be `POST /api/types/create` for testing these changes?

CiaranMn:
> Should `POST /api/schemas/create` be `POST /api/types/create` for testing these changes?

@thehabbos007 oops, yes, thank you

thehabbos007 (approved):
Looks good to me, tested locally and looked through endpoints.
Good emails are hidden unless it's the `/me` endpoint and on signup/login.

Linting fails, but that should be easy fixes

teenoh (commented):
Thanks for this @CiaranMn. Tested locally and it works! 

Although I noticed fetching the types for user Alice returns an empty array `/api/users/alice/types` even after running seed-db. I assume it's mean to return a type since we create that in `seed-db.ts`

CiaranMn:
> Although I noticed fetching the types for user Alice (`/api/users/alice/types`) returns an empty array even after running seed-db. I assume it's mean to return a type since we create that in `seed-db.ts`

Thanks @teenoh - we don't actually create any types in `seed-db.ts`, we just create the collection to hold types, and the user (but no types belonging to the user).

teenoh:
> 
> Thanks @teenoh - we don't actually create any types in `seed-db.ts`, we just create the collection to hold types, and the user (but no types belonging to the user).

Ahh I see. Thanks for the clarification

teenoh (dismissed):

site/src/pages/api/users/[shortname]/blocks/index.api.ts
```diff
@@ -0,0 +1,35 @@
+import { createBaseHandler } from "../../../../../lib/handler/baseHandler";
+import { User } from "../../../../../lib/model/user.model";
+import { formatErrors } from "../../../../../util/api";
+import { ExpandedBlockMetadata } from "../../../../../lib/blocks";
+
+export type ApiBlocksByUserResponseQueryParams = {
+  shortname: string;
+};
+
+export type ApiBlocksByUserResponseResponse = {
```
Comment (teenoh) [inline site/src/pages/api/users/[shortname]/blocks/index.api.ts:10]: Sorry I'm dropping this after approving. While integrating, I noticed in some places we use the format `ApiXYZResponse`, while in other places we use `ApiXYZResponseResponse`. Not sure why the latter has two `Response`, but I think we should stick to 1 pattern

Comment (CiaranMn) [reply site/src/pages/api/users/[shortname]/blocks/index.api.ts:10]: Thanks @teenoh - this is just a typo, I'll remove it

Comment (CiaranMn) [reply site/src/pages/api/users/[shortname]/blocks/index.api.ts:10]: Have renamed this in the next PR (on the frontend UI)

PR #133, opened 2022-01-26
[SITE] Fix multiple auth code checks when using magic links

kachkaev:
We had a new reference of `handleLogin` on each render, which caused multiple calls to `handleSubmit` in `VerificationCodeScreen` → `useEffect`. This PR makes the reference stable, thus reducing the number of API calls to one. As a result, users no longer see _Verification code has already been used_ when clicking on magic login links.

[Asana task](https://app.asana.com/0/1201095311341924/1201698506081400/f) _(internal link)_

## Before

https://user-images.githubusercontent.com/608862/151210215-1f27d797-7686-48ef-8784-0b9fa943c995.mov

Screenshot 2022-01-26 at 17 00 33

## After

https://user-images.githubusercontent.com/608862/151210241-cbb3860e-3ca0-45f0-85be-2713d2a83aa9.mov

Screenshot 2022-01-26 at 17 01 06


CiaranMn (approved):
Thanks Alex - code looks good, tested and works

PR #135, opened 2022-01-26
[SITE] Schema editor frontend

CiaranMn:
This PR introduces a basic schema editor to the site. It's mostly lifted from the one in [HASH](https://github.com/hashintel/hash/tree/main/packages/hash/frontend/src/components/entityTypes/SchemaEditor) with a few tweaks:
1. Use MUI components in this repo
2. Disable sub-schema support (unsure if we want to encourage defining things as sub-schemas rather than separate entities)
3. New wrapping page + plumbing with the API
4. Add ability to link a property field to a schema.org property

**Suggestion for reviewers**
The components in `site/src/components/entityTypes/SchemaEditor/` are basically the same as the ones that have been already reviewed elsewhere. I'd focus on new functionality:
- `updatePropertySchemaOrgLink`
- the wrapping page at `pages/[shortname]/types/[title].tsx`
- the canonical URI at `pages/types/[entityTypeId].tsx` which redirects to the one above, unless you add `?json` or request it with an  `accepts: "application/json"` header, in which case you'll get the raw JSON
- changes t `site/src/lib/jsonSchema.ts`

**To test**
- Create an entity type via the API following instructions in the README or #128 
- visit `/types/[entityTypeId]`, you should be sent to the form
- amend some stuff and refresh to check that changes persist correctly
- visit the raw json at `/types/[entityTypeId]?json` and check it makes sense

**Known issues**
- There is some unused code related to sub-schemas I need to clear up afterwards.
- There's a patch for `ajv` in `site/patches` rather than at the workspace root: running `patch-package` from the root does not generate a patch, I think because the edited code is in `site/node_modules`. I might be doing something wrong with the library, unsure.

**Screenshots**

Screenshot 2022-01-26 at 18 51 35

---

Screenshot 2022-01-26 at 18 51 40


site/src/components/ConfirmationAlert.tsx
```diff
@@ -0,0 +1,48 @@
+import { FunctionComponent, ReactNode } from "react";
```
CiaranMn, line 1: This is unused - it is to confirm deleting sub-schemas. Will either use it or remove it in a follow up.

site/src/components/entityTypes/SchemaEditor/Inputs.tsx
```diff
@@ -0,0 +1,74 @@
+import Checkbox from "@mui/material/Checkbox";
```
CiaranMn, line 1: Untouched except to swap out checkbox and textfield components

site/src/components/entityTypes/SchemaEditor/SubSchemaItem.tsx
```diff
@@ -0,0 +1,135 @@
+import { useRouter } from "next/router";
```
CiaranMn, line 1: Unused - to remove or use in a followup

site/src/components/entityTypes/SchemaEditor/schemaEditorReducer.ts
```diff
@@ -0,0 +1,393 @@
...
+  | Action<"togglePropertyIsArray", { propertyName: string }>
+  | Action<"togglePropertyIsRequired", { propertyName: string }>
+  | Action<
+      "updatePropertyDescription",
+      { propertyName: string; newPropertyDescription: string }
+    >
+  | Action<
+      "updatePropertySchemaOrgLink",
+      { propertyName: string; newSchemaOrgUri: string }
+    >
```
CiaranMn, line 26: this is new, the other actions are untouched

site/package.json
```diff
@@ -13,7 +13,7 @@
     "dev:seed-db": "node --max-old-space-size=2048 node_modules/.bin/ts-node --files --project tsconfig.json scripts/seed-db",
     "generate-blocks-data": "node --max-old-space-size=2048 node_modules/.bin/ts-node --files --project tsconfig.json scripts/generate-blocks-data",
     "generate-sitemap": "node --max-old-space-size=2048 node_modules/.bin/ts-node --files --project tsconfig.json scripts/generate-sitemap",
-    "postinstall": "yarn generate-sitemap && yarn generate-blocks-data",
+    "postinstall": "yarn generate-sitemap && yarn generate-blocks-data && patch-package",
```
kachkaev, line 16:
```suggestion
    "postinstall": "patch-package && yarn generate-sitemap && yarn generate-blocks-data",
```

site/patches/ajv+8.9.0.patch
```diff
@@ -0,0 +1,15 @@
...
+--- a/node_modules/ajv/dist/core.js
++++ b/node_modules/ajv/dist/core.js
+@@ -549,8 +549,8 @@ function checkKeyword(keyword, def) {
+     (0, util_1.eachItem)(keyword, (kwd) => {
+         if (RULES.keywords[kwd])
+             throw new Error(`Keyword ${kwd} is already defined`);
+-        if (!KEYWORD_NAME.test(kwd))
+-            throw new Error(`Keyword ${kwd} has invalid name`);
++        // if (!KEYWORD_NAME.test(kwd))
++        //     throw new Error(`Keyword ${kwd} has invalid name`);
```
kachkaev, line 12: It’d be helpful to introduce a comment explaining the change

site/src/components/entityTypes/SchemaEditor/SchemaEditor.tsx
```diff
@@ -0,0 +1,334 @@
...
+import { SchemaPropertiesTable } from "./SchemaPropertiesTable";
+import { JsonSchema } from "../../../lib/jsonSchema";
+import { TextInputOrDisplay } from "./Inputs";
+import {
+  schemaEditorReducer,
+  SchemaEditorReducerAction,
+} from "./schemaEditorReducer";
+// import { SubSchemaItem } from "./SubSchemaItem";
+import { Link } from "../../Link";
+// import { Button } from "../../Button";
```
kachkaev, line 30:
```suggestion
import { Link } from "../../Link";
```

site/src/components/entityTypes/SchemaEditor/SchemaEditor.tsx
```diff
@@ -0,0 +1,334 @@
...
+  //   event.preventDefault();
+  //
+  //   if (!newSubSchemaName.trim()) {
+  //     return;
+  //   }
+  //
+  //   addSubSchema(newSubSchemaName);
+  //   setNewSubSchemaName("");
+  // };
+
```
kachkaev, line 249:
```suggestion
```

(alternatively we can add a comment saying when to uncomment)

kachkaev (commented):
Left a couple of optional comments based on the diff; testing the change locally 👀

site/src/pages/[shortname]/types/[title].page.tsx
```diff
@@ -0,0 +1,170 @@
...
+      .updateEntityType({ schema: JSON.stringify(schema) }, entityTypeId)
+      .then(({ data }) => {
+        if (data) {
+          return [data.entityType.schema];
+        }
+        throw new Error("Could not update entity type");
+      });
+
+  if (!entityType) {
+    return null;
```
Comment (kachkaev) [inline site/src/pages/[shortname]/types/[title].page.tsx:84]:
```js
import Error from "next/error";
```

```suggestion
    return <Error statusCode={404} />;
```

Otherwise `/@hash/types/adsf` gives us a blank page. It’d be ideal to move fetching to `getServerSideProps` so that we can return `{ notFound: true }` but that’s not something we want to do now.

site/src/pages/types/[entityTypeId].page.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const { req, res } = context;
+
+  const { db } = await connectToDatabase();
+
+  const entityTypeId = context.query.entityTypeId as string;
+  const entityType = await EntityType.getById(db, { entityTypeId });
+
+  const jsonFlag = new URL(`${FRONTEND_URL}${req.url}`).searchParams.get(
+    "json",
+  );
```
Comment (kachkaev) [inline site/src/pages/types/[entityTypeId].page.tsx:25]:
```suggestion
  const jsonFlag = Object.prototype.hasOwnProperty.call(context.query, "json");
```

The original statement returns `undefined` for `?json`, which makes the condition below a bit cryptic.

We can replace  `Object.prototype.hasOwnProperty.call` with [`Object.hasOwn`](https://github.com/tc39/proposal-accessible-object-hasownproperty) soon as part of ES2022.

site/src/pages/types/[entityTypeId].page.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return {
+      redirect: {
+        permanent: false,
+        destination: `/${entityType.schema.author}/types/${entityType.schema.title}`,
+      },
+    };
+  } else {
+    return {
+      notFound: true,
+    };
```
Comment (kachkaev) [inline site/src/pages/types/[entityTypeId].page.tsx:46]: We return a short text message in one case and a 404 page in another. It’d be great to align them.

If we want `/types/xyz` to act as an API, we can move it into the `/api/` folder and add a [rewrite](https://nextjs.org/docs/api-reference/next.config.js/rewrites). This will give `res.json()` as part of `NextApiHandler`, so we won’t have to worry about headers.

site/src/pages/[shortname]/types/[title].page.tsx
Comment (kachkaev) [reply site/src/pages/[shortname]/types/[title].page.tsx:84]: Actually we should introduce the loading state here too. Otherwise, we’ll see 404 while the entity is loading.

kachkaev (commented):
It works! 🎉 Great stuff @CiaranMn!

Screenshot 2022-01-26 at 22 30 25
Screenshot 2022-01-26 at 22 30 18


Happy to approve after you reply to comments or address them 🙂
I’d say all are optional at this point.

site/package.json
CiaranMn: Done

site/patches/ajv+8.9.0.patch
CiaranMn: Done

site/src/components/entityTypes/SchemaEditor/SchemaEditor.tsx
CiaranMn: I've added a comment above these and grouped them

site/src/components/entityTypes/SchemaEditor/SchemaEditor.tsx
CiaranMn: I've added a comment - I did start removing this code yesterday but it was going to involve digging through more components, and it might just have to be all restored soon. So I've added comments/`@todos` for now

site/src/pages/[shortname]/types/[title].page.tsx
Comment (CiaranMn) [reply site/src/pages/[shortname]/types/[title].page.tsx:88]: Done

site/src/pages/types/[entityTypeId].page.tsx
Comment (CiaranMn) [reply site/src/pages/types/[entityTypeId].page.tsx]: Not sure I agree that the proposal is any clearer, but I've added it

site/src/pages/types/[entityTypeId].page.tsx
Comment (CiaranMn) [reply site/src/pages/types/[entityTypeId].page.tsx:43]: I think for things that aren't expected to return web pages, we don't need to return a 404 page.

I'd like to keep the URL for types as `/types/xxx` rather than adding an intervening `/api/`. Keeping it here makes it clear what's going on, rather than having to achieve that through redirects elsewhere.

I don't think being able to remove a line setting the header is worth the additional complexity.

Comment (kachkaev) [reply site/src/pages/types/[entityTypeId].page.tsx:43]: IC.

Next renders `return { notFound: true }` as an HTML page. So we probably want to replace that with `res.statusCode = 404; res.write("Schema not found."); res.end();` for consistency with another `if/else` branch. That does not have to be done now though.

kachkaev (approved):

site/package.json
```diff
@@ -13,7 +13,7 @@
     "dev:seed-db": "node --max-old-space-size=2048 node_modules/.bin/ts-node --files --project tsconfig.json scripts/seed-db",
     "generate-blocks-data": "node --max-old-space-size=2048 node_modules/.bin/ts-node --files --project tsconfig.json scripts/generate-blocks-data",
     "generate-sitemap": "node --max-old-space-size=2048 node_modules/.bin/ts-node --files --project tsconfig.json scripts/generate-sitemap",
-    "postinstall": "yarn generate-sitemap && yarn generate-blocks-data",
+    "postinstall": "patch-package && yarn generate-sitemap && yarn generate-blocks-data",
```
kachkaev, line 16: Hmm just noticed that this call is inside a workspace. I think we should investigate why `ajv` is not hoisted and them move the script to the root `package.json`. That can be done later.

site/src/pages/types/[entityTypeId].page.tsx
Comment (CiaranMn) [reply site/src/pages/types/[entityTypeId].page.tsx:43]: > Next renders return { notFound: true } as an HTML page

Right, but the user only reaches that if they're in the HTML branch - the intention of this route is that you either
1) get JSON (and a plain 404) if you've requested JSON, otherwise
2) sends you to the HTML page for the type, or a 404 page if it doesn't exist

I'll get this in and we can refine afterwards - thank you!

Comment (kachkaev) [reply site/src/pages/types/[entityTypeId].page.tsx:43]: Ah right – I forgot about `jsonFlag` we have in the first `if`. All makes sense not, sorry for the noise! 🤦

PR #140, opened 2022-01-27
[SITE] User Pages Frontend

teenoh:
This PR

- adds the frontend for user pages and connects it with the api created in #128 
- adds a modal in the user's profile page for creating schemas

The code still needs some final cleanups in terms of style/structure and I plan to do that in a follow up PR

---
  - https://app.asana.com/0/0/1201727643575060
  - https://app.asana.com/0/0/1201706915162194


site/src/components/Link.tsx
```diff
@@ -8,11 +8,13 @@ import NextLink, { LinkProps as NextLinkProps } from "next/link";
...
 import { styled } from "@mui/material/styles";
 import { Button } from "./Button";
+import { FRONTEND_URL } from "../lib/config";
 
 export const isHrefExternal = (href: string | UrlObject) =>
   typeof href === "string" &&
   (href === "/discord" ||
-    !/^(mailto:|#|\/|https:\/\/blockprotocol\.org)/.test(href));
+    !/^(mailto:|#|\/|https:\/\/blockprotocol\.org)/.test(href)) &&
+  !href.includes(FRONTEND_URL);
```
teenoh, line 17: updated this because it treated absolute local urls as external links. An example of where we have to use an absolute local url is `schema.$id` property in entityTypes

nonparibus:
So to confirm, the workflow for creating schemas here is to go to a user's own profile page (which may be a variable link... `blockprotocol.org/@something`), switch to the schemas tab, and then click 'Create'?

Can we also ensure that the 'Create' screen/modal is accessible via a permalink that will be universal irrespective of the logged-in user's namespace? e.g. `blockprotocol.org/schemas/new` or something that makes sense?

This should not block this PR, but is required today - we expect to want to link to it directly from within the docs/blog posts.

thehabbos007:
image
There seems to be 2 @s in the editor, and the link also redirects to "/@@asat"

site/src/components/Link.tsx
kachkaev:
```suggestion
  !href.startsWith(FRONTEND_URL);
```

teenoh:
> So to confirm, the workflow for creating schemas here is to go to a user's own profile page (which may be a variable link... `blockprotocol.org/@something`), switch to the schemas tab, and then click 'Create'?
> 
> Can we also ensure that the 'Create' screen/modal is accessible via a permalink that will be universal irrespective of the logged-in user's namespace? e.g. `blockprotocol.org/schemas/new` or something that makes sense?
> 
> This should not block this PR, but is required today - we expect to want to link to it directly from within the docs/blog posts.

Noted @nonparibus. 
What will be the flow for non-logged in users?
Open Link (/schemas/new) -> Redirect to Auth page => Redirect back to (/schemas/new) after successful auth ?
At the moment users have to be authenticated before they can view the create schema modal.

teenoh:
> image There seems to be 2 @s in the editor, and the link also redirects to "/@@asat"
> 
> Edit: this only happens after creating a new Schema. Subsequent visits to existing schemas do not show this

Thanks for catching, will update

CiaranMn (approved):

PR #143, opened 2022-01-27
New block dev package: mock-block-dock

CiaranMn:
This PR introduces a package for making it easier to develop blocks, `mock-block-dock` - which is essentially a dummy embedding application.

It provides an in-memory datastore, and passes the basic Block Protocol functions to blocks, e.g. `createEntities`, `createLinks`, etc, so that block authors don't have to provide their own mock embedding application.

Usage:
```jsx
<MockBlockDock>
  <TestBlock {...props} />
</MockBlockDock>
```

As part of the PR, I've removed the existing mock functions when rendering a block on the site, in favour of this package.

To test:
1. Run the site locally (`yarn dev`)
2. Go to `/hub` and click on a block (e.g. `Image`)
3. Add an image, check it works
or
1. Go to some block in the `hash` repo, e.g. image, video, embed
2. `yarn add mock-block-dock`
3. in the dev renderer, remove the functions passed to the component and wrap it in this package instead

PR #144, opened 2022-01-27
[FE] Rebuild blocks on blockprotocol.org

akash-joshi:
This PR is a follow-up to the previous PRs in the hash repo which add examples to various blocks: https://github.com/hashintel/hash/pull/239, https://github.com/hashintel/hash/pull/233 and https://github.com/hashintel/hash/pull/227. The timestamps inside of the block json files have been updated to cause a rebuild of the blocks and render the example.

`initialWidth` is also updated in a place to increase the width the HASH embed starts with.

### DIFF
--- a/site/src/components/pages/hub/BlockDataContainer.tsx
+++ b/site/src/components/pages/hub/BlockDataContainer.tsx
@@ -43,7 +43,6 @@ export const BlockDataContainer: VoidFunctionComponent<
   const [blockState, setBlockState] = useState<Record<string, unknown>>({
     accountId: "test-account-id",
     entityId: "test-entity-id",
-    initialWidth: 600,
   });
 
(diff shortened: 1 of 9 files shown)


site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -43,7 +43,7 @@ export const BlockDataContainer: VoidFunctionComponent<
   const [blockState, setBlockState] = useState<Record<string, unknown>>({
     accountId: "test-account-id",
     entityId: "test-entity-id",
-    initialWidth: 300,
+    initialWidth: 600,
   });
```
CiaranMn, line 47: I've removed this in #143, `initialWidth` is not a common property to all blocks - can you update it in the example for the embed instead please?

CiaranMn (changes requested):

CiaranMn (approved):

PR #147, opened 2022-01-27
[SITE] Fix API URL reaching client

CiaranMn:
We have a variable, `FRONTEND_URL`, which determines various things including:
- cookie domain
- origin of links sent to users
- endpoint that the API client will call

The variable can be set via user-supplied environment variable, and falls back to the Vercel deployment origin if this isn't present. 

For the third use case, the variable needs to reach the client - this requires prefixing it with `NEXT_PUBLIC_`, which the PR does.

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -19,7 +19,7 @@ The public-facing [blockprotocol.org](https://blockprotocol.org) website serves 
     - `MONGODB_DB_NAME`: the name of the database (for example `local`)
     - `MONGODB_USERNAME`: the database username
     - `MONGODB_PASSWORD`: the database password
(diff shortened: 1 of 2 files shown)


kachkaev:
I guess we can do something like this next time:

https://github.com/hashintel/hash/blob/59b8325e795834a319ce629684e9045014da33a3/packages/hash/frontend/next.config.js#L23-L26

`NEXT_CONFIG_*` becomes an implementation detail of Next and is only used in two files `next.config.js` and some `src/shared/public-env.ts`. The latter reads from `process.env` and exports normal vars. If we need to make yet another variable public, we won't have to change README, deployment config, etc.

PR #157, opened 2022-01-28
[DB|BUILD] Create MongoDB indexes separately from seed script

thehabbos007:
This PR splits up DB seeding and index creation.
Indexes creation is an idempotent operation, so it's harmless to run these at each build.

Both "yarn build" and "yarn dev" call down to "yarn db:indexes" that creates indexes.
This change relates to [asana task (internal)](https://app.asana.com/0/1201095311341924/1201715945634529/f)


thehabbos007:
Seems that index creation is not idempotent when the indices are created with conflicting options

site/package.json
```diff
@@ -4,13 +4,14 @@
   "private": true,
   "license": "MIT",
   "scripts": {
-    "build": "yarn build-blocks && yarn generate-blocks-data && yarn generate-sitemap && next build",
+    "build": "yarn build-blocks && yarn generate-blocks-data && yarn generate-sitemap && yarn db:indexes && next build",
     "build-block": "./scripts/build-blocks.sh",
     "build-blocks": "find ../hub -type f | xargs ./scripts/build-blocks.sh",
     "clean": "rimraf ./.next/",
-    "dev": "yarn generate-sitemap && yarn generate-blocks-data && next dev",
+    "db:indexes": "node --max-old-space-size=2048 node_modules/.bin/ts-node --files --project tsconfig.json scripts/db-indexes",
```
kachkaev, line 11: This gives us a mix of `action:subject` and `subject:action` scripts. Ideally, we want to stick with the former for consistency. Not sure how exactly to change the naming at this point, but we might need to clean things up a little in a separate PR.

kachkaev: In an ideal world, some of our scripts are hierarchical, so running `x` implies running `x:a` and `x:b`. In turn, `x:a` can involve `x:a:n` and `x:a:m`. We already do this for `lint:*` and `fix:*` scripts and can do so too in other scenarios.

site/scripts/db-indexes.ts
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable import/first */
+import dotenv from "dotenv";
+
+dotenv.config({ path: ".env.local" });
```
kachkaev, line 4: Let’s use [`dotenv-flow`](https://www.npmjs.com/package/dotenv-flow) instead. It matches [how Next.js reads env vars](https://nextjs.org/docs/basic-features/environment-variables).

kachkaev: We can also avoid `eslint-disable` I guess. To do this, we can start the script with `import "../shared/environment"`, which will in turn call:

```ts
import { config } from "dotenv-flow";

export const monorepoRootDir = path.resolve(__dirname, "../../..");
config({ silent: true, path: monorepoRootDir });
```

Another way of solving this can be adding `exe` to `package.json` → `scripts`:

```
  "exe": "ts-node --require dotenv-flow/config --transpile-only"
```

Then we can `yarn exe scripts/a/b/c`.

site/scripts/db-indexes.ts
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable import/first */
```
kachkaev, line 1: My suggestion would be to use verb-like file names for scripts, like `scripts/do-something` or `scripts/do-something`. This will help us understand what each script does.

site/scripts/db-indexes.ts
```diff
@@ -0,0 +1,55 @@
...
+    await func();
+  } catch (error) {
+    if (error instanceof Error) {
+      console.error(`❌ Error trying to create index, ${error.toString()}`);
+    }
+  }
+};
+
+// Actions done in the following script should be _idempotent_ as they run in dev/prod
+void (async () => {
```
kachkaev, line 24: Minor suggestion:

```ts
const script = async () => {
 // ...
}

void script();
```

This will be similar to what we have in `next-app/pages/api/*` with `handler` functions.

When we upgrade to native ESM modules, we’ll be able to replace `void script()` with `await script()` thanks to top-level await. Thus, we’ll be able to compose the scripts without using exports. That’s because top-level await [does not timeout](https://stackoverflow.com/q/68694714/1818285).

Until we‘ve unlocked top-level await, defining `const script` can help us composing them if needed. If we need to call a script from another script, we’ll just add `export default script` at the end of the file. 

I’m already using ESM scripts with top-level await [in a side project](https://github.com/kachkaev/tooling-for-how-old-is-this-house) and they are quite cool! Hope we’ll get there soon, but at least let’s get prepared 🙂

site/scripts/db-indexes.ts
```diff
@@ -0,0 +1,55 @@
...
+dotenv.config({ path: ".env.local" });
+import { User, UserDocument } from "../src/lib/model/user.model";
+import {
+  VerificationCode,
+  VerificationCodeDocument,
+} from "../src/lib/model/verificationCode.model";
+import { connectToDatabase } from "../src/lib/mongodb";
+import { EntityType } from "../src/lib/model/entityType.model";
+
+const catchAndLog = async (func: () => Promise<any>) => {
```
kachkaev, line 13:
```suggestion
const catchAndLog = async (func: () => Promise<void>) => {
```

site/scripts/db-indexes.ts
kachkaev: Adding `yarn exe` here: https://github.com/blockprotocol/blockprotocol/pull/169

thehabbos007: How should we handle the existing `scripts` in the site project, then? 
Should we remove them and put them at the root package.json using `exe`, or keep them as is and make naming more consistent?

kachkaev: That’s a good question! Both options have pros and cons.

**If all scripts are in monorepo root**

- 👍 There is only one `yarn exe` and one `/scripts/` folder
- 👍 Commands are shorter (`yarn exe scripts/a/b/c` instead of `yarn workspace package-name exe scripts/a/b/c`)
- 👍 Shell autocompletion is more likely to work for script paths because they are relative to CWD
- 👍 We don’t need to exclude `packages/a/scripts` from publishing (I assume scripts are only for repo maintenance, they are not the same as publishable CLI commands in NPM packages).
- 🤔 We have to keep more `devDependencies` in root `package.json`; those dependencies will be quite mixed
- 🤔 If `devDependencies` are quite big (e.g. Puppeteer), it’s harder to do selective installs (e.g. to build a docker image out of a couple of packages). Not sure we need to think about this in this repo, the question is more related to [HASH](https://github.com/hashintel/hash).

**If all scripts are in individual packages**

Same as above, but with 👍 and 🤔 swapped. Plain count of pros and cos does not mean that some option is better.

I once worked on a monorepo where all scripts were put into `packages/our-scripts`, so they had their own `package.json`. It had cons of both options, so I can’t say it was any better 😅

To be fair, it should be quite simple to jump between global and package-scoped scripts. We can try any option in a consistent way, see how it works for us and then swap the approach if we are not happy 🙂

site/scripts/create-db-indexes.ts
```diff
@@ -0,0 +1,57 @@
+/* eslint-disable import/first */
+import dotenv from "dotenv";
+
+dotenv.config({ path: ".env.local" });
```
kachkaev, line 4: Let’s use `yarn exe ...` – it’s in `main` already. If you want to keep the script inside `site`, we can create a copy of `exe` in a local `package.json`. This will save us from using `dotenv` here and repeating node options in `package.json`. We won’t have to add a new `package.json` script per each script file too, so it will be easier to maintain the growing number of scripts.

thehabbos007: Sure that sounds good. Will get around to it soon! Thank you a lot :)

thehabbos007:
@kachkaev is this what you had in mind?

site/package.json
```diff
@@ -4,17 +4,16 @@
   "private": true,
   "license": "MIT",
   "scripts": {
-    "build": "yarn build-blocks && yarn generate-blocks-data && yarn generate-sitemap && next build",
+    "build": "yarn build-blocks && yarn exe scripts/generate-blocks-data && yarn exe scripts/generate-sitemap && yarn exe scripts/create-db-indexes && next build",
     "build-block": "./scripts/build-blocks.sh",
     "build-blocks": "find ../hub -type f | xargs ./scripts/build-blocks.sh",
     "clean": "rimraf ./.next/",
-    "dev": "yarn generate-sitemap && yarn generate-blocks-data && next dev",
+    "dev": "yarn exe scripts/generate-sitemap && yarn exe scripts/generate-blocks-data && yarn exe scripts/create-db-indexes && next dev",
```
kachkaev, line 11: We can try composing scripts outside `package.json` as this is more flexible. I can try one small idea in a follow-up once you’ve merged.

site/scripts/create-db-indexes.ts
```diff
@@ -0,0 +1,53 @@
+import { User, UserDocument } from "../src/lib/model/user.model";
+import {
+  VerificationCode,
+  VerificationCodeDocument,
+} from "../src/lib/model/verificationCode.model";
+import { connectToDatabase } from "../src/lib/mongodb";
+import { EntityType } from "../src/lib/model/entityType.model";
+
+const catchAndLog = async (func: () => Promise<void>) => {
```
kachkaev, line 9: It might be useful to add a small comment here explaining why we catch and log instead of thowing and terminating the whole script.

kachkaev (approved):
Looking good! Thanks @thehabbos007!

PR #159, opened 2022-01-28
[SITE] Fixes: Profile page card text wrapping. Fix a few links

CiaranMn:
A few small fixes around schemas and links:

1. Make the `@username` on block pages a link ([internal task](https://app.asana.com/0/1201668052739245/1201726402115280/f)) 

2. Break schema titles ([internal task](https://app.asana.com/0/1201668052739245/1201734892191469/f))

**Before**
Screenshot 2022-01-28 at 16 40 50

**After**
Screenshot 2022-01-28 at 17 02 49

3. Fix link in spec to type that now exists (check [link](https://blockprotocol.org/types/43c682fe-8b01-4b26-b426-83edd8562ee0) in prod)

4. The `?json` link only works on a server render because it uses the  `res`. I've just make this open in a new window as a fix.

---
  - https://app.asana.com/0/0/1201726402115286


kachkaev:
@MaggieAppleton I wonder if we can reduce the number of hyphens by removing the conflict between the header and the pill. E.g. somehow like this:

151594980-8f998eb3-c461-4d4c-a39d-4f3b36a734ac

(made in Preview.app 😅)

teenoh:
> 3\. link

Agree with @kachkaev. Can probably place "schema" on the left

site/src/components/pages/user/utils.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * Insert shy hypen indicators at capital letters within single words to improve wrapping ability
+ */
+export const shy = (text: string) =>
```
kachkaev, line 4: I’d call this function `insertShyHyphensAtCaptitalLeters`, but that’s too minor. 🙂 Maybe we can do this later if we need to hoist reuse the function elsewhere.

kachkaev (approved):

PR #160, opened 2022-01-28
[SITE] Fix sidebar stickiness in /spec and /docs

kachkaev:
Follows #125

[Asana task](https://app.asana.com/0/1201668052739245/1201726402115290/f) _(internal link)_

https://user-images.githubusercontent.com/608862/151612715-9ddb3cb4-dc48-4629-b24c-2442c0935fad.mp4


site/src/components/PageSidebar.tsx
```diff
@@ -381,24 +383,34 @@ export const Sidebar: VFC<SidebarProps> = ({
         }}
       />
       <Box
-        paddingRight={3}
         sx={{
           maxHeight: isSticky ? "100vh" : undefined,
-          overflow: isSticky ? "auto" : undefined,
```
CiaranMn, line 387: Can we not retain something like this, or move it to the `Box` below?

Not a fan of having `overflow: scroll` on permanently if we can avoid it

kachkaev: Not sure why, but I did not see a scrollbar on tall screens depsite `overflow: "scroll"`. I changed it to `"auto"` just in case. If you still see some oddities in the UI, could you please share a screenshot? Happy to take another look.

CiaranMn: Looks good now, thanks!

CiaranMn (approved):

PR #164, opened 2022-01-31
[SITE] Redirect /faq to /docs/faq

CiaranMn:
The canonical url for the FAQ page is `/docs/faq`, but for cases where people accidentally visit `/faq` instead (either by their or our error), it's handy to redirect `/faq` to `/docs/faq`. This PR does that.


kachkaev (approved):

MaggieAppleton:
Thanks for that fix! I should have double-checked the URL before tweeting 😅  saves us from all future mistakes though

CiaranMn:
> Thanks for that fix! I should have double-checked the URL before tweeting 😅 saves us from all future mistakes though

No problem - an easy mistake to make!

PR #165, opened 2022-01-31
Introduce `check-dependency-version-consistency`

kachkaev:
[Asana task](https://app.asana.com/0/0/1201720874843702/f) _(internal)_

@nathggns and I discovered [`check-dependency-version-consistency`](https://www.npmjs.com/package/check-dependency-version-consistency) when working on https://github.com/hashintel/hash/pull/206. I accidentally introduced a duplicate ProseMirror dependency version and Nate spotted it by looking into `yarn.lock`. Using `check-dependency-version-consistency` will protect us from similar human errors in the future. I am adding it now as I am planning to bump dependencies in `block-template` soon and it’d be handy to have an extra CI check in place before then.

For the new check to pass, I had to bump a couple of versions by calling `yarn fix:dependency-version-consistency; yarn fix:yarn-deduplicate`.

Our existing `fix:*` and `lint:*` scripts are named after the tools we invoke. So the new scripts would have to be `fix:check-dependency-version-consistency` and `lint:check-dependency-version-consistency`. Maintainers are thinking of renaming the package or at least the CLI script, which has allowed me to go for a shorter name: https://github.com/bmish/check-dependency-version-consistency/issues/321.

Unfortunately, `check-dependency-version-consistency` does not check local packages yet – see https://github.com/bmish/check-dependency-version-consistency/issues/332. The tool is still useful for external packages, we just need to be aware of this (temporary?) limitation.


site/src/components/Footer.tsx
```diff
@@ -148,7 +148,7 @@ const Socials = (
       }}
       startIcon={<Icon className="fa fa-star" />}
     >
-      Star us on Github
+      Star us on GitHub
```
kachkaev, line 151: Bonus: Github → GitHub (not worth its own PR)

CiaranMn (approved):

PR #167, opened 2022-01-31
Make Schema Table Responsive

akash-joshi:
## What this Does

Users inside blockprotocol can create or edit schemas for various entities. However, the current view of the schema pages breaks in mobile view. This PR is responsible for resolving the styling issues, and making the schema editor and viewing table responsive.

- [Make the text display responsive](https://github.com/blockprotocol/blockprotocol/pull/167/files#diff-309a59f94b4cdf913ac3ca9228a73fd75b2436c0ec3ed0a525abbaf39a3a8736R32)
- [Add overflow-x to parent](https://github.com/blockprotocol/blockprotocol/pull/167/files#diff-03a91238518a41200c3aadcccca251ca5002dd0da450e8b9ec637fb944b2572cR284)
- [Add min-width to input fields](https://github.com/blockprotocol/blockprotocol/pull/167/files#diff-3f0032dfe916053af5adc7019965db8e5014d77e8f93e3cde14fc3d1547ea0f5R907)

## How to Test this?

- Open up a user's schema or a different user's schema.
- Test responsiveness for various screen sizes.

chrome_n08SH5F3ZB
chrome_eu1LGw6Oxf

---
  - https://app.asana.com/0/0/1201303965906637


site/src/components/theme.ts
```diff
@@ -981,6 +981,7 @@ export const theme = createTheme({
       },
       styleOverrides: {
         root: {
+          minWidth: "125px",
```
CiaranMn, line 984: I'm not sure about setting this globally - we might want a smaller input, e.g. for a number. How about passing this through from the schema inputs instead?

CiaranMn (changes requested):

site/src/components/theme.ts
akash-joshi: Alright, have updated the TextInput field instead of making this a global here: https://github.com/blockprotocol/blockprotocol/pull/167/files#diff-309a59f94b4cdf913ac3ca9228a73fd75b2436c0ec3ed0a525abbaf39a3a8736R73

site/src/components/entityTypes/SchemaEditor/Inputs.tsx
```diff
@@ -27,7 +28,22 @@ export const TextInputOrDisplay: VoidFunctionComponent<{
   }, [value]);
 
   if (readonly) {
-    return <span>{value}</span>;
+    return (
+      <Box
```
CiaranMn, line 32: This would be better as `Typography`, which is meant to hold text (and will make it a `span` by default, as it was before)

akash-joshi: Have replaced `Box` with `Typography` in the latest commit: https://github.com/blockprotocol/blockprotocol/pull/167/commits/d75062cd57a5710a025383c972e66ee0f8441a74.

Now, it renders it as a `p` instead of a `span`, but I'm letting it be since the `width` property doesn't apply to a span without `display: block`.

Additionally, I've added `overflow-wrap: anywhere` to allowing breaking long unspaced words like URLs.

akash-joshi: Should be done in the latest commit: https://github.com/blockprotocol/blockprotocol/pull/167/commits/a7cbf6ccc8b764da321125aa70686a52895f0322

PR #168, opened 2022-01-31
Upgrade dependencies in `mock-block-dock` and `block-template`

kachkaev:
This PR closes #162 by upgrading `block-template` dependencies. We now have webpack 5 in the toolchain, which provides a slightly better DX and opens opportunities for further improvements. Modern JS syntax like `??` seems to be working in the blocks – it compiles without errors.

---

We still get ESLint errors for the modern syntax, but that’s unrelated to the published blocks. Our repo still on `eslint` v7, which uses a pretty old version of `espree` parser. This will be addressed separately.

Screenshot 2022-02-09 at 00 04 43

---

There is another problem in packages, which is probably outside this PR scope. `yarn dev` for `mock-block-dock` is partially failing in both on `main` and in this branch.

<details>
<summary>Before (webpack 4)</summary>
<pre><code>yarn run v1.22.17
$ concurrently -n webpack,webpack-dev-server -c green,cyan "yarn build:dev -- --watch --verbose" "yarn run-dev-server"
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
$ cross-env NODE_ENV=development webpack-dev-server -d --port 9090 --config dev/webpack.config.js
$ clean; cross-env NODE_ENV=development webpack --mode development --watch --verbose
[webpack] /bin/sh: clean: command not found
[webpack] 
[webpack] webpack is watching the files...
[webpack] 
[webpack] 
[webpack] Insufficient number of arguments or no entry found.
[webpack] Alternatively, run 'webpack(-cli) --help' for usage info.
[webpack] 
[webpack] Hash: f51f0734ccb9579023e8
[webpack] Version: webpack 4.46.0
[webpack] Time: 16ms
[webpack] Built at: 02/09/2022 1:32:48 AM
[webpack] 
[webpack] LOG from webpack.buildChunkGraph.visitModules
[webpack] <t> prepare: 0.060083ms
[webpack] 
[webpack] ERROR in Entry module not found: Error: Can't resolve './src' in '/Users/ak/-/projects/hash/github/blockprotocol/packages/mock-block-dock'
[webpack] resolve './src' in '/Users/ak/-/projects/hash/github/blockprotocol/packages/mock-block-dock'
[webpack]   using description file: /Users/ak/-/projects/hash/github/blockprotocol/packages/mock-block-dock/package.json (relative path: .)
[webpack]     Field 'browser' doesn't contain a valid alias configuration
[...]

PR #169, opened 2022-01-31
Remove duplicate `yarn.lock`, configure `yarn exe` for scripts

kachkaev:
This PR helps unblock #169 

I’ve noticed us having disjoint dependency versions in `.github/workflows/algolia/package.json` and other `package.json` files. The former file is not reached by `check-dependency-version-consistency` (#165) because `.github/workflows/algolia` is not a workspace. Thus, when we upgrade dependencies, we have to either ignore `.github/workflows/algolia/package.json` or update both `yarn.lock` files. This is not sustainable.

This PR makes sure we have only one `yarn.lock` file in the whole project. To run the Algolia script, I created `yarn exe` for convenience. See https://github.com/blockprotocol/blockprotocol/pull/157#discussion_r795958433 for context.


package.json
```diff
@@ -33,10 +34,14 @@
     ]
   },
   "devDependencies": {
+    "@swc/core": "^1.2.135",
+    "@swc/helpers": "^0.3.2",
```
kachkaev, line 38: Enables fast transpilation in `ts-node`
https://github.com/TypeStrong/ts-node#bundled-swc-integration

package.json
```diff
@@ -33,10 +34,14 @@
...
     "@typescript-eslint/eslint-plugin": "4.29.0",
     "@typescript-eslint/parser": "4.29.0",
+    "algoliasearch": "^4.12.0",
```
kachkaev, line 41: Moved from `.github/workflows/algolia`, which was not a workspace

package.json
```diff
@@ -33,10 +34,14 @@
...
     "check-dependency-version-consistency": "^1.6.0",
     "concurrently": "^6.5.1",
+    "dotenv-flow": "3.2.0",
```
kachkaev, line 44: Helps us load `.env`, `.env.local` etc. automatically. See https://github.com/blockprotocol/blockprotocol/pull/157#discussion_r795958433

packages/create-block-app/create-block-app.js
```diff
@@ -20,7 +20,8 @@ const templatePackageName = "block-template";
     process.exit();
   }
 
-  const blockPath = process.argv[3] ?? blockName;
+  // @todo: replace with `process.argv[3] ?? blockName` after upgrading ESLint
+  const blockPath = process.argv[3] ? process.argv[3] : blockName;
```
kachkaev, line 24: Getting syntax error when running `yarn lint:eslint`. Might be to do with outdated built-in babel. I’ll update ESLint to fix that, which might be handy before #168 anyway:

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.4.0

YOUR TYPESCRIPT VERSION: 4.5.4

Please only submit bug reports when using the officially supported version.

=============
```

kachkaev: Same problem with `} catch {` below

scripts/sync-algolia-index.ts
```diff
@@ -111,8 +115,8 @@ const generateAlgoliaRecords: () => AlgoliaRecord[] = () => {
 
 const syncAlgoliaIndex = async () => {
   const client = algoliasearch(
-    process.env.ALGOLIA_PROJECT,
-    process.env.AGOLIA_WRITE_KEY,
+    process.env.ALGOLIA_PROJECT ?? "",
+    process.env.AGOLIA_WRITE_KEY ?? "",
```
kachkaev, line 119: Typescript finally checks this file. It complained about assigning `undefined` to `string`. Might be worth adding `envalid` and checking vars there for a better error message. I can do that in a follow-up.

scripts/sync-algolia-index.ts
```diff
@@ -149,14 +153,18 @@ const syncAlgoliaIndex = async () => {
...
+      `Error while running Algolia Update: ${
+        // @ts-expect-error -- TS is a bit too strict here. We can‘t use instanceof Error because Algolia errors are plain objects
+        error?.message ?? error
+      }`,
+    );
   }
 };
 
-void main();
+void script();
```
kachkaev, line 170: Justification: https://github.com/blockprotocol/blockprotocol/pull/157/files#r795996557

package.json
thehabbos007: SWC :rocket:

thehabbos007:
Just out of curiosity, without supplying a path to a script with `yarn exe` is that environment a REPL that is usable to do ad-hoc testing?

kachkaev:
Hmmm never thought about REPL in this context, but yeah, it looks so! `yarn exe` is essentially a wrapped version of `node`. We currently run `ts-node` in it, but it will become something like `node --require=dotenv-flow/config --loader ts-node/esm/transpile-only --experimental-specifier-resolution=node` in the future when we unlock native ESM in scripts (https://github.com/TypeStrong/ts-node/issues/1007).

thehabbos007 (approved):
Looks good to me and works locally :)

PR #171, opened 2022-02-01
[FE] Fix props update on page switching in hub

akash-joshi:
## What this Does

For the current paragraph and header in demo in the block hub, the props being passed to the demo block don't update on switching to the hub and to the respective blocks. This results in the header data appearing inside the para block or vice-versa.

This PR fixes the issue by passing a unique `entityId` with each block type. This fixes the issue because `mock-block-dock` only updates `latestBlockEntity` [when either the entities or block entityId change](https://github.com/blockprotocol/blockprotocol/blob/main/packages/mock-block-dock/src/MockBlockDock.tsx#L57).

---
  - https://app.asana.com/0/0/1201726402115271


packages/mock-block-dock/dev/DevApp.tsx
```diff
@@ -5,10 +5,28 @@ import { MockBlockDock } from "../src/MockBlockDock";
 
 const node = document.getElementById("app");
 
-const DevApp = () => (
-  <MockBlockDock>
-    <TestBlock entityId="test-entity-1" name="World" />
-  </MockBlockDock>
-);
+const DevApp = () => {
```
akash-joshi, line 8: This change was to eliminate the possibility that it was the props being passed which were not updating.

It's good to have this in here to display `mock-block-dock` updating to prop changes.

CiaranMn: I'm not sure about including this - it feels like something that would be better as a test, because it's something of  a secondary use of `mock-block-dock`.

Why do we want to display it updating to prop changes here?

akash-joshi: I think it's a good to have to display `mock-block-dock`s prop updating abilities as well, since that would be something users would be wondering about.

If not this, we could add a small note on how prop updating happens in `mock-block-dock` maybe? Let me know what you think.

CiaranMn: A small note sounds good - could you add something to the README that explains something like "If the component's props are changed from outside `MockBlockDock` (e.g. updated from some other state store), the mock entity store will be updated to include the new properties"

akash-joshi: Have reverted this change and added the note here: https://github.com/blockprotocol/blockprotocol/pull/171/files#diff-9a539852eacb254f155dd2d9ebf9b135cd5130b0e5945a7b7d7d027aaa0628f9R39

teenoh (approved):
Thanks, @akash-joshi. Tested and it's working well

packages/mock-block-dock/README.md
```diff
@@ -36,6 +36,13 @@ It will also pass `linkGroups` and `linkedEntities`, which will be populated onc
 
 `MockBlockDock` is automatically included in [block-template](https://www.npmjs.com/package/block-template), which you can copy via [create-block-app](https://www.npmjs.com/package/create-block-app)
 
+## Updating Props
+
+Any props passed to the child of `MockBlockDock` are automatically updated whenever:
+
+1.  The list of entities passed to the `mockDataStore` change.
+1.  The entityId passed to the child changes.
+
```
CiaranMn, line 45: We already mention updating props above so I've suggested adding this note there rather than introducing a new section talking about updating props (which repeats some of what is above).
```suggestion

The block will also be re-rendered with new properties if you update them on the child directly (e.g. if you are supplying the block component wrapped by `MockBlockDock` with props from some outside state).

`MockBlockDock` is automatically included in [block-template](https://www.npmjs.com/package/block-template), which you can copy via [create-block-app](https://www.npmjs.com/package/create-block-app)

```

CiaranMn (changes requested):
Thanks Akash - a suggestion on the README

CiaranMn (approved):

PR #174, opened 2022-02-01
Allow users to upload avatars through S3

thehabbos007:
This PR allows users to upload avatars to their profiles.
Images are uploaded to AWS S3 on a bucket.
Existing AWS env vars are used for connection and `BP_AWS_S3_BUCKET_NAME` has been added to specify name of bucket to upload to.

Images have been limited to be 5 MB.

**New endpoint**: POST /api/uploadUserAvatar` [requires cookie authentication]
Uploads a user avatar and apply it to logged in profile page.
- Multipart Request Body:
  - `image`: The image file to be uploaded, extension must be one of jpg, jpeg, png, gif or svg.
- Response Body:
  - `fullUrl`: Url pointing to the newly uploaded user profile.

---
  - https://app.asana.com/0/0/1201734261758485


thehabbos007:
@kachkaev because of a non-resolved issue (with a PR open) on https://github.com/aws/aws-sdk-js-v3/pull/ PR 2700 (not linked as to not create a backreference) I had to pin to an older version of the library which makes `yarn-deduplicate` complain.
I am not sure how to fix this. Do you have a good way to handle this issue [seen here](https://github.com/blockprotocol/blockprotocol/runs/5034795208?check_suite_focus=true)?

site/src/pages/api/uploadUserAvatar.api.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      // @todo should we delete previous user avatar before replacing it?
+
+      await user.update(db, {
+        userAvatar: { url: fullUrl, s3Key },
+      });
+
+      res.status(200).json({ fullUrl });
+    } else {
+      throw new Error("No file given. Please upload a file.");
```
CiaranMn, line 42: There's a `formatErrors` function that standardises how errors are returned - we should give them all status codes too (where anticipated / user error) so they don't come through as `500`:
```suggestion
      res.status(400).json(
        formatErrors({
          msg: "No file given. Please upload a file.",
          param: "image",
        }),
      );
```

site/src/lib/middleware/imageUploadMultipart.ts
```diff
@@ -0,0 +1,40 @@
...
+    },
+    maxFiles: 1,
+    maxFileSize: 5 * 1024 * 1024, // bytes = 5MB
+  });
+
+  const contentType = req.headers["content-type"];
+  if (contentType && contentType.indexOf("multipart/form-data") !== -1) {
+    form.parse(req, (err, _fields, files) => {
+      const file = Object.values(files)[0];
+      // we are only interested with the _first_ file, so other fields are ignored.
```
CiaranMn, line 29: I would return an error somewhere to alert the user to bad input, either:
(a) return an error here if more than one file is provided, or
(b) return an array of files here and let each API route decide how many files it wants (and return an error if there are too many)

site/src/lib/middleware/imageUploadMultipart.ts
```diff
@@ -0,0 +1,40 @@
...
+  const form = new formidable.IncomingForm({
+    fileWriteStreamHandler: () => {
+      return passThrough;
+    },
+    maxFiles: 1,
+    maxFileSize: 5 * 1024 * 1024, // bytes = 5MB
+  });
+
+  const contentType = req.headers["content-type"];
+  if (contentType && contentType.indexOf("multipart/form-data") !== -1) {
```
CiaranMn, line 26:
```suggestion
  if (contentType?.includes("multipart/form-data")) {
```

CiaranMn (changes requested):
Thanks Ahmad - I tested this out and it worked with small image files, but not with a couple of bigger ones (example avatar attached). It never makes it to the callback on `form.parse`.

avatar

site/src/lib/middleware/imageUploadMultipart.ts
CiaranMn: Not sure what `maxFiles` is doing on the formidable config as adding 2 files to the form doesn't trigger anything

thehabbos007:
Hmm, this is quite interesting behavior from the formidable library. I would expect an `error` in case of files too big, but it's silent and doesn't emit any error at all. I'll investigate.

thehabbos007:
Formidable has been switched out with busboy, and with that a change from Streams to Buffers. Busboy is in-memory first unlike Formidable. And switchign to buffers means any logic errors won't sink the request into a waiting loop like before.

The new error checks are more clear now, and everything is async instead of callback-based. File validation happens from the middleware now. I've also removed the hack around S3, as it's not needed when dealing with buffers. Everything should be simpler now :)

cc: @CiaranMn

CiaranMn (changes requested):
Thanks Ahmad, this is working well for me now with various images.

One thing that we could improve is: at the moment, it's irrelevant what body property the user passes an image in via - as long as there's a single file, it will be attached to `body.upload`. 

This isn't a big deal for this endpoint at the moment, but if we ever change this route or add others which require more than a single field, it might lead to them getting mixed up.

We could alternatively return from `imageUploadMultipart` keyed by the user-supplied field name, and use those as the keys on the body.

PR #177, opened 2022-02-02
Add searchbar for mobile

akash-joshi:
## What this Does

PR #125 added search to docs pages. However, the mobile version of the searchbar for the `docs` and `spec` pages wasn't implemented immediately. This PR implements those, and makes a few other changes in styles wherever improvements could be made.

Changes are documented in GH comments below.

## How to test this

1. Checkout this branch locally
2. Check both the mobile and desktop versions of the searchbar and test for bugs.

> Demo Note: The pages take time to build for me locally while I'm recording the screen.

chrome_i8qqoJAB64

---
  - https://app.asana.com/0/0/1201706447225517


site/src/components/Navbar.tsx
```diff
@@ -158,6 +165,10 @@ export const Navbar: VFC<NavbarProps> = ({
     }
   }, [isDesktopSize, displayMobileNav]);
 
+  const closeDrawer = useCallback(() => {
```
akash-joshi, line 168: Creates and passes down a function for closing the sidebar. This is used to close the sidebar when the user taps on a search result.

.github/workflows/algolia-upload-index.yml
```diff
@@ -3,6 +3,7 @@ name: Algolia Upload Index
 env:
   ALGOLIA_PROJECT: ${{ secrets.ALGOLIA_PROJECT }}
   AGOLIA_WRITE_KEY: ${{ secrets.AGOLIA_WRITE_KEY }}
+  ALGOLIA_INDEX: ${{ secrets.ALGOLIA_INDEX }}
```
akash-joshi, line 6: Read `ALGOLIA_INDEX` from secrets instead of hard-coding the value in the script.

site/src/components/Navbar.tsx
```diff
@@ -396,13 +407,20 @@ export const Navbar: VFC<NavbarProps> = ({
...
           }}
         >
-          <Box
-            sx={{
-              overflow: "auto",
-            }}
-          >
-            <MobileNavItems onClose={() => setDisplayMobileNav(false)} />
-          </Box>
+          {(router.asPath.includes("spec") ||
```
akash-joshi, line 410: Displays Search bar in mobile view only if the current view is spec or docs.

site/src/components/Navbar/MobileNavItems.tsx
```diff
@@ -270,6 +276,12 @@ export const MobileNavItems: VFC<MobileNavItemsProps> = ({ onClose }) => {
 
   return (
     <List>
+      {(asPath.includes("spec") || asPath.includes("docs")) && (
```
akash-joshi, line 279: Displays searchbar on mobile only if the path is in `spec` or `docs`

site/src/components/pages/docs/Search/index.tsx
```diff
@@ -115,11 +126,13 @@ export default function Search({ variant, closeDrawer }: SearchProps) {
     event.preventDefault();
     event.stopImmediatePropagation();
 
-    (document.querySelector(`.search-bar input`) as HTMLElement).focus();
-
-    const element = document.querySelector(".docs-list-sidebar");
+    const desktopSearchSelector = document.querySelector(
```
akash-joshi, line 129: Only focus on input if user is in desktop mode.

site/src/components/pages/docs/Search/index.tsx
```diff
@@ -156,11 +169,21 @@ export default function Search({ variant, closeDrawer }: SearchProps) {
     }
   };
 
+  const searchResultContainerStyles: SxProps<Theme> = {
```
akash-joshi, line 172: Styling related changes from here.

.github/workflows/algolia-upload-index.yml
CiaranMn: Superseded by #187

site/src/components/Navbar/MobileNavItems.tsx
CiaranMn: This check might lead to false positives for other pages that happen to include the letters "spec" or "docs" - could you make it so that isn't possible?

site/src/components/pages/docs/Search/index.tsx
```diff
@@ -20,8 +27,12 @@ import SearchItem, {
 } from "./SearchItem";
 import { Link } from "../../../Link";
 
+const isDev = process.env.NODE_ENV === "development";
+
 const client = algoliasearch("POOWZ64DSV", "96dc0442fd27b903440955dc03e5e60e");
-const index = client.initIndex("blockprotocol_testing");
+const index = client.initIndex(
+  isDev ? "blockprotocol_testing" : "blockprotocol",
+);
```
CiaranMn, line 35: Superseded by #187

site/src/components/Navbar.tsx
CiaranMn: I don't think you need this, as `onClose` already exists and does the same thing - you could pass that to `Search` from `MobileNavItems` instead.

CiaranMn (changes requested):
Thanks Akash - looks good and works well. A few bits of cleanup suggested.

CiaranMn (approved):

PR #178, opened 2022-02-02
Smoke-test npm packages

kachkaev:
This PR helps complete https://github.com/blockprotocol/blockprotocol/pull/168 and also simplifies further work on the npm packages. It introduces a simple smoke test for `npx create-block-app`, which includes `npm install`, `npm run dev`, `npm run lint:tsc` and `npm run build` inside a new block folder.

## In scope

- Two new scripts:

  - `scripts/packages/publish-to-local-registry.ts`
     It publishes `block-template`, `blockprotocol`, `create-block-app` and `mock-block-dock` to `http://localhost:4873`.  Registry address is hardcoded to save us from accidental publishing to the real registry (at least for now).

  - `scripts/packages/smoke-test-create-block-app.ts`
      It pretty much follows the instructions at https://blockprotocol.org/docs/developing-blocks. This script can work against both local and public registry, which means that we can also use it for e2e testing of just-published packages.

- New CI job which runs both scripts and then uploads the block folder as an artifact

- Small tweaks in packages (see diff comments)

I am using [Verdaccio](https://github.com/verdaccio/verdaccio) to spin up the local NPM registry. [CRA uses it too](https://github.com/facebook/create-react-app/search?q=Verdaccio) for a similar kind of smoke tests. I’ve chosen to run Verdaccio in a Docker container as it saves us from some extra setup and folder cleaning. The `docker run` command we have in the Yaml can be launched locally as is.

## Out of scope

- The scripts are not using `yarn create block-app`, `yarn install`, etc. It is unlikely that we’ll hit discrepancies between Yarn and NPM at this stage, so we can add Yarn to smoke-testing a bit later. We can also include different versions of Node, Yarn and NPM for some extra robustness.

- There are no assertions inside `smoke-test-create-block-app`. We can check block name and other metadata, but that’s not a priority right now. Just running lifecycle commands with exit code 0 is probably enough to start with. 

- I haven’t updated `CONTRIBUTING.md` so that I can publish this PR by the end of today. I can write something up in a follow-up PR to get this one in faster.

### DIFF
--- a/packages/block-template/webpack-dev-server.config.js
+++ b/packages/block-template/webpack-dev-server.config.js
@@ -20,7 +20,6 @@ module.exports = {
   devServer: {
     hot: true,
     contentBase: __dirname,
-    open: process.env.BROWSER !== "none",
     headers: {
       "Access-Control-Allow-Origin": "*",
       "Access-Control-Allow-Methods": "GET, POST, PUT, DELETE, PATCH, OPTIONS",
@@ -32,6 +31,7 @@ module.exports = {
         "sentry-trace",
       ],
     },
+    open: process.env.BROWSER !== "none",
   },
   resolve: {
     extensions: [
--- a/packages/block-template/webpack-dev-server.config.js
+++ b/packages/block-template/webpack-dev-server.config.js
@@ -20,6 +20,7 @@ module.exports = {
   devServer: {
     hot: true,
     contentBase: __dirname,
+    open: process.env.BROWSER !== "none",
     headers: {
       "Access-Control-Allow-Origin": "*",
       "Access-Control-Allow-Methods": "GET, POST, PUT, DELETE, PATCH, OPTIONS",
--- a/packages/create-block-app/create-block-app.js
+++ b/packages/create-block-app/create-block-app.js
@@ -7,6 +7,7 @@ const child_process = require("child_process");
 const fs = require("fs");
 const pacote = require("pacote");
 const path = require("path");
+const untildify = require("untildify");
 
 const exec = promisify(child_process.exec);
 
@@ -23,7 +24,7 @@ const templatePackageName = "block-template";
   // @todo: replace with `process.argv[3] ?? blockName` after upgrading ESLint
   const blockPath = process.argv[3] ? process.argv[3] : blockName;
 
-  const resolvedBlockPath = path.resolve(blockPath);
+  const resolvedBlockPath = path.resolve(untildify(blockPath));
 
   try {
     fs.statSync(resolvedBlockPath);
--- a/packages/create-block-app/create-block-app.js
+++ b/packages/create-block-app/create-block-app.js
@@ -72,6 +72,11 @@ const templatePackageName = "block-template";
 
   fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, undefined, 2));
 
+  console.log(`Testing testing (remove before merging)`);
+  console.log(`Testing testing (remove before merging)`);
+  console.log(`Testing testing (remove before merging)`);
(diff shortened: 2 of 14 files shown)


packages/block-template/webpack-dev-server.config.js
```diff
@@ -31,6 +31,7 @@ module.exports = {
         "sentry-trace",
       ],
     },
+    open: process.env.BROWSER !== "none",
```
kachkaev, line 34: Replaces `--open` in webpack CLI. `BROWSER=none` matches CRA behavior: https://create-react-app.dev/docs/advanced-configuration/ (I saw it somewhere else too).

I use this flag inside `smoke-test-create-block-app` script to avoid annoying window jumps.

packages/create-block-app/create-block-app.js
```diff
@@ -7,6 +7,7 @@ const child_process = require("child_process");
 const fs = require("fs");
 const pacote = require("pacote");
 const path = require("path");
+const untildify = require("untildify");
```
kachkaev, line 10: Gives us support for paths like `~/projects/blocks/hello`

kachkaev:
I tried new CI job in #168 – it does what it should 🕺

Screenshot 2022-02-02 at 23 15 48

scripts/packages/publish-to-local-registry.ts
```diff
@@ -0,0 +1,72 @@
...
+    NPM_CONFIG_REGISTRY: npmRegistry,
+    PATH: process.env.PATH,
+  },
+  extendEnv: false,
+  stdio: "inherit",
+} as const;
+
+const script = async () => {
+  logStepStart("Login into local registry");
+
```
CiaranMn, line 30: Could potentially bail here if `defaultExecaOptions.env.NPM_CONFIG_REGISTRY` doesn't contain `"localhost"`, and warn people that this script will _unpublish_ things too, just in case 😅  - I'm thinking of someone coming along and thinking 'oh a script to publish the packages, I'll change the config to publish for real'

CiaranMn (dismissed):
Great checks to have! 

One optional comment / bit of paranoia.

scripts/packages/publish-to-local-registry.ts
kachkaev: We use `NPM_CONFIG_REGISTRY: process.env.NPM_CONFIG_REGISTRY` in [scripts/packages/smoke-test-create-block-app.ts](https://github.com/blockprotocol/blockprotocol/pull/178/files#diff-633ed29d74de8a8abed1965dc8ffb2151fdece412f0678592e1b2534098c7856). In the initial version of this script we have a hardcoded value: `NPM_CONFIG_REGISTRY: npmRegistry`.

Does this address your concern?

CiaranMn: Not really - my concern is that the value in `defaultExecaOptions` are changed, and the script run. 

The function itself has nothing to prevent it being run using different options, which might access the real registry, and unpublish a package.

kachkaev: Oh IC. Right, I can change that. So we’ll have hardcoded `npmRegistry` plus early exit if the value is not `localhost:N`. Makes sense!

CiaranMn: It's protecting against a very unlikely event, but since there's a near-zero cost to doing so I'd feel better for it - thank you!

kachkaev: https://github.com/blockprotocol/blockprotocol/pull/178/commits/e43481817c1b5daff119526aba51c49c0c4934cc

CiaranMn (approved):

PR #179, opened 2022-02-02
Only scroll users to anchor when certain conditions are met

akash-joshi:
## What this Does

The current behavior for scroll when moving between the homepage, and the `/spec` page of the blockprotocol website results in the user being scrolled to the first anchor. This results in the user missing key information including headers and a GitHub callout.

This PR attempts to fix this issue by only allowing a frontend scroll when certain conditions are met.

Changes are described in GH comments below.

## How to Test this

1. Go to the homepage and navigate to the spec page. This shouldn't cause a scroll down.
2. Clicking on sidebar items should scroll as normal.
3. Opening a # URL in any tab should scroll to the anchor normally. 
4. On clicking a top-level heading of a different page from the sidebar, scroll to the top, as it already does in the website.

> Note, in the existing code, an empty hash `#` and no hash are treated the same.

---
  - https://app.asana.com/0/0/1201668052739232


site/src/pages/_app.page.tsx
```diff
@@ -58,10 +58,24 @@ const MyApp = ({
     void refetchUser();
   }, [refetchUser]);
 
+  const updatePreviousRoute = (url: string) => {
```
akash-joshi, line 61: Store the URL of the previous page on moving to a new route.

site/src/components/MdxPageContent.tsx
```diff
@@ -38,12 +38,30 @@ export const MdxPageContent: VFC<MdxPageContentProps> = ({
...
+
   useEffect(() => {
     const headingWithCurrentAnchor = headings.find(
       (heading) => heading.anchor === anchor,
     );
 
-    if (headingWithCurrentAnchor) {
+    const previousRoute = sessionStorage.getItem("previousRoute");
+
+    const shouldScrollToAnchor =
```
akash-joshi, line 50: Scroll to anchor only if the below conditions are met.

site/src/components/MdxPageContent.tsx
```diff
@@ -38,12 +38,30 @@ export const MdxPageContent: VFC<MdxPageContentProps> = ({
...
+      // if anchor is empty and we haven't scrolled, prevent it
+      (anchor === "" && scrolledOnce.current) ||
+      // if anchor is not empty, always allow scroll
+      anchor !== "" ||
+      // OR if previous path is either a docs or spec page
+      (previousRoute?.includes("/docs") && router.asPath?.includes("/docs")) ||
+      (previousRoute?.includes("/spec") && router.asPath?.includes("/spec"));
```
CiaranMn, line 57: I think you just need this condition and not the ref or anything else - what is the other stuff for?

CiaranMn (changes requested):
Thanks Akash - I think this needs tweaking/simplifying: the behavior on the preview deployment is inconsistent in that clicking on a new section (e.g. Block Types, Embedding application) will sometimes scroll right to the top of the page, sometimes to the header - we just want to scroll to the header if switching inside docs/spec routes - a suggestion/question below.

akash-joshi:
Hey @CiaranMn, the scrolling behavior one sees is consistent with the one in the blockprotocol website. Let me explain. If you're already on a spec/docs page:
1. Clicking on the header of the same page in the sidenav scrolls you to that anchor.
2. Clicking on the header of a different page in the sidenav scrolls you to the top.

As a part of this PR, I tried to be consistent with the existing behavior observed and didn't try to make additional changes to the behavior.

site/src/components/MdxPageContent.tsx
akash-joshi: The first condition that I've added here is to enable the behavior seen in point 1 of this comment: https://github.com/blockprotocol/blockprotocol/pull/179#issuecomment-1028647755

CiaranMn: Understood - we don't need to scroll to the top if switching inside /docs or inside /spec - there is no new information at the top.

CiaranMn (changes requested):

site/src/components/MdxPageContent.tsx
akash-joshi: @CiaranMn sorry but I don't think I've understood this comment well. Do you mean that this PR is good to go if I remove the 1st condition, ie, `(anchor === "" && scrolledOnce.current)` from this list?

CiaranMn: I mean we don't need the behavior you describe: "Clicking on the header of a different page in the sidenav scrolls you to the top." - clicking on the header of a different page in the sidenav should scroll you to the header.

We should only scroll to the very top if arriving at /docs from a non-docs page, or arriving at /spec from a non-spec page: not when moving within /docs or within /spec.

akash-joshi: Ahh okay, that will probably require more changes within the codebase. Will have a look.

CiaranMn: It's the same as the current behavior, except scrolling to the very top if moving within /docs or within /spec - I think you're almost there as you can tell if we're moving within /docs or within /spec

site/src/components/MdxPageContent.tsx
```diff
@@ -22,50 +22,71 @@ export const MdxPageContent: VFC<MdxPageContentProps> = ({
...
+  const currentHeading = useRef<Heading | undefined>(undefined);
 
   useEffect(() => {
-    setAnchor(router.asPath.split("#")[1] ?? "");
-  }, [router]);
+    setHeadings([]);
 
-  useEffect(() => {
-    const headingWithCurrentAnchor = headings.find(
-      (heading) => heading.anchor === anchor,
```
akash-joshi, line 43: The inconsistency in scrolling behavior was coming from this line. Since the anchor of the first element always remains the same (#), this would mean that the scroll is never triggered if the user is already at the first anchor of a different spec/docs page.

site/src/components/MdxPageContent.tsx
```diff
@@ -22,50 +22,71 @@ export const MdxPageContent: VFC<MdxPageContentProps> = ({
...
-  }, [serializedPage]);
-
-  const [currentHeading, setCurrentHeading] = useState<Heading>();
   const [detectHeadingFromScroll, setDetectHeadingFromScroll] =
     useState<boolean>(true);
 
-  const [anchor, setAnchor] = useState<string>(
-    router.asPath.split("#")[1] ?? "",
-  );
+  const currentHeading = useRef<Heading | undefined>(undefined);
```
akash-joshi, line 28: Converting `currentHeading` into a ref and removing the state hook for `anchor` helped as well, since they were causing more effect triggers than needed when updated.

site/src/components/MdxPageContent.tsx
```diff
@@ -22,50 +21,71 @@ export const MdxPageContent: VFC<MdxPageContentProps> = ({
...
+        // if anchor is empty and we haven't scrolled, prevent it
+        (anchor === "" && scrolledOnce.current) ||
+        // if anchor is not empty, always allow scroll
+        anchor !== "" ||
+        // OR if previous path is either a docs or spec page
+        (previousRoute?.includes("/docs") &&
+          router.asPath?.includes("/docs")) ||
+        (previousRoute?.includes("/spec") && router.asPath?.includes("/spec"));
+
+      // headings takes a while to load in, which can set this off
```
CiaranMn, line 60: This comment can be removed - it relates to an earlier implementation

site/src/components/MdxPageContent.tsx
```diff
@@ -22,50 +21,71 @@ export const MdxPageContent: VFC<MdxPageContentProps> = ({
...
+        scrolledOnce.current = true;
+      }
 
-        if (detectHeadingFromScrollTimer) {
-          clearTimeout(detectHeadingFromScrollTimer);
+      if (headingWithCurrentAnchor && shouldScrollToAnchor) {
+        if (
+          !currentHeading.current ||
+          headingWithCurrentAnchor.element !== currentHeading.current.element
+        ) {
```
CiaranMn, line 69: These conditions could be merged

CiaranMn (changes requested):
Thanks @akash-joshi - one more slight inconsistency in that if you click on another header without having scrolled, it will scroll to the top. e.g. click between various top-level headers in /spec without scrolling, it will jump between scrolling to the top, and scrolling to the header.

Can you investigate this please? Might be something to do with the move of a condition from state to a ref.

site/src/components/PageSidebar.tsx
```diff
@@ -187,11 +187,7 @@ const SidebarPage: VFC<SidebarPageProps> = ({
               setSelectedAnchorElement(ref);
             }
           }}
-          onClick={() => {
```
akash-joshi, line 190: The default behavior of `Link` components is to scroll to top: https://nextjs.org/docs/api-reference/next/link#disable-scrolling-to-the-top-of-the-page.

This prevents scroll to top if a user is in the docs or spec set of pages.

site/src/components/MdxPageContent.tsx
```diff
@@ -22,38 +21,59 @@ export const MdxPageContent: VFC<MdxPageContentProps> = ({
...
-    if (headingWithCurrentAnchor) {
       if (
-        !currentHeading ||
-        headingWithCurrentAnchor.anchor !== currentHeading.anchor
+        headingWithCurrentAnchor &&
+        shouldScrollToAnchor &&
+        (!currentHeading.current ||
+          headingWithCurrentAnchor.element !==
+            currentHeading.current.element) &&
+        document.body.contains(headingWithCurrentAnchor.element)
```
akash-joshi, line 70: This effect was getting stale `headings` data between renders. The condition of `contains` prevents the app from trying to scroll to a stale heading.

site/src/components/PageSidebar.tsx
akash-joshi: @CiaranMn I think this is the definite answer to the behavior we were observing here.

CiaranMn (approved):
Excellent, thank you 🥳

PR #173, opened 2022-02-01
Replace `postinstall` with `prepare` in `mock-block-dock`

kachkaev:
This PR should fix #161.

Unlike `postinstall`, `prepare` only runs in development. See https://docs.npmjs.com/cli/v8/using-npm/scripts.

Unfortunately, Yarn [does not automatically run `prepare` in workspaces](https://github.com/yarnpkg/yarn/issues/3911). We cannot use `yarn workspaces run prepare`, but that would require dummy `prepare` scripts in all packages. Alternatively, we can `yarn lerna run prepare`, but it [has not been maintained](https://github.com/lerna/lerna/issues/2703) for some time.

To keep this PR small, I am running `yarn workspace mock-block-dock run prepare` in top-level `package.json` → `prepare`. We can change the approach later, but it should be good enough for now 🙂

Depending on how our repo setup evolves, we might want to replace `prepare` with `prepublishOnly`. For example, if we use Turborepo, `yarn build` can be placed into a DAG before `yarn lint` etc. This will make `yarn install` a bit faster.

PR #184, opened 2022-02-02
Temporarily disable `post-commit` git hook

kachkaev:
We run `yarn install` after `git commit`, which is useful when cherry-picking. Our `prepare` script got beefier in #173, which made commits quite slow because of a git hook.

When we introduce a monorepo-management tool like Turborepo, we’ll be able to remove package builds from `prepare` and bring the hook back.

### DIFF
--- a/.husky/post-commit
+++ b/.husky/post-commit
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-yarn install
--- a/.husky/post-commit
+++ b/.husky/post-commit
@@ -1,7 +1,7 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-## This hook was temporary emptied in https://github.com/blockprotocol/blockprotocol/pull/184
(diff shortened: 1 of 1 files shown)


.husky/post-commit
```diff
@@ -1,4 +1,7 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn install
+## This hook was disabled in https://github.com/blockprotocol/blockprotocol/pull/184
+## @todo: Revert when whe cleanup `prepare` script in `package.json`
```
teenoh, line 5: small typo
```suggestion
## @todo: Revert when we cleanup `prepare` script in `package.json`
```

teenoh (approved):

.husky/post-commit
kachkaev: Thanks for spotting this typo @teenoh! I’ve added the fix to https://github.com/blockprotocol/blockprotocol/pull/178/files#diff-b6aebccfc7f539deea6c7023d96b327318a788d5193a403074d8e4568f753413

PR #186, opened 2022-02-03
Allow filtering blocks on more parameters

akash-joshi:
## What this Does

The existing blocks API allows users to get a list of all available blocks, and filter them based on:
1. `q`: Which filters by the query string on a set of blockData keys.
2. `json`: Which filters based on the structure of json passed in by the user.

This PR allows filtering based on 3 more data points:
1. `author`: Filter blocks based on block author.
2. `name`: Filter blocks based on either `displayName` or the package name.
3. `license`: Filter blocks based on the type of license that they use.

## How to test this

- Checkout the branch
- Call the API locally, mix and match parameters to see how they combine.

> Note: The Asana task mentions that "Query parameters should be composable / combinable". This PR assumes that all filters will be ANDed, and hence, the order of query parameters doesn't matter. 

---
  - https://app.asana.com/0/0/1201472294283824

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -281,8 +281,13 @@ The following routes require a valid API key sent in an `x-api-key` header:
 
 - Request Params
 
-  - `q`: an optional text query to search for blocks with a matching name or author. If not provided, all blocks are returned.
+  - `author`: an optional text query to search for blocks with a matching author.
+  - `license`: an optional text query to search for blocks with a matching license. Note: The license name needs to be matching.
+  - `name`: an optional text query to search for blocks with a matching name.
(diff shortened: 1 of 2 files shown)


CiaranMn (changes requested):
Thanks Akash - can you update the description of this endpoint in `site/README.md` with any necessary changes please?

akash-joshi:
Have made changes to the README. This bit "If all of the optional params above are **not** provided, the result will contain all of the blocks." could probably have better wording but I'm not able to figure out how.

CiaranMn (approved):

PR #188, opened 2022-02-03
Better handle block hub JSON files. Fix switching between blocks.

CiaranMn:
This PR fixes a couple of assumptions we made about the JSON files describing blocks:
1. Better handles a missing `timestamp` 
2. Allows repo URLs that don't end in .git

It also reverts a change made in #171 to change the `entityId` of a block - `MockBlockDock` doesn't handle this properly and more changes are needed there to prevent crashes if the `entityId` of the block entity is changed from outside.

### DIFF
--- a/site/src/components/pages/hub/BlockDataContainer.tsx
+++ b/site/src/components/pages/hub/BlockDataContainer.tsx
@@ -79,7 +79,7 @@ export const BlockDataContainer: VoidFunctionComponent<
       );
 
     return [result, errorMessages];
-  }, [text, schema, metadata.name]);
+  }, [text, schema]);
 
   return (
     <>
--- a/site/src/components/pages/hub/BlockDataContainer.tsx
+++ b/site/src/components/pages/hub/BlockDataContainer.tsx
@@ -58,8 +58,8 @@ export const BlockDataContainer: VoidFunctionComponent<
   /** used to recompute props and errors on dep changes (caching has no benefit here) */
   const [props, errors] = useMemo<[object | undefined, string[]]>(() => {
     const result = {
-      accountId: `test-account-${metadata.name}`,
-      entityId: `test-entity-${metadata.name}`,
+      accountId: `test-account-1`,
+      entityId: `test-entity-1`,
       getEmbedBlock,
     };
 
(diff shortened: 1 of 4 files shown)


site/src/lib/blocks.ts
```diff
@@ -58,9 +58,9 @@ export const readBlocksFromDisk = (): ExpandedBlockMetadata[] => {
       author: metadata.packagePath.split("/")[0].replace(/^@/, ""),
       icon: getBlockMediaUrl(metadata.icon, metadata.packagePath),
       image: getBlockMediaUrl(metadata.image, metadata.packagePath),
-      lastUpdated: buildConfig.find(
-        ({ workspace }) => workspace === metadata.name,
-      )?.timestamp,
+      lastUpdated:
+        buildConfig.find(({ workspace }) => workspace === metadata.name)
+          ?.timestamp ?? null,
```
CiaranMn, line 63: NextJS cannot serialize `undefined` returned from `getStaticProps`, and this is called from one

site/src/pages/[shortname]/[blockSlug].page.tsx
```diff
@@ -256,23 +256,23 @@ const BlockPage: NextPage<BlockPageProps> = ({
...
-                {`Updated ${
-                  blockMetadata.lastUpdated
-                    ? formatDistance(
-                        new Date(blockMetadata.lastUpdated),
-                        new Date(),
-                        {
-                          addSuffix: true,
-                        },
-                      )
-                    : "Recently"
```
Comment (CiaranMn) [inline site/src/pages/[shortname]/[blockSlug].page.tsx:273]: "Updated Recently" in the absence of a timestamp is not necessarily true

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -58,8 +58,8 @@ export const BlockDataContainer: VoidFunctionComponent<
   /** used to recompute props and errors on dep changes (caching has no benefit here) */
   const [props, errors] = useMemo<[object | undefined, string[]]>(() => {
     const result = {
-      accountId: `test-account-${metadata.name}`,
-      entityId: `test-entity-${metadata.name}`,
+      accountId: `test-account-1`,
```
CiaranMn, line 61: Reverts changes made in #171

nathggns (approved):
Code looks sensible, but I don't have an instance of this running to actually test this functionality wise.

PR #189, opened 2022-02-03
Compose some site scripts to reduce complexity of `package.json`

kachkaev:


scripts/shared/logging.ts
```diff
@@ -1,4 +1,4 @@
-let currentMarkEndOfPhase: (() => void) | undefined = undefined;
+let currentMarkEndOfStep: (() => void) | undefined = undefined;
```
kachkaev, line 1: I renamed `phase` to `step` in #178, but forgot to change this local var

site/package.json
```diff
@@ -93,6 +93,8 @@
     "@types/react-slick": "^0.23.7",
     "@types/uuid": "^8.3.4",
     "babel-plugin-inline-react-svg": "2.0.1",
+    "chalk": "^4.1.2",
+    "execa": "^5.1.1",
```
kachkaev, line 97: We need to switch to `"type": "module"` in `package.json` to upgrade to chalk 5 and execa 6 (they are ESM-only). We can do this when https://github.com/vercel/next.js/pull/33637 is merged and we upgrade Next.js to latest.

site/scripts/dev.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  await (
+    await import("./generate-blocks-data")
+  ).default;
+
+  await (
+    await import("./create-db-indexes")
+  ).default;
+
+  await execa("next", ["dev"], { stdio: "inherit" });
```
thehabbos007, line 19: This might not be an actual problem, but it seems that `execa` buffers `stdout` by default [seen here](https://github.com/sindresorhus/execa#maxbuffer). I am wondering if we are going to hit that limit realistically? Probably not, but worth keeping in mind.

site/scripts/generate-sitemap.ts
```diff
@@ -1,12 +1,19 @@
...
-console.log("✅ Generated site map");
+  writeFileSync(
+    path.join(process.cwd(), `site-map.json`),
+    JSON.stringify(sitemap, null, "\t"),
+  );
+
+  console.log("✅ Site map generated");
+};
+
+export default script();
```
benwerner01, line 19: To help enforce constancy between the logging messages's formatting using `chalk` and use of emojis, what if we create a wrapper function like this:
```TypeScript
const runScript = async (
  script: () => Promise<void>,
  config: { startMsg: string; successMsg: string; errorMsg: string; }
) => {
  try {
    console.log(chalk.bold(`${config.startMsg}...`));
    await script();
    console.log(`✅ ${config.successMsg}`);
  } catch (error) {
    console.log(`❌ ${config.errorMsg}: `, error)
  }
}
```
(this could also be simplified to just accepting a `scriptName: string` argument, and then logging `Running script ${scriptName}...`, `✅ Completed script "${scriptName}"`, etc)

This would let developers focus on writing the functionality of the scripts:

```TypeScript
...
const script = async () => {
  const sitemap = generateSiteMap();

  writeFileSync(
    path.join(process.cwd(), `site-map.json`),
    JSON.stringify(sitemap, null, "\t"),
  );
};

export default runScript(script, {
  startMsg: "Generating sitemap",
  successMsg: "Site map generated",
  errorMsg: "Could not generate sitemap"
});
```

site/scripts/dev.ts
kachkaev: Good point! It’s 100MB by default though, so I hope it will be enough 😅

Alternatively, we can launch Next.js server here directly, i.e. in the same process. That would mean that we switched to a [custom server](https://nextjs.org/docs/advanced-features/custom-server) though – not sure if it’d be a good thing to do.

thehabbos007: Wonder what will happen at the buffer limit. I assume the buffer simply wraps around and still outputs to `stdout`, so perhaps it's a non-issue

site/scripts/generate-sitemap.ts
kachkaev: Thanks Ben! I tried creating similar abstractions for scripts in some side projects, but found them more costly than beneficial. Imagine we have a script that generates a file and let’s say we don’t want to override the result unless `FORCE=true` is provided. So our messages will look something like:

```
[bold] Generating our file...
[magenta] File /x/y/z generated.
```

```
[bold] Generating our file...
[gray] File /x/y/z already exists use FORCE=true to overwrite it.
```

```
[bold] Generating our file (with FORCE=true)...
[magenta] File x/y/z generated.
```

As you can see, `startMsg` and `successMsg` vary. File path (`/x/y/z`) can vary too if it’s in some configurable `DATA_DIR_PATH` directory. This is just one possible scenario of what a script might want to do – the reality will be full of diversity.

My current thinking is that just using standard logging functions plus `chalk` is simple, flexible and solid. There is not much added repetition and the cost of forgetting to log something is quite small. On the other hand, we don’t need to learn how to use a new wrapper abstraction or add features to it.

Here are some example scripts I wrote for a side project: https://github.com/kachkaev/tooling-for-how-old-is-this-house/tree/4e1c420a88fb6c3d75fef17729eeed9c720fadc3/src/scripts. There are tens of them, all ‘wrapper-free’.

The only abstraction I find useful these days is [`ScriptError`](https://github.com/kachkaev/tooling-for-how-old-is-this-house/blob/76bb8ca0d867472757a355695cdfd629b37e84f2/src/shared/helpersForScripts.ts#L182-L193). It is 100% like a normal `Error`, but if a script throws it all the way up, Node does not print the call stack. It’s an alternative to writing this, but with a more familiar syntax.

```ts
console.log(chalk.red("Some friendly message explaining why script failed"));
process.exit(42);

// ↓

throw new ScriptError("Some friendly message explaining why script failed");
```

WDYT?

benwerner01: Thanks Alex, that makes sense to me - you're right we shouldn't sacrifice flexibility when abstracting how we log messages in the scripts.

One way to avoid developers from having to worry about usage of `chalk` and emojis could be an abstraction such as `ScriptError` as you suggest, but I don't feel strongly either-way about this. I'm happy for this to be left as is.

thehabbos007 (changes requested):
Nice to have the scripts be orders in code instead of `package.json`.
I hit a snag with the `seed-db.ts` script. It seems that since we are using this connection helper https://github.com/blockprotocol/blockprotocol/blob/f8de754ae20329586a6767e0122d5b632e467f43/site/src/lib/mongodb.ts#L7-L13 there is an implicit cached client which is invalidated when `.close()` is called on the client. This means composing scripts will lead to errors like that. 

Perhaps we can introduce a `noCache` argument or invalidate the cached client somehow.

package.json
```diff
@@ -39,6 +39,7 @@
     "@typescript-eslint/eslint-plugin": "4.29.0",
     "@typescript-eslint/parser": "4.29.0",
     "algoliasearch": "^4.12.0",
+    "chalk": "^4.1.2",
```
benwerner01, line 42: Is there a reason `chalk` is listed in both `package.json` files?

kachkaev:
Great spot @thehabbos007! I hope I have fixed this in https://github.com/blockprotocol/blockprotocol/pull/189/commits/a3c3a86b77307bc5f6fbc64be64b604f858d98d9

PR #192, opened 2022-02-03
Add `envalid` to improve script feedback when env vars are missing

kachkaev:
We were having issues with [Algolia indexing](https://github.com/blockprotocol/blockprotocol/actions/workflows/algolia-upload-index.yml) recently and it was unclear whether it was because of missing or incorrect env vars. This PR introduces [`envalid`](https://www.npmjs.com/package/envalid?activeTab=versions), which improves script feedback. We can add this library to other scripts as needed.

## Before 🤔

Screenshot 2022-02-03 at 19 33 49

## After 💡

Screenshot 2022-02-03 at 19 33 02

## Feature summary

```ts
const env = envalid.cleanEnv({
  EMAIL: envalid.email({ default: "hello@example.com" }),
  EMAIL_TRANSPORT: envalid.str({ choices: ["aws", "none"], default: "aws", devDefault: "none" }),
  PORT: envalid.port({ default: 3000 }),
});
```

```ts
console.log(typeof env.PORT === 3000); // true

// with PORT=-42 → 💥

console.log(env.EMAIL); // always a valid email

console.log(env.EMAIL_TRANSPORT); // "none"
console.log(env.isProduction); // false

// with NODE_ENV=production
console.log(env.EMAIL_TRANSPORT); // "aws"
console.log(env.isProduction); // true

// with EMAIL_TRANSPORT=oops → 💥
```

## Caveats

- By default, `envalid` calls `process.exit` instead of throwing. This makes script output cleaner because it does not contain the call stack. We can’t use `try / catch` though. If we do want to use `try / catch`, we can implement a [custom reporter](https://github.com/af/envalid#error-reporting), which would throw. This may be useful in server-side code, but I doubt about scripts.

- We need to use `import * as envalid from "envalid"`. Writing `import envalid from "envalid"` does not produce a TS error but crashes. This may change if we switch to native ESM modules.

- Specifying `choices` does not narrow the return type (i.e. `env.EMAIL_TRANSPORT` is `string`, not `"aws" | "none"`). This might be fixed via https://github.com/af/envalid/pull/176

---

[Asana task](https://app.asana.com/0/0/1201766858773607/f) _(internal link)_

### DIFF
--- a/scripts/sync-algolia-index.ts
+++ b/scripts/sync-algolia-index.ts
@@ -2,6 +2,7 @@ import fs from "fs";
 import path from "path";
 import matter from "gray-matter";
 import algoliasearch from "algoliasearch";
+import * as envalid from "envalid";
 
 import siteMap from "../site/site-map.json";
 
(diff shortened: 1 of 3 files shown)


CiaranMn (approved):

scripts/sync-algolia-index.ts
```diff
@@ -91,8 +94,16 @@ const generateAlgoliaRecords: () => AlgoliaRecord[] = () => {
...
+  const specFiles = getFileInfos(
+    path.resolve(monorepoDirPath, "site/src/_pages/spec"),
+    [],
+    "spec",
+  );
+  const docsFiles = getFileInfos(
+    path.resolve(monorepoDirPath, "site/src/_pages/docs"),
+    [],
+    "docs",
+  );
```
kachkaev, line 106: @CiaranMn also fixed paths, which broke when I moved this script:
https://github.com/blockprotocol/blockprotocol/runs/5064081770?check_suite_focus=true

CiaranMn (approved):

PR #202, opened 2022-02-08
Display Block Variants in Block Protocol

akash-joshi:
> Previous PR: https://github.com/hashintel/hash/pull/279

## What this Does

This PR adds a UI for displaying block variants and switching between them in the demo page for blocks on the blockprotocol website. This allows users to switch between variants of blocks, and check their differences, along with examples provided where available. At the time of this PR, only the embed and header blocks have variants associated with them.

Changes described in GH Comments below.

## How to test this

- Visit either the embed or header block
- Switch between variants and check if the data source and block have updated.

chrome_sNBuvIaqCV
chrome_zW8CMaqgLN

---
  - https://app.asana.com/0/0/1201540186206729


packages/blockprotocol/core.d.ts
```diff
@@ -9,6 +9,7 @@ type DistributedOmit<T, K extends PropertyKey> = T extends T
 export type BlockVariant = {
   description?: string | null;
   displayName?: string | null;
+  examples?: JSONObject[] | null;
```
akash-joshi, line 12: Update the variants type with examples array.

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -29,15 +37,23 @@ export const BlockDataContainer: VoidFunctionComponent<
   BlockDataContainerProps
 > = ({ metadata, schema, blockModule }) => {
   const [blockDataTab, setBlockDataTab] = useState(0);
-  const [blockTab, setBlockTab] = useState(0);
+  const [blockVariantsTab, setBlockVariantsTab] = useState(0);
   const [blockModalOpen, setBlockModalOpen] = useState(false);
+  const [alertSnackBarOpen, setAlertSnackBarOpen] = useState(false);
 
-  const example = metadata.examples?.[0];
+  const example = {
```
akash-joshi, line 44: Form the initial example as a combination of the default example, the example of the first variant, and the variant's properties.

The priority of assignment is in ascending order, ie, downwards.

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -55,6 +71,46 @@ export const BlockDataContainer: VoidFunctionComponent<
...
+        setAlertSnackBarOpen(true);
+        setBlockVariantsTab(previousBlockVariantsTab.current);
+      }
+    }
+
+    return () => {
+      if (metadata.variants) {
+        const blockVariant = metadata.variants[blockVariantsTab];
+
+        propertiesToRemove.current = Object.keys({
```
akash-joshi, line 106: Create a list of properties which need to be removed between variant switches.

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -55,6 +71,46 @@ export const BlockDataContainer: VoidFunctionComponent<
     prevPackage.current = metadata.packagePath;
   }, [example, metadata.packagePath, text]);
 
+  useEffect(() => {
+    if (
+      metadata.variants &&
+      previousBlockVariantsTab.current !== blockVariantsTab
+    ) {
```
akash-joshi, line 78: If the user switches variants, delete old properties from the parsed text, append the new example and properties, and set it to the `text` state.

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -112,7 +168,7 @@ export const BlockDataContainer: VoidFunctionComponent<
               marginLeft: "4px",
               height: "unset",
               boxShadow: theme.shadows[1],
-              width: "48% !important",
```
akash-joshi, line 115: This broke the mobile tabs UI, so I fixed it.

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -264,6 +307,20 @@ export const BlockDataContainer: VoidFunctionComponent<
...
+            open={alertSnackBarOpen}
+            autoHideDuration={4000}
+            anchorOrigin={{
+              horizontal: "right",
+              vertical: "bottom",
+            }}
+            onClose={() => setAlertSnackBarOpen(false)}
+          >
+            <Alert severity="warning">
+              Please fix the errors in the <b>Data Source</b> before proceeding.{" "}
```
akash-joshi, line 321: Display an alert to fix block errors in case the user tries to switch variants with a broken data source.

This is because we have to parse the existing text while applying the new properties of a variant to the `text` state.

site/src/components/pages/hub/BlockDataTabs.tsx
```diff
@@ -12,14 +19,26 @@ export const BlockDataTabs: VoidFunctionComponent<BlockDataTabsProps> = ({
   setBlockDataTab,
   modalOpen,
 }) => {
-  return (
+  const theme = useTheme();
+  const isMobile = useMediaQuery(theme.breakpoints.down("md"));
+
+  return isMobile ? (
+    <Select
```
akash-joshi, line 26: Create a custom select on mobile to match the design.

site/src/components/pages/hub/BlockDisplayTabs.tsx
```diff
@@ -0,0 +1,82 @@
...
+export const BlockDisplayTabs: VoidFunctionComponent<BlockDisplayTabsProps> = ({
+  blockTab,
+  metadata,
+  setBlockTab,
+}) => {
+  const theme = useTheme();
+  const isMobile = useMediaQuery(theme.breakpoints.down("md"));
+
+  return isMobile ? (
+    <Select
```
akash-joshi, line 27: Create a custom select on mobile to match the design.

site/src/components/pages/hub/BlockVariantsTabs.tsx
```diff
@@ -0,0 +1,80 @@
...
+};
+
+export const BlockVariantsTabs: VoidFunctionComponent<
+  BlockVariantsTabsProps
+> = ({ blockVariantsTab, metadata, setBlockVariantsTab }) => {
+  const theme = useTheme();
+  const isMobile = useMediaQuery(theme.breakpoints.down("md"));
+
+  return isMobile ? (
+    <Select
```
akash-joshi, line 25: Display a block's variants in a `Select` for mobile and in tabs for desktop.

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -29,31 +38,81 @@ export const BlockDataContainer: VoidFunctionComponent<
...
+        for (const propertyToRemove of propertiesToRemove.current) {
+          delete parsedText[propertyToRemove];
+        }
+
+        const nextText = {
+          ...blockVariant?.properties,
+          ...blockVariant.examples?.[0],
+          ...parsedText,
+        };
+
```
teenoh, line 95: I think we'd achieve the same goal if we do

```suggestion
        const nextText = {
            ...parsedText,
          ...blockVariant?.properties,
          ...blockVariant.examples?.[0],
        };
```

The necessary properties in `parsedText` will be overridden

akash-joshi: I guess this deserves a comment in the code. Some of the variants are also different by having a missing key in the `properties` field. Eg, all but the default variant in embed block have an `embedType` key. `embedType` prevents all embeds but the type specified in its value.

In cases like these, we would want the old keys to be removed. This is true for both the `examples` and `properties` field.

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -29,31 +38,81 @@ export const BlockDataContainer: VoidFunctionComponent<
...
+  }, [metadata, text]);
+
+  useEffect(() => {
+    if (
+      metadata?.variants &&
+      previousBlockVariantsTab.current !== blockVariantsTab
+    ) {
+      const blockVariant: BlockVariant | undefined =
+        metadata?.variants[blockVariantsTab];
+
```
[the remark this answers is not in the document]
akash-joshi: Fair assessment. Have moved the declaration out of the condition to simplify the code in both instances: https://github.com/blockprotocol/blockprotocol/pull/202/commits/15cf78116a023f6cee487dcb1f801c2a8d905d26

packages/blockprotocol/core.d.ts
kachkaev: I guess it’s fine to change the spec here, especially in a non-breaking way. We are still pre-0.1.0, so we don’t have to be too strict. We can release a new package version after making a few small batch updates like that.

site/src/components/pages/hub/BlockDataContainer.tsx
teenoh: In that case, then we probably don't need parsedText when switching between variants

```
variantA = { embedType: "twitter", url: "https://twitter.com/..."}
variantB = { embedType: "youtube", url: "https://youtube.com/..."}
```
Since both variants are mutually exclusive, we can just do

```
nextText = variantA 
nextText = variantB
```

instead of 
```
nextText = {
  ...variantA,
  ...variantB
}
```

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -29,31 +38,78 @@ export const BlockDataContainer: VoidFunctionComponent<
...
+  const [alertSnackBarOpen, setAlertSnackBarOpen] = useState(false);
 
-  const example = metadata.examples?.[0];
+  const [text, setText] = useState("{}");
 
-  const [text, setText] = useState(
-    example ? JSON.stringify(example, undefined, 2) : "",
-  );
+  const previousBlockVariantsTab = useRef(-1);
+  const propertiesToRemove = useRef<string[]>([]);
```
teenoh, line 48: I think we can implement the intended feature without this ref

CiaranMn:
@akash-joshi what's going on with this one please? can you re-request reviewers when you've addressed any feedback you intend to and resolved conflicts

site/src/components/pages/hub/BlockDataContainer.tsx
akash-joshi: Thanks for adding this comment Teenoh. @CiaranMn this brings up an interesting point of discussion in this PR. Do we want the custom attributes added by the user to persist between variant switches?

If we only switch to the default example for the variant while ignoring user input, that simplifies the code significantly, but at the cost of losing a user's custom attributes at each variant switch.

site/src/components/pages/hub/BlockDataContainer.tsx
akash-joshi: `previousBlockVariantsTab` is required, but the requirement of `propertiesToRemove` depends on the resolution for this comment: https://github.com/blockprotocol/blockprotocol/pull/202#discussion_r810792505.

CC: @CiaranMn

site/src/components/pages/hub/BlockDataContainer.tsx
CiaranMn: I don't think it's super important either way for the Block Hub demo. 

It's nice to keep the user's custom input, so long as the properties from the new variant consistently overwrite them (not the example - the variant properties itself). 

I'm not sure that the current code reliably does that - the only properties deleted from user input are from the _previous_ variant. If the next variant has some other property set, and the user has supplied their own value for it, the user's value will take precedence (since the parsed text is the last thing spread).

akash-joshi: Fixed this [here](https://github.com/blockprotocol/blockprotocol/pull/202/files#diff-aa262447e474cd1fbf253d93d617df697f68d9ccf4af71e9a5e1aa766bd9cd19R88) as discussed yesterday.

CiaranMn (changes requested):
I've tested this and noticed one thing needing fixing - if you select a variant, switch block, and then switch back to the original block, the same variant tab will be selected but the text will be the default example. The state needs to be reset when switching blocks.

e.g.
1. Visit Embed block: https://blockprotocol-i19jpt7qa-hashintel.vercel.app/@hash/embed
2. Switch to 'Spotify'
3. Select any other block from the carousel below
4. Switch back to Embed
5. 'Spotify' is selected, but the default (non-Spotify) example is shown

akash-joshi:
@CiaranMn the latest commit (https://github.com/blockprotocol/blockprotocol/pull/202/commits/80c6b0a8075282bd8ef8d048a99da4f518f72d08) should fix this.

CiaranMn (approved):

PR #203, opened 2022-02-08
BP site Material UI upgrade and theme file refactor

benwerner01:
This PR addresses several pieces of tech debt related to the Material UI setup of the BP Site. Changes introduced as part of this PR include:
- upgrading MUI from version `5.2.1` to `5.4.0`
- splitting up the `theme.ts` file into multiple files to improve the readability of the MUI theme declaration and component style overrides
- moving the theme declaration from the `site/src/components` directory one level up to the `site/src` directory
- deprecating the use of the `variants` field to override the styling of a component, which has now been converted to make use of the `styleOverrides` field instead.
  - For more information on this transition [see this MUI GitHub issue](https://github.com/mui/material-ui/issues/30412)
  - For more information on the thinking behind and use of the `styleOverrides` field to override styling of a component, see [this MUI blog post](https://mui.com/blog/callback-support-in-style-overrides/#using-callback-in-styleoverrides)

Note: this PR does not seek to modify any styling of the BP Site

---
  - https://app.asana.com/0/0/1201781684170579
  - https://app.asana.com/0/0/1201531481067019


site/src/theme/palette.ts
```diff
@@ -0,0 +1,64 @@
...
+    40: "#C5D1DB",
+    50: "#9EACBA",
+    60: "#64778C",
+    70: "#4D5C6C",
+    80: "#37434F",
+  },
+  grey: undefined,
+  black: "#0E1114",
+  white: "#FFFFFF",
+};
```
kachkaev, line 57:
```suggestion
} as const;
```

(should make the typing more narrow)

site/src/theme/shadows.ts
```diff
@@ -0,0 +1,12 @@
...
+/** @todo see if it's possible to use keys "sm" | "lg" instead of the array index */
+
+export const shadows: ThemeOptions["shadows"] = [
+  "none",
+  "0px 2px 8px rgba(39, 50, 86, 0.04), 0px 2.59259px 6.44213px rgba(39, 50, 86, 0.06), 0px 0.5px 1px rgba(39, 50, 86, 0.10)",
+  "0px 9px 26px rgba(61, 78, 133, 0.03), 0px 7.12963px 18.37px rgba(61, 78, 133, 0.04), 0px 4.23704px 8.1px rgba(61, 78, 133, 0.05), 0px 0.203704px 0.62963px rgba(61, 78, 133, 0.06)",
+  "0px 20px 41px rgba(61, 78, 133, 0.04), 0px 16px 25px rgba(61, 78, 133, 0.03), 0px 12px 12px rgba(61, 78, 133, 0.02), 0px 2px 3.13px rgba(61, 78, 133, 0.01)",
+  "0px 51px 87px rgba(50, 65, 111, 0.07), 0px 33.0556px 50.9514px rgba(50, 65, 111, 0.0531481), 0px 19.6444px 27.7111px rgba(50, 65, 111, 0.0425185), 0px 10.2px 14.1375px rgba(50, 65, 111, 0.035), 0px 4.15556px 7.08889px rgba(50, 65, 111, 0.0274815), 0px 0.944444px 3.42361px rgba(50, 65, 111, 0.0168519)",
+  "0px 96px 129px rgba(61, 78, 133, 0.13), 0px 48.6px 56.2359px rgba(61, 78, 133, 0.08775), 0px 19.2px 20.9625px rgba(61, 78, 133, 0.065), 0px 4.2px 7.45781px rgba(61, 78, 133, 0.04225)",
+] as any;
```
kachkaev, line 12: Why do we need to use `as any` here? This is potentially dangerous, especially if there are breaking changes in MUI later on. If we can’t add a proper typing, it is a bit safer to use `// @ts-expect-error -- explanation` instead.

site/src/theme/typography.ts
```diff
@@ -0,0 +1,90 @@
+import { ThemeOptions } from "@mui/material";
+import { customColors } from "./palette";
+import { defaultTheme } from "./util";
+
+const FALLBACK_FONTS = [`"Helvetica"`, `"Arial"`, "sans-serif"];
```
kachkaev, line 5: This is just an opinion but, I’d just use 

```suggestion
const fallbackFonts = [`"Helvetica"`, `"Arial"`, "sans-serif"];
```

here. If there is no difference between this const and the next one, why use different notations? This is something a newjoiner in JS would 100% ask and I don’t think we’d be able to provide a good answer. It’s better to save time on this and talk about something more interesting instead 😁

Personally, I’m a fan of banning `UPPER_CASE` variable names in codebases via [`@typescript-eslint/naming-convention`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/naming-convention.md). But this is an optional suggestion because we haven’t agreed on this stuff 🙂

site/src/theme/components/index.ts
```diff
@@ -0,0 +1,48 @@
...
+  MuiCheckbox,
+  MuiTabs,
+  MuiTab,
+  MuiBreadcrumbs,
+  MuiFormControl,
+  MuiInputLabel,
+  MuiInputBase,
+  MuiOutlinedInput,
+  MuiTextField,
+};
```
kachkaev, line 48: I wonder if we could do:

```suggestion
export const componentThemeOptions: ThemeOptions["components"] = {
  MuiCssBaseline: MuiCssBaselineThemeOptions,
  MuiTypography: MuiTypographyThemeOptions,
  // ...
}
```

A symbol called `MuiCssBaseline` placed into `compoents/MuiCssBaseline` looks too much like a component, which might become a mini-trap for some code autocompletion tasks. Using `import { MuiTypographyThemeOptions } from "./MuiTypographyThemeOptions"` / `import { MuiTypographyThemeOptions } from "./mui-typography-theme-options"` is heavier, but less ambiguous to a reader.

Either way, it’d be also nice to sort symbols alphabetically (there is a VS Code command for that).

site/src/theme/components/index.ts
```diff
@@ -0,0 +1,48 @@
...
+import { MuiPaper } from "./MuiPaper";
+import { MuiButton } from "./MuiButton";
+import { MuiContainer } from "./MuiContainer";
+import { MuiIcon } from "./MuiIcon";
+import { MuiSkeleton } from "./MuiSkeleton";
+import {
+  MuiListItemButton,
+  MuiListItemIcon,
+  MuiListItemText,
+} from "./MuiListItem";
```
kachkaev, line 15: We’ve grouped a few things in `MuiListItem.ts` and `MuiTabs` and these are the only exceptions at the moment. Perhaps, we should split these files or group other files too? If we need a principle for grouping, we can use official MUI component categories: https://mui.com/components/

kachkaev (approved):
Thanks for the PR Ben, especially for some useful links in the description! I clicked around the deployment and could not find any regressions so far – great stuff! 

Here are some optional suggestions I’ve got:

site/src/theme/components/index.ts
benwerner01: My thinking behind these groupings was that these are styling overrides for components that are often used together. But I realise this is hard to explain and would easily appear as exceptions. I'll split these into separate files.

Regarding your suggestion of using the categories, I like the idea but fear this will result in rather large files (once we start implementing more custom input designs for instance). Perhaps what we can do instead is organise the component overrides by these groups using directories instead, i.e. we would have `src/theme/components/DataDisplays/MuiListItemButtonThemeOptions.ts`.

What do you think?

site/src/theme/typography.ts
benwerner01: Don't have strong feelings on this, have changed it to `fallbackFonts` here https://github.com/blockprotocol/blockprotocol/pull/203/commits/d8976f9c232272a858d6ab87f6b2b5520a855853

site/src/theme/palette.ts
benwerner01: Good spot, addressed in https://github.com/blockprotocol/blockprotocol/pull/203/commits/08703bdefc0a1e855764cea9ceb90b301a82a47e

site/src/theme/shadows.ts
kachkaev: This worked for me locally:

```ts
export const shadows: ThemeOptions["shadows"] = [
  "none",
  "0px 2px 8px rgba(39, 50, 86, 0.04), 0px 2.59259px 6.44213px rgba(39, 50, 86, 0.06), 0px 0.5px 1px rgba(39, 50, 86, 0.10)",
  "0px 9px 26px rgba(61, 78, 133, 0.03), 0px 7.12963px 18.37px rgba(61, 78, 133, 0.04), 0px 4.23704px 8.1px rgba(61, 78, 133, 0.05), 0px 0.203704px 0.62963px rgba(61, 78, 133, 0.06)",
  "0px 20px 41px rgba(61, 78, 133, 0.04), 0px 16px 25px rgba(61, 78, 133, 0.03), 0px 12px 12px rgba(61, 78, 133, 0.02), 0px 2px 3.13px rgba(61, 78, 133, 0.01)",
  "0px 51px 87px rgba(50, 65, 111, 0.07), 0px 33.0556px 50.9514px rgba(50, 65, 111, 0.0531481), 0px 19.6444px 27.7111px rgba(50, 65, 111, 0.0425185), 0px 10.2px 14.1375px rgba(50, 65, 111, 0.035), 0px 4.15556px 7.08889px rgba(50, 65, 111, 0.0274815), 0px 0.944444px 3.42361px rgba(50, 65, 111, 0.0168519)",
  "0px 96px 129px rgba(61, 78, 133, 0.13), 0px 48.6px 56.2359px rgba(61, 78, 133, 0.08775), 0px 19.2px 20.9625px rgba(61, 78, 133, 0.065), 0px 4.2px 7.45781px rgba(61, 78, 133, 0.04225)",
  ...defaultTheme.shadows.slice(6),
] as ThemeOptions["shadows"]; // MUI expects to have exactly 26 shadows [... more explanation text ...]
```

Using `as` makes the compiler a bit less strict, so the chance of runtime errors increases a bit. I guess it’s fine to use typecasting here, but worth adding a comment.

site/src/theme/components/index.ts
kachkaev: Yeah it’s hard to define a group based on ”often used together” because ”often” is a vague category. I’m happy with any simple rule that a reader can easily reverse-engineer. If keeping each override in a separate file does make the file structure look too artificial, I guess we can go for that.

site/src/theme/shadows.ts
teenoh: @benwerner01 you are correct. Thanks for the suggestion @kachkaev, I think that works although I'm concerned anyone might mistakenly use any of shadows 7 ... 26 which aren't in our design system. To avoid that, we should make shadows 7 - 26 the same as shadow 6. That way, if someone mistakenly tries to use shadow 7 or 8 instead of 6, it still renders the right shadow

site/src/theme/components/MuiIconButtonThemeOptions.ts
```diff
@@ -0,0 +1,38 @@
+import { Components } from "@mui/material";
+import { customColors } from "../palette";
+import { defaultTheme } from "../util";
```
teenoh, line 3: With the callback approach for setting styleOverrides, we don't need to import the defaultTheme anymore. 
See an example here => https://mui.com/blog/callback-support-in-style-overrides/

site/src/theme/components/MuiInputLabelThemeOptions.ts
```diff
@@ -0,0 +1,20 @@
...
+    disableAnimation: true,
+    shrink: true,
+  },
+  styleOverrides: {
+    root: {
+      position: "unset",
+      left: "unset",
+      top: "unset",
+      transform: "unset",
+      fontSize: 15,
```
teenoh, line 15: I think it should be possible to set this based on our typography config. A simple approach I can think off is probably exporting the different font sizes from MuiTypography override file and using that here

site/src/theme/components/MuiOutlinedInputThemeOptions.ts
```diff
@@ -0,0 +1,43 @@
...
+      },
+      "&.Mui-error": {
+        "& .MuiOutlinedInput-notchedOutline": {
+          borderColor: customColors.red[600],
+        },
+      },
+    },
+    input: {
+      padding: defaultTheme.spacing(1.5, 2),
+      fontSize: 18,
```
teenoh, line 30: https://github.com/blockprotocol/blockprotocol/pull/203/files#r802217645

site/src/theme/components/index.ts
teenoh: Sadly I can't seem to see the grouped setup in the changelog. I agree with @benwerner01 though.
I consider components "often used together" as sub-components of base components in our design system. When we have our design system components properly setup and documented (probably using storybook), I think that can aid future contributors in making changes.

site/src/theme/components/MuiButtonThemeOptions.ts
```diff
@@ -0,0 +1,303 @@
...
+            {
+              purple: { borderColor: customColors.purple[600] },
+              teal: { borderColor: customColors.teal[500] },
+              gray: { borderColor: customColors.gray[50] },
+              warning: { borderColor: customColors.orange[500] },
+              danger: { borderColor: customColors.red[700] },
+              inherit: {},
+            }[color]),
+        });
+      } else if (variant === "secondary") {
```
teenoh, line 122: I feel it's a bit difficult to locate certain blocks when going through this file. What if we place a comment like this above each if block

```suggestion
      //===== SECONDARY VARIANT========
      if (variant === "secondary") {
```

Also I'm not sure we need the `else` statement since a variant can only have one value at a time. That way we can add spaces between each `if` block and it's easier to see where it starts/ends

site/src/theme/components/MuiButtonThemeOptions.ts
```diff
@@ -0,0 +1,303 @@
...
+            }[color]),
+        });
+        Object.assign(disabledStyles, {
+          ...(color === "purple" && {
+            background: customColors.purple[200],
+            color: customColors.purple[100],
+          }),
+        });
+        Object.assign(focusVisibleAfterStyles, {
+          borderRadius: 34 + buttonFocusBorderOffset,
```
teenoh, line 111: We should probably make 34 a constant

teenoh (commented):
Thanks for this @benwerner01. I tested too and did not notice any regressions. I dropped some suggestions

site/src/theme/shadows.ts
benwerner01: @teenoh sure, I've repeated the final shadow 20 times instead of deferring to the default values, and @kachkaev I've followed your suggestion to improve the type safety. Here's the commit: https://github.com/blockprotocol/blockprotocol/pull/203/commits/977725688f05593bac0be118beb325399462283d

site/src/theme/components/MuiIconButtonThemeOptions.ts
benwerner01: Good spot, thank you Teeno! This removes the need for importing `defaultTheme` and `customColors` in most circumstances, which is great improvement. Have made this change here: https://github.com/blockprotocol/blockprotocol/pull/203/commits/1a271c7165a5922f3f98741d3f3d94e69b8d122b

site/src/theme/components/MuiButtonThemeOptions.ts
benwerner01: Although this has no noticeable impact on performance, using `else if` instead of `if` lets the interpreter skip having to check all subsequent conditions to see whether they hold true. I'm inclined to keep this as is, just to avoid ambiguity to readers why the runtime environment has to check all conditions when just one condition can be met.

Regarding the comments, I think they're a good idea. Have addressed theme here: https://github.com/blockprotocol/blockprotocol/pull/203/commits/28a5d2391bbc891bd2bd8d98c6cc92a029824a67

site/src/theme/components/MuiButtonThemeOptions.ts
benwerner01: I think that's a good idea, but am not sure what to best call this as I did not implement it. @kachkaev could you please advice?

site/src/theme/components/MuiInputLabelThemeOptions.ts
benwerner01: Good idea, if I'm understanding correctly we can just do `fontSize: theme.typography.bpSmallCopy.fontSize`. Have done this, and made similar changes to other files where I think it makes sense, here: https://github.com/blockprotocol/blockprotocol/pull/203/commits/33f414cba3942c30f6da2c909c1e3b55e98729a7

site/src/theme/components/index.ts
benwerner01: In light of the changes mentioned in [this thread](https://github.com/blockprotocol/blockprotocol/pull/203#discussion_r802743114), I'm inclined to not order them alphabetically, but group them based on the categories from the MUI website. Have done this here: https://github.com/blockprotocol/blockprotocol/pull/203/commits/be20367564a0569581effc13a48b028cc3d3730b

site/src/theme/components/MuiButtonThemeOptions.ts
teenoh: Got it. The comments solve the initial issue I stated, so that works

site/src/theme/components/MuiButtonThemeOptions.ts
benwerner01: Have called this `buttonBorderRadius` in this commit: https://github.com/blockprotocol/blockprotocol/pull/203/commits/a71b9f2d754a24c1af3ba6da170e650d4998c91a

siriwatknp:
@benwerner01 Hi, my name is Jun from MUI team. I found this issue linked to https://github.com/mui/material-ui/issues/30412, so I want to clarify on the deprecation of `variants`.

I might have misunderstood some of the `variants` purpose (you can read more in https://github.com/mui/material-ui/issues/30412#issuecomment-1018467855), so to summarize, we don't plan to deprecate the `variants` at this point. Sorry for the confusion and really appreciate seeing this PR try out the callback in styleOverrides (feedbacks are very welcome)

benwerner01:
@siriwatknp Hi Jun 👋 It appears there’s been some confusion on our side regarding MUI’s future intentions for the `variants` & `styleOverrides` fields, so thank you for the clarification!

I appreciate that [Olivier's comment](https://github.com/mui/material-ui/issues/30412#issuecomment-1018467855) provides some insight, but I’m wondering whether there is some existing or planned documentation for when the `variants` field and when the `styleOverrides` field should be used to define theme styling? At the moment this remains somewhat unclear to me, as both fields can be used for that purpose.

Is it that the `variants` field should be used as much as possible to leverage the performance benefits of static resolution, whereas the `styleOverrides` API is intended for "for more niche use cases of customizations” (as put by @oliviertassinari)? I appreciate this might be something that is still being worked out, but it would be helpful context for us when defining the MUI theme for current and future projects.

Regarding feedback on the `styleOverrides` field, one of the main reasons I was keen to migrate towards it is because of the flexibility that is provided by defining complex styling using a callback functions, as we have now done for [our Button overrides](https://github.com/blockprotocol/blockprotocol/blob/main/site/src/theme/components/inputs/MuiButtonThemeOptions.ts) for example.

Another nice feature is being able to access the top-level `theme` object as one of the function's arguments, which makes it more convenient to access things like theme colours, MUI convenience functions, etc. when defining a MUI theme - especially now that we've split our theme object into multiple files.

Thanks for all the hard work that has gone into maintaining and modernising MUI, it’s an outstanding component library I’ve had the pleasure of using for quite some time now :)

siriwatknp:
> Is it that the variants field should be used as much as possible to leverage the performance benefits of static resolution

I don't think there is a difference in terms of performance (I even think that `styleOverrides` would yield a better performance because it is passed directly to emotion or styled-components whereas `variants` need to map the prop and value on every render to apply the style.

> whereas the styleOverrides API is intended for "for more niche use cases of customizations

I don't think so, I think it is the other way around. The `styleOverrides` should be the primary way of theming because
- it is more powerful than `variants` and you don't have to learn a new syntax, it is the same old function.
- You also get typed safe by accessing `ownerState`.
- `ownerState` exposes the internal state we used inside the components, so you can do complex customization in case you need it.

The only downside is that you can't extract `styleOverrides` (as a function) to a static file and serve via static storage for theming. However, I think that it is rare or even not practical (not sure).

Anyway, this is only my opinion. Please don't take this as final conclusion. I will try to find a conclusion but it will take time (need to see community's feedback). Thanks a lot!

PR #204, opened 2022-02-08
Move Next API related lib files into dedicated api directory

benwerner01:
This PR moves the Next API related `lib` files into a dedicated `api` directory to help distinguish them from the ones that are used by the Next frontend


site/src/lib/api/awsSes.ts
```diff
@@ -2,8 +2,8 @@ import * as aws from "@aws-sdk/client-ses";
 import nodemailer from "nodemailer";
 import { convert } from "html-to-text";
 import SESTransport from "nodemailer/lib/ses-transport";
-import { isProduction } from "./config";
-import { mustGetEnvVar } from "../util/api";
+import { isProduction } from "../config";
+import { mustGetEnvVar } from "../../util/api";
```
kachkaev, line 6: What is the difference between util/api and lib/api?

thehabbos007 (commented):
Thanks for changing the folder structure, should be more straight-forward now :)

I merged in a PR not too long ago with another middleware file https://github.com/blockprotocol/blockprotocol/blob/main/site/src/lib/middleware/multipartUploads.ts 
and there will be some merge conflict, I'm sorry!

teenoh (approved):
Thanks for this @benwerner01 !

site/src/lib/api/awsSes.ts
benwerner01: Good question, having had a quick skim of what is in both directories I don't think we have a good distinction at the moment. I don't think this relates to the purpose of this PR, so I think it should be addressed separately, but it might be worth removing the `util` directory, and move its contents into `lib` for now. We can always create utility files in there for functionality that is shared across the next app. But I'm open to other suggestions/ideas. Let me know what you think.

Either-way, I'm going to merge this PR as I don't think it's the place where changes to the `util` directory structure should be made.

PR #206, opened 2022-02-08
add @jmackie/block-quote

JMackie80:
add @jmackie/block-quote


hub/@jmackie/quote.json
```diff
@@ -0,0 +1,7 @@
+{
+  "workspace": "@jmackie/block-quote",
```
CiaranMn, line 2: Could you remove this line please?

We use this for cases where the block source is inside a [yarn workspace](https://classic.yarnpkg.com/lang/en/docs/workspaces/), which isn't the case for your block.

CiaranMn (changes requested):
Thanks for your contirbution, @JMackie80! 

I'm excited to get your block up on the Block Hub - there's one small change needed to the provided JSON, and then it's good to go.

hub/@jmackie/quote.json
JMackie80: Thanks. Sorry about that.

CiaranMn (changes requested):
Thanks @JMackie80 - we needed to add some code to handle block source being in a subfolder but not a repo. You can now specify the subfolder using the `folder` path in the provided JSON - I've suggested how to do this.

I noticed a couple of things in the `package.json` for your block which it would be good to change also - the [filenames](https://github.com/JMackie80/blocks/blob/main/src/packages/quote/package.json#L77) for the icon and image do not match those in the folder - I think they should instead be this:
```json5
  "blockprotocol": {
    ...
    "icon": "public/quote.svg",
    "image": "public/block-preview.svg",
    ...
  }
```

Could you make these changes and let us know when done please?

hub/@jmackie/quote.json
```diff
@@ -0,0 +1,6 @@
+{
+  "repository": "https://github.com/JMackie80/blocks.git",
+  "branch": "main",
+  "distDir": "dist",
```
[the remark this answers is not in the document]
JMackie80: Oh, ok, thanks!

JMackie80:
@CiaranMn updated the image file names. Thanks.

CiaranMn (approved):
Excellent, thank you very much @JMackie80 - let's get your block up!

CiaranMn:
Now live - https://blockprotocol.org/@jmackie/quote 🥳

PR #207, opened 2022-02-09
Add shortname index on user collection

CiaranMn:
I noticed that user pages were sometimes loading slowly. This is because we didn't have an index on `shortname` in the user collection.

I've already created this in the database directly - this PR adds the index creation to our script so that we can (a) easily see it's present, and (b) have it recreated if for some reason it's ever removed.

### DIFF
--- a/site/scripts/create-db-indexes.ts
+++ b/site/scripts/create-db-indexes.ts
@@ -31,7 +31,7 @@ const script = async () => {
   await catchAndLog(async () => {
     await db
       .collection<UserDocument>(User.COLLECTION_NAME)
-      .createIndex({ shortname: 1 });
+      .createIndex({ shortname: 1 }, { unique: true, sparse: true });
   });
 
   await catchAndLog(async () => {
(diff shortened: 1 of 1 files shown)


site/scripts/create-db-indexes.ts
```diff
@@ -28,6 +28,12 @@ const script = async () => {
       .createIndex({ email: 1 }, { unique: true, sparse: true });
   });
 
+  await catchAndLog(async () => {
+    await db
+      .collection<UserDocument>(User.COLLECTION_NAME)
+      .createIndex({ shortname: 1 });
```
thehabbos007, line 34: Should the shortnames be unique as well?

CiaranMn: Oops, yes - thought I had. Fixed

thehabbos007 (approved):
Looks good to me. Thank you! 
Good thing you made it a sparse index as well. We'll see if the DB index script logs any errors about the index being different on the server, but hopefully not :)

PR #208, opened 2022-02-09
Handle non-workspace subfolders in user block repos

CiaranMn:
This PR introduces an optional `folder` field on the JSON supplied by users when adding blocks to the hub, to handle blocks in subfolders which are _not_ a yarn workspace.

If present, this folder is switched into before building the block.

**Background**
Our script to pull user's blocks from repos and build them relies on running `yarn install && yarn build` as being the correct commands to build the block.

As the block's `package.json` may not be in the root repo, we provide the option for JSON files provided by users to contain an optional `workspace` field, used to modify the `yarn` commands with the relevant `workspace`.

This does not handle the case where the commands must be run against a `package.json` in a subfolder, which is _not_ in a yarn workspace. 

The need for this highlighted by #206 

**How to test**
1. Add the JSON file from the linked PR, and remove all but one of the JSONs in `hub/@hash/` (to speed up build times)
2. Run `yarn build-blocks`
3. Check they all build successfully (this checks a HASH block in a yarn workspace, the new block in a subfolder, and the `@shinypb` block in the repo root, thus covering the three different ways of running the build command)


kachkaev (approved):
👍 Nice addition!

benwerner01 (approved):
This looks good to me, thanks Ciaran 🙂

PR #210, opened 2022-02-10
Make sure published`block-template` does not contain `dist` folder

kachkaev:
I was [developing a new block](https://blockprotocol.org/docs/developing-blocks) and noticed a small issue with the template. When I was publishing `block-template@0.0.6` I accidentally included `dist` because I built this package locally while testing it: https://unpkg.com/browse/block-template@0.0.6/

This PR makes sure that `dist` folder is not published as part of `block-template` and adds tests to verify this. I will publish a new version of the template after merging this PR.


packages/block-template/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "block-template",
-  "version": "0.0.6",
+  "version": "0.0.7",
```
kachkaev, line 3: TODO: publish after merge

scripts/packages/smoke-test-create-block-app.ts
```diff
@@ -1,10 +1,10 @@
 import execa from "execa";
 import path from "path";
 import tmp from "tmp-promise";
+import fs from "fs-extra";
```
kachkaev, line 4: [slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1643896758456119) (internal thread)

scripts/packages/smoke-test-create-block-app.ts
```diff
@@ -79,6 +79,19 @@ const script = async () => {
...
+
+    for (const relativePath of [".npmignore", "dist"]) {
+      if (
+        await fs.pathExists(path.resolve(resolvedBlockDirPath, relativePath))
+      ) {
+        throw new Error(
+          `Unexpected to find \`${relativePath}\` in the published block template`,
+        );
+      }
+    }
```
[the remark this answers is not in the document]
kachkaev: Parallel promises will be faster by an order of a millisecond here, but will make the output less deterministic. This means that you may end up seeing different errors when running the same test twice.

Let's keep `for of` here for readability and output stability. `Promise.all` has its merit, especially when we do network requests (they are much-much slower). BTW when we have too many promises to resolve, we can use [`p-queue`](https://www.npmjs.com/package/p-queue) — it’s a pretty neat library!

packages/block-template/package.json
kachkaev: https://unpkg.com/browse/block-template@0.0.7/

PR #213, opened 2022-02-11
Fix `completeSignup` error

benwerner01:
This PR fixes an error that is thrown when the `completeSignup` Next API endpoint is called.

### DIFF
--- a/site/src/lib/api/model/user.model.ts
+++ b/site/src/lib/api/model/user.model.ts
@@ -234,6 +234,7 @@ export class User {
     updatedProperties: Partial<UserProperties>,
   ): Promise<User> {
     if (
+      this.shortname &&
       updatedProperties.shortname &&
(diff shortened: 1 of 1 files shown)


site/src/lib/api/model/user.model.ts
```diff
@@ -234,6 +234,7 @@ export class User {
     updatedProperties: Partial<UserProperties>,
   ): Promise<User> {
     if (
+      this.shortname &&
       updatedProperties.shortname &&
       updatedProperties.shortname !== this.shortname
     ) {
```
benwerner01, line 240: This check should only be performed when the user already has a `shortname`. Otherwise it prevents new users from setting their `shortname` for the first time as part of the `completeSignup` Next API endpoint.

kachkaev (approved):

PR #217, opened 2022-02-14
Improve hover performance of cards on hub page, and updated animation

nathggns:
This replaces the box shadow transition on block cards on `/hub` with two psuedo elements that animate in and out using opacity. This is because box shadow cannot be transitioned entirely on the GPU, but opacity can be. This should improve the performance of the transition. 

See https://tobiasahlin.com/blog/how-to-animate-box-shadow/ for more details.

It also reduces the scale of the hover transition to make it less jarring

---
  - https://app.asana.com/0/0/1201742926145330


MaggieAppleton:
Nice! This looks good to me. Animation is nice and snappy.

Only slight adjustment I'd make it changing the rate of scaling on the cards from 1.05 to a smaller scale of 1.03. I had this on my laundry list of tiny tweaks to make on BP. Just enough scaling to make it clear the card is being hovered over, but without too much dramatic animation.

Box shadow trick is a cool hack 👍

kachkaev (approved):
Nice one!

I wonder if we can also avoid repainting other cards on hover – looks like we have a reflow event both in Chrome and Safari:

https://user-images.githubusercontent.com/608862/153915963-81cc3e7a-8c94-4260-a688-51dae3ed83d4.mp4


https://user-images.githubusercontent.com/608862/153915979-e4bc26a0-76ae-4b8a-b5f7-b6f824a931c5.mp4

As you can see from the Frame Rate chart in the first screencast, rendering is already more performant on hover. So what I am suggesting is a stretch goal 🙂

kachkaev:
The new solution is a bit more memory-intense because it requires more layers to be kept in memory on idle.

Before / after `willChange`:

https://user-images.githubusercontent.com/608862/153926571-8e2a3ca4-b530-4912-a73d-18855e5428bc.mp4

https://user-images.githubusercontent.com/608862/153926575-e38a8830-ea24-42c9-b86c-f146948a6188.mp4

I was hoping we could avoid `willChange` _and_ reflow by using absolute positioning of some elements etc. But maybe it’s not worth the time now. Perhaps, the original solution scales a bit better because it won’t require more idle memory as we increase the number of cards. WDYT? I’ll approve just in case.

kachkaev (approved):

nathggns:
@kachkaev Decided to remove `will-change` and leave the paint issue you identified unaddressed as its not clear this is contributing to any kind of performance issue, and this could actually make performance worse depending on GPU memory and the number of blocks on the page.

kachkaev (approved):

PR #219, opened 2022-02-15
Set default data source to empty JSON

akash-joshi:
## What this Does

This PR fixes the error currently observed in the Divider block at [blockprotocol.org/@hash/divider](https://blockprotocol.org/@hash/divider) by setting the default data source to an empty JSON `{}`.

## Task Link

- [Divider block JSON parsing errors](https://app.asana.com/0/1201707629991380/1201805956989968/f)


thehabbos007 (approved):
Thank you Akash, this is what I was expecting from the ticket :)

benwerner01 (approved):
Thanks @akash-joshi , this does fix the parsing error for me 👍

PR #220, opened 2022-02-15
Allow direct links to blocks & schemas of a user

akash-joshi:
## What this Does

The user profile page on the BlockProtocol site currently has tabs to navigate between a user's profile, their blocks, and available schemas. However, one presently cannot link to the blocks or schemas tabs directly. This PR creates those pages by refactoring existing code, allowing users to be able to share links to their blocks or schemas pages freely.

## What Does this Change

- [Add slugs to tabs](https://github.com/blockprotocol/blockprotocol/pull/220/files#diff-ea85b256b8482c1d52bf722baf3a828dbf2918d105c4400ec1b19a8607776851R8) for use during routing.
- [Create a `UserPageComponent`](https://github.com/blockprotocol/blockprotocol/pull/220/files#diff-34bc1d1615cd3a42118e7265c1dfc012e8a76c4d77d0529688c4ff95fcb1dfe3R1) for reuse across all user pages.
- Create seperate pages for blocks and schemas with the default tab set [to the respective page](https://github.com/blockprotocol/blockprotocol/pull/220/files#diff-a703eeb67bf29ca9dfa259afcba3e877663639b2209765b08f40765a49e7437fR62).

## How to Test this

- Go to a user's profile
- Switch between various tabs and look at the URL indicator
- Try to go to a specific URL directly

chrome_J0eWKUP4M0

---
  - https://app.asana.com/0/0/1201768159094198


site/src/components/pages/user/UserPageComponent.tsx
```diff
@@ -0,0 +1,350 @@
...
+            sx={{
+              flex: 1,
+            }}
+          >
+            {/* TAB HEADER */}
+            <TabHeader
+              activeTab={activeTab}
+              setActiveTab={(nextTab) => {
+                setActiveTab(nextTab);
+                return router.push(
```
akash-joshi, line 166: Routing between tabs happens here

site/src/components/pages/user/UserPageComponent.tsx
```diff
@@ -0,0 +1,350 @@
...
+
+type UserPageComponentProps = UserPageProps & {
+  initialActiveTab: typeof TABS[number]["value"];
+};
+
+export const UserPageComponent: VoidFunctionComponent<
+  UserPageComponentProps
+> = ({ blocks, entityTypes, user, initialActiveTab }) => {
+  const [activeTab, setActiveTab] =
+    useState<typeof TABS[number]["value"]>(initialActiveTab);
```
akash-joshi, line 50: `initialActiveTab` to be passed down from parent component.

site/src/pages/[shortname]/blocks.page.tsx
```diff
@@ -0,0 +1,67 @@
+import React from "react";
+import { NextPage, GetServerSideProps } from "next";
+
```
Comment (teenoh) [inline site/src/pages/[shortname]/blocks.page.tsx:3]: Instead of creating 2 new files, I think it's better we leverage [Rewrites](https://nextjs.org/docs/api-reference/next.config.js/rewrites) in Nextjs to point `/[shortname]/blocks` and `/[shortname]/schemas` routes to `/shortname` route. I think we already use that for [docs](https://github.com/blockprotocol/blockprotocol/blob/main/site/next.config.js#L22)

CiaranMn (changes requested):
I'm not sure about having to introduce a new page for each tab - the alternative would be to keep the code much as it already is, but rename `[shortname]/index.page.tsx` to `[shortname]/[[tabName]].tsx`, then you can pick up the tab name from the query and load whichever component (or the default component if the user just visited `[shortname]/`).

Can we talk about this one?

akash-joshi:
New changes with the latest commit:
- [Move blocks pages to `/blocks`](https://github.com/blockprotocol/blockprotocol/pull/220/files#diff-37a94694f983c5a515ba0156ac4c11fccc2ee9c7a5ed469af3a15024b07fe6b3R15)
- [Updated existing links](https://github.com/blockprotocol/blockprotocol/pull/220/files#diff-97f263c9a516a4d73af7ea92727fe53cee1ac1ad6b835cdb78fdac97b41ebedeR22) to point to new `/blocks` URL.
- [Return new `/blocks` path](https://github.com/blockprotocol/blockprotocol/pull/220/files#diff-ef6b34f7ca0be2a76177cfb93e2f30a49546f3becd1311ed3a4815adc260abb2R110) as `slug`.
- Replaced [packagePath with slug](https://github.com/blockprotocol/blockprotocol/pull/220/files#diff-896218bb0f1453e8fb6deb0d79f1e638ff3762f8a02eb8fa62cd884f8f23364dR88) wherever required.
- [Get activeTab from router](https://github.com/blockprotocol/blockprotocol/pull/220/files#diff-34bc1d1615cd3a42118e7265c1dfc012e8a76c4d77d0529688c4ff95fcb1dfe3R49) instead of maintaining it in a state.
- [Create single page for profile tabs](https://github.com/blockprotocol/blockprotocol/pull/220/files#diff-1ed99ec657b21c17e18966d43769d1cde4da6cf404eb2d3f665495621f9e4336R1)

site/src/components/pages/user/UserPageComponent.tsx
CiaranMn: This doesn't need a `return`

site/src/components/BlockCard.tsx
```diff
@@ -80,12 +80,12 @@ export const BlockCard: VFC<BlockCardProps> = ({ loading, data }) => {
...
     version,
     lastUpdated,
-    packagePath,
+    slug,
     icon,
   } = data;
 
   return (
-    <Link href={`/${packagePath}`}>
+    <Link href={`/${slug}`}>
```
CiaranMn, line 88: I think calling this `slug` is a bit confusing as it's a whole path, not a single slug - can we call it `blockPagePath` or something instead? It makes it easier to see what's being linked to without having to go look up what `slug` this is.

If you add the `/` at the point of constructing it you could pass it directly rather than interpolating everywhere as well.

CiaranMn (changes requested):
Thanks @akash-joshi - new approach looks good.

I've had a play and I think the one improvement needed is that now it's solely router-based, it takes a second to switch tabs - can we introduce local state again to make it snappier? State can be set from the router query and also optimistically updated when the user switches tab.

site/src/components/pages/user/UserPageComponent.tsx
akash-joshi: @CiaranMn this is for satisfying `no-floating-promises`.

site/src/pages/[shortname]/blocks.page.tsx
Comment (akash-joshi) [reply site/src/pages/[shortname]/blocks.page.tsx]: Thank you Teenoh! This has now been implemented.

akash-joshi:
Thanks @CiaranMn, have implemented the above changes. Addressed your `return` comment here: https://github.com/blockprotocol/blockprotocol/pull/220#discussion_r812617495

site/src/components/pages/user/UserPageComponent.tsx
```diff
@@ -27,44 +28,39 @@ import { Button } from "../../Button";
...
-    }
-
     const matchingTab = TABS.find((tab) => tab.slug === profileTabs?.[0]);
 
-    if (!matchingTab) {
-      if (process.browser) {
-        void router.replace(`/@${user.shortname}`);
-      }
-      return TABS[0].value;
+    if (!matchingTab && typeof window !== "undefined") {
```
CiaranMn, line 57: `typeof window !== "undefined"` should be unnecessary as `useEffect` does not run on the server.

site/src/components/pages/user/UserPageComponent.tsx
CiaranMn: Understood - this is not really in the spirit of `no-floating-promises` because this is moving the promise to be unhandled elsewhere, rather than actually handling it. But not important in this case.

CiaranMn (approved):
Looks good, faster now, thanks!

site/src/components/pages/user/UserPageComponent.tsx
akash-joshi: Should be done now :D

CiaranMn (approved):

PR #223, opened 2022-02-15
Add Create Schema card to dashboard

teenoh:
This PR does the following
- Extracts "Create Schema" Modal into a reusable component
- Adds "Create a Schema" and "View Profile" cards to the dashboard page
- Adds hover styles to dashboard card
- Adds a minHeight to small buttons. This prevents the height from changing when it's in loading state
- Adds a route `/dashboard/create-schema` or `/dashboard?create-schema=true` 
- Take unauthenticated users to the login screen when auth-walled screens are accessing and redirect them back on successful login

Demo

https://user-images.githubusercontent.com/16928688/154247142-b7bd3033-73bc-4a3d-8bf8-87529298ff77.mp4

---
  - https://app.asana.com/0/0/1201743072238309


site/src/components/pages/dashboard/DashboardCard.tsx
```diff
@@ -1,31 +1,166 @@
...
   colorGradient,
   description,
   link,
+  variant,
+  icon,
 }) => {
+  if (variant === "secondary") {
+    return (
+      <DashboardCardSecondary
+        {...{ title, description, link, variant, icon }}
```
akash-joshi, line 132: This doesn't seem very readable. How about assigning `{ title, description, link, variant, icon }` to an object and spreading that object here?

site/src/components/pages/dashboard/DashboardCard.tsx
```diff
@@ -1,31 +1,166 @@
...
+      />
+    );
+  }
+
   return (
-    <Link href={link.href} sx={{ marginBottom: 4 }}>
+    <CardWrapper href={link.href} onClick={link.onClick}>
       <Box
         sx={{
+          // @todo update theme config to include microShadow
```
akash-joshi, line 141: Is there a task for this, or a reason it isn't added as a part of this PR?

akash-joshi (commented):
@teenoh it would be helpful to have a demo gif in the PR description to review the UI 😄

site/src/components/pages/dashboard/DashboardCard.tsx
teenoh: Thanks. Updated here 8423919

site/src/components/pages/dashboard/DashboardCard.tsx
teenoh: I will create a task for that. The reason it isn't added here is because updating the boxShadows in our theme config will also mean updating all places we use boxShadows in the project. This is because the shadows are array index based
```
// theme config
boxShadows: [none, "<baseShadow>", "<smallShadow>"]

// usage
<Box sx={{ boxShadow: 1 // baseShadow }} />
```
Adding a new shadow, will result in updating usages of boxShadow since they most likely now have a new index

```
boxShadows: [none, "microShadow", "<baseShadow>", "<smallShadow>"]
// New usage
<Box sx={{ boxShadow: 2 // baseShadow }} />
```

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+  onClose: () => void;
+};
+
+export const CreateSchemaModal: FC<CreateSchemaModalProps> = ({
+  open,
+  onClose,
+}) => {
+  const [newSchemaTitle, setNewSchemaTitle] = useState("");
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState("");
```
benwerner01, line 21: I think this should be `undefined` by default, and support `ReactNode` incase we want to introduce formatting in the message. 
```suggestion
  const [error, setError] = useState<ReactNode>();
```

nonparibus:
Haven't looked to see if this has already been addressed, but if we're going to put the "create schema" form in a modal, can we please ensure there's a universal direct link to the "create schema" page we can point any user to from the docs (without knowing their specific/individual namespace)?

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+        if (apiError.response?.data.errors) {
+          setError(apiError.response.data.errors[0].msg);
+        } else {
+          // @todo properly handle this
+          setError("An error occured");
+        }
+      } else {
+        const schemaTitle = data?.entityType.schema.title;
+        void router.push(`/@${user.shortname}/types/${schemaTitle}`);
+      }
```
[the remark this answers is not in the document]
teenoh: Is it possible for `error` and `data` to be undefined/null at the same time?

nonparibus:
Bumping this
(would be a good one to showcase at board)

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+        },
+      });
+      setLoading(false);
+      if (apiError) {
+        if (apiError.response?.data.errors) {
+          setError(apiError.response.data.errors[0].msg);
+        } else {
+          // @todo properly handle this
+          setError("An error occured");
+        }
```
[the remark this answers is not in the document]
teenoh: Thanks for catching. Updated to use `error.message` in 107e6d5dadddab

site/src/components/Modal/CreateSchemaModal.tsx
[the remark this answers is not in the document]
teenoh: Updated in 107e6d5dadddab

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+        <Typography
+          sx={{
+            mb: 4,
+            fontSize: 16,
+            lineHeight: 1.5,
+            width: { xs: "90%", md: "85%" },
+          }}
+        >
+          {` Schemas are used to define the structure of entities - in other
+            words, define a ‘type’ of entity`}
```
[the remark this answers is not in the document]
teenoh: Updated in 107e6d5dadddab

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+      if (!user || user === "loading") {
+        return;
+      }
+
+      if (newSchemaTitle === "") {
+        setError("Please enter a valid value");
+        return;
+      }
+
+      setLoading(true);
```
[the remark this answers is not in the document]
teenoh: Added in 107e6d5dadddab

site/src/components/pages/dashboard/DashboardCard.tsx
```diff
@@ -1,31 +1,183 @@
...
+  variant?: "primary" | "secondary";
+};
+
+type CardWrapperProps = {
+  onClick?: () => void;
+  href?: string;
+};
+
+const CardWrapper: FC<CardWrapperProps> = ({ children, onClick, href }) => {
+  const sharedStyles: BoxProps["sx"] = useMemo(
```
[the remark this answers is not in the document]
teenoh: Updated in 107e6d5dadddab069e33955

site/src/components/Modal/CreateSchemaModal.tsx
teenoh: I don't think there's a need to support that until we eventually introduce formatting the message. Will update for consistency

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+          <TextField
+            sx={{ mb: 3 }}
+            label="Schema Title"
+            fullWidth
+            helperText={error}
+            value={newSchemaTitle}
+            onChange={(evt) => {
+              if (error) {
+                setError("");
+              }
```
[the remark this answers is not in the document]
teenoh: That makes sense. Ideally, I think API errors should be rendered at a different point and not under a particular input (validation errors should be there instead). That way errors like network request ones can be persisted until the request is retried. Will update to use the existing approach for consistency

site/next.config.js
```diff
@@ -33,6 +33,11 @@ module.exports = withFonts(
           destination: "/contact/submitted",
           permanent: true,
         },
+        {
+          source: "/dashboard/create-schema",
+          destination: "/dashboard?create-schema=true",
+          permanent: true,
+        },
```
teenoh, line 40: @dw, Create Schema page can be accessed by going to  `/dashboard?create-schema=true` or better still `/dashboard/create-schema`. If the user is not authenticated, they are taken to the login screen and redirected back to the page on successful login

site/src/components/Modal/CreateSchemaModal.tsx
teenoh: Fixed here => https://github.com/blockprotocol/blockprotocol/pull/223/commits/cc8b0a31aafe03be4c128127b8c83292084ea4d9

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+              handleSchemaTitleChange(evt.target.value);
+            }}
+            required
+            error={Boolean(error)}
+          />
+
+          <Button loading={loading} size="small" squared type="submit">
+            Create
+          </Button>
```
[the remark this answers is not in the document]
teenoh: Fixed here => https://github.com/blockprotocol/blockprotocol/pull/223/commits/cc8b0a31aafe03be4c128127b8c83292084ea4d9

site/src/components/Modal/CreateSchemaModal.tsx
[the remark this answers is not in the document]
teenoh: Fixed here => https://github.com/blockprotocol/blockprotocol/pull/223/commits/cc8b0a31aafe03be4c128127b8c83292084ea4d9

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  return (
+    <Modal
+      open={open}
+      onClose={onClose}
+      contentStyle={{
+        top: "40%",
+      }}
+    >
+      <Box sx={{}}>
```
benwerner01, line 89:
```suggestion
      <Box>
```

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+        <Typography
+          sx={{
+            mb: 4,
+            fontSize: 16,
+            lineHeight: 1.5,
+            width: { xs: "90%", md: "85%" },
+          }}
+        >
+          {`Schemas are used to define the structure of entities - in other words,
+          define a ‘type’ of entity`}
```
benwerner01, line 108:
```suggestion
          Schemas are used to define the structure of entities - in other words,
          define a ‘type’ of entity
```

site/src/components/pages/authWall.tsx
```diff
@@ -22,7 +22,7 @@ const AuthWallWrapper: VoidFunctionComponent<{
   }
 
   if (!user?.isSignedUp) {
-    void router.push("/");
+    void router.push(`/login?redirectPath=${router.asPath}`);
```
benwerner01, line 25: Fun fact:
```suggestion
    void router.push({
      pathname: "/login",
      query: { redirectPath: router.asPath },
    });
```

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+        </Typography>
+        <Box component="form" onSubmit={handleCreateSchema}>
+          <TextField
```
benwerner01, line 111: I think it'd be a good idea to focus the `TextField` input when the modal is opened. This would involve capturing the `inputRef` of the TextField using `useRef`, and calling `inputRef.current.focus()` in a `useEffect` hook that triggers when the modal is opened.

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+            sx={{ mb: 3 }}
+            label="Schema Title"
+            fullWidth
+            helperText={helperText}
+            value={newSchemaTitle}
+            onChange={(evt) => {
+              if (apiErrorMessage) {
+                setApiErrorMessage(undefined);
+              }
+              handleSchemaTitleChange(evt.target.value);
```
benwerner01, line 121:
```suggestion
            onChange={({ target }) => {
              if (apiErrorMessage) {
                setApiErrorMessage(undefined);
              }
              handleSchemaTitleChange(target.value);
```

site/src/components/Modal/CreateSchemaModal.tsx
```diff
@@ -0,0 +1,140 @@
...
+            }}
+            required
+            error={displayError}
+          />
+
+          <Button
+            disabled={displayError}
```
benwerner01, line 128: This shouldn't just be disabled when an error is being displayed, it should be disabled when there is any error in the input. When the modal is loaded for the first time for instance, the button should be disabled because the input is empty. Currently it is not:
image

I think this should be as simple as:
```suggestion
            disabled={isSchemaTitleInvalid}
```

An existing example where we're doing this would be in `CompleteSignupScreen.tsx`, I realise we aren't doing this everywhere atm.

benwerner01 (approved):
Thanks for addressing my suggestions, this makes the way in which we are handling inputs/errors much more consistent with the other forms currently in the codebase.

I've left some more suggestions which would be nice to address as part of this PR, but can also be done in follow-up so don't block merging.

site/next.config.js
```diff
@@ -32,6 +32,11 @@ const nextConfig = {
         destination: "/contact/submitted",
         permanent: true,
       },
+      {
+        source: "/dashboard/create-schema",
+        destination: "/dashboard?create-schema=true",
+        permanent: true,
+      },
```
benwerner01, line 39: Btw I think this is fine for now, but I suspect the preference for `/dashboard/create-schema` was not just to be a redirect, but to be the actual endpoint of the modal (so that the browser URL is more readable and doesn't contain query parameters). (Correct me if I'm wrong @nonparibus )

I think this could be accomplished by relocating `dashboard.page.tsx` to `/dashboard/[...slugs].page.tsx`, and opening the modal when `router.query.slugs === ["create-schema"]`.

But appreciate that's a lot more work than a simple redirect, so we can do this down the line.

teenoh: Updated to use the dynamic routes approach here => 6309ee3

site/src/components/Modal/CreateSchemaModal.tsx
teenoh: I assumed the required check on the input field will catch that edge case, but I get your point. Updated here => 6309ee3

site/src/components/Modal/CreateSchemaModal.tsx
teenoh: Updated here => 6309ee3

site/src/components/Modal/CreateSchemaModal.tsx
teenoh: Updated here => 6309ee3

site/src/components/Modal/CreateSchemaModal.tsx
teenoh: Fixed here => 6309ee3

site/src/components/Modal/CreateSchemaModal.tsx
benwerner01: To be clear this suggestion wasn't about functionality, it was about having the button be visibly disabled (the styling should be different when it is disabled). I realise the user can't submit the form when the input is invalid thanks to the required field 👍

site/next.config.js
benwerner01: Nice, this could be further extended to clear the `/create-schema` part of the url when the modal is closed, and to add it when the "create a schema" button is clicked on the dashboard page, but this is fine for now.

benwerner01 (approved):
Looks good, thanks for addressing my suggestions 🙂

PR #225, opened 2022-02-17
Fix mock-block-dock breaking on entityId change

akash-joshi:
## What this Does

Presently, switching between the header and paragraph block on the blockprotocol demo page doesn't change the example its demoing. While #171, it revealed a bug in `mock-block-dock`, where changing the `entityId` passed to its child resulted in a cascading effect, which eventually lead to an error.

This PR adds stability to the internal state of `mock-block-dock`, by utilizing callbacks and state declarations, to make the state more readable, and also fix the above-mentioned bug.

Changes described in GH Comments below.

## Known Issues

- [Repeated switches between blocks causes image block to not load default data and uploading a new file displays the error "Unreachable".](https://app.asana.com/0/1201707629991380/1201857086153366/f)

## Task Link

- [Update mock-block-dock to be reactive to child prop changes](https://app.asana.com/0/1201707629991380/1201820398013222/f).

## Next Ups

- Update the version of `mock-block-dock` on npm with the latest version.
- After using the latest version of `mock-block-dock`, update `BlockDataContainer` to pass block name in `entityId` to get the latest data in the mock entity store.


packages/mock-block-dock/dev/DevApp.tsx
```diff
@@ -1,14 +1,26 @@
...
-  <MockBlockDock>
-    <TestBlock entityId="test-entity-1" name="World" />
-  </MockBlockDock>
-);
+const DevApp = () => {
+  const [entityId, setEntityId] = useState(1);
+
+  return (
+    <>
+      <button type="button" onClick={() => setEntityId(entityId + 1)}>
```
akash-joshi, line 15: This change to the dev setup allowed me to consistently test for the buggy behavior being observed while switching blocks.

packages/mock-block-dock/src/MockBlockDock.tsx
```diff
@@ -21,41 +23,54 @@ export const MockBlockDock: VoidFunctionComponent<MockBlockDockProps> = ({
...
-  const latestBlockEntity = useMemo(
-    () =>
-      entities.find((entity) => entity.entityId === children.props.entityId),
-    [entities, children.props.entityId],
+      return nextMockData;
+    },
+    [blockSchema, children.props],
+  );
+
+  const [mockData, setMockData] = useState<MockData>(
```
akash-joshi, line 63: Move `mockData` to state, to make it more stable. The previous code was running on each render, causing multiple versions of the same entity to be pushed into the `entities` array.

Additionally, the computing code has been moved to a callback function above.

packages/mock-block-dock/src/MockBlockDock.tsx
```diff
@@ -21,41 +23,54 @@ export const MockBlockDock: VoidFunctionComponent<MockBlockDockProps> = ({
...
+    getMockData(initialMockData),
   );
+
+  useEffect(() => {
+    setMockData(getMockData(initialMockData));
+  }, [children.props.entityId, getMockData]);
```
akash-joshi, line 69: Update mockData whenever child entityId changes.

packages/mock-block-dock/src/useMockDatastore.ts
```diff
@@ -44,16 +46,38 @@ export const useMockDatastore: UseMockDataStore = (
     entities: [],
     links: [],
     entityTypes: [],
+    blockEntityId: "",
   },
 ) => {
   const [entities, setEntities] = useState<MockDataStore["entities"]>(
     initialData.entities,
   );
+  const [latestBlockEntity, setLatestBlockEntity] = useState(
```
akash-joshi, line 55: Moved `latestBlockEntity` to `useMockDataStore` since it was very tightly dependent on the values being passed to, and returned from it.

packages/mock-block-dock/src/useMockDatastore.ts
```diff
@@ -44,16 +46,38 @@ export const useMockDatastore: UseMockDataStore = (
...
+    entities.find((entity) => entity.entityId === initialData.blockEntityId),
+  );
   const [links, setLinks] = useState<MockDataStore["links"]>(initialData.links);
   const [entityTypes, _setEntityTypes] = useState<MockDataStore["entityTypes"]>(
     initialData.entityTypes,
   );
 
+  useEffect(() => {
+    setEntities(initialData.entities);
+  }, [initialData.entities]);
```
akash-joshi, line 65: Update entities whenever `initialData` entities change.

packages/mock-block-dock/src/useMockDatastore.ts
```diff
@@ -44,16 +46,38 @@ export const useMockDatastore: UseMockDataStore = (
...
+      (entity) => entity.entityId === initialData.blockEntityId,
+    );
+
+    if (!nextLatestBlockEntity) {
+      nextLatestBlockEntity = initialData.entities.find(
+        (entity) => entity.entityId === initialData.blockEntityId,
+      );
+    }
+
+    setLatestBlockEntity(nextLatestBlockEntity);
```
akash-joshi, line 78: Update `latestBlockEntity` whenever either initialData entities, or the internal entities state changes.

CiaranMn: `latestBlockEntity` is derived from the props and state so I don't think it needs to be in state itself - it can just be recalculated on each render (or inside `useMemo`)

packages/mock-block-dock/src/useMockDatastore.ts
CiaranMn: I think this makes the purpose `useMockDatastore` less well defined - it would be better if we kept it to returning data and functions to manipulate the data, and leave it up to consumers to do what they want with the data (e.g. find a certain entity in it) - i.e. leave this calculation outside it. Then `useMockDatastore` doesn't have to know about blocks or `latestBlockEntity`.

packages/mock-block-dock/src/MockBlockDock.tsx
CiaranMn: could `mockData` be set by the return of `useMemo` instead to make it stable? Then you wouldn't need this state or the effect below.

CiaranMn (changes requested):
Thanks @akash-joshi, this is a tricky one!

I've made a few suggested changes. In this PR can you also:
- bump the version number in `mock-block-dock` and the packages that depend on it
- make the change to `BlockDataContainer` which this is enabling

This will make it easier to test all at once.

packages/mock-block-dock/src/MockBlockDock.tsx
```diff
@@ -21,41 +21,54 @@ export const MockBlockDock: VoidFunctionComponent<MockBlockDockProps> = ({
...
 
   const { entities, entityTypes, links, functions } =
     useMockDatastore(mockData);
 
-  const latestBlockEntity = useMemo(
-    () =>
-      entities.find((entity) => entity.entityId === children.props.entityId),
-    [entities, children.props.entityId],
-  );
+  const latestBlockEntity = useMemo(() => {
```
akash-joshi, line 63: Generating `latestBlockEntity` from `entities` or `mockData.entities`, based on whether the new entities have been generated.

site/src/components/pages/hub/BlockDataContainer.tsx
```diff
@@ -58,8 +58,8 @@ export const BlockDataContainer: VoidFunctionComponent<
   /** used to recompute props and errors on dep changes (caching has no benefit here) */
   const [props, errors] = useMemo<[object | undefined, string[]]>(() => {
     const result = {
-      accountId: `test-account-1`,
-      entityId: `test-entity-1`,
+      accountId: `test-account-${metadata.name}`,
+      entityId: `test-entity-${metadata.name}`,
```
akash-joshi, line 62: Changes the entityId for each block, causing mockStore to update.

CiaranMn (approved):
Thanks Akash, looks good and works well!

PR #216, opened 2022-02-14
Upgrade Next.js to v12

kachkaev:
By-product of https://github.com/blockprotocol/blockprotocol/pull/215

Related material:
- https://nextjs.org/blog/next-12
- https://nextjs.org/docs/upgrading#upgrading-from-11-to-12

[Asana task](https://app.asana.com/0/1201707629991380/1201825402555639/f) _(internal link)_

PR #229, opened 2022-02-18
Simplify site setup by removing `next-fonts` and `next-images`

kachkaev:
We switched from Next 11 to Next 12.1 recently: #216 & #228 + https://github.com/hashintel/hash/pull/320. When upgrading Next in the HASH repo, I was able to simplify our setup by reviewing some of the Next plugins. This PR does a similar review for [blockprotocol.org](https://blockprotocol.org).

[Asana task](https://app.asana.com/0/0/1201815866759884) _(internal link)_

It turns out that we can get rid of both `next-fonts` and `next-images` 🎉

The authors of `next-fonts` advise against using their own plugin – see https://www.npmjs.com/package/next-fonts. Just keeping the fonts in `/public` is good enough in a modern Next app.  We lose `optimizeFonts: true`, but this does not change much for us:

https://blockprotocol.org (prod)

Screenshot 2022-02-18 at 20 02 31

https://blockprotocol-h917irgdt-hashintel.vercel.app (this PR)

Screenshot 2022-02-18 at 20 02 24

Fonts change very rarely, so it’s better to pre-optimize them once than to compress them each time we built our app. Lack of hashes in file paths is not a problem because default Next.js headers for `/public` contain `cache-control: public, max-age=0, must-revalidate`.

I guess that we had `next-image` to potentially import SVGs as React components. All our current SVGs are coming from `/public` as ``, which is actually great. If we ever need inline SVGs, we can refer to this article: https://dev.to/dolearning/importing-svgs-to-next-js-nna. One of the options would be to use `svgr` like in https://github.com/hashintel/hash/pull/320, so I doubt that we’ll have to bring `next-images` back. Ideally, it’d be great to avoid inlining SVGs at all because they impact app performance: https://www.youtube.com/watch?v=Lh9q3h2khlc&t=652s.

I was also going to upgrade `@next/bundle-analyzer` to `12.1.0`, but did not find it in our site’s `package.json` 😁 We can add this plugin later if we need it – there is no practical point doing this now.


site/package.json
```diff
@@ -53,10 +53,8 @@
     "mime-types": "^2.1.34",
     "mock-block-dock": "0.0.5",
     "mongodb": "^4.2.2",
-    "next": "12.1.0",
-    "next-connect": "^0.11.0",
-    "next-fonts": "^1.5.1",
-    "next-images": "1.8.1",
+    "next": "^12.1.0",
+    "next-connect": "^0.12.1",
```
kachkaev, line 57: Using `^` to keep things consistent in this `package.json`. We can drop all `^`s in a separate PR I guess.

teenoh (approved):
LGTM, Thanks @kachkaev!

PR #231, opened 2022-02-23
Improve block preparation process

kachkaev:
[Asana task](https://app.asana.com/0/0/1201858717936353) _(internal link)_

This PR replaces `build-blocks.sh` with `prepare-blocks.ts` and improves block preparation process in a number of ways.  I renamed _build_ with _prepare_ is communicate the process a bit better. We do build blocks, but this is an optional step in a bigger process, which has other steps too (like downloading from GitHub, installing dependencies, copying files etc.).

- Caching is re-introduced. We still rely on Vercel to store it, which may or may not work in all cases. Docs: https://vercel.com/docs/concepts/deployments/build-step#caching
   If we get too many cache misses, we can consider building blocks in a GitHub Action which gives more control. That will complicate our pipelines though, so worth trying Vercel first. When there is a full cache hit, Vercel build drops from 9-10 mins to 3-4 mins.
   
  Screenshot 2022-02-23 at 16 54 37

- We use `commit` instead of `branch` in the block info file. This makes built block artifacts more stable even if cache is reset.

- We support `npm` in  addition to `yarn`.

- We support blocks that don’t have a build step (it was compulsory).

- We check stability of `yarn.lock` / `package-lock.json` (this can be turned off with an env var).


.eslintignore
```diff
@@ -53,6 +53,6 @@ dist
 .vercel
 
 # workspace specific
+site/blocks-data.json
 site/public/blocks
 site/site-map.json
```
kachkaev, line 58: Just cleaning up ignore files. Found `site/yarn.lock` which we no longer have.

site/scripts/copy-blocks-to-ci-cache.ts
```diff
@@ -0,0 +1,37 @@
+import chalk from "chalk";
```
kachkaev, line 1: This command is pretty much a copy-pasted version of `site/scripts/copy-blocks-from-ci-cache.ts`. This is not very DRY, but I don’t think it’s worth introducing abstractions to deduplicate this code. It is temporary (we won’t store block cache in Vercel forever).

site/README.md
```diff
@@ -73,25 +73,19 @@ When running in development environments, avatars go to the `dev/avatars/(user.i
...
-
-- `curl`
-- `jq`
-- `md5sha1sum`
-- `rsync`
-
-These can be installed by your cli pkg mngr of choice (use `brew` on macOS).
+Before serving any blocks, they need to be prepared (i.e. built in most cases).
+Blocks can be registered in the repo's `/hub` with a build-config.
+The build-script `yarn exe yarn exe site/scripts/prepare-blocks.ts` allows to prepare blocks.
```
CiaranMn, line 78: Might as well tidy this up while we're here
```suggestion
The build-script `yarn exe yarn exe site/scripts/prepare-blocks.ts` prepares blocks.
```

site/README.md
```diff
@@ -73,25 +73,19 @@ When running in development environments, avatars go to the `dev/avatars/(user.i
...
 
 ```sh
-# build one or more blocks
-yarn build-block ./hub/@hash/paragraph.json
-# build all blocks
-yarn build-blocks
+# prepare all blocks
+yarn exe yarn exe site/scripts/prepare-blocks.ts
+
+# prepare one or more blocks
```
CiaranMn, line 84: may match none
```suggestion
# prepare blocks matching a filter (in this example, any in the `hub/@hash` folder)
```

site/scripts/prepare-blocks.ts
```diff
@@ -0,0 +1,483 @@
...
+} as const;
+
+interface StoredBlockInfo {
+  repository: string;
+  commit: string;
+
+  distDir?: string;
+  folder?: string;
+  workspace?: string;
+}
```
CiaranMn, line 43: There's an existing type called `BuildConfig` [here](https://github.com/blockprotocol/blockprotocol/blob/335f799a2fb242256e256925b0c3496f5b7c267b/site/src/lib/blocks.ts#L14)  - we should share one type across these files.

You'll see that `readBlocksFromDisk` is [relying](https://github.com/blockprotocol/blockprotocol/blob/335f799a2fb242256e256925b0c3496f5b7c267b/site/src/lib/blocks.ts#L94) on there being a `branch`, leading to things like the below - we'll need to update this. Potentially to include the commit sha in the URL (we could always hide it from the actual displayed link as it'll be pretty ugly - e.g. just make the display text the repo root, with the link actually going to the specific commit & folder).

The `timestamp` is also relied on to generate a 'last updated at' display on the block page (where timestamp exists) - we should at least remove this for now. Obviously ideally we'd get the last updated time from the commit or similar.

Screenshot 2022-02-23 at 18 43 38

CiaranMn (changes requested):
Thanks Alex - good to have this in TypeScript with the new additional checks / stability.

There were a couple of places in the hub that are relying on fields which this PR removes (`timestamp`, `branch`) - I've flagged a couple of instances. Sharing the type for the build config across the build script and the runtime code should flush any others out.

site/README.md
kachkaev: 👍 Addressed in https://github.com/blockprotocol/blockprotocol/pull/231/commits/257b032aace29c19ed0ac16e1e8683cc6a8809aa

site/README.md
kachkaev: Addressed in 74d69756f058ddd27683891898df597c8cbd9268

site/scripts/prepare-blocks.ts
kachkaev: Good spot, thanks Ciaran! I merged the two typings in 1b97e39f2cb045d6ec59b65b044fe3cb2f102d25. I kept `lastUpdated` in the UI, but hardcoded it to null for now inside `site/src/lib/blocks.ts`. We can extract commit metadata via GitHub API in another PR and bring it back.

Links seem to work now:

Screenshot 2022-02-25 at 09 59 34
Screenshot 2022-02-25 at 09 59 51

thehabbos007:
Might have some more comments, but since we're pinning to a commit sha, would it make sense to link to that hash in the link somehow? https://github.com/blockprotocol/blockprotocol/blob/335f799a2fb242256e256925b0c3496f5b7c267b/site/src/pages/%5Bshortname%5D/blocks/%5BblockSlug%5D.page.tsx#L322
For example it shows 
image
this for a block in the hub - which would be slightly misleading if it was pinned to an older commit than main (or master in this case).

kachkaev:
> this for a block in the hub - which would be slightly misleading if it was pinned to an older commit than main (or master in this case).

Our current links are similar to what’s on NPM. But I agree that it’d be good to have hashes one day – maybe when we sort our block versioning?

site/scripts/prepare-blocks.ts
CiaranMn: I'd still be tempted to include the commit sha in the URL - then you can see the source which is actually powering the block on the hub (@thehabbos007's comment). We don't have to worry about versions since we always show one version on the hub, and we know what exact commit that is built from. 

`getRepositoryUrl(`${repositoryUrl}#${commitHash}`)` should give us the link with commit automatically, which we can then truncate for display purposes to only the repo itself (so we don't have horribly long display URLs).

I don't think this is vital, but it seems a cheap thing to do which will make sure people who care to inspect the source will see the source that is actually being used - are there reasons not to?

kachkaev: I’ve added commit hashes to links. So for example Callout block refers to 

https://github.com/hashintel/hash/tree/ec9e50601629207f1fde76a817b02aa8733762fe/packages/blocks/callout

This is how it is rendered:
Screenshot 2022-03-03 at 15 16 05

Not ideal, but might work for now. It might be better to have something like this in the future:

> Origin 🔗
>
> :octocat: hashintel/hash
> #️⃣ ec9e50601629207f1fde76a817b02aa8733762fe
> 📁 packages/blocks/callout

↑ That’s assuming that we still hard-wire our hub to git repos instead of having a publish process where build files end up in our storage.

CiaranMn (approved):

PR #235, opened 2022-02-24
Fix handling of variants/examples combinations in block hub

CiaranMn:
#202 introduced the ability to handle block 'variants' in the block hub.

There were a couple of cases not handled which this PR addresses:
1. Where a block has an empty `variants` array - the index access is now checked
2. When a block has variants, but no example defined for the variant - this now falls back to the root example

We also intend to start using `name` instead of `displayName` for the variant's name. This PR handles blocks that still use the old `displayName` (we can remove it after a while).


teenoh (dismissed):
Thanks @CiaranMn

teenoh (approved):

PR #237, opened 2022-02-25
Update gray colours to add new gray & sync with designs

nathggns:
Our gray colours are out of date from the Figma, and a new gray has been added in between 50 and 60 (so 60, 70 and 80 are now 70, 80, and 90 and there is a new 60). This requires updating existing instances of these colours. 

[Asana Task](https://app.asana.com/0/1201707629991380/1201837037344195/f) (_internal_)

PR #233, opened 2022-02-24
BP Site styling tweaks

benwerner01:
This PR makes the following style fixes:
- [BP hub: fix centering of sub-heading](https://app.asana.com/0/1201472294283826/1201768159094211)
- [Remove underline from Footer navigation links and change hover color](https://app.asana.com/0/0/1201781321575581)
- [Ensure bottom padding on pages is at least equal to top (see attached)](https://app.asana.com/0/0/1201589261424242)
- [fix Footer "HASH" button hover color](https://app.asana.com/0/0/1201781321575583)
- [Centre align text on block hub](https://app.asana.com/0/0/1201781323994489)
- [Adjust hover scale size on block cards](https://app.asana.com/0/0/1201781323994492)
- [Fix login button in mobile menu](https://app.asana.com/0/0/1201822597391098)
- [Change link colour in blue callout blocks to blue](https://app.asana.com/0/0/1201781323994521)
- [Spacing between headings in docs and specification](https://app.asana.com/0/1201472294283826/1201781323994534)
- [Dashboard typography](https://app.asana.com/0/0/1201781323994524)
- [Padding on block focus page tabs](https://app.asana.com/0/0/1201792603369699)
- [[Duplicate] Hub page: long tabs have no padding (but should)](https://app.asana.com/0/0/1201778873029633)

Not addressed in this PR:
- [BP Site block hub styling tweaks](https://app.asana.com/0/1201095311341924/1201880695835354)
- [BP Site homepage styling tweaks](https://app.asana.com/0/1201095311341924/1201880695835351)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201778474952592

PR #243, opened 2022-02-28
Update neglected gray values

benwerner01:
Follow-up PR from https://github.com/blockprotocol/blockprotocol/pull/237/files updating some `gray` values which weren't updated as part of the initial refactor.


nathggns:
@benwerner01 

Oh man, my bad. I guess my regex search didn't consider that they could be referenced by strings. Great catch & apologies for this. 

There's another one that's been missed I think which was committed after my branch (but didn't conflict so I didn't notice) was created in https://github.com/blockprotocol/blockprotocol/pull/233 

I *think* the usage here needs updating to 70. https://github.com/blockprotocol/blockprotocol/blob/main/site/src/pages/dashboard.page.tsx#L98

benwerner01:
> Oh man, my bad. I guess my regex search didn't consider that they could be referenced by strings. Great catch & apologies for this.

@nathggns not your fault, that could've happened to anyone! Ideally we would be consistent across the codebase about wether we use numbers or strings to access colours in the palette. I've changed all these uses of strings to be numbers as part of this PR, and will do my best to only use numbers from now on.

> There's another one that's been missed I think which was committed after my branch (but didn't conflict so I didn't notice) was created in #233

Good spot, thanks for letting me know. Have confirmed from the original task and you're right this should now be `palette.gray[70]`. Have made that change as part of this PR 👍

nathggns (approved):
Great work!

PR #245, opened 2022-03-02
Remove non-MUI CSS: Tailwind preflight and global.css

CiaranMn:
The design system for blockprotocol.org is implemented via customising MUI theme and components, and styling should be exclusively defined via that route.

An early version of the site used Tailwind (via twind) and other css. Tailwind has a set of 'preflight' css which removes styling on elements, including borders on buttons and inputs, and margin on headings. This is affecting small pockets of the site, including user blocks on the hub ([example](https://blockprotocol-jhbsr7q59-hashintel.vercel.app/@colel/blocks/simple-product-link)), and removing it now saves us from building up other places in which we unknowingly rely on non-MUI styling to achieve the desired visual look.

This PR:
- disables Tailwind `preflight` (we can entirely remove tailwind when we implement new designs for the schema editor, the last thing using it)
- removes `global.css` (mostly CSS relating to selectors which are no longer present)
- fixes a bug whereby leading elements inside `InfoCardWrapper` did not have top margin removed, which is causing the title on [the first spec page](https://blockprotocol.org/spec) to have top margin it shouldn't.

To test: browse the preview link once up and check all looks correct.


MaggieAppleton (approved):

benwerner01 (approved):
Looks good to me, thanks Ciaran 👍

PR #250, opened 2022-03-04
Speed up profile page on Vercel

kachkaev:
This PR speeds up TTFB for pages like `/@hash` or `/types/UUID`. This is done via:

- **Pro-active page warm-ups with a new GitHub workflow**
  This trick significantly  reduces TTFB. See diff comment for details. Related discussions:
  - https://github.com/vercel/vercel/discussions/6878
  - https://github.com/vercel/vercel/discussions/5422
  
  Pipeline example: https://github.com/blockprotocol/blockprotocol/actions/runs/1956402602
  
  Screenshot 2022-03-09 at 09 07 08
   
   Event docs (used in a new GitHub workflow):
   - https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#deployment_status
   - https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule
  - https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#deployment_status

- **Replacing `getServerSideProps` with `getStaticProps` + 60s cache on the profile page**
   The effect of this change is quite small compared to Lambda warm-ups. However, we still benefit from fewer calls to Mongo DB by not calling `getServerSideProps` on every request.

   I haven’t switched from `getServerSideProps` to `getStaticProps` in `site/src/pages/types/[entityTypeId].page.tsx` because we use `req` and `res` there.

---
  - https://app.asana.com/0/0/1201837296904631

### DIFF
--- a/.github/workflows/algolia-upload-index.yml
+++ b/.github/workflows/algolia-upload-index.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/warm-up-vercel.yml
+++ b/.github/workflows/warm-up-vercel.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Fetch URLs that use AWS Lambda
         run: |
           cat << EOF > curl-format.txt
-            %{url}\n
+            %{url_effective}\n
               Status code: %{http_code}\n
               Time: %{time_total}\n
             \n
--- a/.github/workflows/warm-up-vercel.yml
+++ b/.github/workflows/warm-up-vercel.yml
@@ -1,9 +1,13 @@
+## Vercel uses AWS Lambda. When we call getStaticProps in fallback mode or getServerSideProps,
+## a cold Lambda function may take 5-15 seconds to execute. This pipeline warms up Lambda
+## containers for new deployments and keeps them up by fetching affected URLs every 10 minutes.
+
 name: Warm-up Vercel
 
 on:
   deployment_status:
   schedule:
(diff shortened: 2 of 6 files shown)


.github/workflows/algolia-upload-index.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
kachkaev, line 13: We had a mix of `ubuntu-latest` and `ubuntu-20.04` in our yamls. I switched to using `ubuntu-20.04` everywhere to avoid possible breaking changes when `ubuntu-22.04` becomes the default.

.github/workflows/warm-up-vercel.yml
```diff
@@ -0,0 +1,32 @@
+## Vercel uses AWS Lambda. When we call getStaticProps in fallback mode or getServerSideProps,
+## a cold Lambda function may take 5-15 seconds to execute. This pipeline warms up Lambda
+## containers for new deployments and keeps them up by fetching affected URLs every 10 minutes.
```
kachkaev, line 3: 👆 👀

.github/workflows/warm-up-vercel.yml
```diff
@@ -0,0 +1,32 @@
...
+
+name: Warm-up Vercel
+
+on:
+  deployment_status:
+  schedule:
+    - cron: "*/10 * * * *"
```
kachkaev, line 10: This event will work only after we merge the PR. If anything is broken, I’ll submit a quick follow-up PR with a fix.

List of workflow runs: https://github.com/blockprotocol/blockprotocol/actions/workflows/warm-up-vercel.yml

CiaranMn (approved):

PR #251, opened 2022-03-04
Refactor profile tabs navigation

kachkaev:
This PR follows #248 and improves client-side tab navigation. In particular:

- It fixes browser history buttons (←→).
- It eliminates calls to `getServerSideProps` when navigating between the tabs.
- It allows ctrl+clicks on tabs to open then in a new tab.

To achieve this:

- I’ve wrapped tabs with `next/link` with `shallow` = `true`.
- I’ve removed local state and `initialTab`. Thus, `router.query` is the only source of truth now.

Out of scope:

- remove `TABS[].value` (it seems to replicate `slug`)
- improve component and file structure for the profile pages

I tried replacing `site/src/pages/[shortname]/[[...profileTabs]].page.tsx` with `{index,blocks,schemas}.page.tsx`, but this did not work quite well. Even if I re-export the component and `getServerSideProps` in two pages out of three, I can’t eliminate calls to `getServerSideProps` even with shallow routing. We might still consider this later if tab contents become too diverse and we do too much overfetching.

[Asana task](https://app.asana.com/0/0/1201920979480081/f) _(internal)_

---

**UPD**: To avoid using `next/link` in `Tabs`, I introduced a new `BaseLink` component. It focuses only on opening external links in a new tab and is used inside `Link` and `LinkButton`. I also removed `noLinkStyle` from `Link` as we’re not using this prop and it complicates things without any payback.


site/scripts/seed-db.ts
```diff
@@ -66,6 +66,11 @@ const script = async () => {
       preferredName: "Alice",
       email: "alice@example.com",
     },
+    {
+      shortname: "hash",
+      preferredName: "HASH",
+      email: "hash@example.com",
+    },
```
kachkaev, line 73: I was getting 404 when going to `localhost:3000/@hash`, even after seeding the data. Adding this change makes it easy to get started locally for everyone now. Happy to remove the diff I’ve missed some instructions.

site/src/components/pages/user/Tabs.tsx
```diff
@@ -17,88 +19,89 @@ export const TABS = [
     value: "schemas",
     slug: "schemas",
   },
-] as {
-  title: string;
-  value: "overview" | "blocks" | "schemas";
-  slug: "" | "blocks" | "schemas";
-}[];
+] as const;
```
kachkaev, line 22: Simplifies typings

site/src/components/pages/user/Tabs.tsx
```diff
@@ -17,88 +19,89 @@ export const TABS = [
...
-          key={value}
-          label={
-            <Box
-              sx={{
-                display: "flex",
-                alignItems: "center",
-              }}
-            >
-              {title}
-              {value !== "overview" && (
```
kachkaev, line 62: 1/2 Most of the diff in this file helps avoid code like this ...

site/src/components/pages/user/Tabs.tsx
```diff
@@ -17,88 +19,89 @@ export const TABS = [
...
-                    sx={{
-                      color: ({ palette }) =>
-                        value === activeTab
-                          ? palette.purple[600]
-                          : palette.gray[70],
-                    }}
-                  >
-                    {value === "blocks"
-                      ? tabItemsCount.blocks
-                      : tabItemsCount.schemas}
```
kachkaev, line 91: 2/2 ... and this

site/scripts/seed-db.ts
akash-joshi: You're right. I think this is a good addition to avoid local development confusion.

site/src/components/pages/user/Tabs.tsx
```diff
@@ -1,5 +1,7 @@
 import { VFC, FC } from "react";
 import { Box, Tabs, Typography, BoxProps, Tab } from "@mui/material";
+// eslint-disable-next-line no-restricted-imports
```
akash-joshi, line 3: @teenoh can confirm, but you won't have to do this if you use our `Link` with the `noLinkStyles` prop.

site/src/pages/[shortname]/[[...profileTabs]].page.tsx
```diff
@@ -42,36 +55,31 @@ export const getServerSideProps: GetServerSideProps<
...
-  entityTypes,
-  initialActiveTab,
-}) => {
+const UserPage: NextPage<UserPageProps> = ({ user, blocks, entityTypes }) => {
+  const router = useRouter();
+
+  const matchingTab = findTab(router.query.profileTabs);
+
+  // Protect against unlikely client-side navigation to a non-existing profile tab
+  if (!matchingTab) {
```
Comment (akash-joshi) [inline site/src/pages/[shortname]/[[...profileTabs]].page.tsx:73]: Good addition!

akash-joshi (commented):
Looks okay to me, ran this locally as well. Would just like to have this checked out by Teenoh: https://github.com/blockprotocol/blockprotocol/pull/251/files#r820427494

site/src/components/pages/user/Tabs.tsx
kachkaev: AFAIU, our Link component renders the child anchor tag, which we don’t want here. Judging by its code but I may be missing something here. All we want from Link around Tab is the right href and the right ctrl+click logic.

kachkaev: Happy to update the diff if needed.

CiaranMn: The only reason we don't want the anchor tag here is because `Tab` is already rendered as one. If we passed `component: "span"` to `Tab`, we could let the custom `Link` render its anchor. 

But it would also need to accept children and put them within the anchor.

kachkaev: This might not work if Tabs component expects Tab as a direct DOM child. Styles may break. So in the end it may end up more complicated than with next/link.

CiaranMn: The alternative is to remove `<Anchor ref={ref} {...other} />` from our link - NextJS will automatically insert an anchor tag if the children of `<NextLink>` is just plain text. If it is any other element, it will pass them in as-is. I believe this is the behavior we want. Then, we can pass `Tab` to our Link component, and it'll just pass it through.

CiaranMn: Remove it and replace it with the children

CiaranMn: It does: https://github.com/vercel/next.js/blob/ff6b75027b98ffc379282cb807551d1b5e9e7732/packages/next/client/link.tsx#L224

kachkaev: Interesting! That might be a legacy feature because it’s not even mentioned in the docs. It won’t work in our case anyway because `children` in our case is `<Tab />`, not a `string` 🤔 

I am not saying that we can’t use our custom Link component, but I’m not sure I see the value in doing this. AFAIU, we’ll need to add some new prop that would just toggle a plain mode, i.e. fallback to standard `next/link` behaviour.

What we can do instead is consider creating a new component that would combine `Tabs` / `Tab` with `next/link`. We can explore this when we need to support tabs with links on more than one page.

CiaranMn: I was suggesting that if we replace `Anchor` in our own `Link` with just passing `children`, we would not need to change anything else. We could just pass `Tab` to it, since `Tab` already renders as an anchor.

But I don't think it's super important if there are other constraints/issues in doing so.

packages/blockprotocol/core.d.ts
```diff
@@ -62,7 +62,7 @@ export type BlockMetadata = {
   /**
    * The dependencies a block relies on but expects the embedding application to provide
    */
-  externals: {
+  externals?: {
```
kachkaev, line 65: Adding `?` fixed a TSC issue for me around `blocksData as BlockMetadata[]` in `site/src/pages/api/blocks.api.ts`. `blocksData` comes from `site/blocks-data.json`, where some elements don’t have `externals`. This error does not show up in CI because `blocksData` is an empty array when blocks are not built.

site/src/components/pages/user/Tabs.tsx
kachkaev: I tried replacing `Anchor` with `children`, but it made `Link` even more convoluted. To save us from using `no-restricted-imports` in multiple files, I created a new file called `BaseLink` which is a minimal wrapper around `next/link`. It is only responsible for opening all external URLs in a new tab.

CiaranMn:
> Looks okay to me, ran this locally as well. Would just like to have this checked out by Teenoh: https://github.com/blockprotocol/blockprotocol/pull/251/files#r820427494

What was this, Akash? The link isn't working

CiaranMn:
> Looks okay to me, ran this locally as well. Would just like to have this checked out by Teenoh: https://github.com/blockprotocol/blockprotocol/pull/251/files#r820427494

What was this, Akash? The link isn't working

akash-joshi:
@CiaranMn it was related to this change: https://github.com/blockprotocol/blockprotocol/pull/251/files#diff-b3bfb396acc1e9352fc2974a14faf6c886a7e1546ce5f757cf43d413a0e0f52dL5

Since Ben and Alex resolved it, I think this PR's good to go.

CiaranMn (dismissed):

akash-joshi (approved):
Looks Good to Me.

PR #252, opened 2022-03-05
Add @kickstartDS Button component

julrich:
Adds the `Button` component (part of the @kickstartds Design System: https://www.kickstartDS.com), as a PoC to see the potential compatibility.

Happy to debug with you guys, if this maybe doesn't quite work yet (would surprise me if it just works :D).
Also quite possible that I'm using it wrong, just let me know!


CiaranMn (changes requested):
Thanks @julrich! Great to see the first kickstartDS block 😁 

Not necessary to merge this, but we are going to need to expand the spec to cover UI primitives - at the moment, it's not clear how the embedding application is supposed to know when the user clicked the button, since the existing defined operations in the spec are related to creating or updating entities, not reporting actions. We could, for example, add an agreed set of actions for UI primitives such as `onClick`, which follow DOM events.  

**Change requested**
In the meantime, I'd like to get your block published - we had to fix something in #253 to support this, and there's one more thing I'd appreciate your help - in your source repo, could you edit [this file](https://github.com/kickstartDS/blocks/blob/main/packages/blocks/button/webpack-block-metadata-plugin.js) as follows please? 

Screenshot 2022-03-14 at 11 16 15

**Explanation** 
webpack is automatically generating a `main.[hash].js.LICENSE.txt` file, which our code to detect the main source file is not handling.  This means that the generated metadata file after running `yarn build`, `dist/block-metadata.json`, incorrectly contains: `"source": "main.cc850050b93e78f6990d.js.LICENSE.txt"`. We have fixed this in the template in #261. After applying the above fix to your source, it should correctly identify the `source` file as the JS one (not the LICENSE.txt one).

Once the source repo is updated, please update the commit here to reflect the latest. Then it should be good to go.

This is how it'll appear on the hub and on its page:
Screenshot 2022-03-14 at 11 09 59

Screenshot 2022-03-14 at 11 12 39

Thanks!

PR #253, opened 2022-03-08
Accept all npm-supported repository URLs in blocks

akash-joshi:
## What this Does

The blockprotocol website presently only builds blocks which are in the GitHub URL format, ie, `https://github.com/user/repository`. However, users should be free to add any npm-supported repository link, as long as we can parse it. 

Hence, this PR allows blockprotocol to do just that, by utilizing the `npm/hosted-git-info` package. We use it to get the source URL from any repository URL, and use it to download and build the blocks.

## How to test this?

Run `yarn build` locally, or have a look at the CI output below.

---
  - https://app.asana.com/0/0/1201934058330698

### DIFF
--- a/site/scripts/prepare-blocks.ts
+++ b/site/scripts/prepare-blocks.ts
@@ -189,7 +189,7 @@ const ensureRepositorySnapshot = async ({
     console.log(`Repository snapshot ready in ${repositorySnapshotDirPath}`);
     return repositorySnapshotDirPath;
   } finally {
-    await cleanupTarDirPath();
+    await cleanupTarDir();
   }
 };
 
--- a/site/scripts/prepare-blocks.ts
+++ b/site/scripts/prepare-blocks.ts
@@ -163,11 +163,16 @@ const ensureRepositorySnapshot = async ({
       defaultExecaOptions,
     );
 
-    console.log(chalk.green(`Unpacking archive...`));
     const untarPath = path.resolve(untarDirPath, "repo");
+
+    console.log(chalk.green(`Creating output directory...`));
+    await execa("mkdir", ["-p", untarPath]);
+
+    console.log(chalk.green(`Unpacking archive...`));
+
     await execa(
       "tar",
-      ["-xfv", path.resolve(untarDirPath, "repo.tar.gz"), "-C", untarPath],
+      ["-xvf", path.resolve(untarDirPath, "repo.tar.gz"), "-C", untarPath],
       defaultExecaOptions,
     );
 
--- a/site/scripts/prepare-blocks.ts
+++ b/site/scripts/prepare-blocks.ts
@@ -148,7 +148,7 @@ const ensureRepositorySnapshot = async ({
     return repositorySnapshotDirPath;
   }
 
-  const { path: tarDirPath, cleanup: cleanupTarDirPath } = await tmp.dir({
+  const { path: tarDirPath, cleanup: cleanupTarDir } = await tmp.dir({
     unsafeCleanup: true,
   });
 
--- a/site/scripts/prepare-blocks.ts
+++ b/site/scripts/prepare-blocks.ts
@@ -148,42 +148,48 @@ const ensureRepositorySnapshot = async ({
     return repositorySnapshotDirPath;
   }
 
-  const { path: untarDirPath, cleanup: cleanupUntarDirPath } = await tmp.dir({
+  const { path: tarDirPath, cleanup: cleanupTarDirPath } = await tmp.dir({
     unsafeCleanup: true,
   });
 
   try {
     console.log(chalk.green(`Downloading ${tarballUrl}...`));
-    // @todo Consider finding cross-platform NPM packages for curl and untar commands
+    // @todo Consider finding cross-platform NPM packages for curl and tar commands
     await execa(
       "curl",
-      ["-sL", "-o", path.resolve(untarDirPath, `repo.tar.gz`), tarballUrl],
+      ["-sL", "-o", path.resolve(tarDirPath, `repo.tar.gz`), tarballUrl],
       defaultExecaOptions,
     );
 
-    const untarPath = path.resolve(untarDirPath, "repo");
+    const outputDirPath = path.resolve(tarDirPath, "repo");
 
-    console.log(chalk.green(`Creating output directory...`));
-    await execa("mkdir", ["-p", untarPath]);
+    console.log(chalk.green(`Ensuring output directory exists...`));
+    await fs.ensureDir(outputDirPath);
 
     console.log(chalk.green(`Unpacking archive...`));
 
     await execa(
       "tar",
-      ["-xvf", path.resolve(untarDirPath, "repo.tar.gz"), "-C", untarPath],
+      [
+        "--extract",
+        `--file=${path.resolve(tarDirPath, "repo.tar.gz")}`,
+        `--directory=${outputDirPath}`,
+        "--verbose",
+      ],
       defaultExecaOptions,
     );
 
-    const innerDir = await fs.readdir(untarPath);
+    const innerDirName = (await fs.readdir(outputDirPath))[0]!;
+
     await fs.move(
-      path.resolve(untarPath, innerDir[0]!),
+      path.resolve(outputDirPath, innerDirName),
       repositorySnapshotDirPath,
     );
 
     console.log(`Repository snapshot ready in ${repositorySnapshotDirPath}`);
     return repositorySnapshotDirPath;
   } finally {
-    await cleanupUntarDirPath();
+    await cleanupTarDirPath();
   }
 };
 
--- a/site/scripts/prepare-blocks.ts
+++ b/site/scripts/prepare-blocks.ts
@@ -21,6 +21,8 @@ import md5 from "md5";
 import Ajv, { JSONSchemaType } from "ajv";
 import tmp from "tmp-promise";
 import slugify from "slugify";
+import hostedGitInfo from "hosted-git-info";
+
 import { StoredBlockInfo } from "../src/lib/blocks";
 
 const monorepoRoot = path.resolve(__dirname, "../..");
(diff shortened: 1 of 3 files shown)


site/scripts/prepare-blocks.ts
```diff
@@ -154,37 +148,42 @@ const ensureRepositorySnapshot = async ({
...
+    console.log(chalk.green(`Creating output directory...`));
+    await execa("mkdir", ["-p", untarPath]);
+
     console.log(chalk.green(`Unpacking archive...`));
-    const unzipPath = path.resolve(unzipDirPath, "repo");
+
     await execa(
-      "unzip",
-      ["-q", path.resolve(unzipDirPath, "repo.zip"), "-d", unzipPath],
+      "tar",
```
akash-joshi, line 172: Replace `unzip` with the `tar` command to extract the tarball.

site/scripts/prepare-blocks.ts
```diff
@@ -154,37 +148,42 @@ const ensureRepositorySnapshot = async ({
...
       "curl",
-      ["-sL", "-o", path.resolve(unzipDirPath, `repo.zip`), zipUrl],
+      ["-sL", "-o", path.resolve(untarDirPath, `repo.tar.gz`), tarballUrl],
       defaultExecaOptions,
     );
 
+    const untarPath = path.resolve(untarDirPath, "repo");
+
+    console.log(chalk.green(`Creating output directory...`));
+    await execa("mkdir", ["-p", untarPath]);
```
akash-joshi, line 167: Since `unzip` doesn't create the output directory for us, we do that manually here. The `-p` flag allows us to do it in an idempotent way.

site/scripts/prepare-blocks.ts
```diff
@@ -117,26 +119,18 @@ const ensureRepositorySnapshot = async ({
...
-    if (repositoryHref.startsWith("https?://github.com")) {
-      throw new Error(
-        `Cannot handle repository URL ${repositoryHref}. Only GitHub links are currently supported. We will add more services based on the demand.`,
-      );
-    } else {
-      throw new Error(
-        `Cannot handle repository URL ${repositoryHref}. It needs to match https://github.com/org/repo`,
-      );
-    }
+  const tarballUrl = hostedGitInfo
```
akash-joshi, line 122: We generate the `tarball` URL in a one-shot fashion and use it throughout the code below.

site/scripts/prepare-blocks.ts
```diff
@@ -154,37 +148,42 @@ const ensureRepositorySnapshot = async ({
     return repositorySnapshotDirPath;
   }
 
-  const { path: unzipDirPath, cleanup: cleanupUnzipDirPath } = await tmp.dir({
+  const { path: untarDirPath, cleanup: cleanupUntarDirPath } = await tmp.dir({
```
kachkaev, line 151:
```suggestion
  const { path: tarDirPath, cleanup: cleanupTarDirPath } = await tmp.dir({
```

`unzip` was chosen after the `unzip` command. Since we call `tar` instead of `untar`, maybe we don’t need `untar` as a term anywhere?

site/scripts/prepare-blocks.ts
```diff
@@ -154,37 +148,42 @@ const ensureRepositorySnapshot = async ({
...
     unsafeCleanup: true,
   });
 
   try {
-    console.log(chalk.green(`Downloading ${zipUrl}...`));
-    // @todo Consider finding cross-platform NPM packages for curl and unzip commands
+    console.log(chalk.green(`Downloading ${tarballUrl}...`));
+    // @todo Consider finding cross-platform NPM packages for curl and untar commands
```
kachkaev, line 157:
```suggestion
    // @todo Consider finding cross-platform NPM packages for `curl` and `tar` commands
```

site/scripts/prepare-blocks.ts
```diff
@@ -154,37 +148,42 @@ const ensureRepositorySnapshot = async ({
...
     await execa(
       "curl",
-      ["-sL", "-o", path.resolve(unzipDirPath, `repo.zip`), zipUrl],
+      ["-sL", "-o", path.resolve(untarDirPath, `repo.tar.gz`), tarballUrl],
       defaultExecaOptions,
     );
 
+    const untarPath = path.resolve(untarDirPath, "repo");
```
kachkaev, line 164: `const untarPath` → `const tarOutputDirPath` / `const outputDirPath`

site/scripts/prepare-blocks.ts
kachkaev: We don’t need `execa` here:


```suggestion
    await fs.ensureDir(...)
```

https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureDir.md

site/scripts/prepare-blocks.ts
```diff
@@ -154,37 +148,42 @@ const ensureRepositorySnapshot = async ({
...
     await execa(
-      "unzip",
-      ["-q", path.resolve(unzipDirPath, "repo.zip"), "-d", unzipPath],
+      "tar",
+      ["-xvf", path.resolve(untarDirPath, "repo.tar.gz"), "-C", untarPath],
       defaultExecaOptions,
     );
 
-    const innerDir = await fs.readdir(unzipPath);
+    const innerDir = await fs.readdir(untarPath);
```
kachkaev, line 177:
```suggestion
    const innerDirName = (await fs.readdir(...))[0]!;
```

```suggestion
    const itemsInsideOutputDir = await fs.readdir(outputDirPath);
```

The original naming was not perfect – my bad. Because `fs.readdir` returns a list of items, it’d be nice to reflect that in variable name or extract the element we need right away. So plural for an array or `[0]!` on the same line.

I prefer to be explicit about what a fs-related variable contains. Hence suffixes like `xyzDirPath`, `abcFileName` etc. When suffixes are omitted, it’s easier to pass wrong variables around as args.

site/scripts/prepare-blocks.ts
```diff
@@ -154,37 +148,42 @@ const ensureRepositorySnapshot = async ({
...
+    await execa("mkdir", ["-p", untarPath]);
+
     console.log(chalk.green(`Unpacking archive...`));
-    const unzipPath = path.resolve(unzipDirPath, "repo");
+
     await execa(
-      "unzip",
-      ["-q", path.resolve(unzipDirPath, "repo.zip"), "-d", unzipPath],
+      "tar",
+      ["-xvf", path.resolve(untarDirPath, "repo.tar.gz"), "-C", untarPath],
```
kachkaev, line 173: Optional: if `tar` has aliases for `-xvf` and `-C`, let’s use them them instead. So we’ll have: `["--option1", "--option2", ....]`.

I usually avoid option shortcuts in all my code and docs, but I have missed this somehow in this file. Shortcuts are more cryptic, so the code is harder to follow and maintain. Shortcuts are still useful in CLI when we type one-off commands from a keyboard.

kachkaev (changes requested):
Thanks Akash! 👍 The change makes sense, I just left a few minor suggestions to consider.

kachkaev:
We’ll also need to update `site/src/pages/[shortname]/blocks/[blockSlug].page.tsx`

```tsx
                  <Link href={blockMetadata.repository}>
                    {/* Show `github.com/org/repo` instead of full URL with protocol, commit hash and path */}
                    {blockMetadata.repository
                      .replace(/^https?:\/\//, "")
                      .split("/")
                      .slice(0, 3)
                      .join("/")}
                  </Link>
```

ATM, we shorten the URL unconditionally because we assume it’s a GitHub repo. If we say that we support more types of URLs, we should call `.slice(0, 3)` after matching the URL against a regexp. We can probably extract what we do into a local function called, say, `generateRepositoryDisplayUrl`.

kachkaev:
I was not adding support for other repos in #231 because I thought we’d switch to some other approach sooner than this would ben necessary. What is the practical problem we are solving?

CiaranMn:
> Asana task

The context for this is that when adding a JSON to hub/ to add a block, we currently only allow users to input a repository URL in the format https://gituhub.com/namespace/project-slug. This means that URLs like `git@github.com:namespace/repo.git` fail, e.g. [this one](https://github.com/blockprotocol/blockprotocol/pull/252)

site/src/pages/[shortname]/blocks/[blockSlug].page.tsx
```diff
@@ -101,6 +102,18 @@ const parseQueryParams = (params: BlockPageQueryParams) => {
   return { shortname, blockSlug };
 };
 
+// Show `github.com/org/repo` instead of full URL with protocol, commit hash and path
+const generateRepositoryUrl = (repository: string): string => {
```
Comment (akash-joshi) [inline site/src/pages/[shortname]/blocks/[blockSlug].page.tsx:106]: @kachkaev this function is responsible for generating the repository URL.

site/src/pages/[shortname]/blocks/[blockSlug].page.tsx
```diff
@@ -123,25 +136,32 @@ export const getStaticProps: GetStaticProps<
     return { notFound: true };
   }
 
+  // getStaticProps doesn't parse undefined, so returning an empty string instead
+  const repositoryDisplayUrl = blockMetadata.repository
```
Comment (akash-joshi) [inline site/src/pages/[shortname]/blocks/[blockSlug].page.tsx:140]: Decided to generate the display URL in `getStaticProps` since we can generate it in one-shot here, without ensuring any props-based reactivity.

Comment (kachkaev) [reply site/src/pages/[shortname]/blocks/[blockSlug].page.tsx:140]: My preference would be to call this transform as close as possible to where we display the link. However, I’m not strongly pro this. Ideally, initial page props should be orthogonal to each other, i.e. completely independent. Anything we can derive from them can go into components and we can use `React.memo` / `useMemo` if needed for some heavy stuff. That’s not the case here.

Because `repositoryDisplayUrl` can be derived from `blockMetadata`, our page data gets bigger but does not carry more information. Besides, if we decide to change the page logic later, it is more likely that we’ll keep the above lines but not use the result anywhere.

site/src/lib/blocks.ts
```diff
@@ -8,14 +8,14 @@ export type ExpandedBlockMetadata = BlockMetadata & {
   blockPackagePath: string;
   lastUpdated?: string | null;
   packagePath: string;
+  // repository is passed down as a string upon expansion
```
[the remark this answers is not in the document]
akash-joshi: Yes. The original type `ExpandedBlockMetadata` is building on, ie, `BlockMetadata`, is a union type of string and the npm repository format. In this case, it's being simplified to just a string during expansion, so I think this comment is a helpful indicator of that.

site/src/pages/[shortname]/blocks/[blockSlug].page.tsx
```diff
@@ -101,6 +102,18 @@ const parseQueryParams = (params: BlockPageQueryParams) => {
   return { shortname, blockSlug };
 };
 
+// Show `github.com/org/repo` instead of full URL with protocol, commit hash and path
+const generateRepositoryUrl = (repository: string): string => {
+  const repositoryUrlObject = new URL(repository);
+  const displayUrl = `${repositoryUrlObject.hostname}${repositoryUrlObject.pathname}`;
```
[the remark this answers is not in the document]
Comment (akash-joshi) [reply site/src/pages/[shortname]/blocks/[blockSlug].page.tsx:111]: I understand the point about port, which can be utilized via `host` instead of `hostname`. Since we're trying to not display all of the URL information here, as we're hiding the commit and branch for GitHub URLs, doesn't it make sense to hide the extra info, eg, query params, for other URLs as well?

This URL is just used for display data, not as the `href`.

site/src/pages/[shortname]/blocks/[blockSlug].page.tsx
Comment (akash-joshi) [reply site/src/pages/[shortname]/blocks/[blockSlug].page.tsx:143]: > Ideally, initial page props should be orthogonal to each other, i.e. completely independent.

Makes sense.

site/src/pages/[shortname]/blocks/[blockSlug].page.tsx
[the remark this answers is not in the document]
Comment (kachkaev) [reply site/src/pages/[shortname]/blocks/[blockSlug].page.tsx:111]: Oh right, makes sense! 👍

site/src/pages/[shortname]/blocks/[blockSlug].page.tsx
```diff
@@ -123,25 +138,27 @@ export const getStaticProps: GetStaticProps<
     return { notFound: true };
   }
 
+  // getStaticProps doesn't parse undefined, so returning an empty string instead
+
```
Comment (kachkaev) [inline site/src/pages/[shortname]/blocks/[blockSlug].page.tsx:142]: 👀

Comment (akash-joshi) [reply site/src/pages/[shortname]/blocks/[blockSlug].page.tsx]: Arghh, removed :D

kachkaev (approved):
Thanks @akash-joshi 👏👏👏

PR #256, opened 2022-03-09
Tweak blockprotocol types and set to 0.0.7

CiaranMn:
This updates a few `blockprotocol` types as follows:

1. We had a few places where we use a `Record<string, unknown>` - create a re-usable type for that.
2. Use `JSONObject` for values in metadata files where one is required
3. Add `schema` as a required field on block metadata, and mention it in the spec - we previously had specified in the spec as `block-schema.json`, but we don't need it to be a specified name since we can look it up from `block-metadata.json`, and requiring a particular name prevents blocks referencing/re-using external schemas as the value for `schema`.
4. Remove type arguments/generics from CRUD operations, since they were only specifying the shape of the data being passed, EXCEPT FOR `BlockProtocolAggregateEntitiesResult` which is doing double duty for aggregations over types and entities
5. Update return types on functions from `unknown[]` to `BlockProtocolEntity[]` or `BlockProtocolEntityType[]` - these have structures which blocks should be able to rely on.
6.. Throw an error in `mock-block-dock` functions where entities aren't found, so that they return entities reliably (not `undefined`)
7. Correct an inconsistency / typo: `BlockProtocolUpdateLinkAction` -> `BlockProtocolUpdateLinksAction`

**To consider in follow up**
- Can we make `UnknownRecord` all `JSONObject` - i.e. make the potential values `JSONValue` rather than `unknown`? at the moment it encourages blocks to deal with entity values which are potentially `undefined`, functions, etc - do we want this?
- What should the behavior be for BP functions which attempt to find entities but cannot? e.g. for retrieval or update. This change implies an error should be thrown rather than returning null/undefined - we should specify it.


akash-joshi (commented):
Looks fine to me.

kachkaev (approved):

PR #258, opened 2022-03-11
Update HASH blocks

kachkaev:
This PR updates block commit hashes to publish changes in https://github.com/hashintel/hash/pull/398 and https://github.com/hashintel/hash/pull/399

\+ see a [small bonus](https://github.com/blockprotocol/blockprotocol/pull/258#discussion_r824829920)

### DIFF
--- a/site/scripts/prepare-blocks.ts
+++ b/site/scripts/prepare-blocks.ts
@@ -174,7 +174,6 @@ const ensureRepositorySnapshot = async ({
         "--extract",
         `--file=${path.resolve(tarDirPath, "repo.tar.gz")}`,
         `--directory=${outputDirPath}`,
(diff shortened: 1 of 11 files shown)


site/scripts/prepare-blocks.ts
```diff
@@ -174,7 +174,6 @@ const ensureRepositorySnapshot = async ({
         "--extract",
         `--file=${path.resolve(tarDirPath, "repo.tar.gz")}`,
         `--directory=${outputDirPath}`,
-        "--verbose",
```
kachkaev, line 177: We discussed this flag with @akash-joshi in DMs.

This change improves Vercel logs. Before:
image

PR #259, opened 2022-03-11
Handle non-default block exports, detect non-React

CiaranMn:
We currently assume that:
1. Blocks are React components
2. They are the default export from the entry source file

We should implement support for non-React components as they are submitted – until we decide on a prescribed interface for the component. This PR throws an error if something that definitely isn't a React component is encountered, so that the problem is explicit and can be addressed.

We should also handle non-default exports. This PR does that, at least for (1) named `App` exports, and (2) named exports where they are the only one in the file. Since we can't reliably detect other patterns.

[Asana task](https://app.asana.com/0/1200211978612931/1201955153125192/f) (_internal_)

### DIFF
--- a/site/src/pages/[shortname]/blocks/[blockSlug].page.tsx
+++ b/site/src/pages/[shortname]/blocks/[blockSlug].page.tsx
@@ -10,7 +10,7 @@ import {
 import { GetStaticPaths, GetStaticProps, NextPage } from "next";
 import Head from "next/head";
 import { useRouter } from "next/router";
-import React, { useMemo, VoidFunctionComponent } from "react";
+import React, { ComponentType, useMemo, VoidFunctionComponent } from "react";
 import { formatDistance } from "date-fns";
 import { BlocksSlider } from "../../../components/BlocksSlider";
 import {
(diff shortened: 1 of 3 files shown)


nathggns (approved):
We should ensure similar render behaviour in HASH, but great improvement.

site/src/pages/[shortname]/blocks/[blockSlug].page.tsx
```diff
@@ -35,15 +35,36 @@ const blockRequire = (name: BlockDependency) => {
   return blockDependencies[name];
 };
 
-const blockEval = (source: string): BlockExports => {
+const blockComponentFromSource = (source: string): ComponentType => {
+  // this does not guarantee filtering out all non-React components
+  if (!source.includes("createElement")) {
+    throw new Error(
+      "Block is not a React component - please implement rendering support for whatever it is and update this check. Note that we may in future change the recommended implementation for blocks to avoid this issue.",
```
Comment (nonparibus) [inline site/src/pages/[shortname]/blocks/[blockSlug].page.tsx:42]: @CiaranMn Within this error we should provide a link to whichever best public thread (GitHub Issue, Discussion or other) we have that touches on this, in case someone wants to join the conversation. This is a good point to capture them.

PR #260, opened 2022-03-11
Bump TS to 4.6.2, enable `noUncheckedIndexedAccess`

kachkaev:
The main purpose of this PR is to add [`"noUncheckedIndexedAccess"`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) to our repo‘s `tsconfig`. I have also upgraded TS from 4.5 to 4.6 given the opportunity. See [release announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/).

Most cases of unchecked indexes are ‘fixed’ by adding `!`. This means we can still get runtime errors for the existing code, and if we do, we can refactor it case-by-case. Newly written code will be better type-checked from now on. In general, we should avoid `!` as much as possible.

- [Asana task](https://app.asana.com/0/0/1201950599278304) _(internal)_
- [Slack thread](https://hashintel.slack.com/archives/C022217GAHF/p1645736187865989) _(internal)_
- https://github.com/hashintel/hash/pull/401

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9314,7 +9314,12 @@ typescript-json-schema@^0.53.0:
     typescript "~4.5.0"
     yargs "^17.1.1"
 
-typescript@^4.5.5, typescript@~4.5.0:
+typescript@^4.6.2:
+  version "4.6.2"
+  resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.6.2.tgz#fe12d2727b708f4eef40f51598b3398baa9611d4"
+  integrity sha512-HM/hFigTBHZhLXshn9sN37H085+hQGeJHJ/X7LpBWLID/fbc2acUMfU+lGD98X81sKP+pFa9f0DZmCwB9GnbAg==
+
+typescript@~4.5.0:
   version "4.5.5"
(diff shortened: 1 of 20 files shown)


yarn.lock
```diff
@@ -9314,7 +9314,12 @@ typescript-json-schema@^0.53.0:
     typescript "~4.5.0"
     yargs "^17.1.1"
 
-typescript@^4.5.5, typescript@~4.5.0:
+typescript@^4.6.2:
+  version "4.6.2"
+  resolved "https://registry.yarnpkg.com/typescript/-/typescript-4.6.2.tgz#fe12d2727b708f4eef40f51598b3398baa9611d4"
+  integrity sha512-HM/hFigTBHZhLXshn9sN37H085+hQGeJHJ/X7LpBWLID/fbc2acUMfU+lGD98X81sKP+pFa9f0DZmCwB9GnbAg==
+
+typescript@~4.5.0:
```
kachkaev, line 9322:
```
yarn why typescript
...
=> Found "typescript-json-schema#typescript@4.5.5"
info This module exists because "_project_#block-template#typescript-json-schema" depends on it.
...
```

Should be fine till we can upgrade `typescript-json-schema`. Related PR: https://github.com/YousefED/typescript-json-schema/pull/469

nathggns (approved):

PR #263, opened 2022-03-14
Upgrade Person Block

akash-joshi:
## What this Does

This PR is a quick follow-up to https://github.com/hashintel/hash/pull/350, updating the commit on the `person` block to update the person-block demo in hub to the latest commit.

The `verbose` flag [was also removed](https://github.com/blockprotocol/blockprotocol/pull/263/files#diff-1bfd9bc24cbe283f91405cd2d044a05277ad4dc7a566503dfcaa52d5837c1858L177) from the `tar` command on @kachkaev's request to prevent CI log pollution.

---
  - https://app.asana.com/0/0/1201884100967648

### DIFF
--- a/hub/@hash/person.json
+++ b/hub/@hash/person.json
@@ -1,6 +1,6 @@
 {
   "repository": "https://github.com/hashintel/hash.git",
-  "commit": "ec9e50601629207f1fde76a817b02aa8733762fe",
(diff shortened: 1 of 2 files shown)


CiaranMn:
These changes applied in #258, except with a later commit for the Person block

PR #273, opened 2022-03-21
Allow APIs to optionally be made public

akash-joshi:
## What this Does?

This PR does some pre-work for https://github.com/hashintel/hash/pull/407, allowing the BlockProtocol `/blocks` API to be accessible from all clients. This is done so that a client which consumes BlockProtocol blocks with an appropriate API key isn't blocked by CORS errors while trying to fetch the desired data.

## What does this change?

An optional `isPublicApi` option is added to `createBaseHandler` [here](https://github.com/blockprotocol/blockprotocol/pull/273/files#diff-bf353d5aba9cfe2a44444fb44f092418bf23327517de62a90af518a061ac1583R32) to optionally allow an API to be accessible from all CORS sources.

---
  - https://app.asana.com/0/0/1201997200034318

### DIFF
--- a/site/src/lib/api/handler/apiKeyRequiredHandler.ts
+++ b/site/src/lib/api/handler/apiKeyRequiredHandler.ts
@@ -10,4 +10,4 @@ export const createApiKeyRequiredHandler = <
   RequestBody = unknown,
   Response = unknown,
 >(): NextConnect<BaseApiRequest<RequestBody>, BaseApiResponse<Response>> =>
-  createBaseHandler().use(hasValidApiKeyMiddleware);
+  createBaseHandler({ isPublicApi: true }).use(hasValidApiKeyMiddleware);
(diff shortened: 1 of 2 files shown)


site/src/lib/api/handler/apiKeyRequiredHandler.ts
```diff
@@ -10,4 +10,4 @@ export const createApiKeyRequiredHandler = <
   RequestBody = unknown,
   Response = unknown,
 >(): NextConnect<BaseApiRequest<RequestBody>, BaseApiResponse<Response>> =>
-  createBaseHandler().use(hasValidApiKeyMiddleware);
+  createBaseHandler({ isPublicApi: true }).use(hasValidApiKeyMiddleware);
```
akash-joshi, line 13: The assumption I'm making here is that all APIs which need an API key to function should be accessible by external clients.

CiaranMn (approved):

PR #276, opened 2022-03-30
Update Blocks API to return icon URLs and unique IDs 

akash-joshi:
## What this Does

This PR is a follow-up to https://github.com/hashintel/hash/pull/407, to fix the temporary solutions defined there. This PR: 
1. Returns absolute URLs for icons.
2. Returns a unique `componentId` for each block.

So that any BlockProtocol client can use these details directly. The icons are used in the clients to display a block's icon, while the `componentId` is used to uniquely identify a block in the hash app, and can also be used as a key for objects storing block data.

Changes described in the GH comments below.

---
  - https://app.asana.com/0/0/1201988634400654
  - https://app.asana.com/0/0/1202003755787889
  - https://app.asana.com/0/0/1201988634400672

### DIFF
--- a/packages/blockprotocol/core.d.ts
+++ b/packages/blockprotocol/core.d.ts
@@ -27,7 +27,7 @@ export type BlockVariant = {
   /**
    * @deprecated - Use the `name` field instead.
    */
-  displayName: string;
+  displayName?: string;
   properties: JSONObject;
   examples?: JSONObject[] | null;
 };
--- a/site/src/lib/blocks.ts
+++ b/site/src/lib/blocks.ts
@@ -112,10 +112,12 @@ export const readBlocksFromDisk = (): ExpandedBlockMetadata[] => {
         icon: generateBlockFileUrl(metadata.icon, metadata.packagePath),
         image: generateBlockFileUrl(metadata.image, metadata.packagePath),
         source: generateBlockFileUrl(metadata.source, metadata.packagePath)!,
-        variants: metadata.variants?.map((variant) => ({
-          ...variant,
-          icon: generateBlockFileUrl(variant.icon, metadata.packagePath)!,
-        })),
+        variants: metadata.variants?.length
+          ? metadata.variants?.map((variant) => ({
+              ...variant,
+              icon: generateBlockFileUrl(variant.icon, metadata.packagePath)!,
+            }))
+          : null,
         schema: generateBlockFileUrl(metadata.schema, metadata.packagePath)!,
         repository,
         blockPackagePath: `/${metadata.packagePath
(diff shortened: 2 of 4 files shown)


packages/blockprotocol/core.d.ts
```diff
@@ -27,7 +27,7 @@ export type BlockVariant = {
   /**
    * @deprecated - Use the `name` field instead.
    */
-  displayName: string;
+  displayName?: string;
```
akash-joshi, line 30: Additional Change: Make `displayName` in variants an optional property.

site/src/lib/blocks.ts
```diff
@@ -85,6 +87,7 @@ export const readBlocksFromDisk = (): ExpandedBlockMetadata[] => {
 
       const metadata: ExpandedBlockMetadata = {
         // @todo should be redundant to block's package.json#name
+        componentId: `${FRONTEND_URL}/blocks/${packagePath}`,
```
akash-joshi, line 90: I'm generating the `componentId` at build-time, right where we're inserting the `packagePath` as well. @CiaranMn lmk if you would prefer to have this in the API instead.

@nathggns I'm presently generating the `componentId` in a format which matches the current entityStore cache. Lmk if we can use any other unique property, eg, the `name` of the block, directly instead.

site/src/pages/api/blocks.api.ts
```diff
@@ -105,6 +106,21 @@ export default createApiKeyRequiredHandler<null, ApiSearchResponse>()
         .map(([block, _]) => block);
     }
 
+    for (const block of data) {
+      // Generate absolute URLs for block icons
+      if (block.icon && !block.icon.startsWith("http")) {
+        block.icon = `${FRONTEND_URL}${block.icon}`;
```
akash-joshi, line 112: If the `icon` isn't already an absolute URL, attach the `FRONTEND_URL` to it to generate one.

site/src/pages/api/blocks.api.ts
```diff
@@ -105,6 +106,21 @@ export default createApiKeyRequiredHandler<null, ApiSearchResponse>()
...
+      }
+
+      for (const variant of block.variants ?? []) {
+        if (variant.icon.startsWith("public")) {
```
akash-joshi, line 116: This one's a little more tricky. I don't think we have a standard for how the block-variant URLs should be formatted yet. For now, all non-absolute variant icons start with `public/`, hence the condition here.

site/src/pages/api/blocks.api.ts
```diff
@@ -105,6 +106,21 @@ export default createApiKeyRequiredHandler<null, ApiSearchResponse>()
...
+          variant.icon = `${block.icon!.split("public/")[0]}public/${
+            variant.icon.split("public/")[1]
+          }`;
+        }
+      }
+    }
```
CiaranMn, line 122: I appreciate this is a draft PR but: any relative links in the block package will be relevant to the folder in which the block source code is - e.g. what will now be its `componentId`. You also don't need to treat anything with `public` in it differently - the `public`  folder is in the block's folder, not the root `public` folder.

site/src/pages/api/blocks.api.ts
CiaranMn: Relative `icon`s are relative to the block's source folder, not the frontend URL - have a look at some examples.

site/src/pages/api/blocks.api.ts
CiaranMn: They are formatted like normal URLs - absolute ones will start with a protocol.

CiaranMn (changes requested):
Thanks @akash-joshi - can you please update the handling of absolute vs relative URLs? you can detect these without any special handling of `public`, and they should be relative to the block's source folder.

Can you also do the same for `image` and `source` properties?

site/src/lib/blocks.ts
```diff
@@ -33,7 +35,7 @@ const getBlockMediaUrl = (
     return mediaPath;
   }
 
-  return `/blocks/${packagePath}/${mediaPath}`;
+  return `${FRONTEND_URL}/blocks/${packagePath}/${mediaPath}`;
```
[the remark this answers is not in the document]
CiaranMn: Thanks @akash-joshi - much simpler!

I have a couple of suggestions to slightly re-complicate it:
1. Can we rename `getBlockMediaUrl`? There's nothing intrinsically 'media' about it now - it's more `getBlockFileUrl` or something like that.
2. Can we check if `mediaPath` has a leading slash, and remove it if it does? Or check if it doesn't, and add it if it does? e.g. someone has `icon: "/public/icon.svg"`, we are going to end up with two slashes in the absolute URL.

CiaranMn (changes requested):

CiaranMn:
@akash-joshi the build script will need fixing, as it currently assumes that `schema` will be relative (it would ideally handle both relative and absolute urls)

site/src/lib/blocks.ts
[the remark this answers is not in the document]
akash-joshi: Addressed here: https://github.com/blockprotocol/blockprotocol/pull/276/commits/a11459ead23d3ab31d8c11624f0f799842e9cd99

CiaranMn (approved):
Thanks @akash-joshi !

We might want to come back and rename/readjust `readBlockDataFromDisk`, which is no longer simply reading from disk - but since this will only be for people who have included absolute URLs to external files (not in their block package), I think this is good for now.

Thanks again! 🎉

PR #277, opened 2022-03-31
Add serve command to block-template

CiaranMn:
We mention the `yarn serve` command in the `block-template` README, but this command has been lost from the `package.json` scripts somewhere along the way.

This PR restores it. It is useful to serve blocks locally for testing, e.g. in the HASH app.

### DIFF
--- a/packages/block-template/package.json
+++ b/packages/block-template/package.json
@@ -27,7 +27,7 @@
     "lint:tsc": "tsc --noEmit",
     "run-dev-server": "cross-env NODE_ENV=development webpack-dev-server --config webpack-dev-server.config.js",
     "schema": "typescript-json-schema tsconfig.json AppProps --required true --out dist/block-schema.json",
-    "serve": "npx serve -C -l 61234 ./dist/"
+    "serve": "npx serve --cors --listen 61234 ./dist/"
   },
   "dependencies": {
     "blockprotocol": "0.0.7"
--- a/packages/block-template/package.json
+++ b/packages/block-template/package.json
@@ -27,7 +27,7 @@
     "lint:tsc": "tsc --noEmit",
     "run-dev-server": "cross-env NODE_ENV=development webpack-dev-server --config webpack-dev-server.config.js",
     "schema": "typescript-json-schema tsconfig.json AppProps --required true --out dist/block-schema.json",
-    "serve": "serve -C -l 61234 ./dist/"
+    "serve": "npx serve -C -l 61234 ./dist/"
   },
   "dependencies": {
     "blockprotocol": "0.0.7"
@@ -52,7 +52,6 @@
     "react-dom": "^17.0.2",
     "regenerator-runtime": "^0.13.9",
     "rimraf": "^3.0.2",
-    "serve": "^13.0.2",
     "typescript": "^4.6.2",
     "typescript-json-schema": "^0.53.0",
     "webpack": "^5.68.0",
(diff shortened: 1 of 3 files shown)


packages/block-template/package.json
```diff
@@ -26,7 +26,8 @@
     "dev": "concurrently -n webpack,webpack-dev-server -c green,cyan \"yarn build:dev --watch\" \"yarn run-dev-server\"",
     "lint:tsc": "tsc --noEmit",
     "run-dev-server": "cross-env NODE_ENV=development webpack-dev-server --config webpack-dev-server.config.js",
-    "schema": "typescript-json-schema tsconfig.json AppProps --required true --out dist/block-schema.json"
+    "schema": "typescript-json-schema tsconfig.json AppProps --required true --out dist/block-schema.json",
+    "serve": "serve -C -l 61234 ./dist/"
```
kachkaev, line 30: Not sure we need this dependency and the script in each new block. Instead, we can change instructions to use `npx serve dist` – it will work for everyone regardless of what’s in `yarn.lock`.

Adding `serve` dependency pins the version and the port. This means ‘you need to upgrade’ warnings sooner or later. Hardcoding the port would make it more difficult to manage multiple local blocks.

I was actually thinking of removing `serve` from all HASH blocks and running `npx serve` in the blocks directory. Local block paths would be something like `localhost:500/paragraph/dist/`, `localhost:500/header/dist/` etc.

kachkaev: So maybe we just need one of these commands in docs:

```
npx serve --cors --listen 1234 dist
npx serve --cors dist

npx serve -C -l 1234 dist
npx serve -C dist
```

The port will be pinned to 5000 if we omit the option.

CiaranMn: Why not alias it in as a script?

CiaranMn: I've updated to remove `serve` as a dependency. I've kept a specified port because in practice the default is often going to be in use (it recently [changed to 3000](https://github.com/vercel/serve/pull/680)), and the fallback changes on each run. This means that it is harder to use as a source for a block in HASH app as it will keep changing where it's served from.

packages/block-template/package.json
```diff
@@ -26,7 +26,8 @@
     "dev": "concurrently -n webpack,webpack-dev-server -c green,cyan \"yarn build:dev --watch\" \"yarn run-dev-server\"",
     "lint:tsc": "tsc --noEmit",
     "run-dev-server": "cross-env NODE_ENV=development webpack-dev-server --config webpack-dev-server.config.js",
-    "schema": "typescript-json-schema tsconfig.json AppProps --required true --out dist/block-schema.json"
+    "schema": "typescript-json-schema tsconfig.json AppProps --required true --out dist/block-schema.json",
+    "serve": "npx serve -C -l 61234 ./dist/"
```
kachkaev, line 30:
```suggestion
    "serve": "npx serve --cors --listen 61234 ./dist/"
```

[optional]

Commands in package.json scripts, markdown blocks as well as `execa` args are more often read than written. Avoiding shortcuts makes them easier to understand as there’s no need to refer to the manual.

kachkaev (approved):

packages/block-template/package.json
kachkaev: Oh I forgot about `serve`’s shift from port 5000 to port 3000. I [complained](https://github.com/vercel/serve/pull/680#issuecomment-959812165) about it, but this did not change their decision back in the day 😅

PR #278, opened 2022-04-06
Use `block-scripts` in `block-template`

kachkaev:
- x-ref: https://github.com/hashintel/hash/pull/470
- [Asana task](https://app.asana.com/0/1201095311341924/1201983205685154) _(internal)_

This PR introduces a simple version of [`block-scripts` package](https://www.npmjs.com/package/block-scripts), which is supposed to make our `block-template` lighter and easier to maintain. I have kept the existing logic and focused mainly on package separation to keep the PR size small. Once we’ve merged this PR, we can publish `block-scripts@0.0.1` and use it https://github.com/hashintel/hash/pull/470. If everything works, we can improve DX further by tweaking the new scripts.

The main exception to my ‘no changes’ principle was the introduction of `package.json` → `block-scripts` session where users can define the ports. This approach helps avoid repetition of `--listen XXXX` in `dev` and `serve` commands. I have also:
- renamed `src/App.tsx` to `src/app.tsx` in line with our ongoing migration to `kebab-case`
- renamed `src/webpack-dev-server.js` to `src/dev.js` to hide the implementation detail. In theory, we don’t have to use webpack for block bundling.

`block-scripts` is a ESM module, which means that we can use `import` / `export` and top-level await without transpiling. I kept webpack configs as CommonJS (`*.cjs`) for now to avoid too much refactoring.

To test the changes, you can `cd packages/block-template` and run block lifecycle scripts `dev` `build` and `start`. Things should work as before.


packages/block-template/README.md
```diff
@@ -8,12 +8,13 @@ See https://blockprotocol.org/docs/developing-blocks
...
 
-1.  Change into the folder: `cd packages/hash/blocks/<name>`
+1.  Change into the folder: `cd path/to/your-block-name`
 
-1.  Write a React component starting in `App.tsx`. To test it during development:
+1.  Write a React component starting in `app.tsx`. To test it during development:
 
-    - edit `src/webpack-dev-server.js` to give your component some props to test with
-    - run the dev server with `yarn start`
+    - edit `package.json` → `examples` to give your component some props to test with
```
CiaranMn, line 15:
```suggestion
    - edit `package.json` → `blockProtocol` → `examples` to give your component some props to test with. This will also be included as example data in the generated `block-metadata.json`
```

packages/block-scripts/shared/ensure-block-schema-in-dist.js
```diff
@@ -0,0 +1,38 @@
...
+    /** @type {import("typescript-json-schema").PartialArgs} */
+    const settings = {
+      required: true,
+    };
+
+    const schema = TJS.generateSchema(program, "AppProps", settings);
+    await fs.ensureDir(blockDistDirPath);
+    await fs.writeJson(blockSchemaInDistFilePath, schema, { spaces: 2 });
+  } catch (error) {
+    console.warn(`Unable to generate block-schema: ${error}`);
```
CiaranMn, line 36: Maybe this should fallback to generating an empty schema? i.e. `{}`. In case the block accepts no props. And still warn about it – then at least the file is there for loading into HASH.

packages/block-scripts/shared/ensure-block-schema-in-dist.js
```diff
@@ -0,0 +1,38 @@
...
+  const customBlockSchemaFilePath = path.resolve(
+    blockRootDirPath,
+    "block-schema.json",
+  );
+
+  if (await fs.pathExists(customBlockSchemaFilePath)) {
+    await fs.copy(customBlockSchemaFilePath, blockSchemaInDistFilePath);
+    console.log("Using custom block-schema.json");
+    return;
+  }
```
CiaranMn, line 21: 👍

packages/block-scripts/scripts/help.js
```diff
@@ -0,0 +1,19 @@
...
+      "Block Scripts",
+    )} (version ${await getBlockScriptsVersion()})`,
+  );
+  console.log("");
+  console.log(
+    `See ${chalk.underline(
+      "https://blockprotocol.org/docs/developing-blocks",
+    )}`,
+  );
+  console.log("");
```
CiaranMn, line 16: Should this also list the available commands (scripts)?

CiaranMn (commented):
Thanks @kachkaev, great improvement!

Noticed / suggested a few small things

packages/block-scripts/shared/ensure-block-schema-in-dist.js
CiaranMn: We could add this to the README

packages/block-scripts/shared/ensure-block-schema-in-dist.js
kachkaev: 👍 242f74454a0a3755a87de92b3fc78646cf2b2873

packages/block-template/README.md
kachkaev: 👍 317a16d0a8d229984a9793525c06c92fc72037c2 (I wrote `blockprotocol` instead of `blockProtocol`)

scripts/packages/publish-to-local-registry.ts
```diff
@@ -10,6 +10,7 @@ const npmUserAndPassword = "verdaccio";
 const npmEmail = "verdaccio@example.com";
 
 const packageNames = [
```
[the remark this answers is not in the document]
kachkaev: Good idea! I addressed it in 31fa05cfe975cea2b6eb2ce715e5c9262dc6a890

CiaranMn (approved):
Great!

PR #281, opened 2022-04-11
Fix ‘NPM Packages’ CI by downgrading NPM to 8.4.1

kachkaev:
This PR fixes ‘NPM Packages’ CI which has been failing for about two weeks now.

We define `node-version: 16` in `actions/setup-node@v2`, which gives us the latest Node version. The CI started failing after 16.14.0 got auto-replaced with 16.14.2. These two Node versions come with different NPM versions: 8.3.1 and 8.5.0.

Downgrading NPM from 8.5.0 to 8.4.1 (and 8.3.0, 8.4.0) fixes the problem even without having to downgrade Node itself. Upgrading NPM 8.5.0 to 8.6.0 (the latest at the time of writing) does not fix the issue.

Looking at CI logs, here is one subtle difference that looks a bit odd:

`npm 8.3.1 / 8.4.0 / 8.4.1`

```
=========================================================
Unpublish block-template from local registry (if present)
=========================================================
↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓

npm WARN using --force Recommended protections disabled.
- block-template@0.0.8
```

`npm 8.5.0 / 8.6.0`

```
=========================================================
Unpublish block-template from local registry (if present)
=========================================================
↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓

npm WARN using --force Recommended protections disabled.
- block-template
```

(there’s no `@0.0.8` in the newer versions)

This difference in the output might be related to what request payload Verdaccio receives. I haven’t found any obvious [CHANGELOG items](https://github.com/npm/cli/blob/latest/CHANGELOG.md#v850-2022-02-10) to blame, but I haven’t looked for too long to be honest.

I tried to spoonfeeding `npm unpublish --force` with `block-template@0.0.8` but this did not help in the never versions of NPM.

Let’s hope that this problem self-resolves over time and we’ll be able to unpin NPM.

---
  - https://app.asana.com/0/0/1202094259991222

### DIFF
--- a/.github/workflows/check-codebase.yml
+++ b/.github/workflows/check-codebase.yml
@@ -147,8 +147,8 @@ jobs:
 
       - name: Run local registry
         run: |
-          docker pull verdaccio/verdaccio
-          docker run -it --detach --rm --name verdaccio -p 4873:4873  verdaccio/verdaccio
+          docker pull verdaccio/verdaccio:5
+          docker run -it --detach --rm --name verdaccio -p 4873:4873  verdaccio/verdaccio:5
           yarn wait-on --timeout 30000 http://localhost:4873
 
       - name: Publish NPM packages to local registry
--- a/.github/workflows/check-codebase.yml
+++ b/.github/workflows/check-codebase.yml
@@ -141,6 +141,10 @@ jobs:
           node-version: 16.14.0
           cache: yarn
 
+      - run: |
+          npm install --global npm@8.6.0
+          npm --version
+
       - run: yarn install
 
       - name: Run local registry
--- a/.github/workflows/check-codebase.yml
+++ b/.github/workflows/check-codebase.yml
@@ -138,12 +138,8 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
-          node-version: 16.14.0
+          node-version: 16
           cache: yarn
-
-      - run: |
-          npm install --global npm@8.6.0
-          npm --version
 
       - run: yarn install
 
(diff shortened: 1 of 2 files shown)


.github/workflows/check-codebase.yml
```diff
@@ -141,12 +141,15 @@ jobs:
...
+        run: npm install --global npm@8.4.1 ## https://github.com/blockprotocol/blockprotocol/pull/281
+
       - run: yarn install
 
       - name: Run local registry
         run: |
-          docker pull verdaccio/verdaccio
-          docker run -it --detach --rm --name verdaccio -p 4873:4873  verdaccio/verdaccio
+          docker pull verdaccio/verdaccio:5
+          docker run -it --detach --rm --name verdaccio -p 4873:4873  verdaccio/verdaccio:5
```
kachkaev, line 152: Using `:5` instead of `:latest` does not change anything at the moment. Both tags point to the same image.

It seems that [v6](https://www.npmjs.com/package/verdaccio) will be released soon, so let’s get ready for that just in case.

CiaranMn (approved):

PR #285, opened 2022-04-16
Add labelProperty and configProperties to spec

CiaranMn:
Introduces `labelProperty` and `configProperties` to the spec as potential additions for entity schemas, and block schemas, respectively.

Screenshot 2022-04-16 at 14 48 54

Screenshot 2022-04-16 at 14 49 18

### DIFF
--- a/site/src/_pages/spec/1_block-types.mdx
+++ b/site/src/_pages/spec/1_block-types.mdx
@@ -173,7 +173,7 @@ As well as `entityTypeId`, entity schemas SHOULD specify the following field at 
 }
 ```
 
-We will publish a JSON schema vocabulary which allows for validating these custom fields.
+We will publish a JSON schema vocabulary and meta-schema which allows for validating these custom fields.
 
 ### Entity functions
 
--- a/site/src/_pages/spec/1_block-types.mdx
+++ b/site/src/_pages/spec/1_block-types.mdx
@@ -27,6 +27,7 @@ A block package SHOULD contain:
 
   - If the block does not function without certain properties, MUST specify a [required](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.5.3) array naming them
   - SHOULD specify any further constraints or [validation](https://json-schema.org/draft/2020-12/json-schema-validation.html) requirements which improve the block’s functionality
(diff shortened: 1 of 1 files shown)


site/src/_pages/spec/1_block-types.mdx
```diff
@@ -27,6 +27,7 @@ A block package SHOULD contain:
 
   - If the block does not function without certain properties, MUST specify a [required](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.5.3) array naming them
   - SHOULD specify any further constraints or [validation](https://json-schema.org/draft/2020-12/json-schema-validation.html) requirements which improve the block’s functionality
+  - MAY specify an additional field, `configProperties`, which is an array of `string` values which MUST be present as a key on the block schema's `properties`, i.e. each must name a property the block expects. This field is used to indicate which properties a block considers to be part of 'configuring' how it appears or behaves, rather than data users are creating or viewing via the blocks. Embedding applications may choose to display their own UI for setting these values, and could allow users to set values for them for all blocks of a type (rather than on a per-block instance basis).
```
benwerner01, line 30: > MAY specify an additional field, `configProperties`, which is an array of `string` values which MUST be present as a key on the block schema's `properties`

I understand what you're saying here, but I feel its slightly ambiguous whether with "which MUST be present as a key on the block schema's properties" you're referring to the `configProperties` field (which should be root-level), or the keys in its array.

Perhaps the following addition would help avoid this ambiguity:
```suggestion
  - MAY specify an additional root-level field, `configProperties`, which is an array of `string` values which MUST be present as a key on the block schema's `properties`, i.e. each must name a property the block expects. This field is used to indicate which properties a block considers to be part of 'configuring' how it appears or behaves, rather than data users are creating or viewing via the blocks. Embedding applications may choose to display their own UI for setting these values, and could allow users to set values for them for all blocks of a type (rather than on a per-block instance basis).
```

site/src/_pages/spec/1_block-types.mdx
```diff
@@ -153,6 +154,27 @@ Note that this is about translating between different JSON Schemas, and is not t
...
+    "name": {
+      "type": "string"
+    }
+  },
+  "labelProperty": "name"
+  ...
+}
+```
+
+We will publish a JSON schema vocabulary which allows for validating these custom fields.
```
benwerner01, line 176: I think we would need a meta-schema to make programatic validation possible, so may be worth mentioning:
```suggestion
We will publish a JSON schema vocabulary and meta-schema which allows for validating these custom fields.
```

benwerner01 (dismissed):
Thanks Ciaran, these additions look good to me 👍

I had two small thoughts on potential improvements, but neither block merging.

benwerner01 (approved):

PR #293, opened 2022-04-25
Custom initial entities in mock-block-dock. Better displayName from create-block-app

CiaranMn:
A couple of small improvements to block authoring packages.

## create-block-app
Set `displayName` to the name of the block, instead of having all blocks start with `"Template"` as their `displayName`.

## mock-block-dock

**1. Custom dummy entities/types/links in data store**

Allow the users to pass some or all of `initialEntities`, `initialEntityTypes`, and `initialLinks`, to overwrite the dummy data hardcoded in the package.

**2. Better resetting of `initialEntities`**

There's a hook to reset `initialEntities` if it changes, but this was firing on the first render when `initialEntities` hadn't changed. This is bad anyway, but also meant that if a block _immediately_ calls `updateEntities` on render, there can be a race condition where that update is overwritten by another call to reset `initialEntities` from `mock-block-dock`.

### DIFF
--- a/hub/@hash/callout.json
+++ b/hub/@hash/callout.json
@@ -1,6 +1,6 @@
 {
   "repository": "https://github.com/hashintel/hash.git",
-  "commit": "c943e77b6cc1bcb7f4006c23c46bd4d51ccb62ba",
+  "commit": "7e25ced45cceb901da86da7c90cd539050f33ab2",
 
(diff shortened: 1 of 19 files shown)


kachkaev (dismissed):
Nice 👏

kachkaev (approved):

PR #294, opened 2022-04-25
Bump block-scripts and block-template version numbers

CiaranMn:
#293 made `block-scripts` and `block-template` dependent on a newer version of `mock-block-dock`, but it did not increase their own version numbers.

This PR does that – I will publish a new version of each after merging.

### DIFF
--- a/packages/block-scripts/package.json
+++ b/packages/block-scripts/package.json
@@ -1,6 +1,6 @@
 {
   "name": "block-scripts",
-  "version": "0.0.2",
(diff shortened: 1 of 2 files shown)


kachkaev (approved):

kachkaev (changes requested):
We need to run `yarn fix:dependency-version-consistency && yarn fix:yarn-deduplicate`

kachkaev (approved):

PR #296, opened 2022-04-26
Further improve `block-scripts`

kachkaev:
Continues #295 

In scope

- Limit number of webpack chunks to 1 to avoid issues with `require()`

- Watch for changes in `block-schema.json`, `package.json`, `public/**/*" and `README.md` in dev

- Add `block-metadata.json` → `builtAt`. We can use this for hot reloading in the embedding apps (alternative suggestions are welcome)

- Check `block-scripts` package with `tsc` (this is possible despite using `*.js`)

### DIFF
--- a/packages/block-scripts/shared/generate-dist-block-metadata.js
+++ b/packages/block-scripts/shared/generate-dist-block-metadata.js
@@ -12,7 +12,6 @@ export const generateDistBlockMetadata = async (source) => {
 
   const {
     name,
-    displayName,
     version,
     description,
     author,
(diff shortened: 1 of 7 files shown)


packages/block-scripts/shared/generate-dist-block-metadata.js
```diff
@@ -12,6 +12,7 @@ export const generateDistBlockMetadata = async (source) => {
 
   const {
     name,
+    displayName,
```
CiaranMn, line 15: `displayName` is part of the `blockprotocol` object, rather than being a `package.json` field, and was already being added to metadata via `blockprotocol` being spread out - is this change required?

CiaranMn (commented):
Thanks @kachkaev - a question on `displayName`

CiaranMn (approved):

PR #297, opened 2022-04-27
Introduce `eslint-plugin-simple-import-sort`, ensure `node:` protocol in imports

kachkaev:
This PR adds two new ESLint autofixes which aim to reduce future git diffs:

- [eslint-plugin-simple-import-sort](https://www.npmjs.com/package/eslint-plugin-simple-import-sort) auto-sorts all imports except those with side-effects (`import x from "x"` vs `import "x"`). Named imports are also sorted within an import statement. See [slack thread](https://hashintel.slack.com/archives/C022217GAHF/p1641327648013800) (internal)

- [unicorn/prefer-node-protocol](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-node-protocol.md) makes sure we always use `node:path` / `node:os` instead of `path` / `os`. This improves import sorting and makes imports a bit more deterministic in general. We’ve been mixing styles before this PR.


site/src/components/Snippet.tsx
```diff
@@ -13,10 +10,14 @@ import { Box, BoxProps } from "@mui/material";
  *
  * @see https://prismjs.com
  */
-
+import "prismjs";
```
kachkaev, line 13: 👀

PR #298, opened 2022-04-29
Switch to kebab-case in site URLs

kachkaev:
This PR converts `camelCase` website URLs to `kebab-case` ones. This better aligns them with web standards and reduces codebase entropy in the long term. For example, `/api/uploadUserAvatar` is now `/api/upload-user-avatar`.

I have also configured [`unicorn/filename-case`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/filename-case.md) for the affected folders to make sure we keep the entropy low.


package.json
```diff
@@ -40,8 +40,8 @@
     "@swc/helpers": "0.3.2",
     "@types/fs-extra": "^9.0.13",
     "@types/wait-on": "5.3.1",
-    "@typescript-eslint/eslint-plugin": "5.11.0",
-    "@typescript-eslint/parser": "5.11.0",
+    "@typescript-eslint/eslint-plugin": "5.21.0",
+    "@typescript-eslint/parser": "5.21.0",
```
kachkaev, line 44: Removes this warning:

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.6.0

YOUR TYPESCRIPT VERSION: 4.6.3

Please only submit bug reports when using the officially supported version.

=============
```

.eslintrc.js
```diff
@@ -315,6 +319,12 @@ module.exports = {
         ],
       },
     },
+    {
+      files: ["site/src/pages/**"],
+      rules: {
+        "unicorn/filename-case": "error",
+      },
+    },
```
kachkaev, line 327: Makes sure we don’t re-introduce `camelCase` page paths

.eslintrc.js
```diff
@@ -204,6 +204,14 @@ module.exports = {
...
       },
     },
+    {
+      files: [
+        "packages/{block-scripts,block-template,blockprotocol,create-block-app}/**",
+      ],
+      rules: {
+        "unicorn/filename-case": "error",
+      },
+    },
```
kachkaev, line 214: Bonus: makes sure we don’t fallback to `camelCase` in these packages. We still need to convert `mock-block-dock` and `react-block-loader` (out of scope for this PR)

CiaranMn (approved):
Code looks good, tested and all pages / functionality working

PR #304, opened 2022-05-06
Switch to `kebab-case` file names in `site` workspace

kachkaev:
This PR follows https://github.com/blockprotocol/blockprotocol/pull/298 and https://github.com/blockprotocol/blockprotocol/pull/278. It converts all file paths inside `site` to `kebab-case` and makes sure we use this style consistently via [unicorn/filename-case](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/filename-case.md). By using `kebab-case` in all file names, we reduce the entropy in the codebase and don't need to pick between different styles for package names, script file names, page URLs and import paths.

Most changes were made automatically via [`npx kebab-ify-files site/src`](https://www.npmjs.com/package/kebab-ify-files).


nathggns:
@kachkaev Is there anything in particular you'd like us to look out for here? I've not really work on the blockprotocol site much myself (beyond a few performance things) so I'm not certain I'm best placed, but at a quick glance it just seems to be file renames and import updates. Are you worried about anything specifically breaking?

kachkaev:
👋 @nathggns! Just saw your reply, sorry. Yeah, just a quick glance is fine for this PR — I hope that CI would spot new issues, if any. I clicked around the vercel deployment – it seemed to work too.

nathggns (approved):

PR #305, opened 2022-05-06
Use `unicorn/filename-case` in `packages/**`

kachkaev:
This PR follows https://github.com/blockprotocol/blockprotocol/pull/304, https://github.com/blockprotocol/blockprotocol/pull/298 and https://github.com/blockprotocol/blockprotocol/pull/278. It converts all remaining paths inside `packages` to `kebab-case` and makes sure we use this style consistently via [unicorn/filename-case](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/filename-case.md). By using `kebab-case` in all file names, we reduce the entropy in the codebase and don't need to pick between different styles for package names, script file names, page URLs and import paths.

Although the PR affects `mock-block-dock` and `react-block-loader`, I'm not bumping their versions because all changes are internal.


.eslintrc.js
```diff
@@ -205,9 +205,7 @@ module.exports = {
       },
     },
     {
-      files: [
-        "packages/{block-scripts,block-template,blockprotocol,create-block-app}/**",
-      ],
+      files: ["packages/**"],
```
kachkaev, line 208: TODO: Replace with `"unicorn/filename-case": "error"` (global rule) when https://github.com/blockprotocol/blockprotocol/pull/304 is merged. I also need to remove this redundant rule: https://github.com/blockprotocol/blockprotocol/pull/304/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R323-R326

nathggns (approved):

PR #307, opened 2022-05-10
Split bp types for filter operators

CiaranMn:
We have a `BlockProtocolFilterOperatorType` which currently requires a `value` to be passed, but some of the operators (e.g. `IS_EMPTY`) shouldn't require one. 

This PR fixes that.

I haven't bumped the `blockprotocol` version as `0.0.9` hasn't been published yet. I'll do so once this merges.

---
  - https://app.asana.com/0/0/1202202445930675


packages/blockprotocol/core.d.ts
```diff
@@ -164,30 +164,41 @@ export type BlockProtocolDeleteEntitiesFunction = {
...
 };
 
-export type BlockProtocolMultiSort = {
+export type BlockProtocolSort = {
   field: string;
   desc?: boolean | undefined | null;
-}[];
+};
+
+export type BlockProtocolMultiSort = BlockProtocolSort[];
```
benwerner01, line 201: Nice 👍

benwerner01 (approved):
Looks good Ciaran, thank you 👍

PR #309, opened 2022-05-13
Use react-block-loader in block hub. Export mock-block-dock hook

### LINKED ISSUE #299: [REPO] Update root README to document new packages
@CiaranMn Please can you add new package summaries to section 2 of the ‘Repository contents’ bit of the root README.md — it contains 4 pkgs, but we now have a few more that can be added: https://github.com/blockprotocol/blockprotocol/tree/main/packages Cheers.

CiaranMn:
#289 added the `react-block-loader` package – a component to load a block from a URL.

This PR starts using it in the `site`'s Block Hub.

As part of this it also:
- makes ESM output only for `react-block-loader` (it has an ESM dependency)
- renames the exported component from `RemoteBlock` to `BlockLoader`, as it is more consistent
- ~adds a hook export from `mock-block-dock` (to more easily have the mock functions etc to pass to `BlockLoader`)~
  -  now introduced separately in https://github.com/blockprotocol/blockprotocol/pull/318
- adds two missing packages to the root README, and thus closes #299


packages/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -46,131 +41,25 @@ export const MockBlockDock: VoidFunctionComponent<MockBlockDockProps> = ({
   initialLinks,
   initialLinkedAggregations,
 }) => {
-  const mockData = useMemo((): MockData => {
```
CiaranMn, line 49: this has all been extracted into a hook, so this component now only gets stuff from the hook and injects it into the child

packages/mock-block-dock/src/use-mock-block-props.ts
```diff
@@ -0,0 +1,179 @@
+import {
```
CiaranMn, line 1: this is the extracted stuff from `mock-block-dock.tsx`

CiaranMn:
Closing this: 
1. we introduced `mock-block-dock` hook in #318 
2. #343 enables `mock-block-dock` loading custom elements (since it needs to be able to anyway), so there's no need to have a separate loader.

`react-block-loader` still needs updating / fixing, but will do so when introducing it to [HASH](https://github.com/hashintel/hash/tree/main/packages/hash)

PR #312, opened 2022-05-18
Implement block sandboxing on the Þ Hub

kachkaev:
This PR puts all block demos into iframes, thus making sure they are sandboxed. Block styles no longer ’leak’ to EA.

The approach leverages native ESM module support in modern browsers as well as https://esm.sh as an ESM CDN. Just as before, blocks only work if they consist of a single CJS webpack bundle. As a follow-up improvement, we can try building blocks to ESM instead of CJS and use [es-module-shims](https://www.npmjs.com/package/es-module-shims) to leverage [import maps](https://blog.logrocket.com/es-modules-in-browsers-with-import-maps/) in all browsers.

Known issues with blocks:

- Some styles got worse (e.g. we see serif font in Paragraph and Callout).

- Blocks without `block-metadata.json` → `externals` stopped working, so I had to implement `hotfixExternals()` for now (the function explains itself in comments).

- Viewport negotiation is not part of BP yet, so blocks cannot ‘ask’ EA to extend the iframe. I've temporarily configured it to use all available space.

- Embed block no longer shows HASH simulations because of CORS. This needs to be fixed in `core.hash.ai` ([internal asana task](https://app.asana.com/0/1201095311341924/1202305315886959/f)).

   ```txt
    Script at 'https://core.hash.ai/hash-prod-2022-04-15-T1521_33440/analyzerworker.js' cannot be accessed from origin 'null'.
    ```

To test this PR, open https://blockprotocol-fj9huk5ki-hashintel.vercel.app/hub in Firefox, Chrome and Safari and browse each block.

[Asana task](https://app.asana.com/0/1200211978612931/1201796051804395/f) _(internal)_


site/package.json
```diff
@@ -56,7 +56,6 @@
     "lodash": "^4.17.21",
     "md5": "^2.3.0",
     "mime-types": "^2.1.34",
-    "mock-block-dock": "0.0.9",
```
kachkaev, line 59: We now use this dependency in a sandbox via CDN

site/src/components/pages/hub/hub-utils.tsx
```diff
@@ -1,78 +1,2 @@
...
-  provider_name: string;
-  provider_url: string;
-  thumbnail_height: number;
-  thumbnail_width: number;
-  thumbnail_url: string;
-  html: string;
-};
-
-// @todo temporary fallback until block implements workarounds for CORS-blocked oembed endpoints
-export async function getEmbedBlock(url: string): Promise<{
```
CiaranMn, line 35: I think this function is still required – some oEmbed endpoints (e.g. [Twitter](https://developer.twitter.com/en/docs/twitter-for-websites/timelines/guides/oembed-api)) don't have CORS enabled, and so we can't fetch the embed code for tweet links users paste in without fetching it server-side.

This is what happens when you try and update an embed, e.g. click the pencil and paste in a new link to embed – the relevant oEmbed endpoint is called, and many do not allow CORS.

site/src/components/pages/hub/sandboxed-block.tsx
```diff
@@ -0,0 +1,155 @@
...
+  useEffect(() => {
+    postBlockProps();
+  }, [postBlockProps]);
+
+  return (
+    <iframe
+      ref={iframeRef}
+      title="block"
+      srcDoc={srcDoc}
+      sandbox="allow-scripts"
```
CiaranMn, line 140: This is required for many of the blocks to function
```suggestion
      sandbox="alllow-forms allow-scripts"
```

CiaranMn (changes requested):
Thanks @kachkaev – clever use of `srcdoc` and dynamic dependency building. Exciting progress on ESM too!

I think it's worth investigating the issues with the Embed a little further now – it's not just the HASH Embed, but all of them that are broken. Some of the problem seems to be that the grandchildren iFrames (i.e. any `iframe` within the one you are creating) inherit the `null` origin of the parent. 

I say that because although the HASH and YouTube embeds are actually iframes with a `src` of `core.hash.ai`/`youtube.com`, the browser is making requests with a null origin, and fail CORS. We could fix this for HASH specifically, but it points to a potential wider problem.

I assume we cannot give the parent `iframe` sandbox="allow-same-origin"` (which makes all the embeds work), because this would allow JS in it to make requests with the origin `blockprotocol.org`.

If the parent `iframe` had a different origin (e.g. it was being loaded from a different domain), we _could_ give it `allow-same-origin`, so long as that different domain was one that we didn't care about any rogue JS acting with the origin of. I don't know if there's some way to reconcile this with the dynamic `srcdoc` approach. 

I think it might be worth investigating this further so we have a better idea for what will be possible with blocks loading iframes with this approach.

site/src/components/pages/hub/hub-utils.tsx
kachkaev: Hmm yeah you are right. I tried finding a workaround, but no luck so far. It almost feels like a block (or Block Hub) should provide the HTML for the iframe. Then the document will have some origin and therefore CORS will be happy AFAIU. Using `allow-same-origin` and `allow-scripts` diminishes the whole point of iframes:

> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

> When the embedded document has the same origin as the embedding page, it is **strongly discouraged** to use both `allow-scripts` and `allow-same-origin`, as that lets the embedded document remove the sandbox attribute — making it no more secure than not using the `sandbox` attribute at all.

CiaranMn: Using `allow-same-origin` and `allow-scripts` only diminishes the point of iframes when the iframe has the same origin as the parent window. So, if our iframe was loading static HTML from a different origin, it would be fine to give it `allow-same-origin` (if there was nothing exploitable about acting with that origin). More [here](https://security.stackexchange.com/a/254521). So we must give anything loaded with `srcdoc` `allow-same-origin`, and it then is inherited in any nested iframes.

Unfortunately so long as we make our iframe using `srcdoc`, its natural origin would be the same as the parent (i.e blockprotocol.org).

So our options are:
1. Blocks which need to act with an origin must be loaded directly from a different origin into an iFrame, which must have `allow-same-origin` set
2. Our iFrame must be loaded from a different origin, and must have `allow-same-origin` set
3. Blocks which make requests that rely on having a specific origin do not work

site/src/components/pages/hub/sandboxed-block-demo.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const postBlockProps = useCallback(() => {
+    iframeRef.current?.contentWindow?.postMessage(JSON.stringify(props), "*");
+  }, [props]);
+
+  useEffect(() => {
+    postBlockProps();
+  }, [postBlockProps]);
+
+  const aliasBaseUrl = ""; // @todo
```
kachkaev, line 24: @CiaranMn I guess I need some help here. It’s impossible to use `blocksandbox.org` because preview URLs are like `blockprotocol-xxx-hashintel.vercel.app`. I see no place to use the alias, so it’s a chicken and egg situation.

site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
```diff
@@ -0,0 +1,133 @@
...
+  if (!blockMetadata) {
+    res.status(404);
+    res.write("Block not found");
+    res.end();
+    return;
+  }
+
+  const { source } = await readBlockDataFromDisk(blockMetadata);
+  const reactVersion = blockMetadata.externals?.react ?? "17.0.2";
+  const mockBlockDockVersion = "0.0.9";
```
CiaranMn, line 52: This is going to be tedious to maintain, is there a way we can get it automatically?

site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  const html = `
+    <script type="module">
+      const handleMessage = ({ data }) => {
+        if (typeof data !== "string") {
+          return;
+        }
+        globalThis.initialBlockProps = JSON.parse(data);
+        window.addEventListener("message", handleMessage);
```
CiaranMn, line 71: Is this supposed to be removing the event listener instead, rather than adding another? This listener doesn't really do anything after the first render, since the other listener further down picks up re-rendering.

site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
```diff
@@ -0,0 +1,133 @@
...
+        const module = { exports: {} };
+        const moduleFactory = new Function("require", "module", "exports", source);
+        moduleFactory(require, module, module.exports);
+
+        return module.exports;
+      }
+
+      const findComponentExport = (module) => {
+        return module.default ?? module.App ?? module[Object.keys(module)[0]];
+      }
```
CiaranMn, line 106: Can we throw if this isn't found?

CiaranMn (commented):
Lookin good @kachkaev – a few comments

site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
kachkaev: https://github.com/blockprotocol/blockprotocol/pull/312/commits/a1db5a63fcf9e8a186ac790955b28c4e1ed1d401

site/src/components/pages/hub/hub-utils.tsx
CiaranMn: The embed blocks are working for the existing dummy data, but are still broken for retrieving _new_ embed code, since some of the oEmbed endpoints do not have CORS enabled. The Embed block crashes prior to this point anyway because it expects

One potentially temporary fix would be to restore the fetch-embed-code api endpoint, and allow requests from any origin (`createBaseHandler({ isPublicApi: true })`), and then pass this `getEmbedCode` function to the block.

CiaranMn (changes requested):
Thanks @kachkaev - a few more observations. 

I also noticed that the block styling overflow is a little worse here than on blockprotocol.org – e.g. the pencil on the embed block is hidden on load, whereas it's visible (if squashed) on prod

Screenshot 2022-06-07 at 17 22 47

kachkaev:
> pencil on the embed block is hidden on load, whereas it's visible (if squashed) on prod

Observing the same issue [in prod](https://blockprotocol.org/@hash/blocks/embed) (although for a different window widths). Not sure we should fix these kinds of issues as part of this PR. Block style is the responsibility of the block. BP website is not the only app in which the block should work.

Screenshot 2022-06-07 at 19 06 43

site/src/pages/[shortname]/blocks/[block-slug].page.tsx
```diff
@@ -11,62 +11,63 @@ import { formatDistance } from "date-fns";
...
+ *
+ * In STAGING, we will mostly be visiting unique deployment URLs
+ * for testing, so we can use the unique branch URL as the origin.
+ * Note: this means the frame in preview deployments will always be
+ * built from the tip of the branch - if you visit the non-latest preview
+ * deployment AND you have changed the framed code, they may be out of sync.
+ */
+const generateSandboxBaseUrl = (): string => {
+  if (isProduction) {
+    const deploymentUrl = process.env.NEXT_PUBLIC_VERCEL_URL;
```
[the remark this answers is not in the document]
Comment (kachkaev) [reply site/src/pages/[shortname]/blocks/[block-slug].page.tsx]: https://github.com/blockprotocol/blockprotocol/pull/312/commits/d2fc4119811f4de1e1f768367e6aaa8b0b222ba3

site/src/components/pages/hub/hub-utils.tsx
kachkaev: Not sure I understand this approach. Could you please help by pushing or via a follow-up PR?

kachkaev: I’ve removed embed block in https://github.com/blockprotocol/blockprotocol/pull/331

CiaranMn (approved):

PR #317, opened 2022-05-31
Improve `yarn dev` DX in `block-scripts`

kachkaev:
- Use two webpack servers in `yarn dev` to enable block embedding in parallel to `mock-block-dock` usage
- Introduce `block-metadata.json` → `devReloadEndpoint` to improve block DX in EAs
- Remove `block-metadata.json` → `builtAt` (it was an unused precursor to the new `devReloadEndpoint` property)
- Add `pickExistingFilePath` helper function and use it to improve entry point detection. In addition to `src/index.tsx` `block-scripts` can also detect `js,jsx,ts`.

## Related links

- [Asana task](https://app.asana.com/0/1200211978612931/1202073961250686/f) _(internal)_
- https://github.com/hashintel/hash/pull/603 ✨ **demo** ✨


packages/block-scripts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "block-scripts",
-  "version": "0.0.6",
+  "version": "0.0.7",
```
kachkaev, line 3: I’ll publish new versions after the review

packages/blockprotocol/core.d.ts
```diff
@@ -49,6 +49,10 @@ export type BlockMetadata = {
    * A short description of the block, to help users understand its capabilities
```
CiaranMn, line 49: Might as well bump the version of this too, no? Could publish it immediately and save adding to the patch in HASH

packages/block-scripts/scripts/dev.js
```diff
@@ -99,9 +41,14 @@ const script = async () => {
...
-  const compiler = webpack(webpackConfig);
-  const server = new WebpackDevServer(webpackDevServerConfig, compiler);
-  await server.start();
+  const buildWebpackConfig = await generateBuildWebpackConfig("development");
+  const devWebpackConfig = await generateDevWebpackConfig();
+
+  webpack(buildWebpackConfig, () => {});
+
+  const devCompiler = webpack(devWebpackConfig);
+  const devServer = new WebpackDevServer(webpackDevServerConfig, devCompiler);
```
CiaranMn, line 50: Since `yarn dev` is now useful for testing blocks in EAs, we could set `open` in `webpackDevServerConfig` to `false`, and let users open the browsers themselves if they want (rather than spawning windows they might not need).

packages/blockprotocol/core.d.ts
kachkaev: The current size of `blockprotocol` patch is quite large, which means that bumping `blockprotocol` in HASH will potentially require unrelated fixes in https://github.com/hashintel/hash/pull/603. I merged that PR to keep it focused. We can still bump `blockprotocol` package here, but upgrade a version in a separate HASH PR.

packages/block-scripts/scripts/dev.js
kachkaev: I removed `open` option from the dev script in https://github.com/blockprotocol/blockprotocol/pull/317/commits/50657d58fc9f3ad819461f3c2f99328b33aa51c1.

We still have it in two other packages? Shall I get rid of it too for consistency? Not a fan of it TBH.

Screenshot 2022-06-02 at 18 44 21

CiaranMn: Don't mind if you want to get rid of it in those.

I think it's different in that currently the only point in running `yarn dev` in the other two is to then look at the thing in the browser, so it isn't the situation we have now with `block-template` where you may want to run `yarn dev` without visiting the endpoint. But I don't feel strongly about it.

kachkaev: https://github.com/blockprotocol/blockprotocol/pull/317/commits/d9c3c800ff427156825374b6f692753ffbdf87b6

CiaranMn (approved):

PR #318, opened 2022-06-02
Add hook export from mock-block-dock

CiaranMn:
`mock-block-dock` previously only exposed its mock data properties and functions to users by wrapping a component and injecting them as props.

This PR adds the props + functions as an export, so that they can be used in other contexts, e.g. called in response to events or messages.

---
  - https://app.asana.com/0/0/1202380713472403


packages/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -46,131 +41,25 @@ export const MockBlockDock: VoidFunctionComponent<MockBlockDockProps> = ({
   initialLinks,
   initialLinkedAggregations,
 }) => {
-  const mockData = useMemo((): MockData => {
```
CiaranMn, line 49: This has all been moved to the hook file

teenoh (approved):
LGTM!

PR #323, opened 2022-06-05
Add Block Protocol 0.2 core + service packages

CiaranMn:
# Purpose

This PR introduces two new packages, preparing for version 0.2 of the Block Protocol, which splits out the specification into:
- `core`: related to defining blocks, rendering blocks, and _how_ messages are passed between blocks and embedding applications
- services: _what_ messages are passed between blocks and embedders. This PR introduces the first and so far only such service,  `graph`, which incorporates the entity/type system.

Full context on these changes will be available via the website shortly.

# Known issues

Linting is failing because it can't find `@blockprotocol/core` – this is both in the repo and published so I'm not sure what's happening there. Same with the testing of publishing - @kachkaev you fixed this for a previous package, didn't you? Or there was some workaround? Appreciate any help you can offer!

# Next steps

Prepare a feature branch which introduces the rest of the necessary changes for `0.2`:
1. Update `mock-block-dock`
2. Update `react-block-loader`
3. Update blocks on the Block Hub (in https://github.com/hashintel/hash)
4. Update the Block Hub

---
  - https://app.asana.com/0/0/1202361504154773


packages/core/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@blockprotocol/core",
```
kachkaev, line 2: If a folder for `@blockprotocol/x` is `packages/x`, it ends up being a sibling of `packages/create-block-app` and alike. This becomes somewhat confusing. Folder suggestions for `@blockprotocol/x`:

- `packages/blockprotocol--x` (with `--` like for `@types/namespace--package` on NPM)
- `packages/@blockprotocol/x` (requires a adding `packages/@blockprotocol/*` to root `package.json` → `workspaces`)

Either way, it’d be nice to have _some_ way of unambiguously mapping folder names to package names or vice versa.

packages/core/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "version": "0.0.1",
+  "description": "Implementation of the Block Protocol Core specification for blocks and embedding applications",
+  "keywords": [
+    "blockprotocol",
+    "blocks",
+    "core"
+  ],
+  "homepage": "https://github.com/blockprotocol/blockprotocol#readme",
```
kachkaev, line 10:
```suggestion
  "homepage": "https://blockprotocol.org",
```

? (unless with stick with `...#readme` everywhere)

packages/core/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "bugs": {
+    "url": "https://github.com/blockprotocol/blockprotocol/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:blockprotocol/blockprotocol.git",
+    "directory": "packages/core"
```
kachkaev, line 17: Might need changing based on another comment

packages/core/src/core-handler.ts
```diff
@@ -0,0 +1,330 @@
...
+      messageName,
+      serviceName,
+    }: {
+      callback: GenericMessageCallback;
+      messageName: string;
+      serviceName: string;
+    },
+  ) {
+    this.messageCallbacksByService[serviceName] ??= {};
+    this.messageCallbacksByService[serviceName]![messageName] = callback;
```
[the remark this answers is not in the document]
kachkaev: That looks like a valid point. What if `messageName` is `__proto__`?

kachkaev (commented):
Not sure about the NPM publish error in CI, but ESLint errors may be to do with the contents of root `package.json` → `prepare` script. It contains this ATM:

```
    "prepare": "patch-package && husky install && yarn workspace mock-block-dock run prepare && yarn workspace @blockprotocol/site run prepare"
```

I guess we need to add more packages till we’ve switched to Nx / Turborepo.

---

I have only looked at the package metadata and haven’t reviewed the actual code. Not sure I have enough context to do see the big picture, which is needed to assess the details.

packages/core/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "url": "https://hash.ai"
+  },
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
+  "types": "./dist/esm/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "yarn clean; yarn build:cjs; yarn build:esm",
```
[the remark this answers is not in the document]
CiaranMn: Fully converting this to an ESM project involves us sorting out [linting](https://github.com/import-js/eslint-plugin-import/issues/2446) and probably a couple of other things – it'd be good to explore soon but I think it's too much of a distraction while we are trying to land a lot of change in the next week.

packages/core/package.json
CiaranMn: Moved in https://github.com/blockprotocol/blockprotocol/pull/323/commits/1660342ca9dd945376f882e19b2f139d7540b958

packages/core/package.json
CiaranMn: Fixed in https://github.com/blockprotocol/blockprotocol/pull/323/commits/1660342ca9dd945376f882e19b2f139d7540b958

packages/core/package.json
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/323/commits/1660342ca9dd945376f882e19b2f139d7540b958

packages/core/src/core-handler.ts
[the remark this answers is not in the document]
CiaranMn: How it's currently used, passing in a malicious `messageName` requires constructing one of these things, which means having access to the Object prototype anyway. But in theory someone in the future could pass in input from somewhere else when constructing one – I'm not sure why, but it's possible – so I've fixed in https://github.com/blockprotocol/blockprotocol/pull/323/commits/2074701710137d8826c4a7ca96907465bdf9de33.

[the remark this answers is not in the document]
CiaranMn: If someone did pass in untrusted input when constructing one they are also leaving themselves open to the actual callback supplied being malicious, but not much we can do about that.

CiaranMn:
Thanks @kachkaev, makes sense about `yarn prepare` and seems to have fixed everything! 🎉

packages/@blockprotocol/graph/src/graph-service.json
```diff
@@ -0,0 +1,688 @@
...
+    {
+      "name": "aggregateEntitiesResponse",
+      "description": "The response to a request to retrieve an aggregation of entities",
+      "source": "embedder",
+      "respondedToByMessage": null,
+      "sentOnInitialization": false,
+      "data": {
+        "type": "object",
+        "properties": {
+          "operation": { "type": "object" },
```
thehabbos007, line 174: Is this operation also a `{ "$ref": "https://blockprotocol.org/types/LinkedAggregationOperation" }`?

packages/core/package.json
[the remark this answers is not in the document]
CiaranMn: Nothing breaks if we do that, but the output does not have file extensions and is therefore not valid ESM output, and has to be bundled/transpiled again by something else to be valid ESM (as far as I understand it, not very confident in this area).

By 'fully converting to an ESM project' I mean adding `"type": "module"`, adding the extensions on import statements and fixing the resulting linting errors.

packages/@blockprotocol/graph/README.md
```diff
@@ -0,0 +1,104 @@
...
+1.  The starting values for any of the following messages you implement:
+
+- `blockEntity`
+- `blockGraph`
+- `linkedAggregations`
+
+These starting values should also be passed in a `graph` property object, if the block can be passed or assigned properties.
+
+```typescript
+import { GraphBlockHandler } from "@blockprotocol/graph";
```
teenoh, line 68: Shouldn't this be `EmbedderGraphHandler`?

packages/@blockprotocol/graph/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+const graphService = new EmbedderGraphHandler({
```
teenoh, line 70: Shouldn't this be `GraphBlockHandler`?

nathggns:
Reviewing now

packages/@blockprotocol/graph/src/graph-service.json
```diff
@@ -0,0 +1,688 @@
...
+          "results": {
+            "type": "array",
+            "items": { "$ref": "https://blockprotocol.org/types/Entity" }
```
Alfred-Mountfield, line 177: General question for most of these `get` operations:

Do users have to retrieve links _from_ these entities in separate queries?

packages/@blockprotocol/graph/src/react.ts
```diff
@@ -0,0 +1,56 @@
...
+  const [graphService, setGraphService] = useState<GraphBlockHandler>();
+  useLayoutEffect(() => {
+    if (ref.current) {
+      setGraphService(
+        new GraphBlockHandler({ callbacks, element: ref.current }),
+      );
+    }
+  }, [callbacks, ref]);
+
+  return { graphService };
```
nathggns, line 34: Why return this wrapped in an object? Why not just `return graphService` or `return [graphService]` if you want future extensibility? React hooks tend to leave it up to their caller to name fields. https://javascript.plainenglish.io/react-hooks-whats-the-difference-between-returning-an-array-and-an-object-34ccba62b71

packages/@blockprotocol/graph/src/react.ts
```diff
@@ -0,0 +1,56 @@
...
+  ref: RefObject<HTMLElement>;
+}) => {
+  const [graphService, setGraphService] = useState<GraphBlockHandler>();
+  useLayoutEffect(() => {
+    if (ref.current) {
+      setGraphService(
+        new GraphBlockHandler({ callbacks, element: ref.current }),
+      );
+    }
+  }, [callbacks, ref]);
```
nathggns, line 32: Discussing on Slack the bug here, where the graph service isn't reconstructed when the element changes

packages/@blockprotocol/core/src/core-block-handler.ts
```diff
@@ -0,0 +1,55 @@
...
+    });
+  }
+
+  /**
+   * Receives the {@link EmbedderInitMessage} sent by the embedding application,
+   * which is a series of payloads namespaced by service and message name.
+   * Calls the individual callbacks registered for each of these values.
+   *
+   * Useful for HTML blocks receiving messages which are sent on initialization by the app
+   * – other types of blocks can receive these messages synchronously via properties.
```
nathggns, line 31: Is there an example of this somewhere? I don't see how this is done. With the react integration, I don't see how an update to these properties triggers a re-render.

nathggns (commented):
Some initial thoughts– but this is really exciting stuff & builds a strong foundation

packages/@blockprotocol/graph/src/graph-service.json
CiaranMn: At the moment, yes, and there's no easy way of doing so. I think we need some 'get me the graph around this entity' query, or arguments to this which allow you to do so.

packages/@blockprotocol/core/src/core-block-handler.ts
CiaranMn: Yes – this is really for blocks that can't receive props directly but must retrieve them. i.e. an arbitrary pile of HTML

packages/@blockprotocol/graph/src/graph-service.json
CiaranMn: It is, thank you

packages/@blockprotocol/graph/src/graph-service.json
CiaranMn: Tracking [here](https://app.asana.com/0/1201095311341924/1202414058995815) (internal)

packages/@blockprotocol/graph/src/react.ts
CiaranMn: I don't think arrays are great for future extensibility because (a) less obvious what each entry is, (b) must retrieve unneeded variables if you don't want the 1st, 2nd etc.

The blog you link says:
> If our component will only have 1 (or few instances) of a hook, use object return.

What's the problem with having it an object? People can rename it if they want

packages/@blockprotocol/core/src/core-block-handler.ts
CiaranMn: If our component will only have 1 (or few instances) of a hook, use object return. Please let me know if another kind of example or discussion would be useful

packages/@blockprotocol/graph/src/react.ts
nathggns: This is probably a stylistic preference, so will resolve this.

packages/@blockprotocol/graph/README.md
CiaranMn: Yes, thank you - https://github.com/blockprotocol/blockprotocol/pull/323/commits/9b14e2b54414b4a6bcb9e3ff66bd432cfa4e2103

packages/@blockprotocol/graph/src/react.ts
CiaranMn: Thanks very much for the discussion - improved [here](https://github.com/blockprotocol/blockprotocol/pull/323/files/17500ecacb229bb60522a83b9f450b2aca10f260..9b77ba16366e1e528546cb4f18c666007e25980a)

nathggns (dismissed):

packages/@blockprotocol/core/src/core-handler.ts
```diff
@@ -0,0 +1,345 @@
...
+    if (
+      !("requestId" in message) ||
+      !("service" in message) ||
+      !("source" in message) ||
+      !("messageName" in message)
+    ) {
+      return false;
+    }
+    return true;
+  }
```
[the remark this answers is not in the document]
teenoh: Nit: Could also do 

```
return (
    typeof message === "object" &&
    message !== null &&
    [ 
      "requestId", 
      "service", 
      "source", 
      "messageName"
    ].every(key => key in message)

```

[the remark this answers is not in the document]
CiaranMn: I think it's much of a muchness so I'm going to commit the one I can do with a click. :)

packages/@blockprotocol/core/src/core-handler.ts
```diff
@@ -0,0 +1,345 @@
...
+ *
+ * This base class handles those common functions. There are two implementers,
+ * Block and Embedder, to handle the divergent functionality (e.g. init messages).
+ */
+export abstract class CoreHandler {
+  /** a default callback for use where no callback is registered for a message */
+  private readonly defaultMessageCallback?: GenericMessageCallback;
+
+  /** an object containing registered callbacks for messages, by service and message name */
+  private readonly messageCallbacksByService: MessageCallbacksByService;
```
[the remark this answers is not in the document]
CiaranMn: It started out as an object of objects, but I then converted it to an object of maps to avoid a theoretical vulnerability if someone allows someone else to pass in malicious message names.

The root thing could be a map as well, I just didn't convert it.

packages/@blockprotocol/core/src/core-handler.ts
```diff
@@ -0,0 +1,345 @@
...
+  protected constructor({
+    element,
+    sourceType,
+  }: {
+    element: HTMLElement;
+    sourceType: "block" | "embedder";
+  }) {
+    this.responseSettlersByRequestIdMap = new Map();
+    this.messageCallbacksByService = {};
+    this.services = new Map();
```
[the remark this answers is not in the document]
CiaranMn: If you mean why not defaults where declared, no particular reason. Do you think setting as defaults is better?

If you mean why not leave them undefined until being used, I think it's less hassle to have them guaranteed to exist when the instance is created, and if the handler is going to be used at all they'll all be necessary (rather than being able to leave some undefined indefinitely).

packages/@blockprotocol/core/src/core-handler.ts
```diff
@@ -0,0 +1,345 @@
...
+  >(
+    this: CoreHandler,
+    args: SendMessageArgs & { respondedToBy: string },
+  ): Promise<MessageData<ExpectedResponseData, ExpectedResponseErrorCodes>>;
+
+  /**
+   * Sends a message by dispatching a CustomEvent containing the message details.
+   * If the message expects a response, i.e. has a value for respondedToBy,
+   * this will return a promise which will be settled when the response is received.
+   * Otherwise the sent message is returned.
```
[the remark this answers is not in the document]
CiaranMn: Not at the moment, no, other than "it's available and someone might want it for something" – we could return undefined instead

packages/@blockprotocol/graph/src/graph-service.json
```diff
@@ -0,0 +1,690 @@
...
+            "description": "The entityId of the entity to update",
+            "type": "string"
+          },
+          "properties": {
+            "description": "The properties of the entity to update, which will be merged with any other properties on the entity",
+            "type": "object"
+          }
+        },
+        "required": ["entityId", "properties"]
+      },
```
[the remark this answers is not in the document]
CiaranMn: My ideal scenario is that we have a single service definition - a longer version of this file - from which we can generate the types and the documentation (i.e. have markdown descriptions all over the place). 

I say that because it seems more accessible to have people write the thing in JSON, and generate other things from that, rather than having them write it in TypeScript.

packages/@blockprotocol/core/src/core-handler.ts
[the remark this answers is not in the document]
nathggns: > Do you think setting as defaults is better?

Yeah, setting where the param is defined vs setting in constructor is functionally the same, but the first is cleaner

packages/@blockprotocol/core/src/core-handler.ts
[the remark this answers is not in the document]
nathggns: > we could return undefined instead

I think that's my preference.

packages/@blockprotocol/core/src/core-handler.ts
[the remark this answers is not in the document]
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/323/commits/6a1ee261c702e60df6d0493c8faf3e2c2101b2cd

packages/@blockprotocol/core/src/core-handler.ts
```diff
@@ -0,0 +1,345 @@
...
+    }
+  }
+
+  /**
+   * Checks all incoming events, and processes the Block Protocol messages:
+   * 1. Settles outstanding promises where the message is a response to a previously sent message
+   * 2. Calls any callbacks which have been registered for dealing with the message
+   * 3. Processes the init messages - see {@link processInitMessage}
+   */
+  private async processReceivedMessage(
```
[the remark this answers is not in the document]
CiaranMn: Doesn't need to be, removed in https://github.com/blockprotocol/blockprotocol/pull/323/commits/6a1ee261c702e60df6d0493c8faf3e2c2101b2cd

packages/@blockprotocol/core/src/core-handler.ts
```diff
@@ -0,0 +1,345 @@
...
+
+    this.element = element;
+    this.sourceType = sourceType;
+    (this.constructor as typeof CoreHandler).instanceMap.set(element, this);
+
+    this.attachEventListeners();
+  }
+
+  private eventListener = (event: Event) => {
+    void this.processReceivedMessage(event as CustomEvent);
```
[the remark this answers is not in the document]
CiaranMn: I've removed all the `void` and caught the errors thrown from `callCallback` in https://github.com/blockprotocol/blockprotocol/pull/323/commits/6a1ee261c702e60df6d0493c8faf3e2c2101b2cd

For now just logging and throwing again - I think actual usage will tell us more about what's best

packages/@blockprotocol/core/src/core-handler.ts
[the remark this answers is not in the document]
CiaranMn: Updated in https://github.com/blockprotocol/blockprotocol/pull/323/commits/6a1ee261c702e60df6d0493c8faf3e2c2101b2cd

nathggns (approved):
Thanks Ciaran, this is fantastic.

packages/@blockprotocol/core/src/core-handler.ts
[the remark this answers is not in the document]
nathggns: 👍

packages/@blockprotocol/core/src/core-handler.ts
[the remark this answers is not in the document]
nathggns: 👍

packages/@blockprotocol/core/src/core-handler.ts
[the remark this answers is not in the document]
nathggns: 👍

packages/@blockprotocol/core/src/core-handler.ts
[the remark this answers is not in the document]
nathggns: 👍

PR #326, opened 2022-06-08
Update mock-block-dock for 0.2 spec

CiaranMn:
## Purpose

This PR updates `mock-block-dock` for the 0.2 spec changes, and is based off the branch introducing the new services packages, since it relies on them.

I've published a beta version as `0.0.11-beta` so that it can be used to test blocks without interfering with anyone installing it until everything else has been updated.

## Changes
- Tweaks the React hooks introduced in #323 to avoid infinite state change loop
- Expands the test block / app in `mock-block-dock`
- Updates the callbacks in `mock-block-dock` to match the new callbacks
- Updates the shape of entities and types in `mock-block-dock` to match the new structure (with properties nested under `properties`, and schema nested under `schema`, respectively)

## Known issues
We need to handle the ref changing in the React graph hooks – to be handled in #323 

## Next steps

Update blocks to use the new approach, which can be tested using this updated mock-block-dock: 

`yarn add mock-block-dock@0.0.11-beta`

---
  - https://app.asana.com/0/0/1202353766855926


packages/mock-block-dock/dev/test-block.tsx
```diff
@@ -1,14 +1,44 @@
...
+      <h1>
+        Hello {properties.name}! The id of this block is {entityId}
+      </h1>
+      <input
+        type="text"
+        placeholder="This block's entity's 'name' property"
+        value={properties.name}
+        onChange={async (event) => {
+          try {
+            const { data, errors } = await graphService!.updateEntity({
```
nathggns, line 28: Unsure if this is supposed to serve as a good example to other users, or just as a little testing tool, but you've got the potential for race conditions because this is handle async - you probably want to set up a queue or something. That reminds me, is the value in `properties` gauranteed to be updated by the time `updateEntity` is finished. Like what's the order in terms of re-rendering with new `properties` vs the promise from `updateEntity` resolving?

packages/mock-block-dock/dev/test-block.tsx
```diff
@@ -1,14 +1,44 @@
...
 };
 
-export const TestBlock: BlockComponent<AppProps> = ({ entityId, name }) => {
+export const TestBlock: BlockComponent<AppProps> = ({ graph }) => {
+  const {
+    blockEntity: { entityId, properties },
+  } = graph;
+  const blockRef = useRef<HTMLDivElement>(null);
+
+  const { graphService } = useGraphBlockService({ ref: blockRef });
```
nathggns, line 15: I'd argue the signature should be more like

```suggestion
  const { graphService } = useGraphBlockService(ref, { callbacks: { /*...*/ } });
```

Feels cleaner in the event you don't want to pass callbacks

nathggns (dismissed):

packages/@blockprotocol/graph/src/react.ts
```diff
@@ -16,20 +16,17 @@ export type BlockComponent<
...
   useLayoutEffect(() => {
-    if (ref.current) {
-      setGraphService(
-        new GraphBlockHandler({ callbacks, element: ref.current }),
-      );
+    if (ref.current && !graphService) {
+      setGraphService(new GraphBlockHandler({ element: ref.current, ...rest }));
     }
-  }, [callbacks, ref]);
+  }, [graphService, ref, rest]);
```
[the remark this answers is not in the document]
CiaranMn: This has been changed in #323

packages/mock-block-dock/dev/test-block.tsx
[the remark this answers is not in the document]
CiaranMn: Agreed – the fact that it might be null makes it more annoying than the previous function approach, so this would be a good improvement to do soon

packages/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -42,29 +47,84 @@ export const MockBlockDock: VoidFunctionComponent<MockBlockDockProps> = ({
...
+
+  useEffect(() => {
+    if (graphService) {
+      graphService.linkedAggregations({ data: linkedAggregations });
+    }
+  }, [linkedAggregations, graphService]);
+
+  const Component = cloneElement(Children.only(children), propsToInject);
+
+  return <div ref={wrapperRef}>{graphService ? Component : null}</div>;
```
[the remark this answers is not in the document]
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/326/commits/ac20fe088d9f0a65f8e54f1f0cadaafb44736e98

packages/mock-block-dock/package.json
```diff
@@ -35,6 +35,7 @@
     "run-dev-server": "cross-env NODE_ENV=development webpack-dev-server --config dev/webpack.config.js"
   },
   "dependencies": {
+    "@blockprotocol/graph": "0.0.2",
     "blockprotocol": "0.0.10",
```
thehabbos007, line 39: Don't we have to use `@blockprotocol/core` instead of `blockprotocol` here?

CiaranMn: Good spot - this is actually unused. Removed in https://github.com/blockprotocol/blockprotocol/pull/326/commits/175e1113a3122f39ee714b485f7db133a0436fe5

packages/mock-block-dock/src/use-mock-block-props.ts
```diff
@@ -24,79 +25,71 @@ import { MockData, useMockDatastore } from "./use-mock-datastore";
...
     };
 
-    const accountId = blockProperties.accountId ?? "accountId";
-
-    const blockEntity: BlockProtocolEntity = {
-      accountId,
+    const newBlockEntity: Entity = {
       entityId: "block1",
+      entityTypeId,
+      properties: {},
```
thehabbos007, line 68: I'm wondering if empty properties as the initial object is a good idea? It would be invalid properties for any schema that has required fields, and the blocks will have to check if `properties` is an empty object before using it as props. I think `undefined` would make it clear that no props have been set yet.

thehabbos007: For context, in the calculation block, i assume the props can always be deserialized, which fails on initial load
image

CiaranMn: We discussed and landing on making the `properties` field optional, as the spec says, which I've done in [make entityTypeId and properties optional in graph Entity type](https://github.com/blockprotocol/blockprotocol/pull/326/commits/d7b90e57c97e05134cf7fd922608897659400e9e) - along with making `entityTypeId` optional, as per the spec.

We decided not to change the 'empty properties' object, since blocks should be able to handle an empty object if they are capable of being loaded with no props, and if they express `required` properties in their schema, embedding applications should not be loading them at all without the necessary properties present in the object.

PR #329, opened 2022-06-08
Update block scripts to include example-graph.json in build

teenoh:
This PR updates block-scripts to also bundle `example-graph.json` when building blocks. The `example-graph.json` file can be used to store dummy entities, links, entity types, the block will make use of in demo environments like the block hub.


CiaranMn (commented):
The version of `block-scripts` in `block-template` needs bumping (and maybe some yarn fixes after that if any still needed), but otherwise looks good to me

CiaranMn (approved):

teenoh:
> The version of `block-scripts` in `block-template` needs bumping (and maybe some yarn fixes after that if any still needed), but otherwise looks good to me

Thanks @CiaranMn. I've updated that now. I didn't notice any yarn fixes that need to be made though

PR #332, opened 2022-06-08
Override Tailwind :focus outline

luisbettencourt:
Some buttons on BlockProtocol were displaying an outline when clicked due to Tailwind's default `:focus` styling.
This was fixed by overriding the `outline` to `none` in `twind.config.js`. Keyboard navigation still shows an outline so it doesn't seem like an update to the `:focus-visible` selector is needed.

Before:

before

After:

after

[Asana Ticket](https://app.asana.com/0/1201095311341924/1202341139706565/f)

### DIFF
--- a/site/twind.config.js
+++ b/site/twind.config.js
@@ -15,4 +15,9 @@ module.exports = {
       backgroundColor: ["odd", "even"],
     },
   },
+  preflight: {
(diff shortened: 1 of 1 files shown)


nonparibus (approved):
Perfect, thank you.

PR #333, opened 2022-06-08
Package scripts fix for Windows

luisbettencourt:
When trying to run BlockProtocol locally on Windows, some package scripts failed because of the usage of `;` to run multiple commands in the same line. Replacing `;` with `&` should make it compatible with Windows.

Context:
Operators for running multiple commands on one line work differently for Unix-like systems (MacOs, Linux) and Windows.

1. `Bash` (Linux/MacOs): 
<html>
<body>

Syntax | Description
-- | --
A ; B | Run A and then run B
A && B | Run A, if it succeeds then run B
A \|\| B | Run A, if it fails then run B

</body>
</html>

2. `CMD` (Windows) :
<html>
<body>

Syntax | Description
-- | --
A & B | Run A and then run B
A && B | Run A, if it succeeds then run B
A \|\| B | Run A, if it fails then run B

</body>
</html>

[Tables souce](https://docgov.dev/posts/npm-scripts/)

It seems like the `;` operator is the only one that is incompatible with Windows, so I propose 2 possible solutions:

1. We keep using the `;` operator and tell users that are using windows to set their yarn config to run scripts via bash. They would have to run something like: `yarn config set script-shell /bin/bash`;
2. We replace the `;` operators with either:
   - `&`: In Unix all commands will run concurrently and in Windows all commands will run sequentially, regardless if the previous commands was successful or not;
   - `&&`: In both bash and cmd, all commands will run sequentially, and only if the previous one was successful.


luisbettencourt:
Decided to go with the solution of replacing the `;`  operator with `&&` because, as discussed with @CiaranMn, the scripts should be executed sequentially and only if the previous one succeeded. As a bonus this operator should work the same in both Unix systems and Windows (as opposed to `&` which runs concurrently on Unix and sequentially on Windows), and should further prevent weird behaviours.

CiaranMn (approved):
Thanks Luís, looks good!

PR #334, opened 2022-06-09
Call `excludeHiddenBlocks()` in two more places

kachkaev:
This PR follows https://github.com/blockprotocol/blockprotocol/pull/331 and improves UI consistency. Hidden blocks are no longer shown in `<BlocksSlider />` (both on the home page and on the block page). Hidden blocks are still available via direct URLs and are still embeddable.


teenoh (approved):
LGTM. Thanks @kachkaev!

PR #336, opened 2022-06-09
Update block-template and create-block-app for v0.2

CiaranMn:
This PR updates the `block-template` for the new spec (public docs coming next week).

It temporarily updates `create-block-app` to find `block-template` with a beta tag.

Both packages have been published with beta tags. This can be tested _outside the repo folder_ by:
- `npx create-block-app@beta block-name`, check the new template is used, and then `yarn dev` in the block folder.
- `npx create-block-app old-template`, check the old template is still used by default

---
  - https://app.asana.com/0/0/1202396799417882


packages/block-template/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "block-template",
-  "version": "0.0.18",
+  "version": "0.0.17-beta",
```
CiaranMn, line 3: `0.16` is the latest published version

kachkaev:
```suggestion
  "version": "0.0.17-canary.0",
```

Beta, alpha or RC imply certain levels of readiness. We don’t have a classification of these (yet), so it’s better to avoid words that carry extra meaning. `canary` (or `next`) just mean work in progress. 

Adding numbers is also common, so we can stick to using them too.

packages/block-template/src/dev.tsx
```diff
@@ -12,7 +12,14 @@ const node = document.getElementById("app");
 
 const App = () => (
```
kachkaev, line 13: It’s a little confusing to see `App` here while there exists `export const App` in `app.tsx`. How about using this PR as an opportunity to reduce ambiguity?

packages/create-block-app/create-block-app.js
```diff
@@ -40,6 +40,7 @@ const templatePackageName = "block-template";
 
   await pacote.extract(templatePackageName, resolvedBlockPath, {
     registry: process.env.NPM_CONFIG_REGISTRY,
+    defaultTag: "beta",
```
kachkaev, line 43: It’s simple to forget to undo this. How about leveraging `process.env.SOMETHING` instead?

packages/block-template/package.json
CiaranMn: I have updated this, and will update other packages as I publish new versions (rather than publishing identical versions of everything now to give them `canary` tags).

packages/block-template/src/dev.tsx
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/336/commits/8b7f464a28817647081e767105e5d98f5c7fda6f

PR #215, opened 2022-02-11
Investigate `glob` and `fs` / `fs-extra` imports

kachkaev:
Follows #211 & #212 

[Asana task](https://app.asana.com/0/1201707629991380/1201795402978687/f) (internal link)

PR #339, opened 2022-06-10
Replace `fs` with `fs-extra` in `blocks.ts`

kachkaev:
This PR was created in response to https://github.com/vercel/nft/issues/293#issuecomment-1152330914. It follows #313 and relates to #215.

I bumped Next to 12.2 as it includes https://github.com/vercel/nft/pull/285 (it allows us to `node:path` instead of `path` and still get correct file detection).


site/src/lib/blocks.ts
```diff
@@ -1,8 +1,9 @@
 import { BlockMetadata, BlockMetadataRepository } from "blockprotocol";
-// eslint-disable-next-line no-restricted-imports,unicorn/prefer-node-protocol -- https://github.com/vercel/nft/issues/293
-import fs from "fs";
-import glob from "glob";
+import fs from "fs-extra";
+import { globby } from "globby";
 import hostedGitInfo from "hosted-git-info";
+// eslint-disable-next-line unicorn/prefer-node-protocol -- Replace "path" with "node:path" after upgrading next from 12.1.6 https://github.com/vercel/nft/issues/293
+import path from "path";
```
kachkaev, line 6: A newer version of Next should include an updated version of `@vercel/nft`, which should allow us to restore `node:` prefix. I can keep this PR open till then. Context: https://github.com/vercel/nft/pull/285

site/src/lib/blocks.ts
```diff
@@ -84,49 +85,56 @@ const getRepositoryUrl = (
...
-          .split("/")
-          .join("/blocks/")}`,
-        lastUpdated: null, // TODO: derive from block data when provided by the hub
-      };
+  const blockMetadataFilePaths = await globby(
+    path.resolve(process.cwd(), `public/blocks/**/block-metadata.json`),
+  );
+
+  const result: ExpandedBlockMetadata[] = [];
+  for (const blockMetadataFilePath of blockMetadataFilePaths) {
```
kachkaev, line 93: Most of the diff is related to whitespace. I switched from `.map()` to `for of` to be able to use `await` in the loop.

site/src/lib/blocks.ts
kachkaev: This problem is no longer present in Next `12.1.7-canary.44`. I suggest waiting for a stable release of Next.js before finishing this PR.

site/src/lib/blocks.ts
```diff
@@ -51,12 +51,12 @@ const generateBlockFileUrl = (
 const getRepositoryUrl = (
   repository: BlockMetadataRepository | undefined,
   commit: string,
-  path: string | undefined,
+  hubInfoDirPath: string | undefined,
```
kachkaev, line 54: Avoids shadowing of `path`

CiaranMn (approved):
Looks good, tested and blocks working on hub

PR #342, opened 2022-06-10
Add base styles using SCSS modules

nathggns:
Using CSS modules to scope the styles, this adds our base styles to the block-template. 

Requires block authors to apply the `.block` className generated by css modules.

Also includes `.box` and `.small-text` helpers, which only work nested inside `.block`. 

image

https://app.asana.com/0/1200211978612931/1202361526477995/f


packages/block-template/src/base.module.scss
```diff
@@ -0,0 +1,251 @@
...
+    transition: 250ms ease-in-out;
+  }
+
+  button:hover {
+    background-color: var(--brand-primary-hover);
+    cursor: pointer;
+  }
+
+  /* Box element for containers and sections */
+  :global(.box) {
```
kachkaev, line 244: Not a blocker, just curious: can this still create clashes?

kachkaev (dismissed):
Nice! I guess we can also bump `block-scripts` and `block-template` to new `0.0.x` versions if we want to release after merging.

packages/block-template/src/base.module.scss
nathggns: So this will generate something like `.some-random-string .box` - if the EA styles `.box` globally, then yes, it will hit this, but this won't leak from the block to the EA.

packages/block-template/src/sass.d.ts
```diff
@@ -0,0 +1 @@
+declare module "*.scss";
```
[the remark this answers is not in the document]
nathggns: Great suggestion

packages/block-template/src/base.module.scss
kachkaev: I thought `:global` would exclude `.some-random-string` from the output, but nevermind. Let’s see how it goes.

I’m also curious about the nature of `.some-random-string` in general. If it’s a hash just from the local relative file name, we may end up with duplicate random strings used in different blocks. I guess it’s more than that, so let’s try and see!

kachkaev (approved):

packages/block-template/src/base.module.scss
kachkaev: Ah I see what you mean now. Without `:global` it’s `styles.box` but with global it’s `.some-random-string .box`. Got you!

nathggns: Basically just means `.box` works globally without using a dynamic classname, provided its within an element with the dynamic classname

PR #343, opened 2022-06-12
Custom elements: add custom element block template, support to mock-block-dock

CiaranMn:
This PR makes several changes to introduce better custom element / web component support:
1. Add support for rendering custom elements to `mock-block-dock` – as part of this it updates the interface so that block source is consistently passed by prop, rather than having React components sent as a child (which was making both the API and the internals more complicated than they needed to be)
2. Add a custom element base class to `@blockprotocol/graph`
3. Use the custom element base class to add a new custom element `block-template`
4. Handle multiple templates in `create-block-app`

In passing it also makes a couple of general DX improvements:

- Debug mode in `mock-block-dock` to show the mock datastore, and the properties being passed to the block
Screenshot 2022-06-10 at 11 36 53

- Usage instructions via `create-block-app --help`
Screenshot 2022-06-12 at 14 24 45

### How to test

`npx create-block-app@next test-block --template custom-element`
`cd test-block && yarn install && yarn dev`

### Note to reviewers

The react template is largely unchanged – I suggest focusing on the other changes.

---
  - https://app.asana.com/0/0/1202107700561483
  - https://app.asana.com/0/0/1202405291900011


README.md
```diff
@@ -32,10 +32,13 @@ We welcome [contributions](#contributing) to this repository. Within it you will
     - the [explanatory documentation](https://blockprotocol.org/docs) at [src/\_pages/docs](https://github.com/blockprotocol/blockprotocol/tree/main/site/src/_pages/docs)
 
 1.  `packages`: utility packages for constructing Block Protocol blocks. These are all available via `yarn add <package_name>` or `npm install <package_name>`. See the individual README files in each folder for more details.
-    - `block-template`: a template for a React-based block
-    - `blockprotocol`: TypeScript types for Block Protocol properties and functions, as described in the spec
+    - `@blockprotocol/core`: implements the Block Protocol core specification for blocks and embedding applications
+    - `@blockprotocol/graph`: implements the Block Protocol graph specification for blocks and embedding applications
+    - `block-scripts`: scripts used by `block-template`
+    - `block-template`: starter templates for blocks, used by `create-block-app
```
nathggns, line 38:
```suggestion
    - `block-template`: starter templates for blocks, used by `create-block-app`
```

nathggns:
(Not sure why it's saying reviewed here, that was just one comment - need to actually review later)

README.md
```diff
@@ -32,10 +32,13 @@ We welcome [contributions](#contributing) to this repository. Within it you will
     - the [explanatory documentation](https://blockprotocol.org/docs) at [src/\_pages/docs](https://github.com/blockprotocol/blockprotocol/tree/main/site/src/_pages/docs)
 
 1.  `packages`: utility packages for constructing Block Protocol blocks. These are all available via `yarn add <package_name>` or `npm install <package_name>`. See the individual README files in each folder for more details.
-    - `block-template`: a template for a React-based block
-    - `blockprotocol`: TypeScript types for Block Protocol properties and functions, as described in the spec
+    - `@blockprotocol/core`: implements the Block Protocol core specification for blocks and embedding applications
+    - `@blockprotocol/graph`: implements the Block Protocol graph specification for blocks and embedding applications
+    - `block-scripts`: scripts used by `block-template`
+    - `block-template`: starter templates for blocks, used by `create-block-app`
```
kachkaev, line 38: Not sure about this move – I haven’t seen multiple templates co-located in a single package. I’d go with one template per package or even keep templates in the repo instead of npm (like [Next.js examples](https://github.com/vercel/next.js/tree/canary/examples)).

Co-locating templates in single package means that users will see version bump for React template while we’ve only updated web components code. That’s confusing.

Some thoughts on this: [Slack](https://hashintel.slack.com/archives/C02LG39FJAU/p1652379269733239?thread_ts=1652165331.405139&cid=C02LG39FJAU) (internal)

CiaranMn: I took this from `create-next-app`, which has multiple templates co-located in a single package, albeit just JS and TS.

Do we think it's likely that users are going to get confused by this? They aren't interacting with this package directly, after all, but via `create-block-app`. And if we have templates separately in the repo, there will be no version bumps at all.

If we think this is going to be an actual problem, can we split them out in a future PR?

packages/block-scripts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "block-scripts",
-  "version": "0.0.7",
+  "version": "0.0.8-next.1",
```
kachkaev, line 3: What is the reason for using `next`? This creates some confusion between ‘next version’ and Next.js. which we also use. I refer to confusion in conversations rather than the code itself. It’s nice to have a technical vocabulary where each word has only one meaning.

CiaranMn: Per your suggestion [here](https://github.com/blockprotocol/blockprotocol/pull/336#discussion_r893702695)

CiaranMn: We can use `canary` in future if you now would prefer not to use `next`. I suggest we don't do this now as it will mean we have even more tags in published packages.

packages/block-template/templates/custom-element/src/app.ts
```diff
@@ -0,0 +1,37 @@
...
+  /** @see https://lit.dev/docs/components/rendering */
+  render() {
+    return html` <h1>Hello, ${this.blockEntity.properties.name}</h1>
+      <p>
+        The entityId of this block is ${this.blockEntity.entityId}. Use it to
+        update its data when calling updateEntities.
+      </p>
+      <!-- @see https://lit.dev/docs/components/events -->
+      <input
+        @change=${this.handleInput}
```
kachkaev, line 33: Is this where we use decorator syntax? What would the code like without this language feature? Betting on decorators is pretty risky.

kachkaev (approved):

PR #346, opened 2022-06-13
Adding RFC process

Alfred-Mountfield:
This PR introduces / proposes an RFC process for the Block Protocol.

It is **heavily** inspired (as credited in the code) by:
* [Rust](https://github.com/rust-lang/rfcs)
* [React](https://github.com/reactjs/rfcs)

Some highlighted differences/modifications from the other projects
* RFCs are stored in this monorepo in an `rfcs` folder rather than in their own repo.
* Removed mention of a Final Comment Period until we want that overhead, for now it seems okay to leave approval and implementation up to general discretion
* Taken React's more lenient approach to RFCs being updated over time, Rust suggests that RFCs should basically not change after being accepted and instead should be followed-up with new RFCs
* Using GitHub Discussions rather than Issues, we seem to be wanting to utilise the GitHub Discussions features (as is the current way we handle RFCs), so I've changed mentions of issues to that instead, and added a new label "category: rfc" which can be used to filter the PRs/issues view.

For all other information please read the proposed changes, rendered links for convenience:
[Updated top-level README](https://github.com/blockprotocol/blockprotocol/blob/am/rfcs/README.md)
[New RFC Readme](https://github.com/blockprotocol/blockprotocol/blob/am/rfcs/rfcs/README.md)
[New RFC Template](https://github.com/blockprotocol/blockprotocol/blob/am/rfcs/rfcs/0000-template.md)

## TO-DO
- [x] There's a TODO in the markdown to link to the docs for services as it seems to me that it's a good example of something that would need an RFC. Currently I'm unsure if we have a link or if that's coming soon.

## Future steps:
- [ ] Add automated tagging of labels for RFCs


rfcs/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+The most common preparations for writing and submitting an RFC includes talking the idea over on our [community Discord server], or in relevant Issue threads when the idea stems from previous discussion.
+
+As a rule of thumb, receiving encouraging feedback from long-standing project developers, and particularly members of the core team is a good indication that the RFC is worth pursuing.
+
+## What the process is
+
+[What the process is]: #what-the-process-is
+
+In short, to get a major feature added to the Block Protocol, one must first get the RFC merged into the [`/./rfcs`](../rfcs) folder as a markdown file. At that point the RFC is "active" and may be implemented with the goal of eventual inclusion into the Block Protocol.
```
thehabbos007, line 54: nit:
Should this path just be
```suggestion
In short, to get a major feature added to the Block Protocol, one must first get the RFC merged into the [`/rfcs`](../rfcs) folder as a markdown file. At that point the RFC is "active" and may be implemented with the goal of eventual inclusion into the Block Protocol.
```

rfcs/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+The Block Protocols's RFC process strongly owes its inspiration to the [Rust RFC process], and [React RFC process].
+
+[Rust RFC process]: https://github.com/rust-lang/rfcs
+[React RFC process]: https://github.com/reactjs/rfcs
+
+We're open to changing it as needed in response to feedback.
+
+[community Discord server]: https://blockprotocol.org/discord
+[developer discussion forum]: https://internals.rust-lang.org/
```
nonparibus, line 108: I don't think this should be here (but if appropriate, our forum can be found at https://community.hash.ai/)

rfcs/README.md
```diff
@@ -0,0 +1,108 @@
+# Block Protocol RFCs
+
+The "RFC" (request for comments) process is intended to provide a consistent and controlled path for new features to enter the project.
+
+Many changes, including bug fixes and documentation improvements can be implemented and reviewed via the normal GitHub pull request workflow.
+
+Some changes though are "substantial", and we ask that these be put through a bit of a design process and produce a consensus approved by the Block Protocol Core Team.
+
+[Active RFCs](https://github.com/blockprotocol/blockprotocol/discussions/categories/rfc)
+[Proposed RFCs](https://github.com/blockprotocol/blockprotocol/pulls?q=is%3Aopen+is%3Apr+label%3A%22category%3A+rfc%22)
```
nonparibus, line 10: This would cover not just proposed RFCs, but any pending/unmerged change to an RFC, right? (e.g. it may contain low-value submissions, uninteresting standalone typo fixes, etc... although hopefully those wouldn't remain "open" for long). Not sure this needs changing, just clarifying our expectations around this.

rfcs/README.md
Alfred-Mountfield: I'm not sure how that's still here, I swear I did a ctrl + F for rust, good catch!

rfcs/README.md
Alfred-Mountfield: It depends on how we use the label, we could create a few labels to distinguish between the various categories. But generally yes any open PR that is tagged as "rfc" will be shown in this view. I'm not sure that's too unfortunate as the main thing people want to look at are "Active RFCs", which aren't necessarily implemented yet (just have been accepted)

rfcs/README.md
```diff
@@ -0,0 +1,107 @@
...
+- Submit a pull request. As a pull request the RFC will receive design feedback from the larger community, and the author should be prepared to revise it in response.
+- Now that your RFC has an open pull request, use the issue number of the PR to update your `0000-` prefix to that number
+- Build consensus and integrate feedback. RFCs that have broad support are much more likely to make progress than those that don't receive any comments.
+- The core team will discuss the RFC pull request, as much as possible in the comment thread of the pull request itself. Offline discussion will be summarized on the pull request comment thread.
+- RFCs rarely go through this process unchanged, especially as alternatives and drawbacks are shown. You can make edits, big and small, to the RFC to clarify or change the design, but make changes as new commits to the pull request, and leave a comment on the pull request explaining your changes. Specifically, **do not squash or rebase commits after they are visible on the pull request**.
```
CiaranMn, line 64: we should have a branch format that _we_ use for RFCs, because then we can protect the ones we create according to the pattern and prevent ourselves accidentally force pushing

rfcs/0000-template.md
```diff
@@ -0,0 +1,92 @@
+- Feature Name: (fill me in with a unique ident, `my_awesome_feature`)
```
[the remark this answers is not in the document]
Alfred-Mountfield:
```suggestion
- Feature Name: (fill me in with a unique ident such as `my-feature`, this should be used in the name of the file)
```

rfcs/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+## What the process is
+
+[What the process is]: #what-the-process-is
+
+In short, to get a major feature added to the Block Protocol, one must first get the RFC merged into the [`/rfcs`](../rfcs) folder as a markdown file. At that point the RFC is "active" and may be implemented with the goal of eventual inclusion into the Block Protocol.
+
+- Fork the Block Protocol repo https://github.com/blockprotocol/blockprotocol
+- Copy [`rfcs/0000-template.md`](0000-template.md) to `rfcs/text/0000-my-feature.md` (where
+'my-feature' is descriptive. Don't assign an RFC number yet).
```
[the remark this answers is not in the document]
Alfred-Mountfield: Should be sorted by the change above.

rfcs/README.md
Alfred-Mountfield: Lets go with `{initial}/rfc-{branch_name} so e.g. "am/rfc-my-feature"

CiaranMn (dismissed):

CiaranMn (approved):

PR #355, opened 2022-06-14
RFC - Non-Primitive Data Types (Graph)

Alfred-Mountfield:
This PR is still a work-in progress.

We define six primitive data types in the Graph type system. These data types define value spaces for the inner properties of entities, and currently are limited to basic primitives such as "text" or "number". Building robust data models will require the ability to define constraints and define validation mechanisms (think maximum/minimum lengths on properties, or applying regular expressions to define the pattern of a string).

This RFC will capture our thoughts on how to achieve this within the type system utilizing #428 and some new extensions to our Data Type model -- culminating in allowing the creation of non-primitive, "custom" data types.

[Rendered view](https://github.com/blockprotocol/blockprotocol/blob/am/rfc-graph-user-data-types/rfcs/text/0355-graph-non-primitive-data-types.md)


rfcs/text/0355-graph-non-primitive-data-types.md
```diff
@@ -0,0 +1,266 @@
...
+
+To provide a way for these constraints to be composable and re-usable, the Block Protocol type system will allow for data types to _inherit_ from other data types. This will allow for data type constraints to be reusable and composable among non-primitive data types. This inheritance will lean on how the `allOf` keyword works in JSON Schema. All data types referenced in the `allOf` array of a data type schema will provide value-space constraints to the inheritor.
+
+A second kind of composability of data types will also be possible by allowing data types to be defined as choices between data types. Conceptually this is similar to JSON Schema's `oneOf` keyword where a schema can take on the shape of _one of_ a list of schemas. This composability, instead of constraining the value-space, will allow data types to expand their value-space.
+
+Lastly, we will allow for data types to compose through the use of lists. This means that data types can be of `{ type: "array" }` and use the `items` keyword to define their value-space(s).
```
indietyp, line 48: The specific reason why data types can be arrays if they can already be modeled using property types? This introduces some redundancy and I imagine confusion, as in: do I construct arrays in my `property-type` or `data-type`.

nonparibus:
In December 2023, we introduced support for non-primitive data types in HASH, based on an advanced version of this RFC. We will be adding more and iterating on these, before introducing support for user-defined non-primitive data types (aka. "custom data types") in the HASH app: https://github.com/hashintel/hash/pull/3784 This draft PR will most likely be finalized for public comment in follow-up to that work, with HASH's implementation serving as a live sandbox for the ideas discussed within this RFC.

PR #363, opened 2022-06-15
Search Improvements

luisbettencourt:
This task aims to improve the search bar by doing the following:

- [x] Move the search bar into the top-level global navigation;
- [x] Move search results into a pop-over modal;
- [x] Update search result items and separate results in documentation from results in the spec.

modal

[Asana Task](https://app.asana.com/0/1201095311341924/1202398431154722/f)
[Notion](https://www.notion.so/hashintel/Search-Bar-Improvements-on-BP-org-da0de17e46e6417ba5ceb14715f0b734#1b03db34465244518933760df94aa60a)
[Figma](https://www.figma.com/file/Fpmde3E9nJMb4VLkPTYgrM/blockprotocol.org?node-id=3134%3A21678)


nathggns:
Looking at this now

nathggns (changes requested):
Thanks @luisbettencourt, looks like some really great work. Well done.

I had a few blocking comments, mostly around the implications of some react data/state management. Also one non-blocking suggestion to remove some dead code / make the UI slightly more visually appealing, but that's optional.

Let me know if you have any questions about my feedback.

nonparibus:
Can we open the modal with the `/` key instead? I don't think we need the full-blown `cmd+k` command-bar experience for what is _only_ website search.

luisbettencourt:
For both mac and windows?

nonparibus:
> For both mac and windows?

Yes. Any reason why not?

site/src/components/pages/docs/search/index.tsx
```diff
@@ -168,169 +175,148 @@ const Search: React.VoidFunctionComponent<SearchProps> = ({
...
-          value={searchText}
-          onChange={(event: ChangeEvent<HTMLInputElement>) =>
-            setSearchText(event.target.value)
-          }
-          placeholder="Search..."
-          required
-        />
-
-        {searchLoading ? (
+            placeholder="Search The BP"
```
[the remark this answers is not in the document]
nonparibus:
```suggestion
            placeholder="Search..."
```

site/src/components/pages/docs/search/search-suggested-links.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+interface SearchSuggestedLinksProps {
+  setSearchText?: (text: string) => void;
+  closeModal?: () => void;
+}
+
+const SUGGESTED_LINKS = [
+  "Introduction to the Block Protocol",
+  "Quick start guide to developing blocks",
+  "A guide for embedding applications",
```
nonparibus, line 16:
```suggestion
  "Embed blocks within your site or app",
```
We haven't published our actual guide for embedding applications yet, and we only explain what the 'EA' term means on certain pages. I think this is a clearer CTA that relates to our current public content better.

site/src/components/pages/docs/search/search-suggested-links.tsx
```diff
@@ -0,0 +1,118 @@
...
+          </Typography>
+          <Typography
+            sx={{
+              color: theme.palette.teal[600],
+              fontWeight: 500,
+              fontSize: 14,
+              textAlign: "start",
+            }}
+          >
+            Our open-source protocol for creating interoperable, data-driven
```
nonparibus, line 109:
```suggestion
            The open-source protocol for creating interoperable, data-driven
```

nonparibus (changes requested):
Small copy tweaks suggested

nonparibus:
Are we using the MUI modal component here?

Also spotted this weird squiggle on the right-side

Capture 2022-06-20 at 20 59 29@2x

nonparibus (changes requested):
Thanks for the changes. Two spotted issues.

**Issue #1 -** clicking the search-box jumplinks should take folks direct to the relevant pages in the docs we're trying to promote.

This means:
- link 1 (intro) to `/docs`
- link 2 (quickstart) to `/docs/developing-blocks`
- link 3 (embed) to `/docs/embedding-blocks`
- link 4 (FAQs) to `/docs/faq`

At the moment they throw up search results for the full entered queries which don't always even return a match. See below.

Capture 2022-06-21 at 01 37 12@2x

**Issue 2 -** results formatting. Some spacing is missing in the results themselves. See below.

Capture 2022-06-21 at 01 37 20@2x

site/src/components/pages/docs/search/search-list-category.tsx
```diff
@@ -0,0 +1,82 @@
...
+      searchResults.filter(
+        (result) => result._highlightResult?.title?.matchLevel !== "none",
+      ),
+    );
+    setItems(
+      searchResults.filter(
+        (result) => result._highlightResult?.title?.matchLevel === "none",
+      ),
+    );
+  }, [searchResults]);
```
CiaranMn, line 34: To expand on what @nathggns said on this – `sections` and `items` do not need to be in state at all.

They are derived from the prop `searchResults`, which is going to cause a re-render whenever it changes. With the existing approach, when `searchResults` updates the component is going to re-render, and _then_ we call `setSections/Items` with the new values, and then the component renders _again_ because state values have changed. There's no need for this – we only want a single re-render when the search results changes. Whenever we have a value that is always going to change at the same time as an existing prop or state value and is simply a result of applying some formula to that existing value, we don't need to store it in its own state variable.

We could just calculate these values directly on every render: 

```suggestion
  const sections = searchResults.filter(
    (result) => result._highlightResult?.title?.matchLevel !== "none",
  );
  const items = searchResults.filter(
    (result) => result._highlightResult?.title?.matchLevel === "none",
  );
```

with this approach they'll be calculated on every render, even if `searchResult` hasn't changed. e.g. as soon as someone types a letter in the search box, that updates state in `ModalSearch`, and this is a child of `ModalSearch` so it re-renders and these are run again, despite `searchResults` not having changed. 

We can use `useMemo` to avoid this, and only re-calculate sections/items when `searchResults` changes, as below. There probably isn't any practical difference in this case since the search results list is very short and filtering it is not expensive.

```suggestion
  const [sections, items] = useMemo(
    () => [
      searchResults.filter(
        (result) => result._highlightResult?.title?.matchLevel !== "none",
      ),
      searchResults.filter(
        (result) => result._highlightResult?.title?.matchLevel === "none",
      ),
    ],
    [searchResults],
  );
```

CiaranMn (commented):
Thanks Luís, great work! Since Nate is away for the next couple of days I'm following up on one of his suggestions

site/src/components/pages/docs/search/search-list-category.tsx
luisbettencourt: Thank you for the great insight, @CiaranMn!

nonparibus:
Functional UAT pass from me. It would be nice to be able to tab between results using arrow keys, and navigate to a selection using the 'enter' key, but the other aspects of the implementation itself do not now appear to be broken - thanks for making those changes.

CiaranMn (dismissed):
Excellent work, @luisbettencourt !

luisbettencourt:
> Functional UAT pass from me. It would be nice to be able to tab between results using arrow keys, and navigate to a selection using the 'enter' key, but the other aspects of the implementation itself do not now appear to be broken - thanks for making those changes.

No problem, thank you for the feedback! 
I find that I am able to do all that except navigate between results using the arrow keys (tab works). We could add that logic if you think it's worth it.

PR #368, opened 2022-06-15
Remove h4s from the sitemap

nathggns:
The docs sidebar is too granular right now. This makes it a bit easier to work with. 

This also fixes the breadcrumbs by removing the max width on them, and also removing the direct linking to h4s.


nonparibus (approved):
Functional UAT. Haven't reviewed code.

PR #372, opened 2022-06-15
Update block commits

CiaranMn:
This PR updates commits on the HASH blocks so that they get the revised READMEs introduced in https://github.com/hashintel/hash/pull/657 - these READMEs now display on their page on the hub.


kachkaev (approved):
I quickly checked the [preview deployment](https://blockprotocol-law7eaxgp-hashintel.vercel.app/hub) and did not spot any issues.

We can bump commits again once I’ve finished https://github.com/hashintel/hash/pull/653

PR #365, opened 2022-06-15
Merge spec into docs (with redirect) + new design 

nathggns:
https://app.asana.com/0/1201095311341924/1202393601483106/f

image

## Todos (in follow up)

- Search to the bottom
- Circle indicator instead of line indicator for headings (to differentiate between pages/sections)

PR #375, opened 2022-06-15
Fix Algolia search indexing

CiaranMn:
The Algolia search indexing assumed that the `spec` would be in its folder, but it's been moved to a subfolder of `docs` in https://github.com/blockprotocol/blockprotocol/pull/365

This PR fixes the Algolia script to account for this.

The script needs more work, particularly to get better search result specificity and highlighting from Algolia - we may want to index individual _sections_ of documents rather than the whole thing. For a future PR.

**Note**: I have already re-indexed the production index to ensure the correct results are showing. This workflow only runs on `main`. To test it, run it locally, amending the index to `blockprotocol-testing`.

### DIFF
--- a/.github/workflows/algolia-upload-index.yml
+++ b/.github/workflows/algolia-upload-index.yml
@@ -22,4 +22,4 @@ jobs:
       - run: yarn install
 
       - name: Sync Algolia Index
-        run: yarn exe scripts/sync-algolia-index.ts
+        run: yarn exe site/scripts/sync-algolia-index.ts
--- a/.github/workflows/algolia-upload-index.yml
+++ b/.github/workflows/algolia-upload-index.yml
@@ -22,4 +22,4 @@ jobs:
       - run: yarn install
 
       - name: Sync Algolia Index
-        run: yarn workspace @blockprotocol/site scripts/sync-algolia-index.ts
+        run: yarn workspace @blockprotocol/site exe scripts/sync-algolia-index.ts
--- a/.github/workflows/algolia-upload-index.yml
+++ b/.github/workflows/algolia-upload-index.yml
@@ -22,4 +22,4 @@ jobs:
       - run: yarn install
 
       - name: Sync Algolia Index
-        run: yarn exe site/scripts/sync-algolia-index.ts
+        run: yarn workspace @blockprotocol/site scripts/sync-algolia-index.ts
--- a/scripts/sync-algolia-index.ts
+++ b/scripts/sync-algolia-index.ts
@@ -19,6 +19,12 @@ type DocsFrontMatter = {
 
 type BlockProtocolFileTypes = "spec" | "docs";
 
+type FileTypeAndPaths = {
+  inputPath: string;
+  outputPath: string;
+  type: BlockProtocolFileTypes;
+};
+
 const getFileInfos = (
   dirPath: string,
   arrayOfFiles: Array<{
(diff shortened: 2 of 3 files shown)


kachkaev:
WDYT of moving this file from `scripts` to `site/scripts`? It seems to be related to that ‘app’ only. This can help us organise scripts in the future.

scripts/sync-algolia-index.ts
```diff
@@ -27,33 +33,26 @@ const getFileInfos = (
...
+      return;
+    }
+
+    const inputPath = path.join(dirPath, "/", file);
+    const outputPath = `.\\output\\${path.join(
+      "./output/",
+      dirPath,
+      "/",
+      file,
+    )}`;
```
kachkaev, line 53: Hmmm not sure I understand why we have `path.join` inside ``` `` ``` 🤔

CiaranMn:
> WDYT of moving this file from `scripts` to `site/scripts`? It seems to be related to that ‘app’ only. This can help us organise scripts in the future.

https://github.com/blockprotocol/blockprotocol/pull/375/commits/dbaa692c1d50cce63a0eeb63393bed803f8ba601

.github/workflows/algolia-upload-index.yml
```diff
@@ -22,4 +22,4 @@ jobs:
       - run: yarn install
 
       - name: Sync Algolia Index
-        run: yarn exe scripts/sync-algolia-index.ts
+        run: yarn exe site/scripts/sync-algolia-index.ts
```
kachkaev, line 25:
```suggestion
        run: yarn workspace @blockprotocol/site scripts/sync-algolia-index.ts
```

(both variants are generally equal, the main difference is `process.cwd()`

scripts/sync-algolia-index.ts
CiaranMn: I expect we will heavily rewrite this soon as I'd rather not start improving bits like this

site/scripts/sync-algolia-index.ts
```diff
@@ -19,6 +19,12 @@ type DocsFrontMatter = {
 
 type BlockProtocolFileTypes = "spec" | "docs";
```
kachkaev, line 20: Do we still need this?

CiaranMn: It is referred to in 3 places below

.github/workflows/algolia-upload-index.yml
```diff
@@ -22,4 +22,4 @@ jobs:
       - run: yarn install
 
       - name: Sync Algolia Index
-        run: yarn exe scripts/sync-algolia-index.ts
+        run: yarn workspace @blockprotocol/site scripts/sync-algolia-index.ts
```
kachkaev, line 25:
```suggestion
        run: yarn workspace @blockprotocol/site exe scripts/sync-algolia-index.ts
```

Oops, sorry I missed `exe` 😬😅

site/scripts/sync-algolia-index.ts
kachkaev: I’m wondering if there still exists conceptual necessity in this type, given that `spec` is a sub-path of `/docs`

CiaranMn: Ah, right – the designs for results still distinguish between 'spec' results and 'docs' results (in breadcrumbs), so I think it's useful to keep them distinct, yes

kachkaev:
@CiaranMn I’ve decided to merge this PR for you to fix ❌ on `main`. We can create a follow-up PR if anything is still missing!

PR #378, opened 2022-06-16
Styling and copy fixes

CiaranMn:
This PR fixes a few small issues:

### 1. Sidebar width and list item wrapping
Previously, the sidebar was not taking up the width specified in the designs and the code (300px), because the body content to the right was squeezing it out when it shouldn't.

Also, sidebar items were being cut off with an ellipsis, which meant that it was impossible to tell what many were about.
 
**Before**
Screenshot 2022-06-16 at 10 38 15

**After**
Screenshot 2022-06-16 at 10 37 28

**Before**
Screenshot 2022-06-16 at 10 42 24

**After**
Screenshot 2022-06-16 at 10 43 29

### 2. Spec copy tweaks

Clarify language related to custom elements, and the F# blog post.

### 3. Validation error wrapping

Wrap words in the error list in the block hub.

### DIFF
--- a/site/src/_pages/docs/1_developing-blocks.mdx
+++ b/site/src/_pages/docs/1_developing-blocks.mdx
@@ -16,7 +16,7 @@ In practice, most block developers will not need to know the lower-level details
 
 We provide three templates which allow you to define the entry point for your block in different ways:
 
-- `custom-element`: create a block defined as a custom element (or Web Component).
+- `custom-element`: create a block defined as a custom element (also known as Web Components).
 - `html`: create a block defined as an HTML file, with JavaScript added via `<script>` tags.
 - `react`: create a block defined as a React component.
 
@@ -217,7 +217,7 @@ If you want to write blocks **using other technologies or frameworks**, you have
 
 1.  use a `custom-element` template and use different approaches when constructing the element-i.e. use a custom element as a wrapper for anything else you like.
 1.  use an `html` template and import and use your additional libraries inside the `<script>` tag.
-1.  use a React component as a wrapper around something else. As an example, see [this blog post](https://hash.dev/blog/build-blocks-in-any-language) on writing a block in F#, wrapped as a React component.
+1.  use languages which can be transpiled to JavaScript. As an example, see [this blog post](https://hash.dev/blog/build-blocks-in-any-language) on writing a block using F#. This block uses React, you can use transpiled JavaScript in blocks defined as HTML files or custom elements too.
 
 ### I don’t want to use TypeScript
 
--- a/site/src/components/page-sidebar.tsx
+++ b/site/src/components/page-sidebar.tsx
@@ -40,9 +40,7 @@ const SidebarLink = styled(Link)(({ theme }) => ({
   fontSize: 15,
   paddingTop: 8,
   paddingBottom: 8,
-  whiteSpace: "nowrap",
-  textOverflow: "ellipsis",
-  overflow: "hidden",
+  wordBreak: "break-word",
 }));
 
 type SidebarPageSectionProps = {
@@ -434,6 +432,7 @@ export const Sidebar: VFC<SidebarProps> = ({
             "padding-top",
             "padding-bottom",
           ]),
+          wordBreak: "break-word",
         }}
       >
         <Box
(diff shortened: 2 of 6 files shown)


site/src/_pages/docs/1_developing-blocks.mdx
```diff
@@ -217,7 +217,7 @@ If you want to write blocks **using other technologies or frameworks**, you have
 
 1.  use a `custom-element` template and use different approaches when constructing the element-i.e. use a custom element as a wrapper for anything else you like.
 1.  use an `html` template and import and use your additional libraries inside the `<script>` tag.
-1.  use a React component as a wrapper around something else. As an example, see [this blog post](https://hash.dev/blog/build-blocks-in-any-language) on writing a block in F#, wrapped as a React component.
+1.  use languages which can be transpiled to JavaScript. As an example, see [this blog post](https://hash.dev/blog/build-blocks-in-any-language) on writing a block using F#. This block uses React, you can use transpiled JavaScript in blocks defined as HTML files or custom elements too.
```
luisbettencourt, line 220: Non blocking, just think this would read better like this:
```suggestion
1.  use a language which can be transpiled to JavaScript. As an example, see [this blog post](https://hash.dev/blog/build-blocks-in-any-language) on writing a block using F#. This block uses React, you can use transpiled JavaScript in blocks defined as HTML files or custom elements too.
```

luisbettencourt (approved):

site/src/components/page-sidebar.tsx
```diff
@@ -434,6 +432,7 @@ export const Sidebar: VFC<SidebarProps> = ({
             "padding-top",
             "padding-bottom",
           ]),
+          wordBreak: "break-word",
```
benwerner01, line 435: I think there's 2 minor visual issues now that words can break to new lines here:
1. The arrow icon should be more to the left when words break onto a new line to be consistent with the sections where this doesn't happen
2. The focus indicator no longer stretches to the full height of the element wrapping the text of a section

image

I think this could be addressed in follow-up though to help get the fixes in this PR out quicker.

benwerner01 (approved):
Thanks Ciaran, this looks good to me. Spotted 2 minor styling issues but don't think the should block merging this PR.

luisbettencourt (dismissed):

MaggieAppleton:
Good quick fix! I can help do small design tweaks in a follow-up PR around alignment, spacing and arrows. We still have to move the arrows to the right-aligned, and add a distinct secondary style for headers inside top-level sections 

OG design mocks
image

site/src/components/page-sidebar.tsx
CiaranMn: Thanks Ben - this was actually looking pretty bad in some situations, where the line would move outside of the link, so I've fixed it in https://github.com/blockprotocol/blockprotocol/pull/378/commits/be86a7f3006724c89320133e2e4b4babcf00017f

@nathggns I've left the indicator in, but hidden, as there's an effect which is doing some layout stuff with it. Haven't looked into exactly what but this change should fix the visual appearance without affecting whatever the behavior is

CiaranMn:
Thanks @MaggieAppleton - I've added the right-alignment of arrows since it was quick. Here's an updated shot with right-aligned arrows and the left indicator appearing properly

Screenshot 2022-06-16 at 12 00 40

site/src/components/page-sidebar.tsx
```diff
@@ -439,15 +448,8 @@ export const Sidebar: VFC<SidebarProps> = ({
...
-            width: 3,
-            height: 35,
-            backgroundColor: ({ palette }) => palette.purple[600],
             top: selectedOffsetTop === undefined ? 0 : selectedOffsetTop,
-            opacity: selectedOffsetTop === undefined ? 0 : 1,
-            transition: theme.transitions.create(["top", "opacity"], {
-              duration: 150,
-            }),
-            zIndex: 2,
+            opacity: 0,
```
luisbettencourt, line 454: With this approach we lose the indicator's movement animation. It's very subtle anyway, so it's probably not an issue?

CiaranMn: I think we can have it gone for now - if we want it back we will need the height of this indicator to be variable based on the height of the list item it's next to (and update that through resizes) - not sure it's worth it but will defer to @nathggns for later consideration about whether we update this further (after this PR, which will help with readability)

luisbettencourt (approved):

site/src/components/page-sidebar.tsx
nathggns: The old one animated its position between links, which seems to be why it was implemented this way. I guess we don't need that behaviour

nathggns: Will look at removing it now

PR #387, opened 2022-06-21
Show graph service message list. Consistency fixes in graph service spec

CiaranMn:
This PR adds a list of the available graph service messages to its page. We previously only linked out to the JSON, which was harder to read.

In the process of doing this I noticed and fixed the following inconsistencies in `graph-service.json`, which had fallen out of sync with how services are supposed to be defined (per the core spec):
1. the property on each message definition should be `messageName`, not `name` 
2. `errorCodes`, not `errorsCodes` 
3. `respondedToBy` not `respondedToByMessage`
4. A few descriptions were in the wrong place in the message `data` schemas

The PR also adds a type for service definitions to `@blockprotocol/core`, and adds anchor links to `h4`/`h5` messages in the spec, so that messages can link to each other and be linked to.

There is no existing design for this and it could definitely be improved on visually.

Screenshot 2022-06-21 at 08 41 40


site/src/_pages/docs/3_spec/2_graph-service-specification.mdx
```diff
@@ -344,6 +344,7 @@ As well as the formal service JSON specification, the messages can be sent and r
 
 #### <Hidden>Message definitions </Hidden>Specification
 
-##### Graph service JSON definition
+The messages available for exchange in the graph service are defined in the service's JSON definition [here](https://github.com/blockprotocol/blockprotocol/blob/main/packages/%40blockprotocol/graph/src/graph-service.json#L5) and are also
```
benwerner01, line 347: Nit:
```suggestion
The messages available for exchange in the graph service are defined in the [service's JSON definition](https://github.com/blockprotocol/blockprotocol/blob/main/packages/%40blockprotocol/graph/src/graph-service.json#L5) and are also
```

site/src/util/mdx-components.tsx
```diff
@@ -180,23 +181,45 @@ export const mdxComponents: Record<string, ComponentType> = {
...
         mt={HEADING_MARGIN_TOP.H5}
         mb={HEADING_MARGIN_BOTTOM}
         variant="bpHeading5"
         {...props}
-      />
+      >
+        {props.children}
+        <HeadingAnchor anchor={anchor} depth={5} />
+      </Heading>
     );
```
benwerner01, line 223: Just want to flag that giving h3 and h4 tags anchors breaks the current implementation of the sidebar in a way that it no longer displays which section you're currently in:

https://user-images.githubusercontent.com/42802102/174753716-023c8620-f4e3-4952-8078-3ab967044b8f.mp4


I think this is because it relies on reading the anchor to determine which section the user is currently in, and it is not able to currently infer that from h4 and h5 headings. It's therefore perhap worth considering not adding these anchor tags for now, and creating a follow-up task where these are added and integrated into the sidebar.

site/src/components/service-message-list.tsx
```diff
@@ -0,0 +1,187 @@
...
+        <MdxLi>
+          <MdxCode>errorCodes</MdxCode>:{" "}
+          <Typography variant="bpSmallCopy" sx={{ fontWeight: 600 }}>
+            {errorCodes && errorCodes.length > 0
+              ? errorCodes.map((code, index) => (
+                  <>
+                    {index > 0 ? ", " : ""}
+                    <MdxCode key={code}>{code}</MdxCode>
+                  </>
+                ))
```
benwerner01, line 150: I could be wrong, but doesn't the key have to be on the fragment here?
```suggestion
              ? errorCodes.map((code, index) => (
                  <React.Fragment key={code}>
                    {index > 0 ? ", " : ""}
                    <MdxCode>{code}</MdxCode>
                  </React.Fragment>
                ))
```

site/src/components/service-message-list.tsx
```diff
@@ -0,0 +1,187 @@
...
+  const isArray = propertySchema.type === "array";
+  const { type: innerType, $ref } = isArray
+    ? (propertySchema.items as JsonSchema)
+    : propertySchema;
+
+  return $ref ? (
+    <MdxA href={$ref} target="_blank">
+      {$ref.match(/[^/]+(?=\/$|$)/)?.[0]}
+      {isArray ? "[]" : ""}
+    </MdxA>
```
benwerner01, line 26: I might be misremembering, but I think our link component opens external URLs in a new tab automatically.
```suggestion
    <MdxA href={$ref}>
      {$ref.match(/[^/]+(?=\/$|$)/)?.[0]}
      {isArray ? "[]" : ""}
    </MdxA>
```

site/src/components/service-message-list.tsx
```diff
@@ -0,0 +1,187 @@
...
+      <>
+        <MdxCode>object</MdxCode>
+        <MdxTable>
+          <thead>
+            <tr>
+              <th>property</th>
+              <th>type</th>
+              <th>required</th>
+              <th>description</th>
+            </tr>
```
benwerner01, line 54: Nit: appreciate the table styling was not introduced in this PR, but might be worth considering giving the table heading a grey background colour, to help distinguish it from the contents a bit more?
image

benwerner01 (approved):
Thanks Ciaran, this PR looks good to me and definitely improves the readability of the API substantially.

Spotted one issue with the sidebar that this PR would introduce, but I'm not sure it should block merging as this PR fixes copy which we may want to get out sooner than the fix.

site/src/util/mdx-components.tsx
benwerner01: Just noticed you allude to this in the description, and appreciate the reason you've introduced the anchor tags is so that the methods can link to each other. Just want to make sure you're aware that this breaks the position indicator anywhere an h4 tag is currently being used.

site/src/components/service-message-list.tsx
CiaranMn: The Link component automatically converts anything containing `blockprotocol.org` into the relative format. I think I'll leave it doing this since these are a special case

benwerner01: Ah good point, I didn't consider that the references start with `blockprotocol.org`

site/src/util/mdx-components.tsx
CiaranMn: I've removed these in https://github.com/blockprotocol/blockprotocol/pull/387/commits/3e8b7528d05f717172be277966b30038d4694135 – I don't think the linking is important enough to figure out the nav right now

site/src/components/service-message-list.tsx
CiaranMn: I've added some grey in https://github.com/blockprotocol/blockprotocol/pull/387/commits/3e8b7528d05f717172be277966b30038d4694135, and also fixed the misplaced `required` which your screenshot reveals – thanks!

benwerner01: Nice, thanks Ciaran 🙂

site/src/util/mdx-components.tsx
benwerner01: Makes sense, this way the sidebar won't seem as buggy until we do have a better implementation of it. Thanks for addressing 👍

PR #391, opened 2022-06-21
Bump all HASH blocks

thehabbos007:
This is a follow up to #390 bumping all existing blocks to make use of the changed webpack config.


CiaranMn (approved):
Excellent! To take one example, the tldraw block is down from 1.5MB to ~250KB

kachkaev (approved):

PR #394, opened 2022-06-22
Fix mock datastore display in hub

teenoh:
Mock datastore doesn't display in block hub because `block-metadata` doesn't include a key that points to `example-graph.json` which contains the mock data. This PR updates block-scripts to include that key and bumps block-scripts version.


CiaranMn (changes requested):
This will cause a crash where blocks do not have an example-graph.json. The key should only have a value if the file exists

teenoh:
> This will cause a crash where blocks do not have an example-graph.json. The key should only have a value if the file exists

Noted... Addressed in [ae8ad32](https://github.com/blockprotocol/blockprotocol/pull/394/commits/ae8ad32f718204489183d4a112ceb3c5d44d6abd)

kachkaev (approved):

PR #396, opened 2022-06-22
Search improvements v2

luisbettencourt:
- Fixed a bug when erasing the search query, the search callback would be called after clearing search results and would cause them to show even when the query was empty.

bugSearch

- Fixed a crash when searching for certain strings (for example `a++` or `***`) - [Asana Task](https://app.asana.com/0/1201095311341924/1202487027725394/f)

- Added animations to the search list:

collapse

- Added arrow navigation between list items

### DIFF
--- a/site/src/components/pages/docs/search/index.tsx
+++ b/site/src/components/pages/docs/search/index.tsx
@@ -54,6 +54,7 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
 }) => {
   const router = useRouter();
   const inputRef = useRef<HTMLInputElement>(null);
+  const searchListItemsRef = useRef<HTMLButtonElement[]>([]);
 
   const [searchText, setSearchText] = useState("");
   const [currentSearchedText, setCurrentSearchedText] = useState("");
@@ -78,6 +79,7 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
             setActiveResult(slicedHits.length);
             setCurrentSearchedText(newSearchText);
             setSearchLoading(false);
+            searchListItemsRef.current = [];
           })
           .catch((err) => {
             // @todo use logger later
@@ -131,10 +133,7 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
   };
 
   useEffect(() => {
-    const items = document.querySelectorAll(
-      ".searchItem",
-      // eslint-disable-next-line no-undef
-    ) as NodeListOf<HTMLButtonElement>;
+    const items = searchListItemsRef.current;
 
     if (activeResult >= items.length) {
       setTimeout(() => inputRef.current?.focus(), 50);
@@ -222,6 +221,7 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
             >
               <SearchList
                 searchResults={searchResults}
+                ref={searchListItemsRef}
                 variant={variant}
                 getHighlight={getHighlight}
                 closeModal={closeModal}
--- a/site/src/components/pages/docs/search/index.tsx
+++ b/site/src/components/pages/docs/search/index.tsx
@@ -54,7 +54,7 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
 }) => {
   const router = useRouter();
   const inputRef = useRef<HTMLInputElement>(null);
-  const searchListItemsRef = useRef<HTMLButtonElement[]>([]);
+  const searchListItemsRefs = useRef<HTMLButtonElement[]>([]);
 
   const [searchText, setSearchText] = useState("");
   const [currentSearchedText, setCurrentSearchedText] = useState("");
@@ -79,7 +79,7 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
             setActiveResult(slicedHits.length);
             setCurrentSearchedText(newSearchText);
             setSearchLoading(false);
-            searchListItemsRef.current = [];
+            searchListItemsRefs.current = [];
           })
           .catch((err) => {
             // @todo use logger later
@@ -133,12 +133,12 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
   };
 
   useEffect(() => {
-    const items = searchListItemsRef.current;
+    const items = searchListItemsRefs.current;
 
     if (activeResult >= items.length) {
       setTimeout(() => inputRef.current?.focus(), 50);
     } else {
-      items?.[activeResult]?.focus();
+      items[activeResult]?.focus();
     }
   }, [activeResult]);
 
@@ -221,7 +221,7 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
             >
               <SearchList
                 searchResults={searchResults}
-                ref={searchListItemsRef}
+                searchListItemsRefs={searchListItemsRefs}
                 variant={variant}
                 getHighlight={getHighlight}
                 closeModal={closeModal}
--- a/site/src/components/pages/docs/search/index.tsx
+++ b/site/src/components/pages/docs/search/index.tsx
@@ -162,7 +162,7 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
           ) {
             event.preventDefault();
             return setActiveResult(
-              activeResult === 0 ? searchResults.length - 1 : activeResult - 1,
+              activeResult === 0 ? searchResults.length : activeResult - 1,
             );
           }
 
--- a/site/src/components/pages/docs/search/index.tsx
+++ b/site/src/components/pages/docs/search/index.tsx
@@ -12,6 +12,7 @@ import React, {
   useRef,
   useState,
 } from "react";
+import { TransitionGroup } from "react-transition-group";
 
 import { Link } from "../../../link";
 import { TextField } from "../../../text-field";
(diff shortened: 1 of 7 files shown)


site/src/components/pages/docs/search/index.tsx
```diff
@@ -64,36 +65,39 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
...
-          .catch((err) => {
-            // @todo use logger later
-            // eslint-disable-next-line no-console
-            console.error(err);
-            setSearchState("failed");
-            setSearchLoading(false);
-          });
-      }, 500),
-    [],
+  const searchOnlineDebounce = useRef(
```
benwerner01, line 68: Just for my understanding, what's the thinking behind a `useMemo` in a `useRef`?

benwerner01: Addressed in https://github.com/blockprotocol/blockprotocol/pull/396/commits/fad381236a1f8b814b57988d5d37006adaac7f53

site/src/components/pages/docs/search/index.tsx
```diff
@@ -143,32 +153,30 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
...
-            }
-
-            if (event.key === "ArrowDown") {
-              event.preventDefault();
-
-              if (activeResult > searchResults.length - 1) {
-                return setActiveResult(0);
-              }
+    <Box
+      onKeyDown={(event: KeyboardEvent<HTMLInputElement>) => {
```
benwerner01, line 157: Nit: arrow down behaves differently than arrow up:

https://user-images.githubusercontent.com/42802102/175256070-ae6857eb-da3d-4fe0-91a0-da2c2121ede5.mov

luisbettencourt: Surprised I didn't catch that, thanks! Addressed in bfc84b2dc36b3d74ea16e1cc7bbb578c707e567f

site/src/components/pages/docs/search/index.tsx
```diff
@@ -127,10 +130,17 @@ export const Search: React.VoidFunctionComponent<SearchProps> = ({
   };
 
   useEffect(() => {
-    if (variant === "desktop") {
+    const items = document.querySelectorAll(
+      ".searchItem",
+      // eslint-disable-next-line no-undef
+    ) as NodeListOf<HTMLButtonElement>;
```
benwerner01, line 136: This isn't a big deal, and there is nothing incorrect about this approach, but I don't think it's best practice to be using query selectors like this in react.

Instead you could use store a list of refs in a `useRef`, which has the benefits of:
- creating an explicit dependency between the child components and this search component by passing the `ref` to the components through their props, rather than relying on them having the `searchItem` class name. If someone accidentally removes `searchItems` class name this would not be caught in CI, whereas if someone were to remove the `ref` passed through props it'd be a type error.
- being more type safe, as the ref would have type `HTMLDivElement | HTMLInputElement` and can only be passed to those elements. That would make the contents of `items` here reliably the elements you were looking for.

site/src/components/pages/docs/search/index.tsx
benwerner01, line 157: I've also spotted a bug where the up button doesn't immediately work after I've entered a search result. Can you reproduce this?

luisbettencourt: Yes, thank you. Addressed in 6c5dcd2298c638618c1b62fb2e49950aee5b1640

site/src/components/pages/docs/search/index.tsx
luisbettencourt: Addressed in e164be2df807ca3043eed8ea39769ac52cd18feb.
I really dislike all the type checking that I've had to add inside the ref callback in `SearchListItem` and `SearchListHeading`, but I still prefer this approach than using `document.querySelectorAll`. If you can think of anyway to make that piece of code cleaner let me know, please.

benwerner01: Thanks for addressing, since you asked how I would clean it up further I've created a small PR off this branch where you can take a look how I would've done it: https://github.com/blockprotocol/blockprotocol/pull/402

If you're happy with it, feel free to approve it and merge it into this PR.

benwerner01: Addressed in https://github.com/blockprotocol/blockprotocol/pull/396/commits/89a3313f4018833368aeda52de4d6183cf96b4fe

site/src/components/pages/docs/search/search-list-category.tsx
```diff
@@ -1,21 +1,31 @@
...
-> = ({ title, icon, searchResults, getHighlight, closeModal }) => {
+const SearchListCategory: FC<SearchListCategoryProps> = ({
+  title,
+  icon,
+  searchResults,
+  getHighlight,
+  closeModal,
+  registerSearchListItemRef,
+}) => {
   const theme = useTheme();
```
benwerner01, line 29: Nit:
```suggestion
```

site/src/components/pages/docs/search/index.tsx
luisbettencourt: Thank you, It's merged

benwerner01 (approved):
Nice work @luisbettencourt , the search modal feels really great to use with these improvements 👍

Have left some minor thoughts/suggestions, but nothing that should block merging.

site/src/components/pages/docs/search/search-list-heading.tsx
```diff
@@ -11,48 +11,49 @@ interface SearchItemHeadingProps {
...
 }
 
-const SearchListHeading: React.VoidFunctionComponent<
-  SearchItemHeadingProps
-> = ({ searchResult, closeModal }) => {
-  const theme = useTheme();
-  const { title, slug } = searchResult;
+const SearchListHeading = forwardRef<HTMLButtonElement, SearchItemHeadingProps>(
+  ({ searchResult, closeModal }, ref) => {
+    const theme = useTheme();
```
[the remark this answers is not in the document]
luisbettencourt: I fell into this trap because on some of our custom components that extend MUI components (`<Button />` for example), using `theme` props inside `sx` doesn't work. I've played with it and think I found solution for it, I'll address this in a new PR.

site/src/components/pages/docs/search/search-list-category.tsx
luisbettencourt: Same as https://github.com/blockprotocol/blockprotocol/pull/396#discussion_r906150285

site/src/components/pages/docs/search/search-list-item.tsx
```diff
@@ -15,82 +15,81 @@ interface SearchItemProps {
...
-const SearchListItem: React.VoidFunctionComponent<SearchItemProps> = ({
-  searchResult,
-  getHighlight,
-  closeModal,
-}) => {
-  const theme = useTheme();
-  const { _highlightResult, slug } = searchResult;
+const SearchListItem = forwardRef<HTMLButtonElement, SearchItemProps>(
+  ({ searchResult, getHighlight, closeModal }, ref) => {
+    const theme = useTheme();
```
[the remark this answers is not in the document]
luisbettencourt: Same as https://github.com/blockprotocol/blockprotocol/pull/396#discussion_r906150285

site/src/components/pages/docs/search/search-list.tsx
```diff
@@ -1,23 +1,27 @@
...
 
-const SearchList: React.VoidFunctionComponent<SearchListProps> = ({
+const SearchList: FC<SearchListProps> = ({
   searchResults,
   variant = "desktop",
   getHighlight,
   closeModal,
+  searchListItemsRefs,
 }) => {
   const theme = useTheme();
```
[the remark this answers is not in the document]
luisbettencourt: Same as https://github.com/blockprotocol/blockprotocol/pull/396#discussion_r906150285

site/src/components/pages/docs/search/search-list.tsx
```diff
@@ -1,23 +1,27 @@
...
   searchResults: AlgoliaResult[];
+  searchListItemsRefs: React.MutableRefObject<HTMLButtonElement[]>;
   variant?: SearchVariants;
   getHighlight: (highlight: AlgoliaHighlightResult) => string;
   closeModal?: () => void;
-}
+};
 
-const SearchList: React.VoidFunctionComponent<SearchListProps> = ({
+const SearchList: FC<SearchListProps> = ({
```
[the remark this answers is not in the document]
luisbettencourt: Right now the maximum number of results is 10. I'm not even sure there's a query that manages 10 hits, but if there is one the modal will stretch outside the window view. Are you suggesting reducing the maximum number of hits, or letting the modal stretch and move the scrollbar outside of the modal?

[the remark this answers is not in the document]
benwerner01: I was suggesting removing the inner scroll-bar that's visible within the screenshot, but am realising I hadn't considered what would happen when the modal exceeds the screen height where a scrollbar may make sense.

Could you set a max-height to the modal that's relative to the screen height and keep the scrollbar, so that it doesn't overflow like this? 

https://user-images.githubusercontent.com/42802102/175566484-c95c8307-352c-4a33-b0af-74cb27ab9a99.mov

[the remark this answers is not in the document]
luisbettencourt: Addressed in a4d90c62407628be111c65ffdb22de5c57895216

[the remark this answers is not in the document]
benwerner01: Perfect, thank you 👍

site/src/components/pages/docs/search/search-list-heading.tsx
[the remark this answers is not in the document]
benwerner01: Okay cool, can you create an asana task for it and link it in this thread? That way we don't lose track of it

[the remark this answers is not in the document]
luisbettencourt: https://app.asana.com/0/1201095311341924/1202505277043423/f

PR #399, opened 2022-06-22
Make HTML blocks work on hub

nathggns:
This makes the updates to the hub/HTML block template to make HTML blocks work in the hub.

Follow ups:

- <s>Seems the initial example props don't work - for the template block as is, the name gets set to undefined until you change it?</s> Fixed by waiting until sandbox iframe has loaded before sending props. Seems there was a race condition there. 
- `@blockprotocol/graph` 0.0.9 needs uploading to npm - cc @CiaranMn  / @kachkaev

### DIFF
--- a/site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
+++ b/site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
@@ -56,6 +56,7 @@ const handler: NextApiHandler = async (req, res) => {
   const { exampleGraph } = await readBlockDataFromDisk(blockMetadata);
 
   const mockBlockDockVersion = packageJson.dependencies["mock-block-dock"];
+  const graphVersion = packageJson.dependencies["@blockprotocol/graph"];
 
   const reactVersion =
     blockMetadata.externals?.react ?? packageJson.dependencies.react;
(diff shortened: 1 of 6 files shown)


site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
```diff
@@ -142,8 +143,11 @@ const handler: NextApiHandler = async (req, res) => {
       
           const entryPoint = blockType.entryPoint.toLocaleLowerCase();
           
-          const rawBlockSource = source;
-          const blockExport = entryPoint === "html" ? rawBlockSource : findBlockExport(loadCjsFromSource(rawBlockSource));
+          let blockExport = findBlockExport(loadCjsFromSource(source));
```
CiaranMn, line 146: I hadn't clocked that the HTML block the template produces is a JS file that exports an HTML string, which is at odds with [how the spec describes HTML blocks as having an HTML entry point](https://blockprotocol.org/docs/spec/core-specification#types-of-block) – I think there are two options:

1. Have HTML blocks be either an HTML file OR a JS file that exports HTML (not crazy about this as it adds more branching, depends on what you think about 2)

2. Update the block template so that what is actually produced is an HTML file with the result of the current `index.js`, not `index.js` itself. 

What do you reckon?

site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
```diff
@@ -142,8 +143,11 @@ const handler: NextApiHandler = async (req, res) => {
...
+          // @todo make this unnecessary
```
CiaranMn, line 147: Do you have thoughts about how best to achieve this?

A couple:
1. Make import in the template the thing you've written here
2. List `@blockprotocol/graph` as a peerDependency, in which case it'll become listed in `metadata.externals` and loaded like the others above

nathggns:
Closing this as a more sophisticated version changing how HTML blocks works is coming

PR #402, opened 2022-06-24
Suggested change for lb/search-improvements-v2

benwerner01:
Refactors the assignment of search list item refs introduced in https://github.com/blockprotocol/blockprotocol/pull/396

### DIFF
--- a/site/src/components/pages/docs/search/search-list-item.tsx
+++ b/site/src/components/pages/docs/search/search-list-item.tsx
@@ -15,8 +15,8 @@ interface SearchItemProps {
   closeModal?: () => void;
 }
 
-const SearchListItem = forwardRef<HTMLButtonElement[], SearchItemProps>(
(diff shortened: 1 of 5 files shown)


site/src/components/pages/docs/search/search-list-item.tsx
```diff
@@ -25,16 +25,7 @@ const SearchListItem = forwardRef<HTMLButtonElement[], SearchItemProps>(
...
-          if (
-            item &&
-            typeof searchListItemsRef !== "function" &&
-            searchListItemsRef?.current &&
-            !searchListItemsRef?.current?.includes(item)
-          ) {
-            searchListItemsRef.current.push(item);
-          }
-        }}
+        ref={ref}
```
benwerner01, line 28: If the `SearchListItem` declares a singular ref with the corresponding type, it can be passed directly here

luisbettencourt:
Way better approach, thank you @benwerner01 !

PR #403, opened 2022-06-24
FAQ page collapsible questions

benwerner01:


CiaranMn:
Since we will now be marking FAQ Q&A explicitly in the markdown, a follow-up task could be: when rendering a markdown page, check if it contains any `<FAQAccordion>`. If it does, include [the necessary JSON-LD](https://developers.google.com/search/docs/advanced/structured-data/faqpage#examples) in the page to allow for rich search results.

I also wonder whether we should just call these things `<FAQ>` rather than `<FAQAccordion>` – if we ever wanted to have a FAQ component that wasn't collapsable, we could make `accordion` a prop you pass to `<FAQ>`.

benwerner01:
> Since we will now be marking FAQ Q&A explicitly in the markdown, a follow-up task could be: when rendering a markdown page, check if it contains any <FAQAccordion>. If it does, include [the necessary JSON-LD](https://developers.google.com/search/docs/advanced/structured-data/faqpage#examples) in the page to allow for rich search results.

@CiaranMn agreed we should generate the JSON-LD for this page. A think another SEO improvement would be to expose our sitemap at an `/sitemap.xml` endpoint, so we can [submit it to google](https://developers.google.com/search/docs/advanced/sitemaps/build-sitemap) (currently I don't think we've submitted anything to them)

nonparibus:
Feedback provided in private. Will provide formal review when those changes are made and this PR is marked as ready for review.

nonparibus (dismissed):

nonparibus (approved):

PR #406, opened 2022-06-27
Allow usage of theme in custom component's sx prop

luisbettencourt:
When using the `(theme) => { ... }` format as a sx prop in custom components that extend MUI components (`<Button />` for example), the styles weren't applied. This was happening because the `sx` prop can be an object, a function, or a boolean and it was being passed as if it was always an object.

The correct approach to pass the `sx` prop is described [here](https://mui.com/system/the-sx-prop/#passing-sx-prop) (thanks @teenoh for the reference).

[Asana Task](https://app.asana.com/0/1201095311341924/1202505277043423/f)


luisbettencourt:
Github's diffchecker is trying to make this PR bigger than it really should be

CiaranMn:
> Github's diffchecker is trying to make this PR bigger than it really should be

You can hide whitespace changes (e.g. the indent changes this introduces) here:
Screenshot 2022-06-27 at 19 04 15

teenoh (approved):
Looks good to me. Thanks @luisbettencourt

PR #407, opened 2022-06-28
Add Github PR Overview block to hub

teenoh:


kachkaev (approved):
There is some glitch with scroll bars, but I don’t think we can fix it as part of this PR.

Screenshot 2022-06-30 at 19 55 02

Might be also worth looking into the icon sizing. The aspect ratio gets broken because we set both `height` and `width` to `2em`.

PR #354, opened 2022-06-14
RFC - Structural Link Constraints (Graph)

Alfred-Mountfield:
This PR is WIP

[Rendered view](https://github.com/blockprotocol/blockprotocol/blob/am/rfc-graph-structural-link-constraints/rfcs/text/0354-graph-structural-link-constraints.md)

PR #408, opened 2022-06-28
RFC - Versioning for Graph Types

Alfred-Mountfield:
This RFC proposes a versioning scheme for the types introduced in #352.

As part of that, it 
- Specifies a URI format for how they should be referenced
- Establishes how we can determine compatibility between Types under the current constraints (this will be affected by #354 and #355)

--- 
[Rendered View](https://github.com/blockprotocol/blockprotocol/blob/am/rfc-versioning-types/rfcs/text/0408-versioning-types.md)

---
  - https://app.asana.com/0/0/1202510174412956


rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+We opt not to constrain the format of the `base_uri` to enable flexibility in non-public use-cases or in domains with constraints that we cannot predict. Despite this, we _suggest_ that the base URIs should have a consistent format for all types hosted on your domain, and should generally be humanly readable in a way that describes your type. For example, one may pick:
+
+```json
+"http://yourdomain/namespace/property-type/type_name"
+```
+
+as the identifier (assuming that the domain decides that type names are unique across a namespace).
+
+Where it's possible to guarantee, this URI should be valid indefinitely.
```
benwerner01, line 50: For my understanding, would this recommendation mean that if a user were to create a `Person` entity type, delete it, and create another `Person` type down the line, that they will have unintentionally prevented themselves from being able to re-use the `http://yourdomain/namespace/entity-type/Person` URI ever again? Or would the version counter not restart from `0`, and the re-created entity type be considered "newer versions" of the prior one? Or is this up to the implementation to decide

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+1.  The `"$id"` field of a Type should be equal to the Type's _versioned URI_.
+1.  When referencing another Type using `"$ref"`, the reference should be equal to the Type's _versioned URI_.
+1.  When used as the key of a property, the _base URI_ should be used.
+
+Example:
+
+```json
+{
+  "kind": "entityType",
+  "$id": "https://blockprotocol.org/@alice/types/entity-type/book",
```
benwerner01, line 80: Possibly I've misunderstood, but shouldn't this URI be versioned given the prior explanation?
```suggestion
  "$id": "https://blockprotocol.org/@alice/types/entity-type/book/v/1",
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+}
+```
+
+## Determining Type Compatibility
+
+A key part of the reasoning for picking this approach depends on the ability to determine compatibility between various types.
+
+For the purposes of this section, the following assumptions are applied:
+
+- Any given Schema X is _compatible with_ another schema Y if all possible values that satisfy Schema X also satisfy schema Y (note the directionality)
```
benwerner01, line 112: Nit: wouldn't this be slightly more consistent with the other bullet?
```suggestion
- Any given Schema X is _compatible with_ another schema Y if and only if all possible values that satisfy Schema X also satisfy schema Y (note the directionality)
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+- The ordering of constraints does not affect the compatibility of schemas
+
+### Data Types
+
+At the present Data Types represent completely disjoint value spaces, and the base primitive types are all **incompatible** with one another. As the [Non-Primitive Data Types RFC](https://github.com/blockprotocol/blockprotocol/pull/355) continues to be specified, it should comment on how checking compatibility will be affected.
+
+### Property Types
+
+1.  Due to how Property Types are used, they implicitly affect the structure of the data in that they require the existence of a key that matches their URI.
+1.  Property Types define their constraints through a `oneOf` field. This `oneOf` is a collection of a number of values (greater than 1) where each value is a variant of the "Property Values" (defined below).
```
benwerner01, line 126: Nit: find this slightly easier to read
```suggestion
1.  Property Types define their constraints through a `oneOf` field. This `oneOf` is a collection of one or more values, where each value is a variant of the "Property Values" (defined below).
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+A Property Type `A` is therefore compatible with another Property Type `B` if and only if
+
+1.  They have the same base URI (which implies they are the same property type but perhaps different versions) and,
+1.  Every element of `A`'s `oneOf` is compatible with one element of `B`'s `oneOf`
```
benwerner01, line 131: Nit:
```suggestion
A Property Type `A` is therefore compatible with another Property Type `B` if and only if:

1.  they have the same base URI (which implies they are the same Property Type but perhaps different versions), and
1.  every element of `A`'s `oneOf` is compatible with one element of `B`'s `oneOf`
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+#### Data Type Reference
+
+A Data Type reference is an inline reference to a data-type. A Data Type reference will be compatible with another Data Type reference if the Data Types that are referred to are also compatible.
+
+> _Without_ non-primitive Data Types this implies that the references are to the same Data Type.
+
+#### Property Type Object
+
+A Property Type Object is a JSON object where
```
CiaranMn, line 143: Narrowing it for the avoidance of doubt – many JSON objects which we don't intend to refer to meet the existing definition
```suggestion
A Property Type Object is JSON object defining a schema which is used as `oneOf` the permitted values for a Property Type, where the schema is of `type: "object`, and within its own `properties` definition:
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+1.  The keys are base URI's to a property type
+1.  The values are defined by either:
+    1.  A reference to a property type
+    1.  An array definition, where
+        1.  the items are defined by a reference to a property type
```
CiaranMn, line 149:
```suggestion
        1.  the `items` are defined by a reference to a property type
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+- A block could mitigate this by "vendoring in" a specific instance of the type and then providing it alongside its source (note: this proposal doesn't actually make suggestions for or against this), but if that iteration of the type is not persisted, identifiable (by some form of version number..), and addressable, then this greatly reduces the reusability of types and would make it easy to lose a specific iteration.
+- This also could cause a lot of problems for the composability of types. Entity types could have their definitions implicitly affected by a property or data type being modified, etc.
+
+## Semantic Versioning
+
+This section is going to largely refer to _semantic versioning_ as defined by [semver.org](https://semver.org/), although it should be noted that it doesn't perfectly map to the ways types operate.
+
+### Effects of versioning on Blocks and Embedding Applications
+
+The Types sit in the middle of producers and consumers of data, and unfortunately various types of changes has differing effects on the compatibility for either side.
```
benwerner01, line 245:
```suggestion
The Types sit in the middle of producers and consumers of data, and unfortunately various types of changes have differing effects on the compatibility for either side.
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+- [Semantic Versioning](https://semver.org)
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- How should we constrain valid URI formats
+  - How should the version pair be formatted? Should we use query parameters?
+  - Should it be possible to go to a URI _without_ expressing a version, if so should returning the latest version be the expected behavior
```
benwerner01, line 337: Potentially this is something we don't need to specify as part of the RFC and leave up to the implementer, but I think we should do this yes (i.e. return the latest version of the type when no version is specified)

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+1.  Each key in the `links` of `A` is also in the `links` of `B`
+1.  If a link in `links` of `A` is defined as an array, then it's also defined as an array in `B` (and therefore the same for direct references)
+1.  If a link in `links` of `A` is defined as an array with a `minItems`, then there is a `minItems` constraint on the respective link in `B` that is equal to or larger to that in `A`
+1.  If a link in `links` of `A` is defined as an array with a `maxItems`, then there is a `maxItems` constraint on the respective link in `B` that is less than or equal to that in `B`
+1.  If a link in `links` of `A` is defined as an array with `ordered` as `true`, then there is an `ordered` as `true` on the respective link in `B`
+1.  Each URI in the `requiredLinks` list of `A` is also in the `requiredLinks` list of `B`
+
+### Link Types
+
+Link Types are only able to be changed through _semantic annotations_ and as such, any Link Type `X`, with the same Base URI as another Link Type `Y` (i.e. a different version of the same link type), are _compatible_.
```
CiaranMn, line 209:
```suggestion
Link Types are only able to be changed through _semantic annotations_ and as such, any Link Type `X`, with the same Base URI as another Link Type `Y` (i.e. a different version of the same link type), are _compatible_. This is likely to change in future.
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+## When to Check Compatibility of Types
+
+Block schemas, and other uses of types (for example graph service methods), can request data from the embedding application by referencing types. Not relying on the version to ensure compatibility means that if a block, or data (associated with a versioned type) in an embedding application, is "outdated", the compatibility can be checked by comparing the two schemas. Data blobs can continue to be validated against the relevant schema.
```
CiaranMn, line 213:
```suggestion
Block schemas, and other uses of types (for example graph service methods), can request particular data structures from the embedding application by referencing types. Not relying on the version of a type to ensure compatibility means that if a block or data (associated with a versioned type) in an embedding application is associated with a different version of a type – or different types altogether  – the compatibility of the data can still be checked by comparing the two schemas. Data blobs can continue to be validated against the relevant schema.
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+This should allow a bit more flexibility within the system when it comes to compatibility of data requirements.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+- This proposal makes the active decision to not embed compatibility information within the version identifier. This contrasts with approaches like semantic versioning, which is explored more in the [Rationale and alternatives](#rationale-and-alternatives) section.
+- The versioning approach does not encapsulate information about the scale of the change, fixing a typo causes a new version in the same way that completely rewriting the constraints does.
```
CiaranMn, line 222:
```suggestion
- The versioning approach does not encapsulate information about the scale of the change: fixing a typo causes a new version in the same way that completely rewriting the constraints does (assuming changes are not batched before a new version URI is created).
```

rfcs/text/0408-versioning-types.md
CiaranMn, line 245:
```suggestion
The Types sit in the middle of producers and consumers of data, and unfortunately various types of changes have differing effects on the compatibility for either side.
```

CiaranMn (commented):
Good stuff! some comments, mostly suggestions for clarification / phrasing

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: Under the guidance of this proposal, the expectation is that publicly accessible URIs are permanent and immutable, that is, they cannot be deleted.

Now if people are operating within a closed sphere of their domain and wish to have things that aren't really _publicly accessible_ then we can't control how they want to use these.

In those circumstances I think they could decide themselves if the URI needs to be reserved in the future or not.

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: You have not misunderstood, I have made a mistake 😁

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+```
+
+## Determining Type Compatibility
+
+A key part of the reasoning for picking this approach depends on the ability to determine compatibility between various types.
+
+For the purposes of this section, the following assumptions are applied:
+
+- Any given Schema X is _compatible with_ another schema Y if all possible values that satisfy Schema X also satisfy schema Y (note the directionality)
+- Any given Schema X is _equivalent to_ another Schema Y iff X is _compatible with_ Y _and_ Y is _compatible with_ X
```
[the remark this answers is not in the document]
Alfred-Mountfield: Was a bit too stuck in maths land, thank you

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: Had to think a little about if one implied the other but I think you are correct

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: I agree, could have been misinterpreted as the values being greater than 1. Thank you

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: I have gone around and made the capitalisation of all of the types consistent. Thank you.

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+1.  Due to how Property Types are used, they implicitly affect the structure of the data in that they require the existence of a key that matches their URI.
+1.  Property Types define their constraints through a `oneOf` field. This `oneOf` is a collection of a number of values (greater than 1) where each value is a variant of the "Property Values" (defined below).
+
+A Property Type `A` is therefore compatible with another Property Type `B` if and only if
+
+1.  They have the same base URI (which implies they are the same property type but perhaps different versions) and,
+1.  Every element of `A`'s `oneOf` is compatible with one element of `B`'s `oneOf`
+
+The "Property Values" is a recursive definition which refers to the sub-schemas defined within the Property Types meta-schema. It defines the element as being one of the following:
```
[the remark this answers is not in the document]
Alfred-Mountfield: Good suggestion, thanks! Kind of forgot we have a perma-link now.

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
Alfred-Mountfield: It's a little obscure, but not inside a Property Type object.

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: I agree that I think we should. I also think we _probably_ do want to include it in the RFC so that a Type Nexus on blockprotocol.org could rely on the presence of that path.

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+The Type System outlined in [RFC 0352](https://github.com/blockprotocol/blockprotocol/blob/am/rfc-graph-type-system/rfcs/text/0352-graph-type-system.md) introduced classes of types that are used to describe the data used and produced by blocks and embedding applications. The types are possibly (ideally often) publicly accessible, and their hierarchy allows for them to be composed together into new types. Due to this, changes in a type can affect the structure of another one. This RFC defines how types are versioned, to mitigate potential problems that would arise without versioning.
+
+# Motivation
+
+[motivation]: #motivation
+
+The Type System is intended to allow for **reusable and shared** descriptions of data. As such, a version of a block should be able to tie itself to a _specific_ description of the structure of the data it requires as if types are able to be updated independently of blocks then this would cause problems. Additionally, if someone (Alice) creates an entity type that refers to someone else's (Bob's) "address" property type, we do not want Bob to be able to change what Alice meant in her entity type by arbitrarily and implicitly updating it.
```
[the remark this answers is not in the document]
Alfred-Mountfield: Sorry yeah I missed another clause, I was trying to say if Blocks _can't_ pin themselves to specific versions _and_ types can be updated independently (which we want), then we encounter problems.

I like your wording, thank you. Although I'm going to omit "as one of its properties" because it's sort of redundant and I _try_ and avoid "property" on its own when talking about Types because we enter zones of ambiguity of if we're talking about JSON stuff properties, property _types_, properties in JSON objects, keys or values, etc.

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
Alfred-Mountfield: I think this is somewhat redundant as I mention these should be treated as immutable. Perhaps "and refer to the same iteration of the type" would be worth clarifying though?

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+## URI Format and Version Numbers
+
+Type URIs, and their version numbers are defined as outlined in the [Guide-level explanation](#guide-level-explanation).
+
+### Schemas
+
+1.  The `"$id"` field of a Type should be equal to the Type's _versioned URI_.
+1.  When referencing another Type using `"$ref"`, the reference should be equal to the Type's _versioned URI_.
+1.  When used as the key of a property, the _base URI_ should be used.
```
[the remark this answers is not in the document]
Alfred-Mountfield: This was a nice unintended side-effect I noticed yesterday about unversioned property names, which is something we had overlooked before.

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+{
+  "https://blockprotocol.org/@alice/types/property-type/name": "The Time Machine",
+  "https://blockprotocol.org/@alice/types/property-type/published-on": "1895-05",
+  "https://blockprotocol.org/@alice/types/property-type/blurb": ...
+}
+```
+
+## Determining Type Compatibility
+
+A key part of the reasoning for picking this approach depends on the ability to determine compatibility between various types.
```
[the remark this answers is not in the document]
Alfred-Mountfield: Actually don't think I agree with this one, we can determine compatibility in this fashion regardless of the approach of versioning. The reasoning of not choosing semver somewhat depends on us being able to do things more dynamically because we can (mostly) determine compatibility between types.

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+For the purposes of this section, the following assumptions are applied:
+
+- Any given Schema X is _compatible with_ another schema Y if all possible values that satisfy Schema X also satisfy schema Y (note the directionality)
+- Any given Schema X is _equivalent to_ another Schema Y iff X is _compatible with_ Y _and_ Y is _compatible with_ X
+- All Type schemas have `additionalProperties: false` unless otherwise stated
```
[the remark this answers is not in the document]
Alfred-Mountfield: Is `additionalProperties: true` required for inheritance? I am not sure it is, we might need to play around with some validators, but I think if you do an `allOf` then the properties inherited by that aren't considered 'additional'

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+A Property Type `A` is therefore compatible with another Property Type `B` if and only if
+
+1.  They have the same base URI (which implies they are the same property type but perhaps different versions) and,
+1.  Every element of `A`'s `oneOf` is compatible with one element of `B`'s `oneOf`
+
+The "Property Values" is a recursive definition which refers to the sub-schemas defined within the Property Types meta-schema. It defines the element as being one of the following:
+
+#### Data Type Reference
+
+A Data Type reference is an inline reference to a data-type. A Data Type reference will be compatible with another Data Type reference if the Data Types that are referred to are also compatible.
```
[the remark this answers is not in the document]
Alfred-Mountfield: It's a good call out, I'm not sure how to phrase it more clearly. A Property type is referenced inside a `properties` blob with an associated key. It's necessarily _inside_ some structure inside the schema:

```json
{
  "kind": "entityType",
  "properties": {
    "http://.../property-type/foo": { "ref": "http://.../property-type/foo/v/1" }
  }
  ...
}
```

Whereas data-types are "$ref" inline so they appear at the top-level of the schema:

```json
{
  "kind": "dataType",
  "$ref": "http://.../data-type/bar/v/1",
  ...
}
```

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
Alfred-Mountfield: Ben suggested this too, thank you. I let my maths education overtake me on this 😁

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+
+```json
+"http://yourdomain/namespace/property-type/type_name"
+```
+
+as the identifier (assuming that the domain decides that type names are unique across a namespace).
+
+Where it's possible to guarantee, this URI should be valid indefinitely.
+
+> ⚠️ New iterations and changes to the type must occur under new version numbers, components of the Block Protocol ecosystem will certainly implement caching and local-persistence logic, and the spec consciously makes no recommendation that these caches are updated.
```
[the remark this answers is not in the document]
Alfred-Mountfield: I have attempted to reword

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: Much better, thank you

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: Have added, although I kept the commas as I think they are correct here?

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: I don't think this makes it clearer, I actually think it's more confusing.

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+The _versioned URI_ of a type is made up of:
+
+```json
+"${base_uri}/v/${version_number}"
+```
+
+We opt not to constrain the format of the `base_uri` to enable flexibility in non-public use-cases or in domains with constraints that we cannot predict. Despite this, we _suggest_ that the base URIs should have a consistent format for all types hosted on your domain, and should generally be humanly readable in a way that describes your type. For example, one may pick:
+
+```json
+"http://yourdomain/namespace/property-type/type_name"
```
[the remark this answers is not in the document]
Alfred-Mountfield: I've tried to make it more obviously an example template and have updated the examples to `example.com` as we haven't decided on a format for blockprotocol.org yet (I think we should do this as part of the discussion of the RFC if it gets merged)

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+The _version number_ of a type is an incrementing positive integer that increases by 1 with each new version, starting from 1.
+
+## Type URIs
+
+The _base URI_ of a type is a _unique_ identifier for the type irrespective of its version.
+
+The _versioned URI_ of a type is made up of:
+
+```json
+"${base_uri}/v/${version_number}"
```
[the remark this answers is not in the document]
Alfred-Mountfield: I also am torn between the options, I was discussing it with @thehabbos007 and we couldn't decide easily. I had a number of reasons, not that I found them particularly convincing
- For self-hosting types, it's a lot easier in most implementations to just have it as part of the route, rather than needing to understand how to interpret query parameters and route more dynamically based on that
- One less special-ish character
- NPM uses `/v/`, I hope that has reasoning behind it that we can benefit from
- I think `/v/` will play nicer with file-systems if people really want to do so

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,344 @@
...
+1.  Entity Types define their constraints on the structure of an entity through their `properties`, where
+    1.  The keys are base URI's to a property type
+    1.  The values are defined by either:
+        1.  A reference to a property type
+        1.  An array definition, where
+            1.  the items are defined by a reference to a property type
+            1.  `minItems` and/or `maxItems` are optionally defined
+1.  Entity Types also define additional constraints on links from their entity through their:
+    1.  `links`, where
+        1.  The keys are _versioned URI's_ to a link type
```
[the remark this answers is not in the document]
Alfred-Mountfield: Have done so

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,355 @@
...
+1.  The values are defined by either:
+    1.  A reference to a property type
+    1.  An array definition, where
+        1.  the `items` are defined by a reference to a property type
+        1.  `minItems` and/or `maxItems` are optionally defined
+1.  The required fields are defined by a `required` list where the elements are base URIs that are a subset of the keys
+
+A property type Object `X` is therefore compatible with another property type Object `Y` if and only if
+
+1.  Each key in the `properties` of `X` is also in the `properties` of `Y`
```
thehabbos007, line 157: Is this the case for properties which are not required?
It seems to me that if the `properties` for `X` referred to here are not present in the `required` list, they do not need to be present in `Y` for compatibility. 

Or at least we need the inverse relationship as well to make it clear that all optional values need to be present on both sides

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,355 @@
...
+
+A data type reference is an inline reference to a data-type. A data type reference will be compatible with another data type reference if the data types that are referred to are also compatible.
+
+> _Without_ non-primitive data types this implies that the references are to the same data type.
+
+#### Property Type Object
+
+A poperty type object is JSON object defining a schema which is used as `oneOf` the permitted values for a property type, where the schema is of `type: "object`, and within its own `properties` definition:
+
+1.  The keys are base URI's to a property type
```
thehabbos007, line 147:
```suggestion
1.  The keys are base URIs to a property type
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,355 @@
...
+
+A definition of an array of "Property Values" `X` is compatible with another definition of an array of "Property Values" `Y`, if and only if
+
+1.  Every element of `X`'s `oneOf` is compatible with at least one element of `Y`'s `oneOf`
+1.  If `X` has a `minItems` constraint specified, then there is a `minItems` constraint in `Y` that is equal to or larger to that in `X`
+1.  If `X` has a `maxItems` constraint specified, then there is a `maxItems` constraint in `Y` that is less than or equal to that in `X`
+
+### Entity Types
+
+1.  Entity types define their constraints on the structure of an entity through their `properties`, where
```
thehabbos007, line 176: `required` (for properties) is missing in this list

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,355 @@
...
+            1.  An empty object (until Link Constraints are implemented)
+            1.  An array definition, where
+                1.  the `items` are an empty object (until Link Constraints are implemented)
+                1.  `ordered` is optionally defined
+                1.  `minItems` and/or `maxItems` are optionally defined
+    1.  `requiredLinks` array, where the elements are _versioned URIs_, and is a subset of the keys in the `links`
+
+An entity type `A` is therefore compatible with another entity type `B` if and only if:
+
+1.  Each key in the `properties` of `A` is also in the `properties` of `B`
```
thehabbos007, line 196: Sort of the same situation here as with the property type section. Should we consider optional values on `A`'s side incompatible with `B`s that don't contain said optional values?

thehabbos007 (commented):
Great job! I have a couple questions here.

Also it seems that some of these concepts could be used for figuring out "constraint schemas" checking, which is nice.

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
benwerner01: Thanks, I think that makes sense (for reference here's the commit https://github.com/blockprotocol/blockprotocol/pull/408/commits/1655ae8c8ab2923f11f6d10499850ad442817ad9)

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
CiaranMn: Yes, it does - I'd read [this](http://json-schema.org/understanding-json-schema/reference/object.html#extending-closed-schemas), including the `unevaluatedProperties` section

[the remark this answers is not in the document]
CiaranMn: For extension via `allOf` that does not allow properties not in the schema or schemas referenced via `allOf`, we need `unevaluatedProperties: false, additionalProperties: true`

rfcs/text/0408-versioning-types.md
benwerner01: Thanks - just want to flag it wasn't just capitalization fixes I suggested, I think the comma placement is incorrect (should be `, and` instead of `and, `). But it's just a nit pick.

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,355 @@
...
+At the present data types represent completely disjoint value spaces, and the base primitive types are all **incompatible** with one another. As the [Non-Primitive Data Types RFC](https://github.com/blockprotocol/blockprotocol/pull/355) continues to be specified, it should comment on how checking compatibility will be affected.
+
+### Property Types
+
+1.  Due to how property types are used, they implicitly affect the structure of the data in that they require the existence of a key that matches their base URI.
+1.  Property types define their constraints through a `oneOf` field. This `oneOf` is a collection of one or more values where each value is a variant of the "Property Values" (defined below).
+
+A property type `A` is therefore compatible with another property type `B` if and only if
+
+1.  they have the same base URI (which implies they are the same property type but perhaps different versions) and,
```
Alfred-Mountfield, line 132:
```suggestion
1.  they have the same base URI (which implies they are the same property type but perhaps different versions), and
```

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: Oops missed that one, thanks for the call-out!

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
Alfred-Mountfield: cc: @thehabbos007 and @TimDiekmann  I was wrong about this when we spoke the other day. We might need to re-evaluate some of our current meta-schemas.

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: It is the case. Because if a property in `X` is not required, it is still possible.

Lets say we have a schema `X` with some properties and an optional property `Foo`. Now let's say another schema `Y` has all the properties, but does not have `Foo`.

{
  "foo": ...,
  ...
}

validates against `X` but does not validate against `Y`.

rfcs/text/0408-versioning-types.md
Alfred-Mountfield, line 132:
```suggestion
1.  they have the same base URI (which implies they are the same property type but perhaps different versions), and
```

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: Same response as above :)

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: Added, good spot

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: Fixed across doc

rfcs/text/0408-versioning-types.md
thehabbos007: Fair enough, just wanted to make sure :)

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
CiaranMn: Sure, sounds good

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
CiaranMn: Looks good

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
CiaranMn: Fair enough

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
CiaranMn: Still not following this, might be worth a sync discussion - surely a data type is referred to from inside a propertyType schema?

```json
{
  "kind": "propertyType",
  "$id": "http://.../property-type/foo/v/1",
  "oneOf: [ {   "$ref": "http://.../data-type/bar/v/1" } ]
}
```
where the data type looks like this:
```
{
  "kind": "dataType",
  "$id": "http://.../data-type/bar/v/1",
  "type": "string"
}

File: rfcs/text/0408-versioning-types.md (hunk 22 above)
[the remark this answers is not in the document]
Comment (CiaranMn) [reply rfcs/text/0408-versioning-types.md:39]: I don't have anything more to add on this, really. Maybe see if you can drum up any other opinions as part of implementation

File: rfcs/text/0408-versioning-types.md (hunk 11 above)
Comment (CiaranMn) [reply rfcs/text/0408-versioning-types.md]: No harm keeping the commas, just thought they slowed it down a bit

File: rfcs/text/0408-versioning-types.md (hunk 12 above)
Comment (CiaranMn) [reply rfcs/text/0408-versioning-types.md:225]: The point I'm trying to make is that what 'causes a new version' is entirely up to whoever is managing the creation of these types. This might be read as implying that fixing a typo _must_ cause a new version.

File: rfcs/text/0408-versioning-types.md (hunk 1 above)
Comment (Alfred-Mountfield) [reply rfcs/text/0408-versioning-types.md:50]: I think that's something we do need to discuss but I'm not sure this PR is the right place for it (open to being corrected on that if we really do need to provision for this in the spec).

I think we can have these in the Github Discussions when this is merged, while we figure out our URI format for blockprotocol.org

File: rfcs/text/0408-versioning-types.md (hunk 12 above)
Comment (Alfred-Mountfield) [reply rfcs/text/0408-versioning-types.md:225]: Right but fixing a typo does mean you _must_ have a new version. Whether or not you include other changes within that doesn't change it.

I explain this explicitly, including something similar to your recommendation, above

> ## What is a "version" of a Type
>
> Any published (accessible) iteration of a Type can be considered a "version" of the Type. Once a specific iteration of a type is made accessible via a URI, it should be considered immutable.
> 
> - This means that any change made to the Type (for example changing the description, adding a field, etc) constitutes reason for a new version.
> - **If multiple changes are made prior to making it accessible (i.e. batching the changes), then a separate version is not needed for each individual change, but they can all be 'released' together.**

File: rfcs/text/0408-versioning-types.md (hunk 1 above)
Comment (CiaranMn) [reply rfcs/text/0408-versioning-types.md:50]: This doesn't block merging

File: rfcs/text/0408-versioning-types.md (hunk 12 above)
Comment (CiaranMn) [reply rfcs/text/0408-versioning-types.md:225]: Right - my recommendation was a callback to that, to reinforce the point that you can fix a typo and then leave that draft version somewhere without creating it. 

It's not important - resolving this

File: rfcs/text/0408-versioning-types.md (hunk 1 above)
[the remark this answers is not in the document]
Comment (Alfred-Mountfield) [reply rfcs/text/0408-versioning-types.md:50]: Sorry I should clarify the _versioned_ URI should refer to the same type indefinitely. I agree about "without version qualifier" or some "/latest" etc.

File: rfcs/text/0408-versioning-types.md (hunk 19 above)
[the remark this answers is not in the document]
Comment (Alfred-Mountfield) [reply rfcs/text/0408-versioning-types.md]: Yes, so the example you gave has the data-type's constraints being applied at the same level as `"kind": "propertyType"` (because "oneOf" basically collapses it in)

If you take this entityType:

```json
{
  "kind": "entityType",
  "$id": "http://.../entity-type/baz/v/1",
  "type": "object",
  "properties": {
    "http://.../property-type/foo/v/1": { "$ref": "http://.../property-type/foo/v/1" }
  }
}
```

Then "http://.../property-type/foo/v/1" does _not_ apply at the top level of the entityType object, its constraints apply to the "http://.../property-type/foo/v/1" _property_ within that object.

Thus the _constraints_ on a property type being the same as the constraints on another property type isn't enough to confirm compatibility because there's an implicit effect on the data (which we have added) which says the property name has to be equal to the `base_uri`.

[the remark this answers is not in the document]
CiaranMn: Right, understood – but the `$ref` isn't inline in my head – my confusion is thinking that there's no scenario in which this earlier example appears:
```json
{
  "kind": "dataType",
  "$ref": "http://.../data-type/bar/v/1",
  ...
}
```
because if we take the property type:
```json
{
  "kind": "propertyType",
  "$id": "http://.../property-type/foo/v/1",
  "oneOf": [ { "$ref": "http://.../data-type/bar/v/1" } ]
}
```
if we evaluate it, isn't the `$ref` is replaced by the schema it refers to, so you never have the `$ref` appear alongside `kind: "dataType"`? I may be misunderstanding how evaluation works and it's (the `$ref`) left lying around.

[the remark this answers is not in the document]
CiaranMn: I don't think this is important to merging the RFC, but might be revealing some misunderstanding I have that's worth clearing up for later.

rfcs/text/0408-versioning-types.md
benwerner01: It may be worth considering allowing for version IDs to be UUIDs, in which case a rule such as "all _versioned_ URIs must be valid indefinitely" will be a lot easier for us to adhere to in a world where users can acquire previously used namespaces, or re-use previous _non-versioned_ URIs.

benwerner01: My main concern was that users will end up with `http://.../Person-3` in their namespace after deleting the (public) entity types `http://.../Person` and `http://.../Person-2`, and there's not a lot we can do about it in the outlined version system. But I'm realising if we don't support deleting public types anyway, then this won't really be an issue as the user will be forced to create a new version of the `Person` type if they wish to change it.

CiaranMn: > It may be worth considering allowing for version IDs to be UUIDs, in which case a rule such as "all versioned URIs must be valid indefinitely" will be a lot easier for us to adhere to in a world where users can acquire previously used namespaces, or re-use previous non-versioned URIs.

I don't think we should maintain URLs with namespaces pointing to a resource created by a user, when a different user now owns that namespace (and has other resources at URLs using it). If we want URIs to be valid indefinitely, I think once someone has used a namespace it can't be used by anyone else.

Alfred-Mountfield: I also see the value in UUIDs but I don't think it outweighs the value of the current approach which @nonparibus identified, where we can programmatically discover later (and earlier) versions by simply incrementing the counter.

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
nonparibus: We should keep in mind that people may want more variations of types than simply the versions. For example, folks might want to use query params to request/serve a localized translation _of_ a specific version of a type for their own i18n.

If we use `/v/` paths for versions we should be clear why, and still consider how we would support layering of additional parameters as well.

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
Alfred-Mountfield: I don't think it's confusion on your end, I think it's me misusing words because I'm lacking a better way of describing it.

- The `$ref` _is_ replaced
- The `oneOf` means it's not 'inline'

I'm just trying to say the constraints within the data-type end up getting applied (if there's only one option in the `oneOf` I guess) as if they had been written inline the data-type object at the top level. That doesn't apply to property types, where there is some more structure implicitly captured that we need to think about for compatibility.

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
Alfred-Mountfield: That's a pretty good argument (in my head) for not talking about query params at the moment at all, and using `/v/` to keep it simple. We _could_ do the i18n thing later on as you say, and at that point query params would be a good fit. I'd probably prefer if we never went down that road because this isn't just about UX but data structure and I think trying to translate keys in actual JSON data is perhaps not a good idea from a technical standpoint. But anyway, we aren't ruling it out as part of this RFC.

rfcs/text/0408-versioning-types.md
nonparibus: [W3C use](https://www.w3.org/2005/05/tr-versions) the root for "the most mature Recommendation of a given technology or the most agreeing-on version". They also use `latest` for the last stable version of a doc, effectively the same thing. In their model `upcoming` denotes a version of a document with WIP changes (which we could use for unreleased edits).

[HASH Core](https://hash.ai/platform/core) uses `stable` for the last cut release of a thing, and its root URL serves _either_ this (if a user does not have edit permissions), or the tip containing unreleased changes (if a user has edit permissions).

I think the W3C model is probably clearer, where root = stable = latest release (with or without a redirect), and another specified keyword can be used to view unreleased edits (possibly depending on permissions managed by an EA).

Alfred-Mountfield: I can add a section _recommending_ that hosts also serve `/v/latest` to redirect to (or directly serve) the most recent published version of a type. I don't think we want to **enforce** it at this point yet, because I'm not sure of the ramifications yet for hosting the URIs on file-systems, etc. If it was clearly very valuable and something we needed for programmatic functionality then it makes sense to require it, but at the moment it seems a nice to have.

I'd also be inclined to not talk about things like `upcoming`, I don't think "unreleased edits" or unstable versions, etc. are particularly useful in the context of schemas, which are much more constrained in behaviour than something like a library or npm package.

Alfred-Mountfield: Despite literally just writing otherwise.. perhaps we do want to provision for unstable/beta/whatever versions of types 🙃 It might be helpful to host temporary URIs for the purposes of development, so a Block Schema can be unstable alongside an unstable release of the Block for beta testing / etc.

Although if we go down that road I think we'd have to spend quite a while thinking about the implications and making sure we don't introduce problems. Probably worth us not provisioning for them now, and instead adding to the spec when we have a better understanding of how this stuff works in practice.

Alfred-Mountfield: I've added a section talking about `/latest`

rfcs/text/0408-versioning-types.md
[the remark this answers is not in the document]
Alfred-Mountfield: I've removed the mention of "inline" because I don't think it adds value to the sentence and is just causing confusion.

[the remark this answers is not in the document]
CiaranMn: Understood, seems reasonable

benwerner01 (dismissed):
Nice work @Alfred-Mountfield - this RFC looks good from my side 👍

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,370 @@
+- Feature Name: versioning-types
+- Start Date: 2022-06-28
+- RFC PR: [blockprotocol/blockprotocol#408](https://github.com/blockprotocol/blockprotocol/pull/408)
+- Block Protocol Discussion: [blockprotocol/blockprotocol#408](https://github.com/blockprotocol/blockprotocol/discussions/408)
+
+# Summary
+
+[summary]: #summary
+
+The Type System outlined in [RFC 0352](https://github.com/blockprotocol/blockprotocol/blob/am/rfc-graph-type-system/rfcs/text/0352-graph-type-system.md) introduced classes of types that are used to describe the data used and produced by blocks and embedding applications. The types are possibly (ideally often) publicly accessible, and their hierarchy allows for them to be composed together into new types. Due to this, changes in a type can affect the structure of another one. This RFC defines how types are versioned, to mitigate potential problems that would arise without versioning.
```
thehabbos007, line 10: This should probably be a link to main
```suggestion
The Type System outlined in [RFC 0352](https://github.com/blockprotocol/blockprotocol/blob/main/rfcs/text/0352-graph-type-system.md) introduced classes of types that are used to describe the data used and produced by blocks and embedding applications. The types are possibly (ideally often) publicly accessible, and their hierarchy allows for them to be composed together into new types. Due to this, changes in a type can affect the structure of another one. This RFC defines how types are versioned, to mitigate potential problems that would arise without versioning.
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,370 @@
...
+
+#### Data Type Reference
+
+A data type reference will be compatible with another data type reference if the data types that are referred to are also compatible.
+
+> _Without_ non-primitive data types this implies that the references are to the same data type.
+
+#### Property Type Object
+
+A poperty type object is JSON object defining a schema which is used as `oneOf` the permitted values for a property type, where the schema is of `type: "object`, and within its own `properties` definition:
```
thehabbos007, line 158:
```suggestion
A property type object is JSON object defining a schema which is used as `oneOf` the permitted values for a property type, where the schema is of `type: "object`, and within its own `properties` definition:
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,370 @@
...
+- Updating a **major** version of a deeply nested type, for instance a data type, will require **major** version changes to all types using it (if they wish to update). This is deeply in contrast to most uses of semver like versioning libraries. In those cases, major versions of sub-dependencies often cause the need to update some internal logic, but these breaking changes to not propagate up the dependency chain. If we used semver for types, we'd sometimes see propagating major version changes, causing very large major versions in the higher-level types with lots of dependencies. This is a problem because the major appeal of semantic versioning would be the assumptions an implementation could make about compatibility in non-major version differences.
+
+## Other Decisions
+
+### Versioned vs Base URIs in Types
+
+- `$schema` needs to be versioned as it uniquely identifies the instance of the schema and makes it referenceable
+- Similarly, uses of `$ref` need to use versioned URIs to they can reference a specific iteration of the schema
```
thehabbos007, line 346:
```suggestion
- Similarly, uses of `$ref` need to use versioned URIs so they can reference a specific iteration of the schema
```

rfcs/text/0408-versioning-types.md
```diff
@@ -0,0 +1,370 @@
...
+An optional field is removed from the Type
+
+```json
+{
+  "P": "2.0.0",
+  "C": "2.1.0"
+}
+```
+
+This allows us to compare the versions depending on the circumstance outlined above, where we're trying to track forwards or backwards compatibility. In addition to being a rather unituitive and confusing solution, this still has a few large problems.
```
thehabbos007, line 334:
```suggestion
This allows us to compare the versions depending on the circumstance outlined above, where we're trying to track forwards or backwards compatibility. In addition to being a rather unintuitive and confusing solution, this still has a few large problems.
```

thehabbos007 (commented):
I agree with Ben's approval, but just adding these small typo comments :)

rfcs/text/0408-versioning-types.md
Alfred-Mountfield: Oops, thank you!

thehabbos007 (dismissed):

thehabbos007 (approved):

PR #348, opened 2022-06-14
Introduce Block Protocol v0.2 and custom element support

CiaranMn:
This PR introduces the new specification, docs and tooling for version 0.2 of the Block Protocol:
- new spec pages
- new doc pages
- package updates:
  - new `@blockprotocol/graph` package for use by block protocol authors and embedding applications
  - new `@blockprotocol/core` package which is used by the graph package
  - update `mock-block-dock` for the new spec

As part of this it introduces proper custom element / web component support:
- `custom-element` as a specified `blockType`
- new `custom-element` template via `create-block-app` 
- `mock-block-dock` support for custom elements

PR #411, opened 2022-06-30
Replace outdated links to `/spec` and `/docs`

kachkaev:
This PR complements https://github.com/hashintel/hash/pull/722. It addresses [Asana task](https://app.asana.com/0/1202482456346912/1202440221033864) (internal).

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ We welcome [contributions](#contributing) to this repository. Within it you will
 
 1.  `site`: the code for [blockprotocol.org](https://blockprotocol.org), including:
 
-    - the [Block Protocol specification](https://blockprotocol.org/spec) at [src/\_pages/spec](https://github.com/blockprotocol/blockprotocol/tree/main/site/src/_pages/spec)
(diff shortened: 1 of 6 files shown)


kachkaev:
@CiaranMn do we need to do anything with this line?

https://github.com/blockprotocol/blockprotocol/blob/9d7964e204be486c02bbb06480fa7de4f52f8d12/site/src/_pages/docs/3_spec/1_core-specification.mdx#L426

It originates from https://github.com/blockprotocol/blockprotocol/pull/348 ([line](https://github.com/blockprotocol/blockprotocol/pull/348/files#diff-80db3a682075d3e41e0be4a40e8f511309360bff51adb840ff280584038c41a0R422))

CiaranMn:
> blockprotocol/site/src/_pages/docs/3_spec/1_core-specification.mdx

Good spot – we should host the JSON at whatever its `$id` is, as we do for the JSON schema mentioned in the graph service.

I've created [a task for this](https://app.asana.com/0/1201095311341924/1202532375884542/f) (internal) which also mentions that we should maintain one copy of these types – e.g. we could have an MDX component `<JSON url=something>` which rendered the JSON at that URL in a code block.

PR #417, opened 2022-07-05
Display non v0.2 blocks in Block Hub

teenoh:
This PR displays non v0.2 blocks, which where hidden in https://github.com/blockprotocol/blockprotocol/pull/337, in the BP site.

### DIFF
--- a/site/src/components/pages/hub/block-data-container.tsx
+++ b/site/src/components/pages/hub/block-data-container.tsx
@@ -36,13 +36,20 @@ type BlockDataContainerProps = {
   schema: BlockSchema;
   sandboxBaseUrl: string;
   exampleGraph: BlockExampleGraph | null;
+  isUnsupportedBlock?: boolean;
 };
 
 const validator = new Validator();
 
 export const BlockDataContainer: VoidFunctionComponent<
   BlockDataContainerProps
-> = ({ metadata, schema, exampleGraph, sandboxBaseUrl }) => {
+> = ({
+  metadata,
+  schema,
+  exampleGraph,
+  sandboxBaseUrl,
+  isUnsupportedBlock,
+}) => {
   const [blockDataTab, setBlockDataTab] = useState(0);
   const [blockVariantsTab, setBlockVariantsTab] = useState(0);
   const [blockModalOpen, setBlockModalOpen] = useState(false);
@@ -230,11 +237,19 @@ export const BlockDataContainer: VoidFunctionComponent<
                   position: "relative",
                 }}
               >
-                <SandboxedBlockDemo
-                  metadata={metadata}
-                  props={props}
-                  sandboxBaseUrl={sandboxBaseUrl}
-                />
+                {isUnsupportedBlock ? (
+                  <>
+                    This block was written for an earlier version of the Block
+                    Protocol specification and cannot currently be displayed in
+                    the Hub.
+                  </>
+                ) : (
+                  <SandboxedBlockDemo
+                    metadata={metadata}
+                    props={props}
+                    sandboxBaseUrl={sandboxBaseUrl}
+                  />
+                )}
               </Box>
             </Box>
           </Box>
--- a/site/src/lib/blocks.ts
+++ b/site/src/lib/blocks.ts
@@ -165,7 +165,8 @@ export const excludeHiddenBlocks = (
   );
 };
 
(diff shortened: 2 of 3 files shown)


site/src/pages/[shortname]/blocks/[block-slug].page.tsx
```diff
@@ -106,6 +107,7 @@ type BlockPageProps = {
   schema: BlockSchema;
   sliderItems: BlockMetadata[];
   exampleGraph: BlockExampleGraph | null; // todo fix typing
+  isUnsupportedBlock: boolean;
```
Comment (CiaranMn) [inline site/src/pages/[shortname]/blocks/[block-slug].page.tsx:110]: Could we just call `isNonV0_2Block` inside `block-data-container`? we're passing this `isUnsupportedBlock` prop through this page but aren't using it

site/src/lib/blocks.ts
```diff
@@ -161,11 +161,15 @@ export const excludeHiddenBlocks = (
...
 ): ExpandedBlockMetadata[] => {
   return blocks.filter(
-    ({ packagePath, protocol }) =>
-      !blocksToHide.includes(packagePath) && parseFloat(protocol) >= 0.2,
+    ({ packagePath }) => !blocksToHide.includes(packagePath),
   );
 };
 
+/** Blocks that aren't compliant with BP V0.2  */
+export const isNonV0_2Block = ({ protocol }: ExpandedBlockMetadata) => {
```
CiaranMn, line 169: we could call this `checkIsUnsupportedBlock` or similar instead, saves renaming it if we change the versions we support

CiaranMn (commented):
Thanks @teenoh, works well! A couple of suggestions to simplify.

site/src/pages/[shortname]/blocks/[block-slug].page.tsx
Comment (teenoh) [reply site/src/pages/[shortname]/blocks/[block-slug].page.tsx]: Agree. Done in [755c63a](https://github.com/blockprotocol/blockprotocol/pull/417/commits/755c63a237bdd58cb3764c31ac7d3293decc4ef7)

site/src/lib/blocks.ts
teenoh: Done in [116e6c4](https://github.com/blockprotocol/blockprotocol/pull/417/commits/116e6c4a04c44e58d9dffcd9663b6cc0e5c5857b)

site/src/components/pages/hub/block-data-container.tsx
```diff
@@ -40,6 +40,11 @@ type BlockDataContainerProps = {
 
 const validator = new Validator();
 
+/** Blocks that aren't compliant with BP V0.2  */
+export const checkIsUnsupportedBlock = ({ protocol }: BlockMetadata) => {
+  return !(parseFloat(protocol) >= 0.2);
```
kachkaev, line 45: We are missing a check for 0.3+ (which would be invalid for the hub). This is only a theoretical issue though, so we can potentially ignore it for now. I'd go for something like:

```
const checkIfBlockIsSupported = ({ protocol }: BlockMetadata) => protocol !== "0.2"
````
	
We don't need to export this fn. If we want to share it, it'd have to go into some other file. This one should only export  the component and its props.

File: site/src/components/pages/hub/block-data-container.tsx [hunk 4]
```diff
@@ -230,11 +235,19 @@ export const BlockDataContainer: VoidFunctionComponent<
                   position: "relative",
                 }}
               >
-                <SandboxedBlockDemo
-                  metadata={metadata}
-                  props={props}
-                  sandboxBaseUrl={sandboxBaseUrl}
-                />
+                {checkIsUnsupportedBlock(metadata.protocol) ? (
```
kachkaev, line 238: Possible mismatch in function arg? `metadata.protocol` vs `metadata`

site/src/components/pages/hub/block-data-container.tsx
```diff
@@ -230,11 +234,19 @@ export const BlockDataContainer: VoidFunctionComponent<
                   position: "relative",
                 }}
               >
-                <SandboxedBlockDemo
-                  metadata={metadata}
-                  props={props}
-                  sandboxBaseUrl={sandboxBaseUrl}
-                />
+                {checkIfBlockIsSupported(metadata) ? (
```
kachkaev, line 237:
```suggestion
                {!checkIfBlockIsSupported(metadata) ? (
```

kachkaev (approved):

PR #420, opened 2022-07-05
Introduce `changesets` for npm version management

kachkaev:


.changeset/block-template-private-true.md
```diff
@@ -0,0 +1,5 @@
+---
+"block-template": patch
+---
+
+Add `"private": true` to all three template `package.json` files
```
kachkaev, line 5: https://github.com/changesets/changesets/discussions/783

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
+name: Release
+on:
+  push:
+    branches:
+      - main
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    if: github.repository == 'blockprotocol/blockprotocol'
```
CiaranMn, line 9: What's this about? Isn't it always?

CiaranMn (approved):

.github/workflows/release.yml
kachkaev: I saw this `if` in one repo that uses `changesets` (not sure I remember which one). This is a protection against running the pipeline in a fork.

PR #421, opened 2022-07-05
Make @blockprotocol/graph an esm package

CiaranMn:
# What this PR does

- Converts `@blockprotocol/graph` to an ESM package
- Splits out its exports into subpaths so that users who don't need the React or custom element code don't have to evaluate those modules, e.g.
```javascript
// now
import { useGraphEmbedderService } from "@blockprotocol/graph/react"
```
```javascript
// before
import { useGraphEmbedderService } from "@blockprotocol/graph"
```

[Asana task](https://app.asana.com/0/0/1202630990806929) (internal)

# Motivation

## Issue 1: ESM dependency

This PR was motivated by an error when attempting to use the `@blockprotocol/graph` package in the [hashintel/hash](https://github.com/hashintel/hash/) frontend. 

The immediate error was caused by the package having an ESM dependency (`lit`, which the `custom-element` base class relies on). While NextJS has some level of support for ESM dependencies, it relies on them being declared as `type: "module"` and being properly set up as an ESM package, which `@blockprotocol/graph` was not – it was configured to have ESM module output, but it wasn't an ESM package in that it didn't declare itself as such, and it didn't follow rules such as import statements including file extensions.

It's worth noting that `blockprotocol/site`, also a NextJS application, was using `@blockprotocol/graph` without issue. My assumption is that this is because `site` itself is `type: "module"`, and therefore chooses the ESM output of dependencies, whereas HASH is not and is using the commonjs output as part of its build process (by calling `require`, which does not work). I haven't confirmed this – it would require looking a bit at NextJS. 

While one solution would potentially be to convert the HASH frontend to be a `type: "module"` NextJS application, it seemed preferable to update this application to work properly without its users having to do things – except if they're using `commonjs` module resolution, in which case they will need to, but this should be a dwindling pool of legacy applications. If any ever surface we need to support we can revisit.

## Issue 2: avoiding client-only modules 

Once converted to an ESM package, a further error was revealed: `lit/reactive-element` has references to `window` which was causing a crash in the NextJS server-side build. This crashed was introduced to `blockprotocol/site` as well as the HASH frontend, even though `blockprotocol/site` was previously using the package fine. I _assume_ this is something to do with the change to an ESM package triggering a different module evaluation strategy, which meant the `lit` code path referencing `window` was hit when it wasn't previously, but my knowledge of commonjs vs ESM module resolution is not good enough to say for sure. 

In any case, the `exports` key in `package.json` provides a solution for this, by allowing exports to be split out into subpaths, rather than having everything exported from a single file. This means that users who aren't going to use the `react` or `custom-element` modules don't have to go through a file which imports them. `exports` is supported in TypeScript as of `4.7`, with some workarounds as detailed below.

# Detailed changes

1. Make `@blockprotocol/graph` a pure ESM package (guided by [this](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm))
2. Split out its exports into individual paths, via an `exports` key in `package.json`, so that the react/custom-element code doesn't need to be evaluated by users who don't need it. This is supported in TypeScript as of 4.7 - see [the release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#esm-nodejs), under the header "`package.json` Exports, Imports". There is an issue in using this in some bundlers, including webpack (which we use for both the site and block templates), with the workaround being to specify `typesVersions` as described in [this issue](https://github.com/microsoft/TypeScript/issues/33079#issuecomment-1169419172), and as seen in the diff for `@blockprotocol/graph/package.json`
3. These changes necessitated turning off `import/no-unresolved`:
  - within the `@blockprotocol/graph` package: we need to have file extensions on imports to be a pure ESM package. These need to be '.js' so that the import statement is correct in the compiled code ([TypeScript does not and is not going to](https://github.com/microsoft/TypeScript/issues/16577#issuecomment-754941937) allow you to specify `.ts` extensions which get rewritten to `.js` in the output). This means that the imports appear 'unresolved' in the code as far as the linting rule is concerned (see [this issue](https://github.com/import-js/eslint-plugin-import/issues/2446#issuecomment-1115215474))
  - within any package that imports `@blockprotocol/graph`, since the linting rule does not support `exports` in package.json ([issue](https://github.com/import-js/eslint-plugin-import/issues/1810))

# Known issues

In https://github.com/hashintel/hash there's an issue when importing useGraphEmbedderService from @blockprotocol/graph/react, in that NextJS's [output file tracing](https://nextjs.org/docs/advanced-features/output-file-tracing) does not include graph-service.json, and yet an import statement for it is preserved.

This is only a problem in production, because Vercel automatically does the minimal production build there, and strips out files it thinks we don't need. I haven't been able to reproduce the issue in a new project, so it seems to be something specific to the setup in HASH. For now I've commented out the import statement causing a problem – the effect of this is that there's no runtime check when a user registers a callback for a message that the message is named validly. The types do check this, however. Restoring this import is tracked [here](https://app.asana.com/0/1202538466812818/1202614421149286/f) (internal).

---


luisbettencourt (dismissed):
Not sure how to test this, but looking at the code everything looks fine (although it's weird to have to include the `.js` extension on relative imports).

packages/@blockprotocol/graph/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@blockprotocol/graph",
-  "version": "0.0.10",
+  "version": "0.0.11",
```
kachkaev, line 3: What’s the purpose of this bump? We could leverage `changesets` instead.

.changeset/convert-graph-to-esm-package.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blockprotocol/graph": patch
```
kachkaev, line 2:
```suggestion
"@blockprotocol/graph": minor
```

packages/@blockprotocol/graph/src/graph-block-handler.ts
```diff
@@ -1,6 +1,13 @@
...
 
-import serviceJsonDefinition from "./graph-service.json";
+/**
+ * There's an issue when importing useGraphEmbedderService from @blockprotocol/graph/react in hashintel/hash:
+ * NextJS's output file tracing does not include graph-service.json, and yet an import statement for it is preserved.
+ * This leads to a 'module cannot be found error'. For now, commenting out the import of the JSON from this file.
+ * @todo restore this when module resolution issue resolved
+ * @see https://app.asana.com/0/1202542409311090/1202614421149286/f
+ */
+// import graphServiceJson from "./graph-service.json" assert { type: "json" };
```
kachkaev, line 10: I guess we can use `readFileSync` instead, which should be picked up by [verce/nft](https://github.com/vercel/nft). A call to this function can be placed in `graph-service-json.ts`. Not a blocker, just something that can save us another release.

packages/@blockprotocol/graph/package.json
kachkaev: Oh I see we refer to a new version in `packages/block-template/templates/custom-element/package.json`. I guess we don't need `.changeset/convert-graph-to-esm-package.md` then.

packages/@blockprotocol/graph/package.json
```diff
@@ -21,16 +21,35 @@
...
   },
-  "main": "./dist/cjs/index.js",
-  "module": "./dist/esm/index.js",
-  "types": "./dist/esm/index.d.ts",
+  "type": "module",
+  "exports": {
+    ".": "./dist/index.js",
+    "./react": "./dist/react.js",
+    "./custom-element": "./dist/custom-element.js",
+    "./graph-service-json": "./dist/graph-service-json.js"
```
kachkaev, line 29: What are the downsides of re-exporting this payload from `index.ts`?

CiaranMn: From memory it caused an issue when trying to import it into a package which didn't need the JSON but had to evaluate the import statement. I can't remember exactly what – probably using `readFileSync` instead of import will eliminate whatever the problem was so I will do that at the same time. [task](https://app.asana.com/0/1202538466812818/1202614421149286/f)

packages/@blockprotocol/graph/src/graph-block-handler.ts
CiaranMn: Good idea, thank you - I will try this in a follow up as @teenoh needs this merged quicker for a PR based on it. 

tracked [here](https://app.asana.com/0/1202538466812818/1202614421149286/f)

teenoh (approved):

PR #427, opened 2022-07-11
Remove `private: true` in block templates with `ignore` config

kachkaev:
Blocked by https://github.com/changesets/changesets/pull/866


kachkaev:
No longer needed since https://github.com/blockprotocol/blockprotocol/pull/480

PR #430, opened 2022-07-12
Hook Service RFC

nathggns:
This RFC describes the hook service, currently described as a forward plan at https://blockprotocol.org/docs/spec/rfcs_and_roadmap#hook 

---
[Rendered view](https://github.com/blockprotocol/blockprotocol/blob/nh/rfc-hook-service/rfcs/text/0430-hook-service.md)

---
  - https://app.asana.com/0/0/1202142633000402


rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
+- Feature Name: `hook-service`
+- Start Date: (fill me in with today's date, 2022-07-12)
+- RFC PR: [blockprotocol/blockprotocol#430](https://github.com/blockprotocol/blockprotocol/pull/430)
+- Block Protocol Discussion: N/A
+
+# Summary
+
+[summary]: #summary
+
+This service standardises the communication between blocks and the embedding application necessary to enable embedding applications to 'take over' the rendering of parts of a block to extend functionality, while ensuring block can remain uncoupled from the embedding application, even where these are implemented by the same organisation.
```
nonparibus, line 10:
```suggestion
This service standardizes the communication necessary between blocks and embedding applications, allowing the latter to 'take over' rendering one or more parts of a block. This allows functionality to be extended, while ensuring blocks can remain uncoupled from embedding applications.
```

Tried to improve clarity.

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+A 'hook' is an injection point blocks provide to embedding applications to allow them to optionally render or modify already rendered views associated with a path to data provided by the graph service's `blockEntity` property. 
+
+The 'optionally' here is key. Blocks cannot require embedding applications to implement a handler for any specific 'hook'. Therefore, each 'hook' must specify a fallback rendering strategy which embedding applications implementing the hook service can use instead of implementing the hook themselves.
```
nonparibus, line 26:
```suggestion
**Note: the word 'optionally' here is key.** Blocks cannot today _require_ embedding applications to implement a handler for any specific 'hook'. Therefore, each 'hook' must specify a fallback rendering strategy which embedding applications implementing the hook service can use instead of implementing the hook themselves.
```

Not sure if we mean 'must' here or 'should'. There may be instances where a block could continue to function without an interface/service being passed in to a single given hook, depending on how these end up getting used.

More broadly we should capitalize [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) terms for clarity in our own RFCs and Service Specifications, as we do in the [Core Specification](https://blockprotocol.org/docs/spec/core-specification).

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+A 'hook' is an injection point blocks provide to embedding applications to allow them to optionally render or modify already rendered views associated with a path to data provided by the graph service's `blockEntity` property. 
+
+The 'optionally' here is key. Blocks cannot require embedding applications to implement a handler for any specific 'hook'. Therefore, each 'hook' must specify a fallback rendering strategy which embedding applications implementing the hook service can use instead of implementing the hook themselves.   
+
+A classical example would be rich text editing. In this case, you may not want to provide the underlying data storing the formatted text, as this may be bespoke to your embedding application. But you may still want to provide the plain text value for the block to make use of – i.e, if the block provides a word count feature.
```
nonparibus, line 28:
```suggestion
A classical example would be rich text editing. In this case, you may not want to provide the underlying data storing the formatted text, as this may be bespoke to your embedding application. But you may still want to provide the plain text value for the block to make use of – i.e., if the block provides a word count feature.
```

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+
+A block implements a hook by sending a `hook` message to the embedding application, with five pieces of information:
+- `node: HTMLElement | null`: The DOM node the embedding application will render a view into, or modifying an already rendered view. 
+- `type: "text" | "image" | "video" | string`: The type of view for the hook. While this can be any string as a fallback must be passed if `type` is not implemented, we believe `text`, `image`, and `video` will be commonly used. Straying from these commonly used types will make your block less portable between different embedding applications, particularly those designed to support all blocks.  
+- `path: string`: The path to the property on the graph service's `blockEntity` property associated with the hook
+- `fallback: (node: HTMLElement, type: "text" | "image" | "video" | string, path: string) => void`: The fallback rendering method which will be used if the embedding application does not implement this particular hook
```
nonparibus, line 34: I don't understand why `fallback` needs sending to the EA. Does the block not just say "I want _type_. Not got _type_? Okay, then. Let's use this other code of mine instead."?

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+- `hookId: string | null`: The ID of the hook as provided in the response when first sending a `hook` message. This should be `null` on first call.  
+
+The embedding application must respond with a `hookResponse` message specifying a `hookId: string` property, which will be provided in future `hook` messages to establish that the hook is simply being updated, and not installed.
```
nonparibus, line 37: Is "installed" the best word here?

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+
+# Prior art
+
+[prior-art]: #prior-art
+
+## `editableRef`
+
+[editableRef]: #editableref
+
+The first exploration of this space was the introduction of `editableRef` in the HASH app, currently powering the rich text blocks.
```
nonparibus, line 84:
```suggestion
A solution similar to the hooks service was first explored in a temporary implementation called `editableRef` within the [HASH](https://hash.ai) embedding application, enabling its _rich text_ blocks.
```

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+
+This was designed to avoid blocks being coupled to the particular rich text editing solution implemented in HASH, and in that it succeed.
+
+`editableRef` is a function provided by the embedding application which a block can call with a DOM node rendered by the block, which they wish the embedding application to render rich text present on the `text` field on the block entity into.
+
+It was also used by the embedding application (HASH) to know that any particular block had an editable text section within it. The particular solution HASH uses for rich text editing needs (or needed) to know at the time blocks are loaded that a user can edit text within them. This worked because a block would use `editableRef` by reading it from the properties passed to it, which means the `editableRef` key would be in the block's schema.
+
+`editableRef` is an unsuitable candidate for standardisation for a number of reasons:
```
nonparibus, line 92:
```suggestion
`editableRef` is an unsuitable candidate for standardization for a number of reasons:
```

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+- It does not indicate which block entity property on which to store the rich text data
+- It means that the type of the value in the `text` field is unknown to the block, and may not match the value in the block's schema
+- It has a one time use – you cannot use it to make multiple editable islands within a block (think a blockquote with a caption)
```
nonparibus, line 95:
```suggestion
- It has a one-time use – you cannot use it to make multiple editable islands within a block (think a blockquote with a caption)
```

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+- The requirement for the use of the feature to show up in the block's schema can make certain kinds of blocks impossible (i.e, blocks that may or may not have editable sections depending on their current state).
```
nonparibus, line 96:
```suggestion
- The requirement for the use of the feature to show up in the block's schema can make certain kinds of blocks impossible (i.e., blocks that may or may not have editable sections depending on their current state).
```

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+- An embedding application can only provide one implementation for `editableRef`, which means there may be a mismatch between the implementation provided by the embedding application and the feature expected by the block. I.e, an embedding application may use `editableRef` to provide image editing features, whereas the block may expect it to be used to provide rich text editing features. 
+
+Another minor reason is its naming is tied into React – i.e, "editableRef".
```
nonparibus, line 99:
```suggestion
Another minor reason is its naming is tied into React – i.e., "editableRef". 
```

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+
+`editableRef` is an unsuitable candidate for standardisation for a number of reasons:
+- It does not indicate which block entity property on which to store the rich text data
+- It means that the type of the value in the `text` field is unknown to the block, and may not match the value in the block's schema
+- It has a one time use – you cannot use it to make multiple editable islands within a block (think a blockquote with a caption)
+- The requirement for the use of the feature to show up in the block's schema can make certain kinds of blocks impossible (i.e, blocks that may or may not have editable sections depending on their current state).
+- An embedding application can only provide one implementation for `editableRef`, which means there may be a mismatch between the implementation provided by the embedding application and the feature expected by the block. I.e, an embedding application may use `editableRef` to provide image editing features, whereas the block may expect it to be used to provide rich text editing features.
```
nonparibus, line 97:
```suggestion
- An embedding application can only provide one implementation for `editableRef`, which means there may be a mismatch between the implementation provided by the embedding application and the feature expected by the block. I.e., an embedding application may use `editableRef` to provide image editing features, whereas the block may expect it to be used to provide rich text editing features. 
```

Not a fan of starting sentences with "i.e.,", but if we must...

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+[unresolved-questions]: #unresolved-questions
+
+- Is hook the best name for this service? Other options include (not exhaustive):
+  - editable/editing/editor service
+  - view service
+  - UI service
+  - plugin service
+  - DOM service
+  - extension service
+- Is the name 'hook' too tied into React and therefore confusing?
```
nonparibus, line 114: No, I don't think so. It has a literal metaphorical meaning that could be understood by non-programmers, nevermind non-React developers.

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+
+- Is hook the best name for this service? Other options include (not exhaustive):
+  - editable/editing/editor service
+  - view service
+  - UI service
+  - plugin service
+  - DOM service
+  - extension service
+- Is the name 'hook' too tied into React and therefore confusing? 
+- Is there any functionality expected to be enabled by this service where an embedding application would need to know statically that the service is being used – i.e, does this need to show up in the schema. If so, how does this work with dynamic properties (i.e, a property on a value in a list)
```
nonparibus, line 115:
```suggestion
- Is there any functionality expected to be enabled by this service where an embedding application would need to know statically that the service is being used – i.e., does this need to show up in the schema. If so, how does this work with dynamic properties (i.e., a property on a value in a list)
```

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+- Is the use of property paths to indicate the relevant value problematic? Do we need to make it easy to generate these property paths (possible using a proxy)?
+- Will blocks want to pass up data which is not present on `blockEntity` (i.e, on a linked entity or aggregation) or not provided by the block service at all
```
nonparibus, line 117:
```suggestion
- Will blocks want to pass up data which either not present on `blockEntity` (i.e., on a linked entity or aggregation) or not provided by the block service at all?
```

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+- Do we want to standardise the "types" of hook available or do we want it to be any arbitrary string? Should we 'special-case' the recommended types by providing specific message types for those, or should all uses of the service use the same message type?
```
nonparibus, line 118:
```suggestion
- Do we want to standardize the "types" of hook available or do we want it to be any arbitrary string? Should we 'special-case' the recommended types by providing specific message types for those, or should all uses of the service use the same message type? 
```

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+  - UI service
+  - plugin service
+  - DOM service
+  - extension service
+- Is the name 'hook' too tied into React and therefore confusing? 
+- Is there any functionality expected to be enabled by this service where an embedding application would need to know statically that the service is being used – i.e, does this need to show up in the schema. If so, how does this work with dynamic properties (i.e, a property on a value in a list)
+- Is the use of property paths to indicate the relevant value problematic? Do we need to make it easy to generate these property paths (possible using a proxy)?
+- Will blocks want to pass up data which is not present on `blockEntity` (i.e, on a linked entity or aggregation) or not provided by the block service at all
+- Do we want to standardise the "types" of hook available or do we want it to be any arbitrary string? Should we 'special-case' the recommended types by providing specific message types for those, or should all uses of the service use the same message type? 
+- Is the recommended list of types of hook sufficient?
```
nonparibus, line 119: To be clear, these are the initial types we expect will be commonly used.

Once we have witnessed real-world usage of the hook types, we will update this section to reflect reality.

nonparibus (changes requested):
A few comments, and our style guide needs following!

rfcs/text/????-hook-service.md
CiaranMn, line 28: I'm not sure about this para – in my head it's not that EAs don't _want_ to provide the data, it's that it's meaningless to the block without knowing what it looks like. On that point I'd say something like:

> A classical example would be rich text editing. In this case, the data structure for formatted text may be bespoke to the embedding application, and impossible for the block to predictably parse or generate. The expected text editing experience may involve custom formatting controls, or special inline actions or commands. By injecting its own rich text editing input, an embedding application can provide all this. 

The second point isn't addressed by the RFC and it isn't clear how it's part of the rationale for the service - how does the block still get the plain text value? I'd leave this point out until we explain how the proposal addresses it.

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+
+A classical example would be rich text editing. In this case, you may not want to provide the underlying data storing the formatted text, as this may be bespoke to your embedding application. But you may still want to provide the plain text value for the block to make use of – i.e, if the block provides a word count feature.
+
+A block implements a hook by sending a `hook` message to the embedding application, with five pieces of information:
+- `node: HTMLElement | null`: The DOM node the embedding application will render a view into, or modifying an already rendered view. 
+- `type: "text" | "image" | "video" | string`: The type of view for the hook. While this can be any string as a fallback must be passed if `type` is not implemented, we believe `text`, `image`, and `video` will be commonly used. Straying from these commonly used types will make your block less portable between different embedding applications, particularly those designed to support all blocks.
```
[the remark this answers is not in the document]
CiaranMn: how about `number`?

do we need to distinguish between `string` and something like `rich-text`?

Use case for these additions: to allow EAs to use their own inputs for common form inputs.

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+
+A classical example would be rich text editing. In this case, you may not want to provide the underlying data storing the formatted text, as this may be bespoke to your embedding application. But you may still want to provide the plain text value for the block to make use of – i.e, if the block provides a word count feature.
+
+A block implements a hook by sending a `hook` message to the embedding application, with five pieces of information:
+- `node: HTMLElement | null`: The DOM node the embedding application will render a view into, or modifying an already rendered view. 
+- `type: "text" | "image" | "video" | string`: The type of view for the hook. While this can be any string as a fallback must be passed if `type` is not implemented, we believe `text`, `image`, and `video` will be commonly used. Straying from these commonly used types will make your block less portable between different embedding applications, particularly those designed to support all blocks.  
+- `path: string`: The path to the property on the graph service's `blockEntity` property associated with the hook
```
CiaranMn, line 33: or something - we don't say anywhere what the EA is supposed to do with this path
```suggestion
- `path: string`: The path on the graph service's `blockEntity`'s `properties` object associated with the hook, which should be used to store any data inputted (where editing is allowed)
```

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+A 'hook' is an injection point blocks provide to embedding applications to allow them to optionally render or modify already rendered views associated with a path to data provided by the graph service's `blockEntity` property. 
+
+The 'optionally' here is key. Blocks cannot require embedding applications to implement a handler for any specific 'hook'. Therefore, each 'hook' must specify a fallback rendering strategy which embedding applications implementing the hook service can use instead of implementing the hook themselves.   
+
+A classical example would be rich text editing. In this case, you may not want to provide the underlying data storing the formatted text, as this may be bespoke to your embedding application. But you may still want to provide the plain text value for the block to make use of – i.e, if the block provides a word count feature.
+
+A block implements a hook by sending a `hook` message to the embedding application, with five pieces of information:
+- `node: HTMLElement | null`: The DOM node the embedding application will render a view into, or modifying an already rendered view.
```
CiaranMn, line 31: I think we need to talk about editing vs readonly views somewhere. We mention editing repeatedly as part of the motivation, but the technical details only talk about providing "a view".

My personal preference would be some kind of system/message argument whereby blocks can specify if this view is supposed to be:
1. Always editable (and shouldn't be shown if the user can't edit)
2. Always readonly
3. Whatever the app wants (e.g. depending on user permissions, or the general mode of the surrounding page)

I don't think leaving it up to the app is sufficient, e.g.:
- a block might want apps to take over displaying videos (if they have a special video player), but not allow users to change what the video is
- a block might want users to be able to edit something in one place and have it displayed in another (via two calls to the hook service, one for editing, one for display)

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+
+[prior-art]: #prior-art
+
+## `editableRef`
+
+[editableRef]: #editableref
+
+The first exploration of this space was the introduction of `editableRef` in the HASH app, currently powering the rich text blocks.
+
+This was designed to avoid blocks being coupled to the particular rich text editing solution implemented in HASH, and in that it succeed.
```
CiaranMn, line 86:
```suggestion
This was designed to avoid blocks being coupled to the particular rich text editing solution implemented in HASH, and in that it succeeded.
```

rfcs/text/????-hook-service.md
CiaranMn, line 96: I'm not sure needing to specify a property in the schema which may or may not show up depending on state is impossible - it could be an optional property? Probably this is part of a wider discussion about how these special values are declared in the schema, as flagged in the questions further down.

rfcs/text/????-hook-service.md
Alfred-Mountfield: e.g., not i.e.

rfcs/text/????-hook-service.md
Alfred-Mountfield: Hooks are also generally a computing term (and webhooks are a thing) so I don't think it's React-centric.

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+  - plugin service
+  - DOM service
+  - extension service
+- Is the name 'hook' too tied into React and therefore confusing? 
+- Is there any functionality expected to be enabled by this service where an embedding application would need to know statically that the service is being used – i.e, does this need to show up in the schema. If so, how does this work with dynamic properties (i.e, a property on a value in a list)
+- Is the use of property paths to indicate the relevant value problematic? Do we need to make it easy to generate these property paths (possible using a proxy)?
+- Will blocks want to pass up data which is not present on `blockEntity` (i.e, on a linked entity or aggregation) or not provided by the block service at all
+- Do we want to standardise the "types" of hook available or do we want it to be any arbitrary string? Should we 'special-case' the recommended types by providing specific message types for those, or should all uses of the service use the same message type? 
+- Is the recommended list of types of hook sufficient?
+- How do we formalise the relationship between the hook service and the graph service?
```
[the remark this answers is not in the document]
Alfred-Mountfield: To me this sounds like saying "how do we represent dependencies between services". The design at the moment mentions and talks about the graph service:

> modify already rendered views associated with a path to data provided by the graph service's `blockEntity` property.

If this crosses the boundary of defining how data is given to the block (e.g. a plain-text entity) then we should figure out how to encode the types of dependencies being introduced. Especially if it's needed for assessing EA/Block compatibility.

rfcs/text/????-hook-service.md
nonparibus: @nathggns Let's remove this as a question.

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+- `fallback: (node: HTMLElement, type: "text" | "image" | "video" | string, path: string) => void`: The fallback rendering method which will be used if the embedding application does not implement this particular hook
+- `hookId: string | null`: The ID of the hook as provided in the response when first sending a `hook` message. This should be `null` on first call.  
+
+The embedding application must respond with a `hookResponse` message specifying a `hookId: string` property, which will be provided in future `hook` messages to establish that the hook is simply being updated, and not installed. 
+
+This `hook` message should be sent every time any one of these pieces of information changes (but not, importantly, if the value `path` refers to changes – embedding applications can handle this). In this way, you can think of the message as setting up the hook.
+
+An embedding application receiving a `hook` message should use this to directly manipulate `node` in order to render the `type` view for the specified `path`. Any embedding application implementing the hook service *must* call the `fallback` for any hooks it does not wish to implement.
```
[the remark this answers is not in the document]
nonparibus: I share this concern (less eloquently/indirectly put in https://github.com/blockprotocol/blockprotocol/pull/430#discussion_r919200301)

rfcs/text/????-hook-service.md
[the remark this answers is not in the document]
nonparibus: I think we need to be clear what we're talking about when we say "type".

We're not (in my mind) talking about data types, like strings, or rich-text -- but the type of _interface_ a user interacts with while dealing with data in these nodes.

They probably will have different interfaces for inputting plain strings vs WYSIWYG-style rich text.

rfcs/text/????-hook-service.md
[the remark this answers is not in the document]
nathggns: Sorry, committing the suggestion marked this as resolved.

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+- RFC PR: [blockprotocol/blockprotocol#430](https://github.com/blockprotocol/blockprotocol/pull/430)
+- Block Protocol Discussion: N/A
+
+# Summary
+
+[summary]: #summary
+
+This service standardises the communication between blocks and the embedding application necessary to enable embedding applications to 'take over' the rendering of parts of a block to extend functionality, while ensuring block can remain uncoupled from the embedding application, even where these are implemented by the same organisation.
+
+In order to enable direct manipulation, this service is likely to be difficult or impossible to implement in Embedding Applications which use common ‘sandboxing’ techniques including `iframe`s.
```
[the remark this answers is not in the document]
nathggns: I see your point. Do you think we ought to remove this line, or clarify it?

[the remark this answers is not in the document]
CiaranMn: I'd make some variant of the point in my second para, i.e. that this service requires the app to share a DOM with the block, and so would not be usable with block types that own an entire sandbox scope, if we ever introduced such a type.

rfcs/text/????-hook-service.md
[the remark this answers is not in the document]
nathggns: @CiaranMn how does a block query if the embedding application implements a service?

rfcs/text/????-hook-service.md
nathggns:
```suggestion
```

rfcs/text/????-hook-service.md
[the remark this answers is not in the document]
CiaranMn: There's nothing specified on this yet – probably we should make it part of the `initResponse` message

rfcs/text/????-hook-service.md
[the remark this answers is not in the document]
nathggns: This was discussed internally at https://hashintel.slack.com/archives/C02LG39FJAU/p1658246784302989 as something which is out of scope for this RFC.

rfcs/text/????-hook-service.md
nathggns: 👍

rfcs/text/????-hook-service.md
nathggns: Fallback has been removed from the proposal. Instead, a `NOT_IMPLEMENTED` error will be returned which blocks can handle in order to implement a fallback.

rfcs/text/????-hook-service.md
```diff
@@ -0,0 +1,134 @@
...
+This proposal fixes all of these problems.
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- Is hook the best name for this service? Other options include (not exhaustive):
+  - editable/editing/editor service
+  - view service
+  - UI service
```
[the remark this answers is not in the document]
nathggns: Added to the list

rfcs/text/????-hook-service.md
nathggns: Do you think I need to add this to the RFC text?

rfcs/text/????-hook-service.md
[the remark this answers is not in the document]
nathggns: I think this has been addressed now. Could you take a look? This text has also been moved.

rfcs/text/????-hook-service.md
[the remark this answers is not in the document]
nathggns: The fallback proposal has been removed, and instead the embedding application must return a `NOT_IMPLEMENTED` error if the hook is not implemented, which the block can handle to implement a fallback. Additionally, the core specification has been extended to require `NOT_IMPLEMENTED` errors for messages to services not implemented by the embedding application.

rfcs/text/????-hook-service.md
nathggns: Discussed offline that this is not really appropriate to be included in the spec. The hook service isn't specifically about "inputting data". The point has been made elsewhere though about how paths should be used.

rfcs/text/????-hook-service.md
nathggns: I explained offline that this is connection to where properties may be one of multiple types, and the hook service is only dynamically implemented when a property is a certain type. Essentially the use of the service is dynamic. 

@CiaranMn do you think the text here needs changing after our discussion?

rfcs/text/????-hook-service.md
nathggns: Updated. I agree the point about text doesn't need to be specifically mentioned here, and the point I was getting at about "representations" of data is now made elsewhere in the RFC.

rfcs/text/????-hook-service.md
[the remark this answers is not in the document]
nathggns: This remains a question (as in reality and also in the RFC) but we have made an internal task to track it – https://app.asana.com/0/1200211978612931/1202636018097196/f

We have agreed answering this question is outside the scope of this RFC.

rfcs/text/????-hook-service.md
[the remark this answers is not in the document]
nonparibus: To be clear/for public record, the thing we've decided will be out-of-scope here is "constraints on hooks" (e.g. how do I selectively disable specific functions within a text-editing interface that may or may not be provided by an embedding application, within a particular hook). One option for how this will work is that function names will be arrived at in a non-prescriptive manner, and supported as independently/organically as the hook `type` names themselves. We can establish precedents in our original implementations without binding future users.

rfcs/text/????-hook-service.md
nonparibus: No. What I mean to imply with my comment above, though, is that I don't think it needs to be a question/something we spend further time discussing now. The line can be removed.

nathggns: Ah, gotcha. Will do!

nonparibus (dismissed):
Approved; see Slack and await further review from @CiaranMn.

rfcs/text/????-hook-service.md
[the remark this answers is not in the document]
CiaranMn: Looks good, thank you

rfcs/text/0430-hook-service.md
```diff
@@ -0,0 +1,148 @@
...
+TODO:
+
+This is the technical portion of the RFC. Explain the design in sufficient detail that:
+
+- Its interaction with other features is clear.
+- It is reasonably clear how the feature would be implemented.
+- Corner cases are dissected by example.
+
+The section should return to the examples given in the previous section, and explain more fully how the detailed proposal makes those examples work.
+
```
CiaranMn, line 77: Suggest using the content above as the reference-level, and therefore moving this up / deleting the template content
```suggestion
```

rfcs/text/0430-hook-service.md
```diff
@@ -0,0 +1,148 @@
...
+## Hook Service
+
+A **hook** is an injection point blocks provide to embedding applications to allow them to optionally render or modify already rendered views associated with a Graph Service property in the block's schema.
+
+**Note: the word 'optionally' here is key.** Blocks cannot today _require_ embedding applications to implement a handler for any specific 'hook'. Instead, they should implement a fallback in the event the embedding application does not implement a hook (see the Errors section below)
+
+A classical example would be rich text editing. In this case, the data structure for formatted text may be bespoke to the embedding application, and impossible for the block to predictably parse or generate. The expected text editing experience may involve custom formatting controls, or special inline actions or commands. By injecting its own rich text editing input, an embedding application can provide all this.
+
+A block implements a hook by sending a `hook` message to the embedding application, with five pieces of information:
```
CiaranMn, line 32: I think the existing content can serve as the reference-level explanation, with a bit more by way of real-world example to fill out the guide-level
```suggestion

In the rich text example, a block would send a message under the service specifying a type of `"text"`, a DOM node in which to inject the rich text editing input, and the property path the data should be stored under. Any data created by the input can be stored by the embedding application in its own format. The block need not receive this data, since it won't understand it anyway. Instead, the application should send the block a value for the property which matches what the block expects in its schema. In the case of rich text, this is likely to be `"string"`, and the application may choose to supply a plain text representation of the rich text created.

# Reference-level explanation

[reference-level-explanation]: #reference-level-explanation

A block implements a hook by sending a `hook` message to the embedding application, with five pieces of information:
```

rfcs/text/0430-hook-service.md
```diff
@@ -0,0 +1,148 @@
...
+
+TODO:
+
+- Why is this design the best in the space of possible designs?
+- What other designs have been considered and what is the rationale for not choosing them?
+- What is the impact of not doing this?
+
+# Prior art
+
+[prior-art]: #prior-art
```
CiaranMn, line 97: Suggest making the discussion of `editableRef` the 'rationale and alternatives' as it includes both an alternative (`editableRef`) and the rationale for moving away from it.
```suggestion
```

CiaranMn (commented):
Thanks @nathggns, I think your updates have addressed my comments – I've suggested some changes to move text around and get rid of the outstanding bits of the RFC template.

rfcs/text/????-hook-service.md
nonparibus: I think my correction is correct. And it's definitely `e.g.` not `i.e.`

nonparibus (dismissed):

nonparibus (approved):

PR #431, opened 2022-07-12
Cleanup GitHub workflows

kachkaev:
This PR is an exact from my WIP Playwright setup. I’ve made a few small tweaks to GitHub workflows and decided to merge them separately to de-risk the main piece of work. See details in comments below.

### DIFF
--- a/.github/workflows/algolia-upload-index.yml
+++ b/.github/workflows/algolia-upload-index.yml
@@ -13,8 +13,8 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
         with:
           node-version: 16
           cache: yarn
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Check codebase
+name: CI
 on: pull_request
 
 jobs:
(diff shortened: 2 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -8,15 +8,14 @@ jobs:
...
 
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
           node-version: 16
           cache: yarn
 
-      - name: Install deps and build (with cache)
-        uses: bahmutov/npm-install@v1
+      - run: yarn install
```
kachkaev, line 18: I’ve copied this yaml from a `changesets` example, but have realised that we don’t need [bahmutov/npm-install](https://github.com/bahmutov/npm-install) here. `actions/setup-node` should take care of the Node cache.

Unfortunately, it’s impossible to check this change immediately, but if this pipeline fails on `main`, it won’t cause any damage.

.github/workflows/ci.yml
```diff
@@ -1,14 +1,14 @@
-name: Check codebase
+name: CI
```
kachkaev, line 1: I renamed this pipeline (including the file) to get shorter output in GitHub UI. The new setup is aligned [with the HASH repo](https://github.com/hashintel/hash/actions/workflows/ci.yml).

.github/workflows/algolia-upload-index.yml
```diff
@@ -13,8 +13,8 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
```
kachkaev, line 17: New major versions of these actions run on Node 16 instead of Node 12.
- https://github.com/actions/checkout/releases
- https://github.com/actions/setup-node/releases

CiaranMn (approved):

PR #432, opened 2022-07-12
Update Footer

teenoh:
Minor update to the site's footer. 

---
  - https://app.asana.com/0/0/1202562814920532


site/src/components/footer.tsx
```diff
@@ -217,59 +217,6 @@ export const Footer: FC<FooterProps> = ({ sx = [], ...boxProps }) => {
...
-              Supported by{" "}
-              <Link
-                href="https://hash.ai"
-                sx={{
-                  position: "relative",
-                  top: -1,
-                  marginLeft: 1,
-                }}
-              >
-                <HASHLogoIcon />
```
kachkaev, line 269: This component is still defined in the codebase, but we no longer use it. I think we can delete it entirely on `main` and then restore from git history, if needed.

kachkaev (changes requested):
See a comment with a minor suggestion 👀

nonparibus (approved):

kachkaev (approved):

PR #434, opened 2022-07-13
Configure Playwright for integration and smoke tests

kachkaev:
This PR introduces integration and smoke (e2e) tests for [blockprotocol.org](https://blockprotocol.org/). This should speed up PR reviews and increase confidence in codebase stability.

See details in [site/README.md](https://github.com/blockprotocol/blockprotocol/pull/434/files#diff-669d2a51f46a925c4e680583bb9a7fc173b8ca2f52d991e8986ab52c20321064)

In scope:

- Playwright dependencies and configuration
- three example tests (1 integration, 1 smoke and 1 universal)
- two new CI jobs
   - Integration tests in `ci.yml` – includes launching a local DB, building and running Next.js + caching
   - Smoke tests in `site-deployment.yml` – includes running Playwright against Vercel deployment

Out of scope:

- real test scenarios (to be added in follow-up PRs)
- using multiple browsers in integration tests (to be enabled once we’ve written key test scenarios)

---
  - https://app.asana.com/0/0/1202605360401418


.gitignore
```diff
@@ -36,8 +36,10 @@ dist
 
 # workspace specific
 site/blocks-data.json
+site/playwright-report
+site/test-results
 site/public/blocks
 site/public/schemas
 site/site-map.json
 
-.parcel-cache
```
kachkaev, line 43: I could  not find any signs of using parcel in the repo, so we should not need this line

.gitignore
```diff
@@ -36,8 +36,10 @@ dist
...
+var
```
kachkaev, line 45: `var` stores service logs during integration tests. I added it to `.gitignore` in case if someone wants to run the commands from the YAML locally. We might also use this local folder for some other stuff.

.github/workflows/site-deployment.yml
```diff
@@ -3,15 +3,20 @@ name: Site deployment
 on:
   deployment_status:
   schedule:
-    - cron: "*/10 * * * *"
+    - cron: "*/10 * * * *" ## Every 10 minutes
+    - cron: "0 4 * * *" ## Every night (UTC time)
```
CiaranMn, line 7:
```suggestion
    - cron: "0 4 * * *" ## 4am every night (UTC time)
```

.github/workflows/site-deployment.yml
```diff
@@ -3,15 +3,20 @@ name: Site deployment
...
 
 jobs:
   warm_up_vercel:
     ## Vercel uses AWS Lambda. When we call getStaticProps in fallback mode or getServerSideProps,
     ## a cold Lambda function may take 5-15 seconds to execute. This pipeline warms up Lambda
     ## containers for new deployments and keeps them up by fetching affected URLs every 10 minutes.
+    ## Context: https://github.com/orgs/vercel/discussions/496
     name: Warm up Vercel
-    if: github.event_name == 'schedule' || github.event.deployment_status.state == 'success'
+    if: github.event.deployment_status.state == 'success' || github.event_name == 'schedule'
```
CiaranMn, line 19: Not sure if this would fire twice at 4am using the previous approach, but probably worth tying it to its intended schedule in any case, so we don't have to worry about it if we introduce more.
```suggestion
    if: github.event.deployment_status.state == 'success' || github.event.schedule == '*/10 * * * *'
```

CiaranMn (approved):
Looks good, a couple of non-blocking suggestions

.github/workflows/site-deployment.yml
kachkaev: I’ve defined `needs: [warm_up_vercel]` in `smoke_tests` to run tasks in a sequence. If we change this condition, the `warm_up_vercel` will be skipped and so `smoke_tests` will probably not get triggered.

GitHub’s `cron` schedule is [quite inaccurate](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule), i.e. the runs can be postponed or even skipped. I see no harm in warming up Vercel functions ≈145 times instead of ≈144*

* ≈ once in 10 mins × 24 hours

kachkaev: I think that once in 5-10 mins is the maximum frequency GitHub can give: https://stackoverflow.com/a/63192190/1818285

PR #443, opened 2022-07-19
Update homepage playwright tests

teenoh:
This PR updates the homepage playwright tests to cover more scenarios. It also

- Fixes an issue with opening the search modal via "/" on firefox
- Updates page header tests.

### DIFF
--- a/site/src/components/carousel.tsx
+++ b/site/src/components/carousel.tsx
@@ -14,7 +14,7 @@ type CarouselProps<T> = {
   edgeBackground?: { left: string; right: string };
   settings?: Partial<Settings>;
   sx?: BoxProps["sx"];
-  "data-testid": string;
+  "data-testid"?: string;
 };
 
 type ArrowProps = {
--- a/site/src/components/carousel.tsx
+++ b/site/src/components/carousel.tsx
@@ -14,6 +14,7 @@ type CarouselProps<T> = {
   edgeBackground?: { left: string; right: string };
   settings?: Partial<Settings>;
   sx?: BoxProps["sx"];
+  "data-testid": string;
 };
 
 type ArrowProps = {
@@ -68,6 +69,7 @@ export const Carousel = <T,>({
   settings = {},
   edgeBackground,
   sx = [],
+  ...props
 }: CarouselProps<T>): ReactElement => {
   const theme = useTheme();
   const isMobile = useMediaQuery(theme.breakpoints.down("md"));
@@ -102,6 +104,7 @@ export const Carousel = <T,>({
 
   return (
     <Box
+      data-testid={props["data-testid"]}
       sx={[
         {
           overflow: "hidden",
--- a/site/src/components/search-nav-button.tsx
+++ b/site/src/components/search-nav-button.tsx
@@ -11,7 +11,7 @@ export const SearchNavButton = () => {
   const theme = useTheme();
 
   useEffect(() => {
-    const onKeyDown = (evt: KeyboardEvent) => {
+    const onKeyDown = (event: KeyboardEvent) => {
       if (evt.key === "/") {
         evt.preventDefault();
         setDisplayModal((prev) => !prev);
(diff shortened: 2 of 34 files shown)


teenoh:
> @teenoh Wouldn't we rather use the original `U+2026` as previously seen here?

I agree. I've reverted back

site/src/components/carousel.tsx
```diff
@@ -102,6 +104,7 @@ export const Carousel = <T,>({
 
   return (
     <Box
+      data-testid={props["data-testid"]}
```
kachkaev, line 107: Do we need this to be a variable? A hardcoded `data-testid` value could work too in this case. If we have multiple block carousels, we can select them via `...parent... >> [data-testid="block-slider (or some other name)"]`.

site/src/components/search-nav-button.tsx
```diff
@@ -11,8 +11,9 @@ export const SearchNavButton = () => {
   const theme = useTheme();
 
   useEffect(() => {
-    const onKeyDown = ({ key }: KeyboardEvent) => {
-      if (key === "/") {
+    const onKeyDown = (evt: KeyboardEvent) => {
+      if (evt.key === "/") {
+        evt.preventDefault();
```
kachkaev, line 16: 👍

site/tests/universal/homepage.test.ts
```diff
@@ -1,5 +1,12 @@
 import { expect, test } from "@playwright/test";
 
+const links = {
+  exploreAllBlocks: "/hub",
+  developingBlocks: "/docs/developing-blocks",
+  embeddingBlocks: "/docs/embedding-blocks",
+  login: "/login",
+};
```
kachkaev, line 8: I'd just hardcode the values in assertion statements. The fewer levels of abstraction we have in tests, the more robust they are. Repetition is fine – we can always use search & replace if we need it.

site/src/components/search-nav-button.tsx
```diff
@@ -11,8 +11,9 @@ export const SearchNavButton = () => {
   const theme = useTheme();
 
   useEffect(() => {
-    const onKeyDown = ({ key }: KeyboardEvent) => {
-      if (key === "/") {
+    const onKeyDown = (evt: KeyboardEvent) => {
```
kachkaev, line 14:
```suggestion
    const onKeyDown = (event: KeyboardEvent) => {
```
(less crypting)

site/tests/universal/page-header.test.ts
```diff
@@ -1,6 +1,6 @@
 import { expect, test } from "@playwright/test";
 
-test("page header navigation works", async ({ page }) => {
+test("links and interactions in page header should work", async ({ page }) => {
```
kachkaev, line 3: How about placing search modal into another `test("")` (same file)? It might be easier to communicate the intent of our app interactions.

teenoh: Sounds good. Updated here => https://github.com/blockprotocol/blockprotocol/pull/443/commits/54e97e8463a588161ddf075be0d38afd4331ced6

site/tests/universal/homepage.test.ts
teenoh: Sounds good. Updated here => https://github.com/blockprotocol/blockprotocol/pull/443/commits/54e97e8463a588161ddf075be0d38afd4331ced6

site/tests/universal/homepage.test.ts
```diff
@@ -26,4 +26,40 @@ test("Home page should contain key elements", async ({ page }) => {
...
+
+  // Block slider
+  await expect(page.locator('[data-testid="block-slider"]')).toBeVisible();
+
+  // New blocks can get added, hence the usage of greater than instead of equal too
+  expect(
+    await page.locator('[data-testid="block-slider"] >> .slick-slide').count(),
+  ).toBeGreaterThan(5);
+
+  const finalCTA = page.locator('[data-testid="final-cta"]');
```
kachkaev, line 48: Optional, minor (can be ignored or addressed in a follow-up): `finalCTA` → `finalCta`, i.e. `strictCamelCase`. This naming style is a bit more readable, especially when there are multiple abbreviations next to each other.

kachkaev (approved):

site/tests/universal/homepage.test.ts
teenoh: Will address in a follow up

PR #446, opened 2022-07-19
Add tests for hub page 

teenoh:
This PR 
- adds tests for the hub page and page's footer
- fixes page-header tests on mobile


site/src/components/mdx/mdx-page-content.tsx
```diff
@@ -108,7 +108,7 @@ export const MdxPageContent: VFC<MdxPageContentProps> = ({
 
   useEffect(() => {
     const onScroll = () => {
-      if (!detectHeadingFromScroll) return;
+      if (!detectHeadingFromScroll || headings.length === 0) return;
 
       let headingAtScrollPosition: Heading = headings[0]!;
```
teenoh, line 113: Added this check since there were scenarios where `headings` was an empty array and that caused an error.

site/tests/universal/hubpage.test.ts
```diff
@@ -0,0 +1,34 @@
...
+    ),
+  ).toBeVisible();
+
+  await expect(
+    page.locator("text=All open-source and free to use"),
+  ).toBeVisible();
+
+  expect(
+    await page.locator('[data-testid="block-card"]').count(),
+  ).toBeGreaterThan(5);
```
kachkaev, line 18: Optional, can be done in a follow-up: lets find one of the cards (Code?) and inspect it a bit further. We can make sure that it's clickable and the URL is of the right shape. We can also check elements like header, description, version and user. We don't have to match version precisely, just by a regexp. We can possibly escape some `data-testid`s by making the markup a bit more semantic (e.g. using header tag instead of div)

kachkaev (dismissed):

site/tests/universal/hubpage.test.ts
teenoh: I agree... I still think we need a way to target certain sections since we can have multiple semantic elements. A good thing to use would be classnames, but most classnames we have are generated and a bit cryptic

kachkaev: Class names are an implementation detail, which is invisible to end users. It's better to interact with something more meaningful to humans such as visible text or a11y attributes.

If this is impossible, plan B is to use `data-testid`. We don't have to assign it to every element. If selecting by text is possible within a region, we can add `data-testid` to the region and then use text / a11y attributes within it.

teenoh: will address changes in a follow up

teenoh: Thanks, I will address this in a follow up

teenoh: Added tests for Code block card here => [6b37b72](https://github.com/blockprotocol/blockprotocol/pull/446/commits/6b37b72077bdbbe841be597dd28ab9f919eedcdd)

site/tests/integration/hubpage.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    "/docs/developing-blocks",
+  );
+
+  await expect(page.locator("text=Read the Quick Start Guide")).toHaveAttribute(
+    "href",
+    "/docs/developing-blocks",
+  );
+});
+
+const codeBlock = blocksData.find(
```
kachkaev, line 34: Optional: `codeBlock` → `codeBlockMetadata`

site/tests/integration/hubpage.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+test("Block Card should contain key elements", async ({ page }) => {
+  expect(
+    codeBlock,
+    "Code block should be prepared before this test",
+  ).toBeDefined();
+
+  if (!codeBlock) {
+    return;
+  }
```
kachkaev, line 46: That’s just to narrow TS type, right? How about throwing an exception `if (!codeBlock)` instead? We might even be able to do this in the root of the file. Not having block metadata does not fail the test but rather fails the setup.

site/tests/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "type": "commonjs",
+  "devDependencies": {
+    "@playwright/test": "1.23.3",
+    "playwright": "^1.23.3"
+  }
```
kachkaev, line 6: What happens if we remove these?

kachkaev: If it's just ESLint, we can silence it. Alternatively, we can see what's needed to remove `"type": "commonjs"` at all.

site/tests/integration/hubpage.test.ts
teenoh: I agree. I initially put it in the test because we don't need it for all tests in that file. will update

site/tests/package.json
teenoh: It's just ESLint. Silenced it in https://github.com/blockprotocol/blockprotocol/pull/446/commits/b6eeea537ddc7374ace9d03daacf553b870d4ac9

site/tests/integration/hubpage.test.ts
teenoh: https://github.com/blockprotocol/blockprotocol/pull/446/commits/b6eeea537ddc7374ace9d03daacf553b870d4ac9

kachkaev (approved):
🚀

kachkaev (dismissed):

.github/workflows/ci.yml
```diff
@@ -173,7 +173,6 @@ jobs:
 
   site-integration-tests:
     name: "Site integration tests"
-    if: ${{ github.event_name == 'pull_request' }}
```
kachkaev, line 176: Thanks @teenoh! Context: https://github.com/blockprotocol/blockprotocol/pull/447#discussion_r926805900

PR #448, opened 2022-07-20
Bump blocks

nathggns:
https://github.com/hashintel/hash/pull/826 updated our header and callout blocks, but the updates aren't yet in the hub. This bumps our block commit shas to fix that.


kachkaev (approved):
Both blocks are still hidden, so we can't test them in a preview release. I assume they've been checked locally in HASH as part of https://github.com/hashintel/hash/pull/826

teenoh (approved):

PR #450, opened 2022-07-21
Add tests for block detail page

teenoh:
This PR adds tests for block detail page. The tests for this page are split into 2 =>
1. An integration test that checks if key elements (e.g author, displayName, version, readme, carousel) are rendered in the block page 
1. An integration test that checks if the appropriate UI is rendered for unsupported blocks
1. A universal test that checks if our sandboxing setup renders the block and if changes to the json editor updates the block


site/tests/integration/blockpage.test.ts
```diff
@@ -0,0 +1,113 @@
+import { expect, test } from "playwright-test-coverage";
+
+import blocksData from "../../blocks-data.json";
```
kachkaev, line 3: This import breaks linting because the file is not checked-in in git: https://github.com/blockprotocol/blockprotocol/runs/7455330240?check_suite_focus=true#step:9:54

You can use `await fs.readJson()` via `fs-promise` instead.

site/tests/integration/blockpage.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+test("Block page should contain key elements", async ({ page, isMobile }) => {
+  await page.goto(codeBlock.blockPackagePath);
+
+  await expect(
+    page.locator(`h1:has-text('${codeBlock.displayName}')`),
+  ).toBeVisible();
+
+  await expect(
+    page.locator(`p:has-text('${codeBlock.description}')`),
```
kachkaev, line 25: Generally speaking, we should escape `'` in values here. I _guess_ we can get away with the selectors as is because `codeBlock` is owned by HASH so we should not anticipate query injections from ourselves 😅 The syntax will be broken though is someone adds `'` to description.

site/tests/integration/blockpage.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  );
+
+  if (isMobile) {
+    await page.locator("text=Data Source").click();
+    await page.locator("text=Block Schema").click();
+  } else {
+    await page.locator("text=Block Schema").click();
+  }
+  await expect(page.locator("#simple-tabpanel-1")).toBeVisible();
+  // await expect(page.locator("#simple-tabpanel-1")).not.toBeEmpty();
```
kachkaev, line 54: 👀

site/tests/integration/blockpage.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  await expect(page.locator("h2:has-text('Repository')")).toBeVisible();
+
+  await expect(
+    page.locator("a:below(h2:has-text('Repository'))").first(),
+  ).toHaveAttribute("href", codeBlock.repository);
+
+  await expect(page.locator("text=Explore more blocks")).toBeVisible();
+
+  // Block slider
```
kachkaev, line 76: This comment is a bit like:
Screenshot 2022-07-21 at 19 25 04

😅 I'd remove all comments that repeat that selectors and ideally use selectors that don't require comments.

site/tests/integration/blockpage.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  await page.goto(unsupportedBlock.blockPackagePath);
+
+  const blockExample = unsupportedBlock.examples[0] as Record<string, string>;
+
+  if (isMobile) {
+    await page.locator(".MuiTabs-root >> text=Source Code").click();
+  }
+
+  // confirm block properties tab contains example data
+  await expect(page.locator("#simple-tabpanel-0 >> textarea")).toHaveValue(
```
kachkaev, line 98: Optional: replace classes and ids with data-testid. This will make selectors look more semantically meaningful and less fragile when we change implementation details. For example, if we update MUI components and their class names have changed, we should not expect the tests to fail as long as the app looks the same for the end users.

site/tests/shared/sleep.ts
```diff
@@ -0,0 +1,4 @@
+export const sleep = (ms: number) =>
```
kachkaev, line 1: `import sleep from "sleep-promise"`


It's a dependecy of the repo root, but can make it a dev depency of site too just in case.

site/tests/universal/blockpage.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  ).toHaveAttribute("href", codeBlockUrl);
+
+  await page.goto(codeBlockUrl);
+
+  await expect(page.locator("iframe[title='block']")).toBeVisible();
+
+  const blockFrameLocator = page.frameLocator("iframe[title='block']");
+
+  // wait till block is rendered
+  // @todo figure out why this doesn't come up on safari
```
kachkaev, line 24: 🤔 Please share the result when you find it out!

site/tests/integration/blockpage.test.ts
teenoh: Linting breaks because blocksData isn't properly typed when imported. Typed it in [5adad49](https://github.com/blockprotocol/blockprotocol/pull/450/commits/5adad49fdf2865ff470f34040be638718ea3a2aa)

site/tests/integration/blockpage.test.ts
teenoh: Uncommented in [5adad49](https://github.com/blockprotocol/blockprotocol/pull/450/commits/5adad49fdf2865ff470f34040be638718ea3a2aa)

site/tests/integration/blockpage.test.ts
teenoh: Agreed. Updated in [5adad49](https://github.com/blockprotocol/blockprotocol/pull/450/commits/5adad49fdf2865ff470f34040be638718ea3a2aa)

site/tests/universal/blockpage.test.ts
teenoh: I increased the timeout to 10000 for this to work on safari and iphone. Although I feel this behaviour is flaky and we'd probably have to watch and see if it fails in subsequent PRs. 

I noticed a different issue when trying to update the block properties via the editor which is peculiar to safari and iphone. Apparently parsing, updating and stringifying the json editor's value always results in an unescaped string in safari. 

Screenshot 2022-07-22 at 03 02 20

site/src/pages/[shortname]/blocks/[block-slug].page.tsx
```diff
@@ -281,44 +281,45 @@ const BlockPage: NextPage<BlockPageProps> = ({
...
+              src={blockMetadata.icon ?? undefined}
+            />
           ) : null}
 
           <Box>
-            <Typography
-              sx={{ display: { xs: "flex", md: "unset" } }}
-              variant="bpHeading1"
-              mt={2}
-            >
```
Comment (teenoh) [inline site/src/pages/[shortname]/blocks/[block-slug].page.tsx:303]: Changed this from 
 
```
<h1>
  <img />
  <h1>Block display name</h1>
</h1>
```
to 

```
<div>
<img />
<h1></h1>
</div>
```

site/tests/shared/sleep.ts
teenoh: Thanks! I removed this for now though since I'm not using it

site/tests/universal/blockpage.test.ts
```diff
@@ -0,0 +1,58 @@
+import { expect, test } from "playwright-test-coverage";
+
+const codeBlockUrl = "/@hash/blocks/code";
```
teenoh, line 3: We hardcode this because `blocks-data.json` is not present in the vercel deployment. There's an assertion in the test to confirm if this is correct, so in the event the block url changes, the test will fail and we can always update

site/tests/universal/blockpage.test.ts
teenoh: Created an  [internal task](https://app.asana.com/0/1202542409311090/1202651551651719) to address this

site/tests/integration/blockpage.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  ).toBeVisible();
+  await expect(
+    page.locator("[data-testid='block-schema-tabpanel']"),
+  ).not.toBeEmpty();
+
+  if (isMobile) {
+    await page.locator("text=Preview").click();
+  }
+
+  if (browserName !== "webkit") {
```
kachkaev, line 71: Optional: add @todo & asana link

site/tests/integration/blockpage.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    ).toHaveValue(blockExample.caption!);
+  }
+
+  // check if readme was displayed
+  await expect(page.locator("article")).toBeVisible();
+
+  await expect(page.locator("h2:has-text('Repository')")).toBeVisible();
+
+  await expect(
+    page.locator("a:below(h2:has-text('Repository'))").first(),
```
kachkaev, line 86: 👍

site/tests/universal/blockpage.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  await page
+    .locator("[data-testid='block-properties-tabpanel'] >> textarea")
+    .fill(JSON.stringify({ ...blockProperties, caption: "New caption" }));
+
+  if (isMobile) {
+    await page.locator("text=Preview").click();
+  }
+
+  await expect(blockFrameLocator.locator("input")).toHaveValue("New caption");
```
kachkaev, line 53: 🔥

site/tests/integration/page-footer.test.ts
```diff
@@ -1,44 +1,48 @@
...
     "href",
     "https://github.com/blockprotocol/blockprotocol",
   );
-
   await expect(
     socialLinksSelector.locator("text=Star us on Github"),
   ).toHaveAttribute("href", "https://github.com/blockprotocol/blockprotocol");
+
+  await page.locator("footer >> a:has-text('Contact Us')").click();
+  await expect(page).toHaveURL("/contact");
```
kachkaev, line 32: We don't have to click on every link if this slows down the test. It's possible to leverage await `expect(locator).toHaveAttribute("href", "/path/to/page")`. We jump between pages in header because it's a universal test – we also test routing with it.

This suggestion is optional.

kachkaev (approved):
Great stuff @teenoh! Let's get this change in and iterate over our tests in a follow-up 🚀

site/tests/integration/page-footer.test.ts
teenoh: I think I'll do this then, since jumping between pages here, leads to timeout issues on safari

kachkaev (approved):

site/tests/integration/page-footer.test.ts
teenoh: Updated in [47592f2](https://github.com/blockprotocol/blockprotocol/pull/450/commits/47592f20cb448265c5baa3a94f80671da8609ae7)

site/tests/integration/blockpage.test.ts
teenoh: Good catch. Added in [47592f2](https://github.com/blockprotocol/blockprotocol/pull/450/commits/47592f20cb448265c5baa3a94f80671da8609ae7)

Issue #234, opened 2022-02-24
[SITE] i18n support


nonparibus:
Community members in China have reached out to us about providing translations of the specification and docs site.

We use NextJS on the Block Protocol website. This makes it easy to support internationalization of content, but some set up work is required. See https://nextjs.org/docs/advanced-features/i18n-routing

We're interested in hearing from community members with interest in helping us get the site ready for i18n, or who wish to provide translations.

nonparibus:
We've had separate expressions of interest regarding translation of the site into Spanish.

m1911star:
Hi there, I would like to help with i18n support, but I wonder if I can decide which option to choose, here is the difference：

- sub-path-routing: this will introduce route like `https://blockprotocol.org/zh-CN/docs` 
- domain-routing: this will need multi domains

I think we can choose option 1？

Also, next.js does not support content translation by default, I decide to use https://github.com/isaachinman/next-i18next, any advice?

nonparibus:
> Hi there, I would like to help with i18n support, but I wonder if I can decide which option to choose, here is the difference：
> 
> 
> 
> - sub-path-routing: this will introduce route like `https://blockprotocol.org/zh-CN/docs` 
> 
> - domain-routing: this will need multi domains
> 
> 
> 
> I prefer option 1, any advice?
> 
> 
> 
> Also, next.js does not support content translation by default, I decide to use https://github.com/isaachinman/next-i18next, any advice?

Definite preference for option 1 our side, too. Thanks for your interest in moving this forward. Cc @kachkaev from our BP core team re: implementation guidance.

m1911star:
Hi, I just finished a prototype with i18n and create a pr. https://github.com/blockprotocol/blockprotocol/pull/451

Now I'm fixing docs page 404 error. 

Also looking for help with complete content translations~ sincerely look forward to your advice~

nonparibus:
Thanks, @m1911star! CCing @0xDing and @tsui66 who both have eyes 👀 on this, in case either is in a position to help with content translations in #451.

nonparibus:
CCing @imchell who raised another big FE/UI topic and may also be interested in the translation piece.

PR #451, opened 2022-07-21
Add i18n Support (#234)

m1911star:
- use https://next-intl-docs.vercel.app/ to add i18n support
- routers like `/zh-CN`, default locale is `en-US`，reference https://nextjs.org/docs/advanced-features/i18n-routing#limits-for-the-i18n-config

issues to be fixed

- [ ] when navigate to `/zh-CN/docs`, page not found

image
image


nonparibus:
Designs have been created in Figma to add a language switcher to the footer.

**Internal links** to:

- [footer dropdown](https://www.figma.com/file/Fpmde3E9nJMb4VLkPTYgrM/blockprotocol.org?node-id=284%3A5795)
- [popped menu state](https://www.figma.com/file/Fpmde3E9nJMb4VLkPTYgrM/blockprotocol.org?node-id=3724%3A120106))

**Preview:**

LanguageSwitcher

LanguageSwitcher_Dropdown

yarn.lock
```diff
@@ -3247,7 +3301,7 @@
 
 "@types/react@*", "@types/react@>=16", "@types/react@^17.0.39", "@types/react@^18.0.0", "@types/react@^18.0.15":
   version "18.0.15"
-  resolved "https://registry.yarnpkg.com/@types/react/-/react-18.0.15.tgz#d355644c26832dc27f3e6cbf0c4f4603fc4ab7fe"
+  resolved "https://registry.npmmirror.com/@types/react/-/react-18.0.15.tgz#d355644c26832dc27f3e6cbf0c4f4603fc4ab7fe"
```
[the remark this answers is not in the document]
m1911star: ah, yes, I'll modify later

m1911star:
@nonparibus

> why you elected to go with next-intl, and the alternatives you considered (if any)

I have considered multi choices, like [next-i18next](https://github.com/i18next/next-i18next) [i18next](https://react.i18next.com/) [next-intl](https://next-intl-docs.vercel.app/)

When I integrate next-i18next and i18next, I have troubles with SSR, while I read https://react.i18next.com/latest/ssr, I tried [next-i18next](https://github.com/i18next/next-i18next), since the site project use `type: module`, I can't use next-i18next.config.js ( If we can change this, next-i18next will be the best choice). The last one next-intl is the minimal choice and ease to use.

Also some points listed above
> move all English text to i18n json files
>     decide on which i18n namespaces we need for the UI so that we don’t load all i18n strings on every page
Yes, and I add `common.json` for common text and `index.json` for `index.page`, so I think the filename will be the root namespace. ( document page will introduce a `document.json` )

>     decide on i18n routing (subdomain-based / cookie-based via middleware / path-based) - I personally think path-based makes most sense.

Yeah, I think path-based will be the suitable choice.

>     design & implement language toggle UI - we'll take this action internally when this PR is a bit further along. We'll locate the language-switcher in the BP site footer.

While I dont have permission to access the figma file, I think when we decide the technical solutions, like which library to use, and how to collaborate content translation and how to update content later, also some ci and cd issues, we can implement the ui part.Of course I just propose a solution and fixing issues in document page, it's up to you whether I implement the ui. Thanks~

nonparibus:
@m1911star Thanks for the notes. @kachkaev on our team is a contributor to [next-18next](https://github.com/i18next/next-i18next) so I'm hoping he'll have some thoughts. We'll get back to your shortly.

We're happy to implement the language-switcher UI component our side, in a separate fast-follow PR, once this pull request (and accompanying translations) progress further.

nonparibus:
@m1911star We've discussed as a core team what the "minimum viable translation" we'd be willing to accept for any given language is, and it's that the _homepage itself_, alongside header/footer navigation, should be translated.

Would you be willing to add these translations to get this PR into a state where we can review/think about the corresponding tech and infrastructure changes required?

We're very keen to work with you on making this happen. 😄

m1911star:
> @m1911star We've discussed as a core team what the "minimum viable translation" we'd be willing to accept for any given language is, and it's that the _homepage itself_, alongside header/footer navigation, should be translated.
> 
> Would you be willing to add these translations to get this PR into a state where we can review/think about the corresponding tech and infrastructure changes required?
> 
> **I suggest we defer translations of inner-pages, including the docs and spec for now, as these are still frequently/rapidly changing.**
> 
> We're very keen to work with you on making this happen. 😄

I just came back from holiday, so I will try to implement homepage, alongside header/footer navigation. On my way.

site/src/locales/en-US/common.json
```diff
@@ -2,5 +2,23 @@
...
     "signup": "Sign Up"
-  }
+  },
+  "Search": "Search",
+  "Block Hub": "Block Hub",
+  "Documentation": "Documentation",
+  "Specification": "Specification",
+  "Publish a Block": "Publish a Block",
+  "Contact Us": "Contact Us",
+  "page_header_text": "A POWERFUL NEW PROTOCOL FOR DEVELOPERS",
```
nonparibus, line 12: **Is it possible to selectively apply CSS based on the i18n profile?** Ideally, we would avoid having all-caps in the translated string here (and instead rely on CSS to text-transform to all-uppercase). However, I am wary of applying text-transforms such as this blanketly to all translations for some of the reasons outlined in [this comment](https://github.com/meetuparchive/sassquatch2/issues/144#issuecomment-118958029).

m1911star: Ok, I'll think about it~

PR #452, opened 2022-07-21
Add login flow tests

kachkaev:
In scope

- Dummy email sender (simplified version of what we have in HASH)
- Login flow via verification code + logout
- Login flow via magic link ([excluding webkit](https://app.asana.com/0/1202542409311090/1202652399221616/f)) + logout
- Various edge case interactions in the login modal (incorrect email, close and reopen, etc)

Out of scope
- Various edge case interactions in the verification code screen of the login modal (TBD after sign up flow)
- `/login` page (separate from modal)


site/tests/integration/login-flow.test.ts
```diff
@@ -0,0 +1,201 @@
...
+const emailInputSelector = '[placeholder="claude\\@example\\.com"]';
+const loginButtonSelector = "button[type=submit]:has-text('Log In')";
+const verificationCodeInputSelector = '[placeholder="your-verification-code"]';
+const accountDropdownButtonSelector = '[data-testid="account-dropdown-button"]';
+
+test("login works for an existing user (via verification code)", async ({
+  page,
+  isMobile,
+}) => {
+  await execa("yarn", ["exe", "scripts/seed-db.ts"]);
```
teenoh, line 16: Let's use [test.beforeEach](https://playwright.dev/docs/api/class-test#test-before-each) for this

kachkaev: I can do that in a follow-up 👍

site/tests/integration/login-flow.test.ts
```diff
@@ -0,0 +1,201 @@
...
+  await expect(loginModal.locator("text=Check your inbox")).toBeVisible();
+
+  const magicLink = await readValueFromRecentDummyEmail("Magic Link: ");
+
+  const page2 = await context.newPage();
+  await page2.goto(magicLink);
+
+  const accountDropdownButton2 = page2.locator(accountDropdownButtonSelector);
+  await expect(accountDropdownButton2).toBeVisible();
+  await expect(accountDropdownButton2).toHaveText("A");
```
teenoh, line 102: feels like we'd just need 1 of these... Can't have text "A" if it isn't visible. Can be addressed in a follow up

site/tests/integration/login-flow.test.ts
```diff
@@ -0,0 +1,201 @@
...
+  );
+
+  await expect(accountDropdownPopover).toBeVisible();
+  await expect(accountDropdownPopover).toContainText("Alice@alice");
+  await accountDropdownPopover.locator(`span:has-text("Log Out")`).click();
+
+  await openLoginModal({ page, isMobile });
+
+  await page2.reload();
+  await openLoginModal({ page: page2, isMobile });
```
teenoh, line 120:
```suggestion
```

I don't think these are needed

site/tests/integration/login-flow.test.ts
```diff
@@ -0,0 +1,201 @@
...
+});
+
+test("login modal screen 1 correctly handles interactions", async ({
```
teenoh, line 123: Do we have a "login modal screen 2"?

kachkaev: yep, it's where we ask to enter verification code

site/src/lib/api/dummy-emails.ts
```diff
@@ -0,0 +1,23 @@
+import fs from "fs-extra";
+import path from "node:path";
+
+const dummyEmailsFilePath = "../var/site/dummy-emails.log";
+
+let firstCall = true;
+
+export const sendDummyEmail = async (rows: string[]) => {
```
teenoh, line 8: 🙌

site/tests/integration/login-flow.test.ts
```diff
@@ -0,0 +1,201 @@
...
+  await verificationCodeInput.fill(
+    await readValueFromRecentDummyEmail("Email verification code: "),
+  );
+  await verificationCodeInput.press("Enter");
+
+  const accountDropdownButton = page.locator(accountDropdownButtonSelector);
+  await expect(accountDropdownButton).toBeVisible();
+
+  await accountDropdownButton.click();
+  await expect(accountDropdownButton).toHaveText("A");
```
teenoh, line 42: probably makes sense for this check to happen before the click...

teenoh (approved):
Thanks for working on this! Dropped some comments but they can be addressed in a follow up

site/tests/integration/login-flow.test.ts
kachkaev: `toBeVisible()` checks whether the element is actually visible on the screen as opposed to be just present in DOM:
https://playwright.dev/docs/actionability#visible

I understand that it's unlikely to pass `toHaveText` without `toBeVisible` but I'd still keep it here just in case.

nonparibus:
> Thanks for working on this! Dropped some comments but they can be addressed in a follow up

Why not err on getting thee changes made?

Have we captured the follow-up as a new ticket?

Approaching things like this without at least cross-linking make it much harder to project manage and QA.

site/tests/integration/login-flow.test.ts
kachkaev: If we are able to open login modal, are logged out. Added a comment in https://github.com/blockprotocol/blockprotocol/pull/455

site/tests/integration/login-flow.test.ts
kachkaev: Fair point! Fixed in https://github.com/blockprotocol/blockprotocol/pull/455

site/tests/integration/login-flow.test.ts
kachkaev: Done in https://github.com/blockprotocol/blockprotocol/pull/455

kachkaev:
@nonparibus We have a few new tickets created in Asana, they are mainly to do with discovered bugs in Webkit. I have addressed @teenoh's neat picks in https://github.com/blockprotocol/blockprotocol/pull/455 – they are not worth individual tickets. These tiny enhancements are not worth switching branches or another 10-min CI wait. Iteration over perfection.

nonparibus:
@kachkaev Sure. We can iterate PRs in-line next to provided comments when the tweaks are indeed small, and quick. Default to ASAP ([internal](https://www.notion.so/hashintel/Default-to-ASAP-49b50083be64466cb678551cbf111914)).

If we're not then please x-link to tickets or be explicit that they'll be addressed in followup commits and then close the loop by x-linking, for the aforementioned ease of PM/QA.

PR #457, opened 2022-07-25
Add integration test for sign up flow

kachkaev:


site/tests/integration/sign-up-flow.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  await expect(
+    page.locator('text=This user has already been taken"'),
+  ).not.toBeVisible();
+
+  await preferredNameInput.fill("Alice");
+
+  // @todo Remove when Safari bug is fixed
+  // (main page scrolls down when modal is closed, navbar disappears)
+  test.skip(
```
teenoh, line 83: I assume test.skip should be used if we want to skip the entire test as opposed to a section. I think it's better to use a conditional to handle the remaining parts of the test

```
if (browserName !== "webkit")
{
  ...
}
```

teenoh: This can be addressed in a follow up

teenoh (approved):

site/tests/integration/sign-up-flow.ts
kachkaev: WDYT of using `test.fixme` instead?

teenoh: Yeah we should definitely use that

PR #460, opened 2022-07-26
Update block scripts to use JSX transform

teenoh:
This PR updates `block-scripts` setup to use the updated JSX transform. This makes it possible to use `JSX` in blocks without importing React.

[Relevant material](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)

---
  - https://app.asana.com/0/0/1202402363828486

### DIFF
--- a/.changeset/block-scripts-jsx-transform.md
+++ b/.changeset/block-scripts-jsx-transform.md
@@ -1,0 +1,6 @@
+---
+"block-scripts": patch
+"block-template": patch
+---
+
+Update block-scripts to use JSX transform
--- a/.changeset/block-scripts-jsx-transform.md
+++ b/.changeset/block-scripts-jsx-transform.md
@@ -1,6 +1,0 @@
----
-"block-scripts": patch
-"block-template": patch
----
-
-Update block-scripts to use JSX transform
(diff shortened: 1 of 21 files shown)


nathggns (dismissed):

packages/block-scripts/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "block-scripts",
-  "version": "0.0.13",
+  "version": "0.0.14",
```
kachkaev, line 3: There is no need to bump package versions – this will be done automatically when #424 is merged

.changeset/block-scripts-jsx-transform.md
```diff
@@ -0,0 +1,6 @@
+---
+"block-scripts": patch
+"block-template": patch
```
kachkaev, line 3: What change have we made in block-template?

teenoh: we updated the version number of block scripts used there

kachkaev: Oh IC. This change makes sense then!

packages/block-scripts/package.json
teenoh: Thanks. I'm leaving as is since I will be publishing this version manually

kachkaev (approved):

PR #463, opened 2022-07-26
Fix login modal scroll in Webkit

kachkaev:
[Asana task](https://app.asana.com/0/1200211978612931/1202629133807661/f) (internal)

The bug was discovered while working on https://github.com/blockprotocol/blockprotocol/issues/452

We hide login modal with a fade-out effect. When we press close, _Email address_ field is unfocused and automatically focused again while the modal is fading. This causes an odd scroll jump in Webkit browsers (both in Playwright and on macOS / iOS).

https://user-images.githubusercontent.com/608862/181071370-436b55f5-9e3f-4161-b167-eb205a31b15c.mp4

Because scrolling the page down hides the top menu, `login-flow` test often fails for Safari and iPhone. It can pass, but only if the header menu is pressed very quickly, before it animates away.

This PR marks the email input field as disabled while the modal is being faded. This makes it non-focusable, which fixes the scrolling bug.

I tried reproducing the same bug for the second login screen (with the code) but could not. Hence, the fix is just for the first screen.

### DIFF
--- a/site/src/components/screens/send-login-code-screen.tsx
+++ b/site/src/components/screens/send-login-code-screen.tsx
@@ -17,6 +17,7 @@ import { TextField } from "../text-field";
 import { VerificationCodeInfo } from "./verification-code-screen";
 
 type SendLoginCodeScreenProps = {
+  disabled?: boolean;
   initialEmail?: string;
   onLoginCodeSent: (params: {
     verificationCodeInfo: VerificationCodeInfo;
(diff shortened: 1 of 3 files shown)


site/src/components/screens/send-login-code-screen.tsx
```diff
@@ -27,14 +28,14 @@ type SendLoginCodeScreenProps = {
...
 export const SendLoginCodeScreen: FunctionComponent<
   SendLoginCodeScreenProps
-> = ({ initialEmail, onLoginCodeSent, onClose }) => {
+> = ({ disabled, initialEmail, onLoginCodeSent, onClose }) => {
   const emailInputRef = useRef<HTMLInputElement>(null);
 
   useEffect(() => {
-    if (emailInputRef.current) {
+    if (emailInputRef.current && !disabled) {
       emailInputRef.current.select();
```
teenoh, line 36:
```suggestion
      emailInputRef.current.focus();
```

Tested and still noticed the same behaviour on safari. I think the problem is with `select`. When I switched to using `focus`, I didn't experience the behaviour

teenoh: It also worked without the `disabled` fix when I tested

kachkaev: Does this `useEffect` fire when you press close? It was only seeing it on modal open. Changing to `emailInputRef.current.focus()` did not work for me locally when I tried various solutions yesterday.

kachkaev: https://github.com/blockprotocol/blockprotocol/pull/463/commits/8ceac91650f48f1e0469848c6c199cc5effcbf01

teenoh (approved):

PR #464, opened 2022-07-26
Fix coverage stats

kachkaev:
I was [documenting](https://app.asana.com/0/1200211978612931/1202644137081309/f) integration tests and noticed two issues with reported coverage. This PR fixes them.

- We used `npx kill-port 3000` in CI, which stopped `next start` with SIGKILL instead of SIGINT. Abrupt process termination meant that the runtime coverage was not reported. I have temporarily replaced `yarn kill-port` with `kill -2` (i.e. SIGKILL with SIGINT). We can switch back when https://github.com/tiaanduplessis/kill-port/issues/48 is closed upstream.
  
  ↑ this change **increased** reported code coverage

- We did not have `@istanbuljs/nyc-config-typescript` in `.nycrc.json`, which caused some source files to be missing from the list of all files. In particular, all `site/src/pages/api/**.ts` files were not included into the report. Adding `@istanbuljs/nyc-config-typescript` made the list of files more complete.

  ↑ this change **decreased** reported code coverage

Overall coverage has **decreased**, but this is expected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -226,7 +226,10 @@ jobs:
 
       - name: Stop frontend ## Updates .nyc_output
         if: ${{ success() || failure() }}
-        run: yarn kill-port 3000
+        run: kill -2 $(lsof -t -i:3000)
+        ## TODO: Replace with kill-port when SIGINT is supported as an argument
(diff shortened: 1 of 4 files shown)


teenoh (approved):
Thanks for fixing this 🎉

PR #466, opened 2022-07-27
Add blockprotocol changeset for @types/react bump

CiaranMn:
We upgraded the `@types/react` dependency in `blockprotocol` in #426, but didn't publish a new version of `blockprotocol`. This is leading to type mismatches in https://github.com/hashintel/hash

This PR adds a changeset for `blockprotocol` so we can publish a new version.

### DIFF
--- a/.changeset/blockprotocol-react-types.md
+++ b/.changeset/blockprotocol-react-types.md
@@ -1,0 +1,5 @@
+---
+"blockprotocol": patch
+---
+
+upgrade @types/react dependency


kachkaev (approved):
The changest will point to this PR, but I guess it's fine because it's possible to trace #426 from here.

When we update dependencies in HASH, we can potentially remove `@types/react` from [`package.json` resolutions](https://github.com/hashintel/hash/blob/3eebdd7e0d91364f26ab4ef1995f85dce2add2ed/package.json#L67) (to be checked locally via `yarn install; yarn why @types/react`).

PR #468, opened 2022-07-27
Configure Sentry

kachkaev:
In scope:

- Sentry setup for Production and Preview Vercel deployments
- Inclusion of logged in users to Sentry scope
- Exception capturing for API routes
- Telemetry section in README

Caveats:

- Enabling Sentry for Preview deployments might clutter [Sentry storage with source maps](https://sentry.io/settings/hashintel/projects/blockprotocol-site/source-maps/). If we hit the limits, we can disable `.*SENTRY.*` variables in Preview environment [on Vercel](https://vercel.com/hashintel/blockprotocol/settings/environment-variables).

- I was unable to use `withSentry()` wrapper for API routes because this method is incompatible with [next-connect](https://www.npmjs.com/package/next-connect). The solution I’ve ended up with is inspired by the [Next.js Sentry example](https://github.com/vercel/next.js/blob/cd3e054f14ce38f4ff57c727a997da2a6e1d05dd/examples/with-sentry/pages/api/test4.js#L7-L12) and a [`next-connect` issue](https://github.com/hoangvvo/next-connect/issues/133).

- I was unable to find direct links from Sentry pages to source code on GitHub. Could be an issue with Sentry ←→ GitHub integration which I don’t have access to.

---
  - https://app.asana.com/0/0/1202670617765417


CiaranMn (approved):
Looks good, thanks @kachkaev

PR #469, opened 2022-07-27
Fix typo in changeset release action

CiaranMn:
The changeset release action has a typo in it which means it does not work - see [here](https://github.com/blockprotocol/blockprotocol/runs/7541374647?check_suite_focus=true).


nonparibus (approved):

kachkaev:
Oops, thanks for fixing!

PR #470, opened 2022-07-27
FAQ subsection titles appearing as slugs in sidebar

luisbettencourt:
Questions with styles in the FAQ section were incorrectly displaying a slug in the sidebar (see `why-arent-the-default-schemaorg-definitions-sufficient`).

image

This PR fixes the parsing of FAQ sections in the docs to correctly handle anchors and HTML styled titles.

image

[Task](https://app.asana.com/0/1200211978612931/1202579651285633/f)

### DIFF
--- a/site/src/util/mdx-utils.tsx
+++ b/site/src/util/mdx-utils.tsx
@@ -17,10 +17,11 @@ type Node = {
 };
 
 type TextNode = {
-  value: string;
+  value: string | TextNode;
 } & Node;
 
-const isTextNode = (node: Node): node is TextNode => "value" in node;
+const isTextNode = (node: Node | string): node is TextNode =>
+  typeof node !== "string" && "value" in node;
 
 type Parent = {
   children: (TextNode | Node)[];
@@ -31,6 +32,7 @@ const isParent = (node: Node): node is Parent => "children" in node;
 type Heading = {
   type: "heading";
   depth: number;
+  anchor?: string;
 } & Parent;
 
 const isHeading = (node: Node): node is Heading => node.type === "heading";
@@ -68,13 +70,13 @@ const getHeadingsFromParent = (parent: Parent): Heading[] =>
           type: "heading",
           /** @todo: don't assume that FAQ accordions are always headings at depth 3 */
           depth: 3,
+          anchor: child.attributes.find(({ name }) => name === "anchor")?.value,
           children: [
             {
               type: "text",
               value:
-                child.attributes.find(
-                  ({ name }) => name === "anchor" || name === "question",
-                )?.value ?? "Unknown",
+                child.attributes.find(({ name }) => name === "question")
+                  ?.value ?? "Unknown",
             },
           ],
         };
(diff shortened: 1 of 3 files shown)


site/src/util/mdx-utils.tsx
```diff
@@ -194,9 +204,11 @@ export const getPage = (params: {
     sections: headings.reduce<SiteMapPageSection[]>((prev, currentHeading) => {
       const newSection = {
         title: getVisibleText(currentHeading),
-        anchor: slugify(getFullText(currentHeading), {
-          lower: true,
-        }),
+        anchor:
+          currentHeading.anchor ||
```
CiaranMn, line 208: It feels like it would be better and less user-error prone to get rid of the need to add custom `anchor` to FAQ entries, and instead just strip out any HTML from FAQ titles here when setting the anchor. But I appreciate this is a pre-existing problem

CiaranMn (dismissed):
Thanks @luisbettencourt – this fixes the bug.

I noticed a couple of things when testing that I've created follow-up tasks for – these are existing issues in main:
1. We need to construct dynamic <title> content differently - see [this discussion](https://github.com/vercel/next.js/discussions/38256) [[internal task](https://app.asana.com/0/1200211978612931/1202684108421153/f)]
2. A hydration error when visiting [this route](https://blockprotocol-g3iv6kvg5.stage.hash.ai/docs/faq#why-arent-the-default-schemaorg-definitions-sufficient) [[internal task](https://app.asana.com/0/1200211978612931/1202684108421158/f)] – slightly annoying in local dev because it causes a crash

site/src/util/mdx-utils.tsx
luisbettencourt: Done in d6e89df82491d6b0f722760175b4f8083a20afab. Let me know if something is missing

site/src/util/mdx-utils.tsx
```diff
@@ -68,13 +71,13 @@ const getHeadingsFromParent = (parent: Parent): Heading[] =>
           type: "heading",
           /** @todo: don't assume that FAQ accordions are always headings at depth 3 */
           depth: 3,
+          anchor: child.attributes.find(({ name }) => name === "anchor")?.value,
```
CiaranMn, line 74: These don't exist any more

site/src/util/mdx-utils.tsx
```diff
@@ -31,6 +33,7 @@ const isParent = (node: Node): node is Parent => "children" in node;
 type Heading = {
   type: "heading";
   depth: number;
+  anchor?: string;
```
CiaranMn, line 36: You no longer need this

site/src/util/mdx-utils.tsx
CiaranMn: Not missing, but there are still provisions for an `anchor` property on headings that no longer exists, including here

CiaranMn (commented):
Thanks @luisbettencourt – good to get rid of the need for the custom `anchor`. There are still references to it in the code that now can be removed

site/src/util/mdx-utils.tsx
luisbettencourt: Removed in 90b9a7c2d74f6b9a3b894ae5913e9a3585295dba

site/src/util/mdx-utils.tsx
luisbettencourt: Removed in 90b9a7c2d74f6b9a3b894ae5913e9a3585295dba

CiaranMn (approved):
Great!

PR #471, opened 2022-07-29
Add the Shuffle block to the Hub

luisbettencourt:
This PR adds the Shuffle block to the BP hub

### DIFF
--- a/hub/@hash/shuffle.json
+++ b/hub/@hash/shuffle.json
@@ -1,0 +1,7 @@
+{
+  "repository": "https://github.com/hashintel/hash.git",
+  "commit": "98818b692742dbadd77eaf824c6d51bb8dec7998",
+
+  "workspace": "@hashintel/block-shuffle",
+  "distDir": "dist"
+}
--- a/hub/@hash/shuffle.json
+++ b/hub/@hash/shuffle.json
@@ -2,6 +2,6 @@
   "repository": "https://github.com/hashintel/hash.git",
   "commit": "98818b692742dbadd77eaf824c6d51bb8dec7998",
 
-  "workspace": "@hashintel/block-shuffle",
+  "workspace": "@hashintel/shuffle",
   "distDir": "dist"
 }
--- a/hub/@hash/shuffle.json
+++ b/hub/@hash/shuffle.json
@@ -1,7 +1,7 @@
 {
   "repository": "https://github.com/hashintel/hash.git",
-  "commit": "98818b692742dbadd77eaf824c6d51bb8dec7998",
+  "commit": "93b643e3f17d16c3371c159dc865f234138faee3",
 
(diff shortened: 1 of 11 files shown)


hub/@hash/shuffle.json
```diff
@@ -0,0 +1,7 @@
+{
+  "repository": "https://github.com/hashintel/hash.git",
+  "commit": "98818b692742dbadd77eaf824c6d51bb8dec7998",
+
+  "workspace": "@hashintel/block-shuffle",
```
benwerner01, line 5: @luisbettencourt from the [deployment build error](https://vercel.com/hashintel/blockprotocol/6z9a9RRn4hmCK5L3XGefYr53LGNB), looks like `@hashintel/block-shuffle` needs to be added as a yarn workspace to the [`hashintel/hash` repo](https://github.com/hashintel/hash)

benwerner01: Actually looks like you just need to rename the workspace from `@hashintel/shuffle` to `@hashintel/block-shuffle`

luisbettencourt: Yes, my bad. Was just waiting for the CI to confirm this

hub/@hash/shuffle.json
```diff
@@ -0,0 +1,7 @@
+{
+  "repository": "https://github.com/hashintel/hash.git",
```
benwerner01, line 2: I think the `examples` field in the package.json of this block should be changed from:
```TypeScript
    "examples": [
      [
        "Alice",
        "Bob",
        "Charlie"
      ]
    ],
```
to 
```TypeScript
    "examples": [
      {
        "items": [
          "Alice",
          "Bob",
          "Charlie"
        ]
      }
    ],
```

CiaranMn (commented):
Thanks @luisbettencourt – I think there are a couple of fixes needed in the block source code, both of which I should've realised before:
1. The `examples` need updating – if you have a look at the [preview page](https://blockprotocol-ft9r2j4o2.stage.hash.ai/@hash/blocks/shuffle), you can see it's trying to read the array as an object (because the block properties are an object) - the first entry in the examples array should in fact be:
```json
{
  "items": [
    "Alice",
    "Bob",
    "Charlie"
  ]
}
```
2. We should name the workspace `@hashintel/block-shuffle` while we're doing (1), since it'll make it consistent with the other block workspaces in that repo

hub/@hash/shuffle.json
benwerner01: Otherwise the array is interpreted as an object, meaning the block doesn't correctly render the example, and the "Block Properties" displays misleading fields that don't match the block schema:

image

luisbettencourt:
Thank you. Opened https://github.com/hashintel/hash/pull/881 to address the changes requested.

PR #474, opened 2022-08-01
Fix flaky block page test on firefox

teenoh:


kachkaev (approved):
Makes sense, let’s try it out! 👍

kachkaev:
@teenoh I just saw a failed test despite a 10-second delay: https://github.com/blockprotocol/blockprotocol/runs/7612388851?check_suite_focus=true#step:9:15

Looks like we need to explore this further.

PR #484, opened 2022-08-04
Add read-only mode to `mock-block-dock`

teenoh:
This PR does the following
- Updates mock-block-dock to use jsx transform
- Updates mock-block-block to allow testing blocks in readonly mode
- Updates the test example blocks to render appropriate UI based on readonly property received 

https://user-images.githubusercontent.com/16928688/182774016-f35db549-99a4-492b-b9b8-200246707194.mov

---
  - https://app.asana.com/0/0/1202643260631377


packages/block-template/templates/custom-element/package.json
```diff
@@ -25,12 +25,12 @@
...
   },
   "dependencies": {
-    "@blockprotocol/graph": "0.0.11",
+    "@blockprotocol/graph": "0.0.12",
     "lit": "^2.2.5"
   },
   "devDependencies": {
     "block-scripts": "0.0.14",
-    "mock-block-dock": "0.0.17",
+    "mock-block-dock": "0.0.18-canary.1",
```
CiaranMn, line 33: this can be `0.0.18`

packages/mock-block-dock/src/use-mock-block-props/use-mock-datastore.ts
```diff
@@ -66,6 +67,17 @@ export const useMockDatastore = (
...
+        if (readonly) {
+          return {
+            errors: [
+              {
+                code: "FORBIDDEN",
+                message: "Operation can't be carried out in readonly mode",
+              },
+            ],
+          };
+        }
```
CiaranMn, line 79: You can save a bit of space by declaring the readonly error at the top of the file and then referencing it, since it is the same in every function, e.g.
```typescript
const readonlyErrorReturn: {
  errors: [{ code: "FORBIDDEN"; message: string }];
} = {
  errors: [
    {
      code: "FORBIDDEN",
      message: "Operation can't be carried out in readonly mode",
    },
  ],
};

...

if (readonly) {
  return readonlyErrorReturn;
}

CiaranMn (commented):
Thanks @teenoh, good stuff! Couple of small suggestsions

packages/block-template/templates/custom-element/package.json
teenoh: Thanks. I planned to update once approved. Will update now

packages/mock-block-dock/src/use-mock-block-props/use-mock-datastore.ts
teenoh: Thanks! Added in [f020a5c](https://github.com/blockprotocol/blockprotocol/pull/484/commits/f020a5c352818f6861429fe3955c3ab80ee3433a)

PR #480, opened 2022-08-02
Split block-template into individual packages

kachkaev:
This PR replaces a single `block-template` package with three packages: `block-template-html`, `block-template-react` and `block-template-custom-element`. `create-block-app` and tests are updated accordingly.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202438000302957

PR #488, opened 2022-08-08
Include read-only mode checkbox in `mock-block-dock` debug UI

teenoh:


packages/mock-block-dock/dev/dev-app.tsx
```diff
@@ -33,9 +33,9 @@ const DevApp: FunctionComponent = () => {
           Increment Entity ID
         </button>
 
-        <label htmlFor="reaodnly-input" style={{ marginLeft: 20 }}>
+        <label htmlFor="external-readonly-input" style={{ marginLeft: 20 }}>
```
CiaranMn, line 36: Do you need this `htmlFor`? The label is wrapping the input, which means it should be identified as belonging to the input automatically. 

It also avoids the `id` potentially clashing with an element in the block.

The same comment applies to the other readonly input.

teenoh: Added the `htmlFor` to fix an eslint error, which the label wrapping the input should actually fix... Will remove

teenoh:
> I suggest rebasing this PR when https://github.com/blockprotocol/blockprotocol/pull/480 is merged. We'll be able to use .changeset/*.md instead of a manual version bump.
Will do. 

> Am I right that we now have an incoming value of readonly and a checkbox that overrides it? Can this introduce two incompatible sources of truth that end up clashing with each other?
Yeah. Users can trigger readonly mode in two ways

- by passing a read-only prop to mock-block-dock
- through the read-only checkbox in `mock-block-dock`'s  debug mode

The readonly checkbox only comes up when `mock-block-dock` is in debug mode. It saves user from having to wire up a read-only toggle themselves when debugging the block in that state. 

Ideally users should only use one of the triggers at a time. In scenario's they try to use both, the read-only prop passed to `mock-block-dock` takes higher precedence

packages/mock-block-dock/dev/dev-app.tsx
teenoh: Updated in [828bd21](https://github.com/blockprotocol/blockprotocol/pull/488/commits/828bd2197ad5f074ec53c2f2491efbdc3b2b42d0)

kachkaev (dismissed):
> Ideally users should only use one of the triggers at a time. In scenario's they try to use both, the read-only prop passed to mock-block-dock takes higher precedence

I see. I don’t have an up-to-date model of `mock-block-dock` in my head, but there is a chance that we’re squeezing multiple concerns into one element of a system. Could be worth thinking about this case again in the future and check if we can compose multiple elements together to avoid ad-hoc overrides.

Just one small suggestion re ESLint comments.

CiaranMn (dismissed):

packages/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -181,6 +193,15 @@ export const MockBlockDock: FunctionComponent<MockBlockDockProps> = ({
     <div style={{ fontFamily: "sans-serif" }}>
       <div style={{ marginBottom: 30 }}>
         <h3 style={{ marginTop: 0, marginBottom: 3 }}>Block</h3>
+        {/* eslint-disable-next-line jsx-a11y/label-has-associated-control */}
```
kachkaev, line 196: 👀

kachkaev:
```suggestion
        {/* eslint-disable-next-line jsx-a11y/label-has-associated-control -- https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/869 */}
```

kachkaev (approved):

PR #492, opened 2022-08-09
Fix multiple re-renders bug in `mock-block-dock`

teenoh:
A bug that was introduced in #488 where `mock-block-dock` has infinite re-renders when the `readonly` prop is undefined. This PR fixes that

### DIFF
--- a/.changeset/mock-block-dock-fix.md
+++ b/.changeset/mock-block-dock-fix.md
@@ -1,0 +1,5 @@
+---
+mock-block-dock: patch
+---
+
+Fix bug introduced in 0.19 that comes up when `readonly` prop is undefined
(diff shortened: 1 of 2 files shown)


.changeset/mock-block-dock-fix.md
```diff
@@ -0,0 +1,5 @@
+---
+mock-block-dock: patch
+---
+
+Fix bug introduced in 0.19 that comes up when `readonly` prop is undefined
```
kachkaev, line 5:
```suggestion
Fix bug introduced in 0.0.19 that comes up when `readonly` prop is undefined
```

kachkaev (approved):

PR #495, opened 2022-08-10
Update changesets

kachkaev:
Follows https://github.com/blockprotocol/blockprotocol/pull/491 and https://github.com/blockprotocol/blockprotocol/pull/493. Hopefully saves us from making some manual edits to releases.

The problem with unwanted bumps of our private `@blockprotocol/site` will likely remain. I added it to the `ignore` lists, but it seems necessary to remove `version` field from `package.json` instead. This will be possible when we upgrade to Yarn 3, which is in turn blocked by https://github.com/dependabot/dependabot-core/issues/1297.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -11,6 +11,7 @@
   "updateInternalDependencies": "patch",
   "ignore": [],
   "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
+    "updateInternalDependents": "always",
     "onlyUpdatePeerDependentsWhenOutOfRange": true
   }
 }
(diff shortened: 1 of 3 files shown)


.changeset/config.json
```diff
@@ -11,6 +11,7 @@
   "updateInternalDependencies": "patch",
   "ignore": [],
   "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
+    "updateInternalDependents": "always",
```
kachkaev, line 14: Might help bump dependent packages if there dependencies have changed ([docs](https://github.com/changesets/changesets/blob/e4204820cdb93d22a3c7a27f452a0ab30db6369a/docs/experimental-options.md#updateinternaldependents-type-out-of-range--always)).

teenoh (approved):

PR #498, opened 2022-08-12
Don't prefetch /discord links

CiaranMn:
In production, NextJS will prefetch the page for any `<Link>` in the viewport to an internal URL.

We have a route called `/discord` that looks to NextJS like an internal URL, but is actually something that is only handled server-side (and they all open in a new window for that reason).

This causes NextJS to attempt to prefetch a page that doesn't exist, littering the console with errors (since there's a Discord link in the footer on every page). 

This PR fixes that by disabling `prefetch` for the Discord links.

### To test

Try hovering your mouse over the Discord link in the footer on https://blockprotocol.org vs the preview build when available (if testing locally, must build and start a production build).


kachkaev (approved):
👏

PR #502, opened 2022-08-16
Update `passport` to 0.6.0

kachkaev:
Blocked by https://github.com/jaredhanson/passport/issues/904

---
  - https://app.asana.com/0/0/1202807084032216


kachkaev:
I don't know enough about passport to confirm if we need that patch or not. We could add it nevertheless next sprint if you think that it's worthy.

PR #503, opened 2022-08-17
Add Type System Package

Alfred-Mountfield:

This introduces a new package (set of packages) for the Type System as part of the implementation of #418. The definition of the ontology types, their builders, and their validation logic, will be written in Rust. This Rust is then compiled to WASM and turned into NPM packages that can be consumed in Web or NodeJS contexts. This should allow us to have consistent definitions of the Type System elements which we can use within Blocks (a Type Editor block for example) as well as in backend components written in both Node and Rust.

It should be possible in the future to generate other bindings to these implementations, potentially Python. It may also be possible to auto-generate JSON schema from the struct definitions themselves, which means that we would be able to keep the Block Protocol spec in sync with our libraries and reference implementations.

## 🔗 Related links

## 🔍 What does this change?

- Adds `packages/type-system`
- Adds a Rust crate for Type System definitions within `packages/type-system/rust`
  - Adds a `cargo-make` setup to manage the various build and testing scripts
  - Adds a `wasm-pack` configuration to turn a `wasm_bindgen` configured crate into an NPM package with TypeScript bindings
  - Adds definitions and validation logic for `BaseURI` and `VersionedUri` of Type System types
- Adds a TypeScript folder to contain the generated NPM packages, and
  - `packages/ts/integration` which contains Jest integration tests which tests the generated NPM packages

- READMEs have been added for the package itself. We likely want to improve discoverability of this package, and add usage instructions for how to import and use the NPM packages.

## ⚠️ Known issues

- `wasm-pack` has a `bundler` target which could allow us to create an ES module. Unfortunately jest is unable to correctly load that when it's made, and nextJS and `create-block-app` both struggle with handling the included `.wasm` asset when bundling the ES module as they mess up the loading mechanism and path. Because of this the current PR generates two different packages that can be used in either Web or NodeJS contexts.
- When passing a `number` through on the TypeScript side to Rust, if you pass a non-integer it gets floored without erroring. (So if you pass `2.3` you end up with the version `2`). A TODO has been added

## 🐾 Next steps

  - Data Types
  - Property Types
  - Entity Types
  - Link Types

## 🛡 What tests cover this?

- Rust unit tests _will_ be added to also check the crate
- There is a setup for WASM headless browser tests but these are currently not utilised, a testing strategy will need to be developed

## ❓ How to test this?

1.  Follow the READMEs

## 📹 Demo

image
Pretty boring demo but this is a VersionedUri being created within a Block


packages/type-system/ts/test-block/src/app.tsx
```diff
@@ -1,75 +0,0 @@
...
-    return <div>Loading...</div>;
-  } else {
-    return (
-      <div className={styles.block} ref={blockRootRef}>
-        <h1>URI: {uri?.toString()}</h1>
-        <button
-          onClick={async () => {
-            if (!uri) {
-              const baseUri = new BaseUri(BASE_URI);
-              uri = new VersionedUri(baseUri, getRandomInt(1000));
```
Alfred-Mountfield: This shows an example of the library working within a block. I deleted it as I assumed we don't want a random stray block sitting around without any automated testing as part of it.

packages/type-system/rust/src/ontology/uri/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+impl FromStr for VersionedUri {
+    type Err = ParseVersionedUriError;
+
+    fn from_str(uri: &str) -> Result<Self, ParseVersionedUriError> {
+        let (base_uri, version) = uri.rsplit_once("/v/").ok_or(ParseVersionedUriError)?;
+
+        // TODO: better error handling
```
Alfred-Mountfield, line 58: At the moment I'm swallowing all of the [useful errors](https://docs.rs/url/latest/url/enum.ParseError.html) from the `Url` crate. I thought it was out of scope of this PR to manage how to expose all of those to TS and how we'd potentially want to handle them.

packages/type-system/rust/src/ontology/uri/native.rs
```diff
@@ -0,0 +1,34 @@
+use url::Url;
+
+use crate::ontology::uri::{error::ParseBaseUriError, BaseUri, VersionedUri};
+
+impl BaseUri {
+    pub fn new(uri: &str) -> Result<BaseUri, ParseBaseUriError> {
+        // TODO: Propagate more useful errors
```
Alfred-Mountfield, line 7: See comment above about swallowing [url errors](https://docs.rs/url/latest/url/enum.ParseError.html)

packages/type-system/rust/src/ontology/uri/native.rs
```diff
@@ -0,0 +1,34 @@
...
+        // TODO: This attempts to parse the string _into_ a valid URL. Perhaps we want to enforce
+        //  that the string is valid (by checking the output is equal to the input). An example:
+        //  "file://loc%61lhost/" is turned into "file:///"
```
Alfred-Mountfield, line 10: This is quite strange behaviour but seems somewhat consistent with the reference implementation of the actual [URL spec](https://url.spec.whatwg.org/). I'm not sure what we want to do here because it does helpfully remove stray slashes and standardises some URLs, but at the same time the example I gave here seems like something we'd probably want errors for (things that _are_ considered errors by the library are [here](https://docs.rs/url/latest/url/enum.ParseError.html)).

packages/type-system/rust/src/ontology/uri/wasm.rs
```diff
@@ -0,0 +1,69 @@
...
+use wasm_bindgen::prelude::*;
+
+use crate::ontology::uri::{error::ParseBaseUriError, BaseUri, VersionedUri};
+
+#[wasm_bindgen]
+impl BaseUri {
+    #[wasm_bindgen(constructor)]
+    pub fn new(uri: &str) -> Result<BaseUri, ParseBaseUriError> {
+        // TODO: Propagate more useful errors
+        // TODO: This attempts to parse the string _into_ a valid URL. Perhaps we want to enforce
```
Alfred-Mountfield, line 11: See comments above

packages/type-system/rust/tests/web.rs
```diff
@@ -0,0 +1,10 @@
+//! Test suite for the Web and headless browsers.
+
+#![cfg(target_arch = "wasm32")]
+
+extern crate wasm_bindgen_test;
+use wasm_bindgen_test::*;
+
+wasm_bindgen_test_configure!(run_in_browser);
+
+// TODO: Decide which tests to write here
```
Alfred-Mountfield, line 10: So `wasm-pack` has this nice feature where it will automatically download a driver for a browser on your system and will compile the module into WASM and actually run these tests (written normally in Rust) inside your browser. That seems quite helpful but I think it basically runs it inside WASM, so it's not testing the bindings, which is why I ended up going with the Jest setup. Perhaps there are still some tests we'd like to add here though.

packages/type-system/ts/README.md
```diff
@@ -0,0 +1,18 @@
+# Type-System TypeScript packages
+
+This folder contains
+- [type-system-node](./type-system-node) (_Autogenerated_) - The Type System package, built to target NodeJS
```
Alfred-Mountfield, line 4: I'm not sure if we want to commit these to version control

packages/type-system/ts/integration/src/uri.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    expect(uri.toJSON()).toBe(expected);
+  };
+
+  test("succeeds on valid version number", () => {
+    testVersionedUri(0, "http://example.com/v/0");
+    testVersionedUri(1, "http://example.com/v/1");
+    testVersionedUri(20.0, "http://example.com/v/20");
+  });
+
+  // @todo: this test ideally should fail, but currently the number is Math.floored into an integer silently
```
Alfred-Mountfield, line 66: Just highlighting this as a current limitation

packages/type-system/rust/build-wasm.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import Dict
+import logging
+
+
+def run_wasm_pack(target, out_dir: Path):
+    logging.debug("Running wasm-pack, targeting %s, for %s", target, out_dir)
+    subprocess.run(["wasm-pack", "build", "--target", target, "--out-dir", str(out_dir), "--out-name", "index", "--scope", "blockprotocol"], check=True)
+
+
+def patch_package_json(file: Path, unique_patch: Dict):
```
Alfred-Mountfield, line 13: There aren't great configuration options for the generated `package.json`s and if we're publishing to NPM I assumed we'd want to patch in some content. I wrote this script to setup the infra to do so but I don't really know what we want to patch in.

packages/type-system/ts/integration/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@blockprotocol/type-system-integration",
+  "version": "0.0.0",
+  "private": true,
+  "description": "Integration tests for the Type System package",
+  "scripts": {
+    "test": "jest"
+  },
+  "dependencies": {
+    "@blockprotocol/type-system-node": "0.0.0"
```
Alfred-Mountfield, line 10: @kachkaev CI seems to be failing because of these
```
error Couldn't find package "@blockprotocol/type-system-node@0.0.0" required by "@blockprotocol/type-system-integration@0.0.0" on the "npm" registry.
```

Should I swap to the `file:/..` syntax? Or is there something else we can do

packages/type-system/ts/README.md
Alfred-Mountfield: (it's worth noting that these are built packages, not plain source code that needs transpiling)

If we don't commit them then I think we'll need to figure out a publishing mechanism before we're able to use these easily in a block. If we do commit it then I think it should be simple to install them by just referencing their link in git

packages/@blockprotocol/type-system-node/tests/uri.test.ts
```diff
@@ -0,0 +1,72 @@
...
+describe("BaseURI Class", () => {
+  const testBaseUri = (input: string, expected: string) => {
+    const uri = new BaseUri(input);
+
+    expect(uri.toString()).toBe(expected);
+    expect(uri.toJSON()).toBe(expected);
+  };
+
+  test("succeeds on valid URIs", () => {
+    testBaseUri("http://example.com/", "http://example.com/");
```
kachkaev, line 12: We can try using `test.each` / `it.each` here instead

packages/type-system/ts/integration/package.json
Alfred-Mountfield: This has been solved by adding the packages to version control

packages/type-system/ts/README.md
Alfred-Mountfield: We've decided it makes sense to add them to version control. When we come to adding CI checks we'll need to check for a diff between the generated content and the committed one.

crates/type-system/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+name = "type_system"
+crate-type = ["cdylib", "rlib"]
+
+[features]
+default = ["console_error_panic_hook"]
+
+[dependencies]
+serde = { version = "1.0.137", features = ["derive"] }
+thiserror = "1.0.32"
+url = "2.2.2"
```
Alfred-Mountfield, line 19: Turns out URL is adding a _**lot**_ (~300-400kb) to the size of the WASM binary. This is a [known issue](https://github.com/servo/rust-url/issues/557) but I don't think they're planning on addressing it.

My current thinking is we go with it until it seems like the size is a problem, at which point we could write a lightweight URL thing that does what we need from it, or we could conditionally [utilise the browser's implementation](https://docs.rs/web-sys/0.3.45/web_sys/struct.Url.html) of `url` in a wasm environment.

Either way we have options and I don't think we need to worry right now.

crates/type-system/build-wasm.ts
```diff
@@ -0,0 +1,79 @@
+import * as path from "node:path";
```
kachkaev, line 1: Moving this file to `crates/type-system/scripts/build-wasm.ts` should fix most of ESLint errors. You can also run `yarn fix` locally to address a few. We can look at the remaining ones afterwards.

.eslintignore
```diff
@@ -62,3 +62,5 @@ site/public/blocks
 site/public/schemas
 site/site-map.json
 site/test-results
+packages/@blockprotocol/type-system-node/src
+packages/@blockprotocol/type-system-web/src
```
kachkaev, line 66: We can copy this to `.prettierignore` too

crates/type-system/.gitignore
```diff
@@ -0,0 +1,6 @@
+/target
+**/*.rs.bk
+Cargo.lock
+bin/
+pkg/
+wasm-pack.log
```
kachkaev, line 6: Can we hoist this to top-level `.gitignore`?

crates/type-system/README.md
```diff
@@ -0,0 +1,19 @@
+# The Block Protocol Type-System
+
+[//]: # (TODO: Introduction)
+
+## Requirements
+- [Rust](https://www.rust-lang.org/tools/install)
```
kachkaev, line 6: Try running `yarn fix:markdownlint` and `yarn fix:prettier` from repo root to fix some formatting issues in this README.

crates/type-system/.gitignore
Alfred-Mountfield: That's been done https://github.com/blockprotocol/blockprotocol/pull/503/commits/9d0cf95cc83e95a1c2ed20e4540796c417207b3b

Let me know if that's not how you intended it to be

.eslintignore
Alfred-Mountfield: Updated the ignore files 
https://github.com/blockprotocol/blockprotocol/pull/503/commits/ac445d6c1b9193d089b951d4da434febc60011fd

crates/type-system/build-wasm.ts
Alfred-Mountfield: That was done https://github.com/blockprotocol/blockprotocol/pull/503/commits/3891af587f3efc63e1fa992637fe8056c3f0fb62

Thanks :)

crates/type-system/README.md
Alfred-Mountfield: Was holding off until I figured out the appropriate ignores.

Applied the fix https://github.com/blockprotocol/blockprotocol/pull/503/commits/6dbd20e25655eb5eaa812509cb81a2ba38e36126

(and removed unnecessary README)

packages/@blockprotocol/type-system-node/tests/uri.test.ts
```diff
@@ -0,0 +1,51 @@
+import { BaseUri, VersionedUri } from "@blockprotocol/type-system-node";
```
kachkaev, line 1: Might be fixable via

```suggestion
import { BaseUri, VersionedUri } from "..";
```

packages/@blockprotocol/type-system-node/tests/uri.test.ts
```diff
@@ -0,0 +1,51 @@
...
+
+describe("BaseURI Class", () => {
```
kachkaev, line 3: `'describe' is not defined` can be fixable via top-level `.eslntrc`. Try finding `overrides` and adding this item there:

```
    {
      files: ["packages/**/*.test.{j,t}s{x,}"],
      env: {
        jest: true,
        node: true,
      },
    },
```

Alfred-Mountfield: Thanks! That was stumping me

packages/@blockprotocol/type-system-node/tests/uri.test.ts
Alfred-Mountfield: As discussed offline, this was a misguided attempt to test the `package.json`.
Changed in https://github.com/blockprotocol/blockprotocol/pull/503/commits/d93057ed98481477f6feeee5c1072c988dc578ca

packages/type-system/rust/build-wasm.py
Alfred-Mountfield: This has been made redundant

packages/@blockprotocol/type-system-node/src/index_bg.wasm.d.ts
```diff
@@ -0,0 +1,19 @@
+/* tslint:disable */
```
thehabbos007, line 1: Should we add `.gitattributes` for the auto-generated files similarly to what we did for #928 (this goes for the other package as well)?

kachkaev: We can yes. What do we get from doing so in this case?

crates/type-system/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+# all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
+# code size when deploying.
+console_error_panic_hook = { version = "0.1.6", optional = true }
+
+# `wee_alloc` is a tiny allocator for wasm that is only ~1K in code size
+# compared to the default allocator's ~10K. It is slower than the default
+# allocator, however.
+#
+# Unfortunately, `wee_alloc` requires nightly Rust when targeting wasm for now.
+wee_alloc = { version = "0.4.5", optional = true }
```
TimDiekmann, line 33: Given the size of the binary is it worth pulling an external allocator for this?

crates/type-system/rustfmt.toml
```diff
@@ -0,0 +1,27 @@
+# General
```
TimDiekmann, line 1: Is this a 1:1 copy from hashintel/hash?

crates/type-system/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+#![feature(lint_reasons)]
```
TimDiekmann, line 1: Do we want to pull in the same lints as for the HASH backend?

crates/type-system/src/lib.rs
```diff
@@ -0,0 +1,11 @@
...
+mod ontology;
+mod utils;
+
+// When the `wee_alloc` feature is enabled, use `wee_alloc` as the global
+// allocator.
```
TimDiekmann, line 6: Not really a fan of this comment as this is exactly what is written in the code below

crates/type-system/Cargo.toml
```diff
@@ -0,0 +1,40 @@
...
+crate-type = ["cdylib", "rlib"]
+
+[features]
+default = ["console_error_panic_hook"]
+
+[dependencies]
+serde = { version = "1.0.137", features = ["derive"] }
+thiserror = "1.0.32"
+url = "2.2.2"
+wasm-bindgen = { version = "0.2.63", features = ["serde-serialize"] }
```
TimDiekmann, line 20: I think it's possible to only pull it for the wasm target:

```toml
[target.'cfg(target_arch = "wasm32")'.dependencies]
wasm-bindgen = { version = "0.2.63", features = ["serde-serialize"] }
```

crates/type-system/src/ontology/uri/wasm.rs
```diff
@@ -0,0 +1,18 @@
...
+use crate::ontology::uri::{error::ParseBaseUriError, BaseUri, VersionedUri};
+use crate::uri::error::ParseVersionedUriError;
+
+#[wasm_bindgen(js_name = parseBaseUri)]
+pub fn parse_base_uri(uri: &str) -> Result<String, ParseBaseUriError> {
+    let url = BaseUri::new(uri)?;
+    Ok(url.to_string())
+}
+
+#[wasm_bindgen(js_name = isValidVersionedUri)]
```
TimDiekmann, line 21: Is there something like `toCamelCase`?

TimDiekmann (changes requested):

crates/type-system/rustfmt.toml
Alfred-Mountfield: It is

crates/type-system/Cargo.toml
Alfred-Mountfield: I think so, yes. We're very likely to address the binary size problems from `url` and generally speaking a lot of people seem to use `wee_alloc` for WASM so I'm happy to stick with convention.

crates/type-system/src/lib.rs
Alfred-Mountfield: Happy to delete it, the comment was from the boilerplate when the project was generated

crates/type-system/src/ontology/uri/wasm.rs
Alfred-Mountfield: not that I've found https://github.com/rustwasm/wasm-bindgen/issues/1818

crates/type-system/src/lib.rs
Alfred-Mountfield: Good idea, have brought most of them in, we can remove if they cause problems

crates/type-system/src/lib.rs
Alfred-Mountfield: Have deleted

crates/type-system/Cargo.toml
Alfred-Mountfield: I've moved most of the wasm stuff under that heading

packages/@blockprotocol/type-system-node/src/index_bg.wasm.d.ts
thehabbos007: They'll be suppressed in diffs and excluded from the language statistics on the GitHub repo

thehabbos007 (commented):
I don't have much to add to the review considering the upcoming PR to change binding code. The package output looks nice. I suppose the wildcard in the root `package.json` takes care of adding it to the Yarn workspace.

What is the plan for the HASH Graph backend, by the way? Will it depend on this type-system crate on a specific git revision, or go through crates.io?

Otherwise build/test passes fine locally here. I'd wait for Alex' final approval :+1:

packages/@blockprotocol/type-system-node/src/index_bg.wasm.d.ts
Alfred-Mountfield: I'm happy to add this if you also think it's a good idea @kachkaev

Alfred-Mountfield:
> What is the plan for the HASH Graph backend, by the way? Will it depend on this type-system crate on a specific git revision

At the moment I think during dev lets just target it via git, in the longer run we should look at publishing this crate (and the npm packages)

packages/@blockprotocol/type-system-node/src/index_bg.wasm.d.ts
kachkaev: Yep, let's do that. Why not?

crates/type-system/scripts/build-wasm.ts
```diff
@@ -0,0 +1,80 @@
+import fs from "fs-extra";
+import * as child_process from "node:child_process";
```
kachkaev, line 2: We can use `execa` instead, which has a slightly more convenient API. No need to do this now, this is just an idea.

packages/@blockprotocol/type-system-node/src/index_bg.wasm.d.ts
Alfred-Mountfield: That's been [done](https://github.com/blockprotocol/blockprotocol/pull/503/files/15ef69e4a87fea093bae318405001dab2f02f9c4..994b6b238ed95057452f443ec3272b14e0a5af49)

I assume that's what you had in mind @thehabbos007?

crates/type-system/scripts/build-wasm.ts
```diff
@@ -0,0 +1,80 @@
...
+    throw error;
+  }
+};
+
+const moveSrcFiles = (packagePath: PathObject, packageName: string) => {
+  const destinationPath = path.resolve(
+    __dirname,
+    "..",
+    "..",
+    "..",
```
kachkaev, line 39: Potential future improvement: might be worth avoiding `__dirname` in ESM modules. It's a built-in constant, which may look confusing (are we in CommonJS? 🤔)

We rarely  need script dir name, so can go for something like:

```ts
export const packageRootDirPath = path.resolve(
  path.dirname(fileURLToPath(import.meta.url)),
  "..",
);
```

(copied from some other file)

crates/type-system/scripts/build-wasm.ts
```diff
@@ -0,0 +1,80 @@
...
+  moveSrcFiles(pkgFolderPath, packageName);
+  cleanUp(pkgFolderPath);
+};
+
+const buildPackages = () => {
+  buildWasmPackage("type-system-web", "web");
+  buildWasmPackage("type-system-node", "nodejs");
+};
+
+buildPackages();
```
kachkaev, line 80: Potential future enhancements:

```
const script = async () => {
  // ... 
};

await script();
```

(see other script files)

We can also switch to async fs functions for consistency with other code. Using sync functions is OK in scripts, but is undesired in API handlers etc. Using one style everywhere is simpler from the cognitive POV. There are fewer ifs in the rules we follow.

packages/@blockprotocol/type-system-node/.gitattributes
```diff
@@ -0,0 +1 @@
+src/* linguist-generated=true
```
kachkaev, line 1: Let's avoid `.gitattributes` in subfolders to make repo config more transparent and easier to manage (at least for now). One source of truth helps simplify the set of rules and see the whole picture at a glance. If any folder can configure itself, that's quite a lot of extra scanning work for developers.

packages/@blockprotocol/type-system-node/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "@blockprotocol/type-system-node",
+  "version": "0.0.0",
```
kachkaev, line 3: If we are ready to publish these packages, I'd version them as `0.0.1`. We can later use Changesets to prepare new versions.

kachkaev (commented):
Looking good! Let's tweak `.gitattributes` and version numbers and :shipit:!

kachkaev:
One more thing: [Broken `yarn lint:tsc`](https://github.com/blockprotocol/blockprotocol/runs/7920802267?check_suite_focus=true#step:9:54) should be fixable by adding `tsconfig.json` to `type-system-web`.

packages/@blockprotocol/type-system-node/package.json
kachkaev: If we want to try merging packages before releasing, we can keep 0.0.0 for now.

crates/type-system/scripts/build-wasm.ts
Alfred-Mountfield: Going to avoid committing this one. I appreciate the argument about consistency but I don't think it's worth the dev effort to swap to an async API here when I need the script to behave synchronously anyway

packages/@blockprotocol/type-system-node/package.json
Alfred-Mountfield: I don't think they're worth publishing in their current state at all as is. I think there'll be a separate PR like "Prep type-system packages for publishing" or something in the future. (I suspect when we've implemented the 4 ontology types and it's not just a convoluted, oversized URI validation library 😁) We can try and revisit merging them at that point, as well as making sure we're happy with the package.json's, descriptions, etc.

packages/@blockprotocol/type-system-node/.gitattributes
Alfred-Mountfield: I've combined and moved to the top-level in a new file as we didn't seem to have one

crates/type-system/scripts/build-wasm.ts
Alfred-Mountfield: Have captured as a @todo

TimDiekmann (dismissed):
Rust code looks good!

Looking forward also to use this `Makefile.toml` for the backend 🎉

packages/@blockprotocol/type-system-node/tests/uri.test.ts
```diff
@@ -0,0 +1,51 @@
+import { isValidVersionedUri, parseBaseUri } from "..";
```
kachkaev, line 1: It' usually `test` instead of `tests`, but we can keep the folder name as is for now.

kachkaev:
@TimDiekmann / @thehabbos007 do you have any outstanding suggestions? Feel free to approve if not 🙂

packages/@blockprotocol/type-system-node/tests/uri.test.ts
Alfred-Mountfield: I'll fix on the next PR, thanks :)

TimDiekmann (approved):

PR #497, opened 2022-08-11
Fix React hydration issues related to # in paths

kachkaev:
[Asana task](https://app.asana.com/0/1200211978612931/1202736533402964/f) (internal links)

The goal of this PR is to fix two Sentry issues (internal links):

- [There was an error while hydrating](https://sentry.io/organizations/hashintel/issues/3476303382/?project=6603021)
- [Hydration failed because the initial UI does not match what was rendered on the server.](https://sentry.io/organizations/hashintel/issues/3476303370/?project=6603021)

Related upstream issue:
- https://github.com/vercel/next.js/issues/25202

To test:
1.  Open docs page
1.  Open dev tools
1.  Navigate to a header and see `#[id]` appended to the URL
1.  Refresh the page
1.  Ensure no errors in console
1.  Repeat for mobile layout (narrow document width)

PR #504, opened 2022-08-17
Fix several Sentry & `yarn dev` warnings

kachkaev:


site/src/_pages/docs/4_faq.mdx
```diff
@@ -191,7 +191,7 @@ We don’t prescribe canonical types for any _thing_, but do recommend that user
 
 </FAQ>
 
-<FAQ question={<div>How does the Block Protocol relate to <i>schema.org</i>?</div>}>
+<FAQ question={<span>How does the Block Protocol relate to <i>schema.org</i>?</span>}>
```
kachkaev, line 194: Prevents `<div>`s inside `<p>`s on `/docs/faq`. Fixes a [hydration error on Sentry](https://sentry.io/organizations/hashintel/issues/3476303370/events/8c51d2bce8a94763a8f85529f272084c/?project=6603021) that still remains after #497

site/src/components/base-link.tsx
```diff
@@ -13,9 +13,9 @@ import { FRONTEND_URL } from "../lib/config";
 
 export const isHrefExternal = (href: string | UrlObject) =>
   typeof href === "string" &&
-  (href === "/discord" ||
-    !/^(mailto:|#|\/|https:\/\/blockprotocol\.org)/.test(href)) &&
-  !href.startsWith(FRONTEND_URL);
+  (href.endsWith("/discord") ||
+    (!/^(mailto:|#|\/|https:\/\/blockprotocol\.org)/.test(href) &&
+      !href.startsWith(FRONTEND_URL)));
```
kachkaev, line 18: Continues #498. Prevents prefetching inside mdx content too.

site/src/components/link.tsx
```diff
@@ -22,13 +22,11 @@ export const BaseLinkWithAnchor = forwardRef<
...
   PlainLinkWithAnchorProps
 >((props, ref) => {
-  const { as, href, replace, scroll, shallow, prefetch, locale, ...other } =
-    props;
+  const { as, href, replace, scroll, shallow, locale, ...other } = props;
 
   return (
     <BaseLink
       href={href}
-      prefetch={prefetch}
```
kachkaev, line 31: We no longer use prefetch anywhere, so don't need to pass it down

site/src/components/mdx/faq.tsx
```diff
@@ -33,7 +33,7 @@ export const FAQ: FunctionComponent<FAQProps> = ({ question, children }) => {
       expanded={expanded}
       sx={{
         boxShadow: "none",
-        "&:first-child": {
+        "&:first-of-type": {
```
kachkaev, line 36: Fixes this warning in a few places:

```
The pseudo class ":first-child" is potentially unsafe when doing server-side rendering. Try changing it to ":first-of-type".
```

site/src/components/navbar.tsx
```diff
@@ -151,12 +158,12 @@ const useScrollingNavbar = (
 export const Navbar: FunctionComponent<NavbarProps> = ({ openLoginModal }) => {
   const theme = useTheme();
   const { pathname } = useRouter();
-  const asPath = useHydrationFriendlyAsPath();
+  const hydrationFriendlyAsPath = useHydrationFriendlyAsPath();
```
kachkaev, line 161: I’ve renamed `asPath` to `hydrationFriendlyAsPath` to avoid confusion between the two. I was still using real `asPath` in mobile navbar components, which kept triggering warnings like this:

```
Warning: Prop `href` did not match. Server: "/docs" Client: "/docs#"
```

We don't have to pass `hydrationFriendlyAsPath` as a prop because `useHydrationFriendlyAsPath()` can be called everywhere in theory. I still use a prop to avoid extra re-renders because of `useEffect` inside each `useHydrationFriendlyAsPath()`. Values may go out of sync with the current hook implementation, which can trigger warnings or inconsistent renders.

site/src/components/navbar.tsx
```diff
@@ -180,6 +187,14 @@ export const Navbar: FunctionComponent<NavbarProps> = ({ openLoginModal }) => {
     document.body.style.overflow = mobileNavVisible ? "hidden" : "auto";
   }, [mobileNavVisible]);
 
+  const handleLoginButtonClick = useCallback<MouseEventHandler>(
```
kachkaev, line 190: Should prevent [Cancel rendering route](https://sentry.io/organizations/hashintel/issues/3500395414/?project=6603021& query=is%3Aunresolved) error (happens when we click Log In and url switches to `#`).

site/src/util/mui-utils.tsx
```diff
@@ -16,7 +16,7 @@ export const parseIntFromPixelString = (pixelString: string): number => {
  * components modal, drawer, menu, popover, dialog.
  */
 export const useScrollLock = (active: boolean) =>
-  useLayoutEffect(() => {
+  useEffect(() => {
```
kachkaev, line 19: Prevents this warning:

```
Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://reactjs.org/link/uselayouteffect-ssr for common fixes.
```

If `useEffect` is not what we need here, we can switch to adding `next/dynamic` around `Modal` and `Popover`. I haven't spotted any visual regressions with the current approach though.

CiaranMn (approved):

PR #506, opened 2022-08-18
Include `timestamp` when messages are sent

teenoh:
This PR ensures a timestamp is included when messages are sent. While implementing a logs viewer in https://github.com/blockprotocol/blockprotocol/pull/507, the logs were not displaying in the right order because the timestamp was being attached when the event was received as opposed to when it was dispatched. This ensures the right timestamp is set.


CiaranMn (approved):
👍

PR #508, opened 2022-08-18
Extract `@local/eslint-config`

kachkaev:
### In scope

- Create new package called `@local/eslint-config`  
  I suggest using `@local` namespace for private packages to avoid confusion with packages from npmjs.com.

- Move `.eslintrc.js` and all ESLint deps to a new package

- Configure root `.eslintrc.json` by referring to the new package

### Out of scope (to be done in a follow-up)

- Remove `.eslintrc.json` from root

- Create ESLint config files and scripts in individual packages 

- Simplify `@local/eslint-config/index.js` (reduce the number of `overrides` by placing them to individual packages)

- Use `turbo` in top-level `fix:eslint` and `lint:eslint` scripts

---
  - https://app.asana.com/0/0/1202809472441891


yusufkinatas (approved):
Tested on my local, works as expected

PR #505, opened 2022-08-17
Update block commit hash

kachkaev:
Testing https://github.com/hashintel/hash/pull/916

To be published once the above PR is merged

PR #509, opened 2022-08-18
Clean up / clarify block metadata, and stop reading it from disk in the site

CiaranMn:
# What is this PR?

This PR is preparation for introducing a new self-service block publishing flow, whereby users will be able to complete a simple form to have their block immediately published.

Block metadata will be stored in a database rather than in the repo/site folders as now, and the built block assets will be served from external object storage rather than the NextJS site as they are now. 

There will be a short period where we continue to support the in-repo metadata alongside the self-service flow.

This PR paves the way for that by:
- Clarifying the different parts of the expanded/re-written block metadata we create from the `block-metadata.json` files, and separate out its creation so that we can use it for blocks that we aren't building from source ourselves or serving from the NextJS site.
- Create a Block model class. The Block class will handle blocks being added by the user at runtime, via a new API. The first publishing flow intended is to link to a npm package, which we will pull, store the metadata in our db and the assets in remote storage. For now, the Block class implements the existing metadata we use for blocks. We can rejig this later (e.g. to distinguish between what is coming from metadata-json and what are additions)
- Stop reading the block metadata from disk in the NextJS site code, and go via the new Block model, since blocks won't all be stored on disk
- renames some variables for clarity, e.g. `blockPackagePath` -> `blockSitePath` (since we also have a `packagePath`). There is a lot more that could be done to clean the 'expanded metadata' up, but needn't happen now.

# How to test it
Check the block hub page and individual block pages still work in the preview deployment.
N.B. any blocks for users which don't actually exist in the preview db will not display.

# Next steps
Further PR to introduce the new block publishing method alongside the existing one.


packages/block-scripts/shared/block-assets-plugin.js
```diff
@@ -31,14 +31,6 @@ export class BlockAssetsPlugin {
         )}/ws`;
       }
 
-      const exampleGraphFileExists = await fs.pathExists(
```
CiaranMn, line 34: This is now handled where we do the other metadata additions / enhancements

packages/block-scripts/shared/block-assets-plugin.js
```diff
@@ -31,14 +31,6 @@ export class BlockAssetsPlugin {
...
-        path.resolve(blockRootDirPath, "example-graph.json"),
-      );
-
-      if (exampleGraphFileExists) {
-        metadataExtra.exampleGraph = "example-graph.json";
-      }
-
```
CiaranMn, line 41: This is now handled where we do the other metadata additions / enhancements

site/src/lib/api/model/block.model.ts
```diff
@@ -0,0 +1,193 @@
+import { Db, ObjectId, WithId } from "mongodb";
+
+import localBlocks from "../../../../blocks-data.json" assert { type: "json" };
+import { ExpandedBlockMetadata } from "../../blocks";
+import { User } from "./user.model";
+
+const BUCKET_BASE_URL =
+  "https://c6499786332a3d2fb35419a7803ab7aa.r2.cloudflarestorage.com/blocks";
```
CiaranMn, line 8: To be moved to environment variable in subsequent PR

site/src/lib/api/model/block.model.ts
```diff
@@ -0,0 +1,193 @@
+import { Db, ObjectId, WithId } from "mongodb";
```
CiaranMn, line 1: This file will change in a subsequent PR, it is mostly stubs at the moment, save what it does to serve the existing local blocks.

site/src/pages/[shortname]/blocks/[block-slug].page.tsx
```diff
@@ -129,9 +131,7 @@ export const getStaticPaths: GetStaticPaths<
   BlockPageQueryParams
 > = async () => {
   return {
-    paths: (await readBlocksFromDisk()).map(
-      (metadata) => metadata.blockPackagePath,
-    ),
+    paths: [],
```
Comment (CiaranMn) [inline site/src/pages/[shortname]/blocks/[block-slug].page.tsx:134]: Doing this now since 'build every block page' will become infeasible once we have many.

site/src/lib/api/model/block.model.ts
```diff
@@ -0,0 +1,124 @@
+import { BlockVariant, JsonObject } from "@blockprotocol/core";
+import { BlockType } from "@blockprotocol/core/dist/esm/types";
+import { Db } from "mongodb";
+
+import localBlocks from "../../../../blocks-data.json" assert { type: "json" };
+import { ExpandedBlockMetadata } from "../../blocks";
+import { User } from "./user.model";
+
+export class Block implements ExpandedBlockMetadata {
```
kachkaev, line 9: On a base level, model is a javascript object with a few methods. These methods may require different sets of dependencies, which potentially makes the total number of model dependencies quite large. Object methods are not tree-shakeable, so calling a single model method in an API route unnecessary pulls all model dependencies. This makes cold lamda starts pretty slow: https://github.com/vercel/community/discussions/496#discussioncomment-3004219

I would avoid using the concept of models as such and stick with [plain data + transform function] as much as we can. We don't need to remove all models in one go, but at least avoid creating new ones with the same fundamental bundle size issue.

kachkaev (changes requested):
Not sure I fully understand our approach, in particular how we approach:

1. preview / production sets of blocks (do we have one DB? how can we preview blocks? https://github.com/blockprotocol/blockprotocol/pull/505)

1. local development (source of blocks)

Integration tests are failing, which is related to point 2. Although we can swap the current approach rather quickly, I’m not sure we get more benefits than problems without specing things out first. Some unresolved questions can be found in [Asana Task](https://app.asana.com/0/1200211978612931/1202539910143054/f) (internal).

site/src/lib/api/model/block.model.ts
CiaranMn: I've removed the Block model and replaced it with [this file](https://github.com/blockprotocol/blockprotocol/blob/7d04f497366ca55288508e975b60c9b7d6271b4f/site/src/lib/api/blocks.ts)

CiaranMn:
> Not sure I fully understand our approach, in particular how we approach:
> 
> 1. preview / production sets of blocks (do we have one DB? how can we preview blocks? https://github.com/blockprotocol/blockprotocol/pull/505)
> 
> 2. local development (source of blocks)
> 
> Integration tests are failing, which is related to point 2. Although we can swap the current approach rather quickly, I’m not sure we get more benefits than problems without specing things out first. Some unresolved questions can be found in [Asana Task](https://app.asana.com/0/1200211978612931/1202539910143054/f) (internal).

Thanks @kachkaev – I've fixed the tests, which were failing because of a tweak to block naming (we were using the name from the block's `package.json` to identify them, e.g. `@hashintel/block-code`, but I've stopped us keeping that around in the repo as it bears no relation to how the block is identified here). They are timing out now, which appears to be some issue with esm.sh unrelated to this PR (try visiting a block page in prod)

Worth noting that the next PR will add 'block metadata in a db' as a parallel approach to the local JSON, so the local JSON files won't be removed yet, but to your questions on environments:
1. There is a preview db and a production db, so any blocks added to a preview deployment via the self-service, 'blocks in db' approach will only appear in preview deployments. In contrast to 'block metadata in the repo', any block added to the preview db from one deployment will be shared across all preview deployments. I think we can live with this. If it becomes a problem, we will have to have separate datastores per preview deployment.
2. Local development has its own db so can do add and remove blocks at will without affect anyone else.

Also worth saying that given that this is preparing for introducing self-service publishing, as part of the next up the publishing process / block hub will have to be resilient to missing metadata or reject blocks with it, because blocks can be added straight to production without having gone through any review process / preview, except any that provide as part of the block publishing flow.

nathggns:
I've just recalled that we use the `package.json` name in HASH for the paragraph block. Is that going to become a problem? https://github.com/hashintel/hash/blob/d5f1d742af4d5ad88012ab2a2d2d3753393bc143/packages/hash/frontend/src/blocks/page/createEditorView.ts#L153

CiaranMn:
> I've just recalled that we use the `package.json` name in HASH for the paragraph block. Is that going to become a problem? https://github.com/hashintel/hash/blob/d5f1d742af4d5ad88012ab2a2d2d3753393bc143/packages/hash/frontend/src/blocks/page/createEditorView.ts#L153

Ugh, yes. Thanks @nathggns. I'll do a PR in `hash` to use something else to identify it.

nathggns:
Looking now

site/src/lib/blocks.ts
```diff
@@ -93,56 +178,46 @@ export const readBlocksFromDisk = async (): Promise<
...
+        "block-metadata.json",
+        "example-graph.json",
       ),
-      lastUpdated: null, // TODO: derive from block data when provided by the hub
-    });
+    );
+
+    const blockDistributionFolderUrl = `${FRONTEND_URL}/blocks/${pathWithNamespace}`;
+
+    const expandedMetadata: ExpandedBlockMetadata = enhanceBlockMetadata(
```
nathggns, line 207:
```suggestion
    const expandedMetadata = enhanceBlockMetadata(
```

nathggns (approved):

site/src/lib/blocks.ts
```diff
@@ -80,9 +106,68 @@ const getRepositoryUrl = (
...
+    repository,
+    schema: generateBlockFileUrl(metadata.schema, blockDistributionFolderUrl)!,
+    source: generateBlockFileUrl(metadata.source, blockDistributionFolderUrl)!,
+    variants: metadata.variants?.length
+      ? metadata.variants?.map((variant) => ({
+          ...variant,
+          icon: generateBlockFileUrl(variant.icon, blockDistributionFolderUrl)!,
+        }))
+      : null,
+    exampleGraph: generateBlockFileUrl(
```
teenoh, line 162: 👍

teenoh (dismissed):
LGTM!

site/src/lib/blocks.ts
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/509/commits/91ee764c4dde800575af1c9fcb383002bfa79ae3

site/src/lib/api/blocks.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const getBlockByUserAndName = (params: {
+  name: string;
+  shortname: string;
+}): ExpandedBlockMetadata | null => {
+  const { name, shortname } = params;
+  // the inferred type of the JSON is incompatible with the TS type because a string union is inferred as any string
+  return (localBlocks.filter(
+    ({ pathWithNamespace }) => pathWithNamespace === `@${shortname}/${name}`,
+  )?.[0] ?? null) as ExpandedBlockMetadata | null;
```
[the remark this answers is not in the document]
CiaranMn: 😅 

https://github.com/blockprotocol/blockprotocol/pull/509/commits/91ee764c4dde800575af1c9fcb383002bfa79ae3

site/src/lib/blocks.ts
```diff
@@ -80,9 +106,68 @@ const getRepositoryUrl = (
   return undefined;
 };
 
+const enhanceBlockMetadata = (
```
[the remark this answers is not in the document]
CiaranMn: Indeed it is

https://github.com/blockprotocol/blockprotocol/pull/509/commits/91ee764c4dde800575af1c9fcb383002bfa79ae3

nathggns (approved):

PR #511, opened 2022-08-19
Update User Dashboard Design & Content

yusufkinatas:

Update the logged-in dashboard to reflect new design. Also refactor the code around it a little bit.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1202765662321688/f) _(internal)_
- [Design](https://www.figma.com/file/Fpmde3E9nJMb4VLkPTYgrM/blockprotocol.org?node-id=3703%3A46095) _(internal)_

## ❓ How to test this?

1. Checkout the branch / view the deployment
2. Login and open dashboard
3. Compare dashboard with Figma

### Card hover state
https://user-images.githubusercontent.com/39553853/185637247-81171163-d233-4174-8296-9f6d3df5b148.mov

### Skeleton loading for Dashboard
https://user-images.githubusercontent.com/39553853/185649335-819f7e1e-3245-4d1b-a189-371723285bef.mov

### Overview 
Screen Shot 2022-08-19 at 17 09 32


site/src/pages/dashboard/[[...slugs]].page.tsx
```diff
@@ -9,81 +8,35 @@ import {
...
   const { preferredName: userName, shortname } = user ?? {};
   const [schemaModalOpen, setSchemaModalOpen] = useState(false);
+  const [userHasApiKey, setUserHasApiKey] = useState<boolean | undefined>(
+    undefined,
+  );
   const router = useRouter();
 
+  useEffect(() => {
+    void apiClient
+      .getUserApiKeys()
```
Comment (teenoh) [inline site/src/pages/dashboard/[[...slugs]].page.tsx:32]: If getUserApiKeys fails for some reason, the page will keep loading. We should handle setting userHasApiKey in that scenario

Comment (teenoh) [reply site/src/pages/dashboard/[[...slugs]].page.tsx:32]: I think it might make more sense to add a `hasApiKey` to the user object serverside, so we don't have to make this call in the client. We can do this in the serialize method here => https://github.com/blockprotocol/blockprotocol/blob/main/site/src/lib/api/model/user.model.ts#L447-L455

site/src/components/pages/dashboard/dashboard-card/dashboard-card.tsx
```diff
@@ -0,0 +1,221 @@
...
+            flexDirection: "column",
+            flex: 1,
+          }}
+        >
+          <Typography
+            sx={{
+              fontFamily: "Apercu Pro",
+              fontSize: "28.128px",
+              lineHeight: "120%",
+              color: "#37434F",
```
teenoh, line 207: We should use variants from the design system and not set fontFamily/fontSize directly

site/src/components/pages/dashboard/dashboard-card/dashboard-card.tsx
```diff
@@ -0,0 +1,221 @@
...
+              mb: 1,
+            }}
+          >
+            {title}
+          </Typography>
+          <Typography mb={2} color="#4D5C6C">
```
teenoh, line 213: we should pick colors from the theme

site/src/components/pages/dashboard/dashboard-card/dashboard-card.tsx
yusufkinatas: This was already existing code, but right, it should be fixed

site/src/components/pages/dashboard/dashboard-card/dashboard-card.tsx
yusufkinatas: Same as above

site/src/components/pages/dashboard/utils.ts
```diff
@@ -8,3 +18,129 @@ export const dashboardPages: { tabTitle: string; tabHref: string }[] = [
...
+        link: {
+          title: "Manage keys",
+          href: "/settings/api-keys",
+        },
+        icon: faKey,
+        variant: "secondary",
+      },
+      {
+        /** @todo show "View your public profile" if user is already completed profile */
+        title: "Complete your public profile",
```
teenoh, line 87: Since we don't have the code in for "Completing profile", let's use "View your public profile" and add a `@todo` to switch back when update profile functionality is in

teenoh (changes requested):
Thanks @yusufkinatas. This looks really good! I added comments around using properties from our theme and minimising api calls. 
Another thing to note is playwright tests are failing because this PR changes some text which tests depend on. You will have to update the tests too

site/src/components/pages/dashboard/dashboard-card/dashboard-card.tsx
teenoh: Ohh I wasn't aware. Thanks for looking at fixing

site/src/pages/dashboard/[[...slugs]].page.tsx
Comment (teenoh) [reply site/src/pages/dashboard/[[...slugs]].page.tsx:32]: It will also be great if we can then have the skeletons showing up the first time we fetch the current user

site/src/components/pages/dashboard/dashboard-card/dashboard-card.tsx
yusufkinatas: Updated at https://github.com/blockprotocol/blockprotocol/pull/511/commits/b0e108599e1954e95939a7b15acce487c24f45fd

site/src/components/pages/dashboard/dashboard-card/dashboard-card.tsx
yusufkinatas: Updated at https://github.com/blockprotocol/blockprotocol/pull/511/commits/b0e108599e1954e95939a7b15acce487c24f45fd

site/src/components/pages/dashboard/utils.ts
yusufkinatas: Updated at https://github.com/blockprotocol/blockprotocol/pull/511/commits/b0e108599e1954e95939a7b15acce487c24f45fd

site/src/pages/dashboard/[[...slugs]].page.tsx
Comment (yusufkinatas) [reply site/src/pages/dashboard/[[...slugs]].page.tsx:32]: Fixed possible infinite loading issue here https://github.com/blockprotocol/blockprotocol/pull/511/commits/546ba3dcd6f8dbdae1f2ae8045e0ab5eeb4f1896.

@teenoh About adding loading state to the whole page depending on user or other params, I feel like it's out of scope of this PR. I think we should develop a re-useable solution to showing loading states based on params at some point. But for now we only have 1 page has skeleton loading state, which is dashboard. And the current loading is just a glimpse. So I feel like we can postpone having a better loading state solution when we start rendering more user related data on the dashboard.

Comment (teenoh) [reply site/src/pages/dashboard/[[...slugs]].page.tsx:32]: Totally understand. In my comment, I was actually referring to having the skeletons in the dashboard page when the user is being fetched (user === 'loading'), but I agree we need to come up with a more reusable solution to show loading states on other pages and that's outside the scope of this PR.
 
In the mean time, I think it should be a bit straightforward to show the skeletons when the user is being fetched in dashboard page.

Currently, reloading the dashboard page looks like this => 

https://user-images.githubusercontent.com/16928688/185929274-4ab49aef-9c8b-4f1c-934c-16106e81a09d.mov

If this takes longer than expected, then we can address in a follow up

Comment (teenoh) [reply site/src/pages/dashboard/[[...slugs]].page.tsx:32]: @CiaranMn what do you think about including `hasApiKey` in the user object?

Comment (CiaranMn) [reply site/src/pages/dashboard/[[...slugs]].page.tsx:32]: My instinct is that it's a bit wasteful to check for a user's API keys whenever we want to retrieve the user, when we really only need it for a couple of pages here – if the aim is to have the data available on the first render, can we make the request to `getUserApiKeys` in [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props) instead?

Comment (yusufkinatas) [reply site/src/pages/dashboard/[[...slugs]].page.tsx:32]: @CiaranMn For that, we need to get the username on serverside. I checked how our authentication flow is written quickly, seems like we only have `blockprotocol-session-id` on serverside. And couldn't see any examples of it's being used in serverside. So I'm not sure how to implement it on `getServerSideProps`, or is it supported by our authentication flow right now.

Comment (CiaranMn) [reply site/src/pages/dashboard/[[...slugs]].page.tsx:32]: Ah, okay. We would need to move the call to get the user to server side then. Possibly we could use something like `getSession` in `session.middleware.ts`. There's a task ([internal link](https://app.asana.com/0/1200211978612931/1201715945634534/f)) for moving the user call to server side.

In the meantime, we could just not vary the API key text based on whether has an API key or not. We don't even allow users to have one more than one key at the moment so it's inaccurate as it is - you can either (a) create or (b) regenerate an API key. @nonparibus preference?

Comment (nonparibus) [reply site/src/pages/dashboard/[[...slugs]].page.tsx:32]: Fine to not vary it for now. Eventually we will definitely want users to be able to have multiple keys but there is no immediate urgency re: this.

teenoh (approved):

PR #512, opened 2022-08-19
Add Data Types to Package System

Alfred-Mountfield:

Expands on #503 by adding Data Types

## 🔗 Related links

## 🔍 What does this change?

- Adds the `DataType` Rust struct
- Adds [Tsify](https://github.com/madonoharu/tsify) to generate a TypeScript interface for the struct
- Adds a `tests/data` module to store test data to use in the unit tests
- Adds the `isValidDataType` function to expose to TS
- Rebuilds the npm packages
- Expands the npm integration tests to test `isValidDataType`
- Driveby: fixes an issue with versioned URI parsing, adds a unit test for it
- Driveby: renames `tests` to `test` in `@blockprotocol/type-system-node`

- Same as last time - READMEs have been added for the package itself. We likely want to improve discoverability of this package, and add usage instructions for how to import and use the NPM packages.
- The TypeScript bindings have improved docstrings now

## ⚠️ Known issues

- some issues with `wasm-pack`, see https://github.com/iotaledger/identity.rs/pull/694
- Rust errors are still being passed over the boundary as objects with memory in WASM, this means they're not particularly ergonomic to use in JS. A follow-up PR will deal with error handling.
- Renaming function parameters in the generated typescript files isn't ergonomic. Because of this I have disabled the lint about snake_case and am using camelCase inside the wasm files 🤮 
- The `@throws` appears above the `@param` in the JsDoc. I don't know if this is avoidable without manually rewriting the files :(

## 🐾 Next steps

  - Property Types
  - Entity Types
  - Link Types

## 🛡 What tests cover this?

- Rust unit tests _will_ be added to also check the crate
- There is a setup for WASM headless browser tests but these are currently not utilised, a testing strategy will need to be developed
- It would seem that generating the NPM packages is not idempotent. This might cause problems when we try to check for diffs with the generated files.

## ❓ How to test this?

1.  Follow the README for the crate
1.  Run `yarn test` inside `@blockprotocol/type-system-node`


crates/type-system/src/ontology/data_type/wasm.rs
```diff
@@ -0,0 +1,24 @@
...
+extern "C" {
+    #[wasm_bindgen(typescript_type = "DataType")]
+    pub type IDataType;
+}
+
+/// Checks if a given {DataType} is valid
+///
+/// @throws {MalformedDataTypeError} if the data type is malformed
+#[wasm_bindgen]
+pub fn isValidDataType(dataTypeObj: &IDataType) -> Result<(), MalformedDataTypeError> {
```
Alfred-Mountfield, line 19: This pattern is probably worth explaining. 

- `Tsify` generates a TypeScript interface (see `derive` on `DataType`). 
- We then tell `wasm_bindgen` that there is an external type `IDataType` which it can find in the TypeScript bindings as `DataType`. 
- We can then in turn use that in the `wasm_bindgen` function here where `IDataType` becomes an alias for `JsValue`.
- This is important because it allows us to call `into_serde`, if we didn't rely on the binding and instead told it the real type, then it would do an unchecked_unwrap which will panic in WASM rather than returning serde errors.
- We need the `extern "C"` thing so that the wasm binding _for the functions_ have the correct interface on the function parameters

crates/type-system/src/ontology/data_type/wasm.rs
```diff
@@ -0,0 +1,24 @@
+#![allow(
+    non_snake_case,
+    reason = "We want camelCase types in TS, and this module is WASM only"
```
TimDiekmann, line 3:
```suggestion
    reason = "We want camelCase functions in TS, and this module is WASM only"
```

TimDiekmann: This feels weird though, can't we rename the function in the WASM attribute?

crates/type-system/src/ontology/uri/mod.rs
```diff
@@ -87,8 +87,7 @@ impl FromStr for VersionedUri {
     type Err = ParseVersionedUriError;
 
     fn from_str(uri: &str) -> Result<Self, ParseVersionedUriError> {
-        let (base_uri, version) = uri.rsplit_once("/v/").ok_or(ParseVersionedUriError)?;
-
+        let (base_uri, version) = uri.rsplit_once("v/").ok_or(ParseVersionedUriError)?;
```
TimDiekmann, line 90: What is the reasoning behind this? Do we allow URIs like `https://example.com/serv/1`?

crates/type-system/src/ontology/data_type/wasm.rs
Alfred-Mountfield: I'm not just talking about functions, I'm on about variable names, etc. The main problem is arguments _within_ functions, we can rename a function but we can't rename the arguments within them. So renaming
```rust
pub fn is_valid_data_type(data_type_obj: &IDataType)
```

will result in 
```typescript
function isValidDataType(data_type_obj: DataType)
```

rather than 
```typescript
function isValidDataType(dataTypeObj: DataType)
```

crates/type-system/src/ontology/uri/mod.rs
Alfred-Mountfield: That's a good point actually, and is a bug. Thank you. I will need to adjust this logic.
The issue is as follows
`http://ex.com/foo/v/1` will be split into
`["http://ex.com/foo", "1"]`
and `Url::parse("http://ex.com/foo")` gives a valid `url`, however
`url.join("/v/1")` will result in `foo` being dropped because it doesn't have a trailing `slash` which means it's treated as a file and is dropped. So you end up with `http://ex.com/v/1`

Can you think of a clean solution to this beyond me modifying the `base_uri` to append `"/"`?
We could perhaps do a regex with capturing groups but I'm not sure if that's cleaner or not

crates/type-system/src/ontology/data_type/wasm.rs
TimDiekmann: Gotcha, makes sense!

Alfred-Mountfield: I guess those aren't quite types.
```suggestion
    reason = "We want camelCase variable names in TS, and this module is WASM only"
```

crates/type-system/src/ontology/uri/mod.rs
TimDiekmann: Hmm, what about `format!("{base_uri}/v/{version}")`? Using `Path` may have weird side effects when on Windows, which could result in `{base_uri}\v\{version}`.

Alfred-Mountfield: I'm a little confused, we're not talking about `Path` or reconstructing the URI. We're talking about _extracting_ the components, hence the `rsplit`

Alfred-Mountfield: My point is that `base_uri` _must_ have a trailing slash or else the `Url` library doesn't treat it as a valid base.

Alfred-Mountfield: I'm using a regex now

crates/type-system/src/ontology/uri/mod.rs
```diff
@@ -87,9 +89,14 @@ impl FromStr for VersionedUri {
     type Err = ParseVersionedUriError;
 
     fn from_str(uri: &str) -> Result<Self, ParseVersionedUriError> {
-        let (base_uri, version) = uri.rsplit_once("/v/").ok_or(ParseVersionedUriError)?;
-
         // TODO: better error handling
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r#"(.*/)v/(\d)*"#).expect("Regex failed to compile");
+        }
```
TimDiekmann, line 95: Do we want to use [LazyLock](https://doc.rust-lang.org/nightly/std/sync/struct.LazyLock.html) instead of `lazy_static`?

TimDiekmann (approved):
LGTM 🎉 

Only one nit, but not blocking this PR

crates/type-system/src/ontology/uri/mod.rs
Alfred-Mountfield: Any preference about these? I'm just following the [regex docs suggestion](https://docs.rs/regex/latest/regex/#example-avoid-compiling-the-same-regex-in-a-loop)

TimDiekmann: I'd probably use `LazyLock`:

```rust
static RE: LazyLock<Regex> = LazyLock::new(|| {
    Regex::new(r#"(.*/)v/(\d)*"#).expect("Regex failed to compile")
});
```

I guess it's not in the regex docs as it's not stabilized yet.

crates/type-system/src/ontology/uri/mod.rs
```diff
@@ -87,9 +89,14 @@ impl FromStr for VersionedUri {
     type Err = ParseVersionedUriError;
 
     fn from_str(uri: &str) -> Result<Self, ParseVersionedUriError> {
-        let (base_uri, version) = uri.rsplit_once("/v/").ok_or(ParseVersionedUriError)?;
-
         // TODO: better error handling
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r#"(.*/)v/(\d)*"#).expect("Regex failed to compile");
```
thehabbos007, line 94: I think this Regex should be changed to the following
```suggestion
            static ref RE: Regex = Regex::new(r#"(.+/)v/(\d+)"#).expect("Regex failed to compile");
```
otherwise the capture group number 2 will capture the right-most number, and the regex would match against `/v/`.

thehabbos007 (changes requested):

crates/type-system/src/ontology/data_type/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+use serde_json;
+#[cfg(target_arch = "wasm32")]
+use {tsify::Tsify, wasm_bindgen::prelude::*};
+
+use crate::uri::VersionedUri;
+
+/// Will serialize as a constant value `"dataType"`
```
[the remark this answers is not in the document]
Alfred-Mountfield: For **serialization** (with serde) it does I believe, however it doesn't work for `tsify`'s type generation.

crates/type-system/src/ontology/uri/mod.rs
Alfred-Mountfield: As discussed offline, we needed more modifications

thehabbos007 (approved):
Thank you Alfie! :rocket:

PR #518, opened 2022-08-22
Change version referencing in spec

nonparibus:
This PR changes the way we refer to the specification versions on the Block Protocol website.

This is a temporary change that will be effective until such time as we expose older versions of the spec via a dropdown/other selector on the live site.


kachkaev (dismissed):

site/src/_pages/docs/3_spec/1_core-specification.mdx
```diff
@@ -1,6 +1,6 @@
-# Core Specification 0.2
+# Core Service
```
kachkaev, line 1: Hmm actually I’m not sure if the Core is a ‘service’. We should probably keep ‘Core Specification’ or switch to something like ‘Core’ /  ‘BlockProtocol Core’.

In the latter case we should probably update URLs too, e.g.:

- `/docs/spec/core`
- `/docs/spec/graph-service`

I can do that (including new redirects) based on the decision.

CiaranMn: Agree with not calling core a 'service' – if we want to do that we need to rework the rest of the document, which specifies how you can define 'services', which are described in a way which is inconsistent with this document.

We can do that, just needs a bit more thinking about how we then frame the different parts of the BP – we no longer have the core spec which defines rendering and comms transport, and service spec which defines comms content, but instead a collection of services, one of which has special status.

nonparibus, line 1:
```suggestion
# Core Specification
```

nonparibus: Per https://github.com/blockprotocol/blockprotocol/pull/518/files#r953445265

nonparibus: Actioned in https://github.com/blockprotocol/blockprotocol/pull/518/commits/6b363a9d924bff39d2f1666a2b73a38d8073b2a2

site/src/_pages/docs/3_spec/1_core-specification.mdx
```diff
@@ -1,6 +1,6 @@
-# Core Specification 0.2
+# Core Specification
```
nonparibus, line 1:
```suggestion
# Core
```

Upon reflection I think this suggestion from Alex is cleaner. The whole thing is already in the 'Specification' section of the sidebar/site.

site/tests/integration/docs.test.ts
```diff
@@ -105,12 +105,10 @@ test("Docs page should contain key elements and interactions should work", async
     page.locator("text=PreviousEmbedding Blocks >> a"),
   ).toHaveAttribute("href", "/docs/embedding-blocks");
 
-  await expect(
-    page.locator("text=NextCore Specification 0.2 >> a"),
-  ).toBeVisible();
+  await expect(page.locator("text=NextCore Specification >> a")).toBeVisible();
```
nonparibus, line 108:
```suggestion
  await expect(page.locator("text=NextCore >> a")).toBeVisible();
```

Follows https://github.com/blockprotocol/blockprotocol/pull/518/commits/b3ee31071fe47c55d5e3096bbf7937573bc15a6f

site/tests/integration/docs.test.ts
```diff
@@ -105,12 +105,10 @@ test("Docs page should contain key elements and interactions should work", async
...
 
   await expect(
-    page.locator("text=NextCore Specification 0.2 >> a"),
+    page.locator("text=NextCore Specification >> a"),
```
nonparibus, line 111:
```suggestion
    page.locator("text=NextCore >> a"),
```

Follows https://github.com/blockprotocol/blockprotocol/pull/518/commits/b3ee31071fe47c55d5e3096bbf7937573bc15a6f

nonparibus:
@kachkaev Will we need to set up any redirects if we rename the `Core` page (or even the `Graph Service` one, which previously also contained `0.2`)? Just wanted to explicitly flag/check.

kachkaev:
The URLs haven't changed

- https://blockprotocol.org/docs/spec/core-specification  
  https://blockprotocol.org/docs/spec/graph-service-specification
  
- https://blockprotocol-ar5luchpv.stage.hash.ai/docs/spec/core-specification  
  https://blockprotocol-ar5luchpv.stage.hash.ai/docs/spec/graph-service-specification
  
If we want to drop ‘specification’ from URLs, we’ll need redirects – yes.

nonparibus:
@kachkaev Would you mind adding these in for me on this PR? Thanks!

site/tests/integration/docs.test.ts
nonparibus: @kachkaev Can you confirm that this and https://github.com/blockprotocol/blockprotocol/pull/518#discussion_r956164262 are correct? (Following https://github.com/blockprotocol/blockprotocol/pull/518#discussion_r956162707) Thanks!

kachkaev (approved):

kachkaev:
@nonparibus is away, so merging. Integration tests are failing, but this is related to #526

PR #519, opened 2022-08-22
Add Property Types to Package System

Alfred-Mountfield:

Expands on #512 by adding Property Types

## 🔗 Related links

- Previous implementations, a lot of this was originally developed on HASH before realising we could contribute it back up
 - https://github.com/hashintel/hash/pull/797
 - https://github.com/hashintel/hash/pull/751

## 🔍 What does this change?

- Adds the `PropertyType` Rust struct and necessary structs it depends on:
  - Adds `DataTypeReference`
  - Adds `ValidateUri` to check complex structs have the correct combination of `VersionedUri`s and `BaseUri`s
  - Adds `PropertyTypeReference`
  - Adds the `repr_shared` module to encapsulate common structs used for JSON schema representations and to aid with serde
    - `Array`
    - `ValueOrArray`
    - `Object` and its intermediary `ObjectRepr` (useful for getting errors during de/serialization)
    - `OneOf` and its intermediary `OneOfRepr` (useful for getting errors during de/serialization)
  -  `PropertyValues`
- Adds the `isValidPropertyType` method across the WASM boundary to validate a JSON struct against the DataType structure
-  Correctly generates bindings for `BaseUri` within the declaration file (fixes its use within other types)
- Adds helper test functions to check serialization
- Adds test data for use in the tests
- Fixes the data-type package integration tests due to uncovered bug
- Adds property-type package integration tests
- Regenerates the wasm packages
- Driveby: Updates `DataType` to generate better TS bindings with more `tsify` attributes
- Driveby: Fixes the `build-wasm.ts` script to correctly error when `wasm-pack` fails

- Same as last time - READMEs have been added for the package itself. We likely want to improve discoverability of this package, and add usage instructions for how to import and use the NPM packages.
- The TypeScript bindings have improved docstrings now

## ⚠️ Known issues

- Rust errors are still being passed over the boundary as objects with memory in WASM, this means they're not particularly ergonomic to use in JS. A follow-up PR will deal with error handling.
- `tsify` has a number of issues I have uncovered
  - https://github.com/madonoharu/tsify/issues/5
  - https://github.com/madonoharu/tsify/issues/4
  - https://github.com/madonoharu/tsify/issues/3
- the artifacts generated by `wasm-pack` (and `wasm-bindgen`) do not seem to be idempotent, we'll need to account for shuffling of functions when it comes to checking they've been built in CI 

## 🐾 Next steps

  - Entity Types
  - Link Types
  - Check

## 🛡 What tests cover this?

- A variety of Rust unit tests have been added 

## ❓ How to test this?

1.  Follow the README for the crate
1.  Run `yarn test` inside `@blockprotocol/type-system-node`


crates/type-system/src/ontology/property_type/wasm.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// #[wasm_bindgen]
+// pub fn newPropertyType(
+//     uri: &str,
+//     title: &str,
+//     description: &str,
+// ) -> Result<IPropertyType, TempError> {
+//     let property_type = PropertyType::new(VersionedUri::from_str(uri), title.to_string(), None,
+// ()); }
```
TimDiekmann, line 42: For what is this needed/needs to be kept?

crates/type-system/src/ontology/uri/mod.rs
```diff
@@ -1,18 +1,20 @@
...
+#[cfg(target_arch = "wasm32")]
+use tsify::Tsify;
 use url::Url;
 
 use crate::uri::error::ParseBaseUriError;
 
+#[cfg_attr(target_arch = "wasm32", derive(Tsify))]
 #[derive(Clone, PartialEq, Eq, Hash)]
-pub struct BaseUri(Url);
+pub struct BaseUri(#[cfg_attr(target_arch = "wasm32", tsify(type = "string"))] Url);
```
TimDiekmann, line 17: To make this more readable we probably want to convert it into a `struct { ... }` (instead of `struct(...)`

TimDiekmann (changes requested):
Awesome 🎉 

Generally, the PR LGTM, only a few small things.

Can we add a `.gitattributes` file covering the `dist/` folders to mark them as autogenerated?

crates/type-system/src/ontology/property_type/wasm.rs
Alfred-Mountfield: This is the part I'm currently modifying (that I spoke about offline and is the reason this PR is still _technically_ in draft). The changes should pretty much just be localised to here though so the rest of the review is intact

crates/type-system/src/ontology/uri/mod.rs
Alfred-Mountfield: I do wish there was a better way to inject this attr :(

I hear you, but I don't think we should swap to a struct instead of a newtype just for readability of its initial definition. Also having it as a newtype is somewhat important for the `tsify` as the generated Typescript type is an alias rather than a class/interface.

TimDiekmann: That makes sense, thanks for clarifying!

Alfred-Mountfield:
> Can we add a .gitattributes file covering the dist/ folders to mark them as autogenerated?

[This was already done](https://github.com/blockprotocol/blockprotocol/blob/main/.gitattributes), but mistakenly wasn't updated to cover the change from `src` to `dist`. I will push a fix

crates/type-system/src/ontology/property_type/wasm.rs
Alfred-Mountfield: I've actually removed it. I was trying to come up with arguments for `newPropertyType` and I'm not sure how to create a more ergonomic input for the "oneOf" yet that isn't just the same as the normal interface. I'll interface with @nathggns as he starts work on the editors

crates/type-system/src/ontology/data_type/mod.rs
```diff
@@ -24,9 +27,11 @@ enum DataTypeTag {
 pub struct DataType {
     #[cfg_attr(target_arch = "wasm32", tsify(type = "'dataType'"))]
     kind: DataTypeTag,
+    #[cfg_attr(target_arch = "wasm32", tsify(type = "`${string}/v/${number}`"))]
```
thehabbos007, line 30: Just want to flag that template literal types would need some sort of validating function to assert structure. Otherwise code using arbitrary non-constant `string`s won't compile. 

I guess we don't currently have a way to write manual typescript methods to add to the package next to the auto generated methods. Is that something we're interested in to have a type assertion validator function for `${string}/v/${number}`?

Detailed explanation:
```typescript
// Imagine this is a user-supplied string we don't have any restriction on
// i.e. it's of type `string`
const maybe$id = "invalidstring/v/";

// We'd need a vaildation functions such as this that asserts the type
const exampleStrIsValidUri = (str: string): str is DataType["$id"] => {
  // This is for brevity. Maybe a regex instead.
  return str.indexOf("/v/") !== -1;
};

// For us to conditionally use it as an $id
if (exampleStrIsValidUri(maybe$id)) {
  const _: DataType = {
    $id: maybe$id,
    // ..
  };
}
```

packages/@blockprotocol/type-system-node/test/data-type.test.ts
```diff
@@ -51,7 +51,7 @@ const primitiveDataTypes: DataType[] = [
 const invalidDataTypes: DataType[] = [
   {
     kind: "dataType",
-    $id: "https://blockprotocol.org/@blockprotocol/types/data-type/text/", // unversioned URI
```
thehabbos007, line 54: Wouldn't it also make sense to keep an invalid data type that uses an unversioned URI for testing?

thehabbos007 (commented):
Looks good to me as well, have one thing that might be worth discussing though :)

crates/type-system/src/ontology/data_type/mod.rs
Alfred-Mountfield: Interesting, lets discuss offline

Alfred-Mountfield: I'm trying out a new approach:
- typing `VersionedUri` by creating a new alias in a private newtype.
- Removing the explicit overrides in the other types, relying on the fact that `Tsify` will keep the original type name if the thing being referred to doesn't implement `Tsify`. So in this instance, **because** `VersionedUri` doesn't implement `Tsify`, any struct with `foo: VersionedUri` will generate a TS interface with `VersionedUri` in the type definition. Which would usually fail TypeScript compilation except we're now adding the newtype under the same name so it resolves nicely.
- Turning `isValidVersionedUri` into a type guard by manually defining the TS function signatures in static strings.
 - This means I was able to get rid of the `allow(non_snake_case)`
   - I'm not sure if we want to repeat this pattern in the other WASM modules, it does mean we have more things to keep in sync, we can't just rely on `wasm_bindgen` creating nice signatures and interpolating the Rust docs in. @TimDiekmann pinging in case this is of interest

https://github.com/blockprotocol/blockprotocol/pull/519/files/0c2ac1574f74c9b09b9898188e210d2399fb0004..becc2fe048b64b0552ef5e4f583e9502551d593b

crates/type-system/src/ontology/uri/mod.rs
```diff
@@ -72,6 +72,7 @@ impl<'de> Deserialize<'de> for BaseUri {
 }
 
 // TODO: can we impl Tsify to turn this into a type: template string
+//  if we can then we should delete wasm::VersionedUriPatch
```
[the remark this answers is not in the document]
Alfred-Mountfield: Yes. If we `Tsify` `VersionedUri` then it will get turned into an interface rather than a simple type alias for a string.

TimDiekmann (approved):
LGTM 🎉

PR #520, opened 2022-08-22
Redirect logged-in users visiting root to `/dashboard`

yusufkinatas:

- Redirects logged-in users to `/dashboard` from `/` via [Next.js Middleware](https://nextjs.org/docs/advanced-features/middleware)

## 🔗 Related links
[Asana task](https://app.asana.com/0/1200211978612931/1202765662321696/f) (internal)

## ❓ How to test this?

1.  Login
2. Try opening root `/` . You should be redirected to `dashboard`
3. Logout
4. Try opening `/`. You should be able to see homepage


site/src/middleware.page.ts
```diff
@@ -0,0 +1,18 @@
+import type { NextRequest } from "next/server";
+import { NextResponse } from "next/server";
+
+export async function middleware(request: NextRequest) {
+  const { cookies } = request;
+
+  // destroying session at `logout.api.ts` removes this cookie
+  const isLoggedIn = !!cookies.get("blockprotocol-session-id");
```
kachkaev, line 8: In theory, `blockprotocol-session-id` may contain an expired session. This will redirect a logged out user to the dashboard. Is there a way to parse a cookie here, thus getting a more accurate result?

yusufkinatas: I checked the  `next-session` library. Looks like they're calling `session.destroy()` when  the cookie is expired, 
which removes the cookie (our logout method calls `session.destroy()` as well, to make sure cookie is removed on logout) https://github.com/hoangvvo/next-session/blob/master/src/memory-store.ts#L10-L27

kachkaev: Are we confident that a present cookie is always a valid cookie? AFAIU, the value may be there, but its decrypted payload can become invalid. That’s in theory, I’m not 100% sure wither it’s the case for `next-cookie`.

yusufkinatas: Tested two scenarios @kachkaev 

- Manually corrupting cookie (adding random characters to it)
  - Broken flow. opening homepage redirects user to `/dashboard`, which redirects to `/login` . So user have to clear cookies, or login again to get back to normal again.
  - I feel like this is ok, because user intended to break the cookie by editing it?
- Cookie expiration
  - Expected flow happens. User can open homepage because library destroys the session & removes the cookie

kachkaev: > Cookie expiration 
> library destroys the session & removes the cookie

What do you mean by cookie expiration here? If it’s expired on the browser level, then it won’t be passed to the server as request headers. What may also happen is: the cookie itself is not expired, but the session that it represents is expired in the DB. We don’t do any session checking in the middleware so I believe we’ll end up with the first scenario then (corrupt cookie).

The current implementation is probably fine, I just want to understand the probability of each scenario.

kachkaev:
Looks like we need to update integration tests since we can no longer see the home page as a logged in user. Covering new lines of code with tests would be useful too.

site/src/middleware.page.ts
yusufkinatas: >What do you mean by cookie expiration here?

I mean the expiration set at [session.middleware.ts](https://github.com/blockprotocol/blockprotocol/blob/cc23b28c3a78852cfeeae910bac6b4a5a976c199/site/src/lib/api/middleware/session.middleware.ts) via `COOKIE_MAX_AGE_SEC ` 

I'm not sure about how `session that it represents is expired in the DB` works on this repo. @benwerner01 Implemented it, maybe we can add a `@todo`, or a follow-up ticket, then ask him for more details about this?

kachkaev: `session.middleware.ts` contains a wrapper for some API calls and it connects to Mongo to retrieve a session. This happens in Node.js / Lambda, i.e. after the new `middleware.page.ts` has already triaged the requests. This means that a previously valid session id that has disappeared from Mongo DB will work the same way as a manually crafted session id.

Redirecting to `/dashboard` and then `/login` in this case is probably fine, especially if the probability of this is low. Ideally, we want to check the auth properly right in `middleware.page.ts`, but I understand that it may be a bit challenging with our current setup.

yusufkinatas: I understand this concern, and created a [follow-up ticket](https://app.asana.com/0/1201095311341924/1202854058619669/f) _(internal)_ to check this with original author of the session system

site/src/middleware.page.ts
```diff
@@ -1,3 +1,5 @@
+/* istanbul ignore file */
+/** @see https://github.com/vercel/next.js/issues/32608#issuecomment-1007439478 about istanbul ignore file */
```
kachkaev, line 2: I would not introduce `istanbul ignore` for now. Merging is not blocked when Codecov check is failing. Keeping a true picture of what’s covered and what’s not is more beneficial at this stage.

yusufkinatas: Disabled it to test if it'll fix failing integration tests, but no luck. Started a thread on slack about this.

kachkaev (approved):
👏

site/src/middleware.page.ts
benwerner01: > What may also happen is: the cookie itself is not expired, but the session that it represents is expired in the DB

@kachkaev I don't think this is a concern unless the requester spoofs the cookie expiration time, as the session expiration time (stored in DB) is the same as the HTTP cookie expiration time (stored in browser). The session expiration time is checked when the cookie is validated in `session.middleware.ts`. Therefore you're right that we would end up with the first scenario (a client-side redirect to `/login`). I think this behaviour is fine.

benwerner01: > I'm not sure about how session that it represents is expired in the DB works on this repo. @benwerner01 Implemented it, maybe we can add a @todo, or a follow-up ticket, then ask him for more details about this?

@yusufkinatas each cookie is associated with a session in the db. The `max-age` of the cookie (stored in browser) and the session expiration time (stored in db) are initialised to the same value. When a cookie is validated in `session.middleware.ts`, it's corresponding session is retrieved from the db, and the expiration time (retrieved from the db) is verified.

In practice, when the expiration time is reached the browser automatically stops sending the cookie with requests. Therefore `session.middleware.ts` is able to treat subsequent requests as unauthenticated ones, and the login flow is triggered.

If the requester spoofs cookie max-age in the browser, `session.middleware.ts` invalidates this cookie when it checks the session expiration time that is stored in the db. It does this by returning an authentication error, and it clears the HTTP-cookie in the browser using the `Set-Cookie` header in the response (I've tested this locally to confirm this is the behaviour). This also triggers the login flow.

Let me know if this clarifies the behaviour.

benwerner01: What this PR aims to accomplish (redirecting logged in users to `/dashboard` from the homepage) is tricky given we want to statically serve the homepage for logged out users, which we don't want to lose.

In order for the homepage to be served statically, we cannot make async requests that block returning the homepage (such as validating a session by retrieving it from the datastore using `getSession(req, res)`).

My instinct would've been to trigger the redirect to `/dashboard` on the client when the user is logged in (i.e. when the `/me` query resolves successfully). This would mean the homepage is statically displayed for logged out users, and for logged in users they would briefly see it and then be immediately redirected. This is suboptimal because logged in users briefly see the homepage.

To my understanding the solution in this PR provides a better behaviour than this: statically serving the homepage for logged out users, and instantly redirecting logged in users (or at least requests where the `blockprotocol-session-id` header is present). I'm happy for us to proceed with this implementation 👍

One thing I'd like to note is that there is a potential for a bug if we modify the behaviour of the session invalidation so that the session can expire before the cookie, and the cookie isn't removed from the browser automatically. This would result in users who's session has expired being redirected from `/` to `/dashboard` to `/login` every time they access the homepage in the same browser. To be clear this is not the current behaviour, but worth keeping in mind if we modify how the session management works down the line.

yusufkinatas: Thanks for the detailed explanation @benwerner01! Happy to hear that current implementation is ok!

PR #522, opened 2022-08-23
Add self-service publishing by linking npm package

CiaranMn:
# Purpose

This PR adds the ability for users to publish blocks to the Block Hub by linking to an existing block published to npm.

It also clears up some inconsistency between the `block-metadata.json` hosted from the site, and the metadata returned by the API, by adding all the extra metadata to `block-metadata.json` when building the blocks in the first place.

## How to use it

I made [test-npm-block](https://www.npmjs.com/package/test-npm-block) to be able to test this.

You will need keys for R2 - find them in 1Password.

add them to `.env.local.`
```
CLOUDFLARE_R2_ENDPOINT=xxxx
CLOUDFLARE_SECRET_ACCESS_KEY=xxxx
CLOUDFLARE_ACCESS_KEY_ID=xxxx
```

The API endpoint is `POST` `/api/blocks/create`, and it requires `blockName` (the unique slug to be used for the block in the Block Hub) and `npmPackageName` in the body.
e.g. (in the console once you are logged in)
```typescript
fetch("http://localhost:3000/api/blocks/create", {
  method: "POST",
  headers: {
    "Content-Type": "application/json",
  },
  body: JSON.stringify({
    blockName: "test-block",
    npmPackageName: "test-npm-block",
  }),
});
```

Once you add this block, check it appears alongside other blocks on the homepage slide, hub, individual block page, etc.

If the request fails, check the response. Maybe you need to choose a different author / block name if it's already been published.

You can also swap out `http://localhost:3000` for the latest preview deployment

## Other points to note
1. Block source is stored in Cloudflare R2, which requires the use of a worker, created by the code in the new `wrangler` folder.
2. Block source is stored by their author (BP username), name (chosen when calling the API), and version (from npm package). 
3. There is a bucket for development (including preview deployments and local), and production. This means that block source from different devs will be uploaded to the same bucket across all dev environments. To avoid clashes from local dev, blocks added from local dev are prefixed by git username. There can be no clashes across preview deployments because they all use the same db, and therefore once a given `author/block-name` pair has been claimed (inserted into the db), no-one else can insert it.

# Next steps

A subsequent PR will add a webhook and register it with npm to be informed of new package versions, and upload new versions as they appear – for now, overwriting the existing package, since we don't handle different versions.

Later we can expose different versions of blocks. When we introduce versioning of blocks (internal [task](https://app.asana.com/0/0/1202539910143057/f)) we will need to:
- store block distributions in a folder per version
- keep a metadata file around per version
- add the ability to request specific versions from the API (and return latest otherwise)
- stop using `componentId` as equivalent to "the folder the block source is in". we will need a new unique id per block that is valid across all versions, and an additional field on the metadata that indicates where the source is stored.

---
  - [internal task](https://app.asana.com/0/0/1202796290587278)


site/scripts/prepare-blocks.ts
```diff
@@ -489,14 +495,44 @@ const script = async () => {
...
+        },
+        timestamps: {
+          lastUpdated: now,
+        },
+        includesExampleGraph: exampleGraphFileExists,
+      });
+
+      const blockMetadataToWrite: BlockMetadataOnDisk = {
+        ...extendedBlockMetadata,
+        unstable_hubInfo: {
```
CiaranMn, line 524: add all the extended metadata here so we have one consistent set of extended metadata everywhere we make it available

site/scripts/prepare-blocks.ts
```diff
@@ -489,14 +495,42 @@ const script = async () => {
       });
 
       const blockMetadata = await fs.readJson(blockMetadataPath);
-      blockMetadata.unstable_hubInfo = {
-        ...hubInfo,
-        checksum: blockInfoChecksum,
-        commit: blockInfo.commit,
-        preparedAt: new Date().toISOString(),
-        name: blockInfo.name,
+
```
CiaranMn, line 498: Instead of a 2-stage process where we add some stuff to the metadata files at `/public/blocks`, which are served via URI, and then add some _more_ metadata to `blocks-data.json`, we just add all the extra metadata to the files in the first place here.

site/src/lib/api/wrangler/index.js
```diff
@@ -0,0 +1,30 @@
+export default {
+  async fetch(request, env) {
```
CiaranMn, line 2: We need these wranglers because we want objects in the bucket to be public, and it is not possible to just set that for the bucket.

site/src/lib/api/blocks/npm.ts
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  const now = new Date().toISOString();
+
+  // for blocks developed locally, add a prefix to the storage URL - the R2 bucket is shared across all dev environments
+  let storageNamespacePrefix = "";
+  if (!isRunningOnVercel) {
+    const exec = promisify(child_process.exec);
+
+    const gitConfigUserNameResult = await exec("git config --get user.name");
```
kachkaev, line 177: We might be able to use `execa` here, especially given that it’s already important. If we cannot, a comment explaining why would be useful.

site/src/lib/api/blocks/r2.ts
```diff
@@ -0,0 +1,97 @@
+import S3 from "aws-sdk/clients/s3";
+import fs from "fs-extra";
+import mime from "mime-types";
+import path from "node:path";
+
+import { isProduction } from "../../config";
+
+const cloudflareEndpoint =
+  "https://c6499786332a3d2fb35419a7803ab7aa.r2.cloudflarestorage.com";
```
kachkaev, line 9: Do we need to hardcode this?

kachkaev (commented):
👍

I think we can simplify a few things, but we can probably merge first and try the new storage in the wild. Eventually, it’d be nice to land on just one way of storing blocks for system coherence.

site/src/lib/api/blocks/npm.ts
```diff
@@ -0,0 +1,278 @@
...
+    pathWithNamespace,
+  )}`;
+  const publicPackagePath = `${publicBaseR2Url}/${remoteStoragePrefix}`;
+
+  const sourceInformation = {
+    blockDistributionFolderUrl: publicPackagePath,
+    npmPackageName,
+    pathWithNamespace,
+    repository:
+      typeof packageJson.repository === "object" && !!packageJson.repository
```
CiaranMn, line 199: `@todo` – repository may also be a string

site/src/lib/api/blocks/npm.ts
```diff
@@ -0,0 +1,278 @@
...
+      {
+        cause: { code: "INVALID_PACKAGE_CONTENTS" },
+      },
+    );
+  }
+
+  // check that the schema actually exists
+  const schemaPath = metadataJson.schema;
+  if (schemaPath.startsWith("http")) {
+    await fetch(schemaPath, { method: "HEAD" });
```
CiaranMn, line 155: `@todo` – this error should be caught and a friendly error message/code provided

site/src/lib/api/blocks/npm.ts
CiaranMn: done in https://github.com/blockprotocol/blockprotocol/pull/522/commits/db9a55ba883429bc77f70d061d2b78d51aba1e80

site/src/lib/api/blocks/r2.ts
CiaranMn: nope - switched to env and documented new envs in https://github.com/blockprotocol/blockprotocol/pull/522/commits/db9a55ba883429bc77f70d061d2b78d51aba1e80

site/src/lib/api/blocks/npm.ts
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/522/commits/360a52e224df3091b4e3f8cca0098a1a7f75421a

kachkaev (approved):
I was able to add three blocks to the hub as `@alice` locally.

Screenshot 2022-08-29 at 17 09 43 (2)

Publishing the same block twice resulted `HTTP 400` instead of `HTTP 200`, which made sense. We can probably improve this by returning a specific error code in a follow-up PR so that it can be displayed in the UI correctly. ATM, duplicate blocks are indistinguishable from misconfigured R2.

I see a few potential ways to simplify the flow and the local dev, which we can discuss separately.

site/src/lib/api/wrangler/README.md
```diff
@@ -0,0 +1,11 @@
+# Wrangler for R2
```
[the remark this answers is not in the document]
CiaranMn: Thanks @kachkaev – I will do this in the next PR, in the interests of getting this merged to allow Yusuf's to target `main`

PR #523, opened 2022-08-23
Switch back to the latest `mockBlockDockVersion` in sandbox

kachkaev:
Reverts blockprotocol/blockprotocol#515

We no longer need to pin MBD because the upstream issue is now resolved: https://github.com/ije/esm.sh/issues/406#issuecomment-1223330666

Extras: see comments

### DIFF
--- a/scripts/packages/e2e-test-create-block-app.ts
+++ b/scripts/packages/e2e-test-create-block-app.ts
@@ -136,7 +136,7 @@ const script = async () => {
       },
     });
 
-    await waitOn({ resources: ["http://localhost:63212"], timeout: 20000 });
+    await waitOn({ resources: ["http://localhost:63212"], timeout: 30000 });
 
     await killProcessTree(devProcess.pid!, "SIGINT");
 
--- a/site/tests/integration/blockpage.test.ts
+++ b/site/tests/integration/blockpage.test.ts
@@ -84,7 +84,9 @@ test("Block page should contain key elements", async ({
   if (browserName !== "webkit") {
     await expect(
       page.frameLocator("iframe[title='block']").locator("input"),
-    ).toBeVisible();
+    ).toBeVisible({
+      timeout: 30000, // @todo Remove after re-engineering block sandbox
+    });
     await expect(
       page.frameLocator("iframe[title='block']").locator("input"),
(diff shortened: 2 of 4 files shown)


site/tests/integration/blockpage.test.ts
```diff
@@ -84,7 +84,9 @@ test("Block page should contain key elements", async ({
   if (browserName !== "webkit") {
     await expect(
       page.frameLocator("iframe[title='block']").locator("input"),
-    ).toBeVisible();
+    ).toBeVisible({
+      timeout: 30000, // @todo Remove after re-engineering block sandbox
```
kachkaev, line 88: I matched the timeout with the one we have in universal tests after noticing that 5s is not always enough for esm.sh CDN.

site/tests/integration/docs.test.ts
```diff
@@ -16,15 +16,15 @@ test("Docs page should contain key elements and interactions should work", async
 
   await expect(page.locator('h1:has-text("Introduction")')).toBeVisible();
 
-  const sidebarSelector = page.locator(
+  const sidebarLocator = page.locator(
```
kachkaev, line 19: `const sidebarSelector` was actually a locator, so I change the variable name for clarity. [Selectors](https://playwright.dev/docs/selectors) also a thing in Playwright.

scripts/packages/e2e-test-create-block-app.ts
```diff
@@ -136,7 +136,7 @@ const script = async () => {
       },
     });
 
-    await waitOn({ resources: ["http://localhost:63212"], timeout: 20000 });
+    await waitOn({ resources: ["http://localhost:63212"], timeout: 30000 });
```
kachkaev, line 139: Helps pass CI on slow GitHub machines (which we may be assigned at random).

- [Successful run](https://github.com/blockprotocol/blockprotocol/runs/7974162736?check_suite_focus=true#step:7:169) (≈ 15s)
- [Failed run](https://github.com/blockprotocol/blockprotocol/runs/7976345609?check_suite_focus=true#step:7:189) (hitting 20s timeout)

Looking at task durations in the above runs shows that the second machine is generally slower.

teenoh (approved):

PR #525, opened 2022-08-23
Add turborepo, use it for yarn build

kachkaev:
This PR replaces `prepare` scripts with calling `turbo run build` as means to warm up the repo. This should save us a few seconds most of the times when we switch between branches. Importantly, this also fixes an issue with mass package publishing ([internal slack thread](https://hashintel.slack.com/archives/C02LG39FJAU/p1661162153010159)). The new flow are opaque to devs, so no changes in docs are necessary.

The only remaining package with `prepare` is `block-template-html`, which probably makes sense as is. This package is at a topological root so keeping `prepare` should not break publishing.

This PR also unlocks further work in  #508. Running `lint:eslint` will become a turbo dependency of running `build`.

We can improve Turborepo config in follow-up PRs – I left this out of scope for now. The most obvious next step would be configuring [global caching](https://turborepo.org/docs/core-concepts/caching), but this is not necessary for an MVP.


site/package.json
```diff
@@ -7,13 +7,13 @@
   "scripts": {
     "build": "yarn exe scripts/build.ts",
     "clean": "rimraf ./.next/",
+    "codegen": "yarn exe scripts/codegen.ts",
```
kachkaev, line 10: Extracted from `prepare`.  `patch-package` is called from the root `package.json` prepare for simplicity. This step is temporary anyway and will be replaced with [native patches](https://yarnpkg.com/cli/patch) in yarn 3.

CiaranMn (approved):
Screenshot 2022-08-24 at 08 44 27

PR #526, opened 2022-08-24
Implement block publishing flow

yusufkinatas:

This adds `Publish a block` interface for logged-in users. So it will be easier to publish new blocks to `Block Hub`. For now, we only support `Publish from npm` method. So users can publish a valid block as a `npm package`, and add that block to `Block Hub` via `Publish from npm` method.

## 🔗 Related links

- [Asana task - publish block](https://app.asana.com/0/1202727124867670/1202796290587274/f) _(internal)_
- [Asana task - list published blocks](https://app.asana.com/0/0/1202796290587276/f) _(internal)_
- [Design](https://www.figma.com/file/Fpmde3E9nJMb4VLkPTYgrM/blockprotocol.org?node-id=3895%3A29099) _(internal)_

## 🔍 What does this change?

- add a wrapper for `MUI <Alert />`, and override theme options for `Alert` component according to design.
  - also add eslint rule to prevent importing `Alert` from `mui`
- add `react-hook-form` to project, for better form validation / state handling
- update`playwright` `1.24.0` -> `1.25.1`
  -  **Reason:** To have `Pick Selector` & `Record new tests from current page state` features on Playwright VSCode Extension. 
  - checked the  [1.25 Changelog](https://playwright.dev/docs/release-notes#version-125) to make sure there is no breaking change.
  - also tried running tests locally to make sure everything works as expected after the update. 
- disable the `Button` component when `loading={true}`
- Improved `ListViewCard` component, now it supports `extraContent`, and could be used with custom content. (see [pages/blocks/index.page.tsx](https://github.com/blockprotocol/blockprotocol/blob/a55bc705aa0eb82f31e33678d59608f0afc5c886/site/src/pages/blocks/index.page.tsx#L78-L91))
- create `formatUpdatedAt` helper function
- add `error?.code` to `ApiClientError`
- add `bpCode` typography variant for `code` style
- Add `green` & new purple variants to `palette`

## 🐾 Next steps

- Implement `SWR` for data fetching logic instead of manually handling following:
  - Setting data to state
  - Handling errors
  - Loading status

## 🛡 What tests cover this?

- Integration tests at `site/tests/integration/block-publish-npm.test.ts`

## ❓ How to test this?

1.  Create a new use
2. Open dashboard
3. Select `Blocks` tab from above
4. Confirm the empty state of the blocks list
5. Publish a new block via npm. Here is a [npm package](https://www.npmjs.com/package/test-npm-block) for testing this functionality.
6. Get redirected back to `Blocks` page after publishing
7. Confirm the filled state of the blocks list


site/src/components/pages/dashboard/top-navigation-tabs.tsx
```diff
@@ -43,6 +42,7 @@ export const TopNavigationTabs: FunctionComponent = () => {
               component="a"
               onClick={(event: MouseEvent) => {
                 event.preventDefault();
+                void router.push(tabHref);
```
yusufkinatas, line 45: So it always opens the tab's href, even if tab is already active. 
Which means, when user is on `/blocks/publish`, clicking the `Blocks` tab will take the user back to the tabHref, which is `/blocks`

site/src/components/pages/user/list-view-card.tsx
```diff
@@ -1,107 +1,103 @@
-import { Box, Typography } from "@mui/material";
```
yusufkinatas, line 1: On this file, basically added `extraContent` instead of `lastUpdated` for customizability. And removed the extra `Box` wrapping everything.

site/src/components/text-field.tsx
```diff
@@ -11,9 +11,7 @@ import { FunctionComponent, ReactNode, useEffect, useState } from "react";
 
 import { FontAwesomeIcon } from "./icons";
 
-type TextFieldProps = {
-  displayErrorOnTouched?: boolean;
-} & MuiTextFieldProps;
```
yusufkinatas, line 16: this was unused

site/src/lib/api-client.ts
```diff
@@ -44,19 +48,21 @@ const axiosClient = axios.create({
 });
 
 export type ApiClientError = AxiosError<{
-  errors?: Partial<ValidationError>[];
+  errors?: Partial<ValidationError & { code?: string }>[];
```
yusufkinatas, line 51: not sure if this is the best way of doing this. we use `error.code` to have predefined error types, on top of `error.message`. Because we need the logic for `depending on the type of the error, show the error message next to the related text field`. Messages could change, they're not reliable. Code's will not change unless we change the logic of BE.

site/src/components/button.tsx
```diff
@@ -80,10 +80,11 @@ export type ButtonProps = {
 // probably rename to BPButton?
 // @todo implement loading
```
teenoh, line 81: This wasn't introduced in this PR, but please can you help remove this. `loading` has been implemented here. The renaming comment is also no longer needed

site/src/theme/typography.ts
```diff
@@ -95,4 +95,15 @@ export const typography: ThemeOptions["typography"] = {
     lineHeight: 1.1,
     color: customColors.gray[90],
   },
+  bpCode: {
+    fontFamily: [`"JetBrains Mono"`, ...fallbackFonts].join(", "),
```
teenoh, line 99: Let's also add `monospace` as a fallback

site/src/components/pages/blocks/block-list-container.tsx
```diff
@@ -0,0 +1,23 @@
+import { experimental_sx as sx, styled } from "@mui/material";
+
+import { BlockFormContainer } from "./block-form-container";
+
+export const BlockListContainer = styled(BlockFormContainer)<{
+  hasBlocks?: boolean;
+}>(
```
teenoh, line 7: nit: I don't think we need to create a new file for component style overrides like this that have few lines and are used in one place. I feel it'll quickly become difficult to find components that cause certain styles since one would have to jump through files just to update a particular style.   I'd prefer to place them in the same file they are used if the styles are small like this or better still have a single `block-form-styles` file that exports all the style-based components needed in block form. 

I think it's better we create a new file for styled components when 
1) It has a lot of styles 
2) It is reused in more than one component

CiaranMn (commented):
Thanks @yusufkinatas!

Could you merge `main` into this branch? Mine has merged now, but with different commits so this has the old ones from that.

There are also some issues I've encountered with registering an npm webhook with Vercel, which means it isn't ready to expose to users – can you feature flag the npm publishing part of this please?

e.g. we could have an environment variable, `NPM_PUBLISHING`, and if that's true the 'publish' button takes the user to the new screen. If it's false/unset we just take them to the same place as before.

the other changes, e.g. blocks tab, can still be exposed as they don't depend on npm publishing

site/src/components/text-field.tsx
```diff
@@ -11,9 +11,7 @@ import { FunctionComponent, ReactNode, useEffect, useState } from "react";
...
+type TextFieldProps = {} & MuiTextFieldProps;
```
teenoh, line 14:
```suggestion
type TextFieldProps = MuiTextFieldProps;
```

site/src/components/pages/hub/block-data-container.tsx
```diff
@@ -364,9 +364,15 @@ export const BlockDataContainer: FunctionComponent<BlockDataContainerProps> = ({
             }}
             onClose={() => setAlertSnackBarOpen(false)}
           >
-            <Alert severity="warning">
-              Please fix the errors in the <b>Data Source</b> before proceeding.{" "}
-            </Alert>
+            <Alert
+              description={
+                <>
+                  Please fix the errors in the <b>Data Source</b> before
```
teenoh, line 370: Not related to this PR but can you do a driveby fix by renaming this to "Block Properties"

```suggestion
                  Please fix the errors in <b>Block Properties</b> before
```

teenoh (commented):
Thanks @yusufkinatas! I dropped some comments which can either be addressed here or in a follow up. Other than those and the conflicts, this looks good to me

yusufkinatas:
> Thanks @yusufkinatas!
> 
> Could you merge `main` into this branch? Mine has merged now, but with different commits so this has the old ones from that.
> 
> There are also some issues I've encountered with registering an npm webhook with Vercel, which means it isn't ready to expose to users – can you feature flag the npm publishing part of this please?
> 
> e.g. we could have an environment variable, `NPM_PUBLISHING`, and if that's true the 'publish' button takes the user to the new screen. If it's false/unset we just take them to the same place as before.
> 
> the other changes, e.g. blocks tab, can still be exposed as they don't depend on npm publishing
@CiaranMn Added the env variable here https://github.com/blockprotocol/blockprotocol/pull/526/commits/ba79a45ba27067a2c38b2b104f45e2dc7e02015e

site/src/components/pages/user/tab-panel-contents-with-blocks.tsx
```diff
@@ -50,11 +51,10 @@ export const TabPanelContentsWithBlocks: FunctionComponent<
         }) => (
           <ListViewCard
             key={name}
-            type="block"
             icon={icon}
             title={displayName!}
             description={description}
-            lastUpdated={lastUpdated}
+            extraContent={formatUpdatedAt(lastUpdated)}
```
CiaranMn, line 57: We might as well reuse the `extraContent` for the blocks displayed here and in the `/blocks` route, unless there's some reason to hide the `version` and `author` here but not there.

site/src/pages/blocks/index.page.tsx
```diff
@@ -0,0 +1,105 @@
...
+                  <Box display="flex" gap={1.5}>
+                    <Link href={`/@${block.author}`}>@{block.author}</Link>
+                    <span>{`V${block.version}`}</span>
+                    <span>{formatUpdatedAt(block.lastUpdated)}</span>
+                  </Box>
+                }
+              />
+            ))
+          ) : (
+            <BlockListEmptyState />
```
CiaranMn, line 97: This will always flash on first load – can we introduce a loading state instead to avoid displaying "You haven't published any blocks" to everyone, even if they have?

CiaranMn (commented):
I tested this out and have a couple of small comments – I particularly think we should get rid of the flash of "you haven't published any blocks" since it can be inaccurate

site/src/pages/blocks/index.page.tsx
yusufkinatas: Thanks for the heads up, forgot about loading state somehow 🤦 
Addressed here https://github.com/blockprotocol/blockprotocol/pull/526/commits/f9968db0cde28c0704902c5535696e9e432e90b7

site/src/components/pages/user/tab-panel-contents-with-blocks.tsx
yusufkinatas: Makes sense, why not?
Addressed here https://github.com/blockprotocol/blockprotocol/pull/526/commits/f9968db0cde28c0704902c5535696e9e432e90b7

PR #527, opened 2022-08-24
Change Base URI parsing logic, and Type System Package improvements

Alfred-Mountfield:
This PR modifies the `BaseUri` parsing logic. We have decided it would be dangerous to convert URIs into standard form, as various URL libraries may have differing interpretations of the spec, and this could be very problematic when trying to resolve equality of types, finding data in a system, etc. Instead, it makes sense for the Block Protocol to accept the URL as given, only checking to ensure it is a valid URL.

**An important characteristic of the current definition is that all Base URIs must have a trailing slash**

This also adds two functions exposed to Typescript to extract the base URI and version of a versioned URI.
Also adds some drive-by fixes such as renaming expect strings to be lower-case, removing unnecessary results, moving the location of clones, etc..


TimDiekmann (approved):

thehabbos007 (approved):
We talked briefly about revisiting the void/throw return value of `isValidBaseUri`. No need to change it here, but we definitely want to look at alternatives to throwing :+1:

PR #533, opened 2022-08-25
Add plural titles to PropertyType

Alfred-Mountfield:
For UI convenience, it would be nice to be able to display a plural name for property types (and entity and link types), this adds an optional field to the type to do just that.

We could also consider implementing this for data types.

- [Asana task](https://app.asana.com/0/1202787635216884/1202848989198301/f) _(internal)_


TimDiekmann (approved):

TimDiekmann (approved):

thehabbos007:
When we bump the dependency in HASH, we should be aware that we need to update `/home/ahmad/dev/hash/hash/packages/graph/hash_graph/lib/graph/src/api/rest/json_schemas/property_type.json` and the underlying GQL resolvers in the workspace api. Do we have a task to capture this?

PR #534, opened 2022-08-25
Remove template on VersionedUri

Alfred-Mountfield:
We had an experimental approach where we were constraining the type of `VersionedUri` a little bit further in TypeScript by using a template literal.

This provided some benefit but it also turns out that it makes the library harder to use when trying to convert between different interfaces, making it unergonomic where you have to destructure the object so you can call `isVersionedUri` even when the type has already been checked, due to coercion. 

This PR removes the template type. We might bring it back in the future, as the change is small and easy to reimplement

### DIFF
--- a/crates/type-system/src/ontology/uri/wasm.rs
+++ b/crates/type-system/src/ontology/uri/wasm.rs
@@ -8,10 +8,10 @@ use crate::{
     uri::{error::ParseVersionedUriError, VersionedUri},
 };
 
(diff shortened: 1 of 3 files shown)


Alfred-Mountfield:
It would seem that this isn't as bad to get around as we thought. Going to close off this PR, we can revisit if it does turn out to be unergonomic as we use the packages more.

PR #528, opened 2022-08-24
Run `lint:tsc` via Turborepo

kachkaev:
Extracted from #508 to keep PR scope focused. Continues https://github.com/blockprotocol/blockprotocol/pull/525.

- saves us from creating dummy `lint:tsc` in some existing and future packages (requirement of `yarn workspaces`)
- speeds up repeated calls to `yarn lint:tsc`:

```
 Tasks:    15 successful, 15 total
Cached:    15 cached, 15 total
  Time:    280ms >>> FULL TURBO
```

PR #537, opened 2022-08-25
Update Prettier and markdownlint, use Prettier cache, cleanup ignore files

kachkaev:
The original goal of this PR was to configure Prettier and markdownlint via Turborepo, similar to what’s done in #528 and #508. Prettier and markdownlint scripts are called from root, so `turbo.json` would contain the following:

```json
    "//#lint:markdownlint": {
      "dependsOn": [],
      "outputs": []
    },
    "//#lint:prettier": {
      "dependsOn": [],
      "outputs": []
    },
```

[docs](https://turborepo.org/blog/turbo-1-3-0#run-and-cache-scripts-from-the-root-of-your-monorepo)

The savings ended up quite small, so I re-scoped my branch to the following changes:
- Update Prettier and markdownlint (including Prettier plugins)
- Enable `--cache` in `lint:prettier` and `fix:prettier` ([introduced](https://prettier.io/blog/2022/06/14/2.7.0.html#cli) in Prettier 2.7)
- Sync and cleanup `.*ignore` files (to reduce diff in #508)


.markdownlint.json
```diff
@@ -5,6 +5,7 @@
     "siblings_only": true
   },
   "line-length": false,
+  "link-image-reference-definitions": false,
```
kachkaev, line 8: This new rules is incompatible with custom header slugs in RFCs, e.g.:

```md
# Summary

[summary]: #summary

One paragraph explanation of the feature.
```

Playground: https://dlaa.me/markdownlint/

rfcs/text/0352-graph-type-system.md
```diff
@@ -2682,14 +2682,14 @@ A more immediately apparent concern is if schemas are versioned, but their use i
 
 ## The Big Picture
 
-As outlined in the [Main Motivation section](#main-motivation), there are some clear shortcomings to how the Graph service works at the moment in regards to how it expresses data structure. It's also important to mention that there is a tough balance between comprehensive functionality for:
+As outlined in the [Motivation section](#motivation), there are some clear shortcomings to how the Graph service works at the moment in regards to how it expresses data structure. It's also important to mention that there is a tough balance between comprehensive functionality for:
```
kachkaev, line 2685: @Alfred-Mountfield soes this change make sense to you? Spotted by a new default rule called [`link-fragments`](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md051).

Alfred-Mountfield: I think it makes sense. I don't see a section called _Main_ motivation so it was likely a mistake. I thought we were ignoring RFCs for lints though to avoid introducing too much overhead for people to get them to pass, especially as they're just technical docs?

nonparibus: +1 to this suggestion. I feel like we'd be better off linting RFCs ourselves (manually) _at the point of acceptance_ when we're ready to merge in to the repo, and it'd be preferable if lint tests weren't run (or more precisely weren't at risk of failing) by users simply iterating on drafts.

kachkaev: We can think about CI triggers separately. AFAIR, dev.azure.com does not run CI for PR drafts but they can be triggered manually. Since Microsoft bought GitHub, they borrowed some useful features from dev.azure.com, so maybe there is an option to configure something similar. I would not focus on this right now since we don’t have external RFCs yet.

Alfred-Mountfield: I think it's more that we can just add RFCs to the ignore files for now

packages/@local/eslint-config/index.js
```diff
@@ -303,7 +303,7 @@ module.exports = {
     {
       files: ["site/**"],
       parserOptions: {
-        tsconfigRootDir: `${__dirname}/site`,
+        tsconfigRootDir: `${__dirname}/../../../site`,
```
kachkaev, line 306: Regression from #508, spotted by @yusufkinatas. We resolved it locally in DMs. Attaching the fix here as it does not seem worth a separate PR.

rfcs/text/0352-graph-type-system.md
CiaranMn: Since we don't currently ignore RFCs, and linting them has actually spotted a broken link, I don't think we need to start ignoring them now.

People can ignore linting failures until they actually want to merge them, so I don't see it as introducing any real overhead.

CiaranMn (approved):

rfcs/text/0352-graph-type-system.md
Alfred-Mountfield: > CI can fail not only because of markdownlint, but prettier too. It needs to be also disabled for RFCs following the above logic

Well I do think there is some argument that we don't necessarily need `prettier` for RFCs either.

But I agree with this:
Since we don't currently ignore RFCs, and linting them has actually spotted a broken link, I don't think we need to start ignoring them now.

PR #507, opened 2022-08-18
`mock-block-dock` improvements

teenoh:
This PR adds some visual and functionality improvements to `mock-block-dock`'s debug mode.

Features
- Improved UI 
- Log view of messages sent between embedder and block with filters
- Light/dark mode theme switcher

Next Up
- Allow editing of block properties from `mock-block-dock` debug UI

https://user-images.githubusercontent.com/16928688/185547861-1f8985ba-1d5c-476e-94e2-c13ac71dda1a.mov

PR #539, opened 2022-08-26
`mock-block-dock` improvements

teenoh:
### Description
This PR builds upon #507 and adds the following improvements to `mock-block-dock` =>

- Debug UI has been improved to match designs
- `bottom-view` has been renamed to `dev-tools`
- `dev-tools` can be resized and its final position is saved
- Block schema is now displayed in the `Properties` tab if present and can be uploaded if not
- Block properties can be updated from the debug UI and validated against the set block schema
- Debug mode can switched on/off

### Next Up
- Handle switching block entity from debug UI
- Update block templates to pass `blockInfo` into `mock-block-dock`

### Demo

https://user-images.githubusercontent.com/16928688/187208845-0a84818b-1f7a-4a04-8542-9908044bd284.mov


package.json
```diff
@@ -29,7 +29,7 @@
     "dev": "yarn workspace @blockprotocol/site dev",
     "dev:seed-db": "yarn workspace @blockprotocol/site dev:seed-db",
     "dev:db": "yarn workspace @blockprotocol/site dev:db",
-    "prepare": "patch-package && yarn workspace @blockprotocol/site patch-package && turbo run codegen && turbo run build \"--filter=!./site/\""
```
teenoh, line 32: This was required to patch `ajv` in `site`. It's no longer needed since the `ajv` patch has been moved to the root `patches` directory. It was moved because we also make use of ajv in `mock-block-dock` and have the apply the same patch there

packages/mock-block-dock/dev/dev-app.tsx
```diff
@@ -88,6 +95,7 @@ const DevApp: FunctionComponent = () => {
         debug
         blockDefinition={blockDefinition}
         blockEntity={blockEntity}
+        blockName="Test Block"
```
teenoh, line 98: There's a new `blockName` prop which is displayed at the top of the debug UI. This property will be pulled from `blockprotocol.displayName` field in the block's package.json file. See [internal link](https://www.figma.com/file/Fpmde3E9nJMb4VLkPTYgrM?node-id=3896:28985#252029946)

teenoh: @CiaranMn what do you think about adding a `blockInfo` prop which is the content pulled from "blockprotocol" key in the block's package.json?

CiaranMn: Yes - I think we should do that in our block templates

packages/mock-block-dock/package.json
```diff
@@ -24,12 +24,16 @@
...
+    "build:cjs": "tsc --p tsconfig.build.json --module commonjs --outDir dist/cjs",
     "build:dev": "yarn clean && cross-env NODE_ENV=development webpack --mode development",
     "build:esm": "tsc --p tsconfig.build.json",
     "clean": "rimraf ./dist/",
-    "dev": "concurrently -n webpack,webpack-dev-server -c green,cyan \"yarn build:dev --watch\" \"yarn run-dev-server\"",
+    "copy-files": "yarn copy-files:cjs && yarn copy-files:esm",
+    "copy-files:cjs": "copyfiles -u 1 ./src/assets/* dist/cjs/",
+    "copy-files:esm": "copyfiles -u 1 ./src/assets/* dist/esm/",
+    "dev": "yarn generate-version-file && concurrently -n webpack,webpack-dev-server -c green,cyan \"yarn build:dev --watch\" \"yarn run-dev-server\"",
+    "generate-version-file": "node -p \"'export const MOCK_BLOCK_DOCK_VERSION = ' + JSON.stringify(require('./package.json').version) + ';'\" > src/version.ts",
```
teenoh, line 36: We need to display the package version in the debug UI. Pulling the version from package.json affects the build output since it also tries to bundle package.json. This creates a `version.ts` file that contains the package's version. It's called when we run `dev` and `build` so that the file is always present

packages/mock-block-dock/package.json
```diff
@@ -39,10 +43,12 @@
     "@emotion/react": "^11.10.0",
     "@emotion/styled": "^11.10.0",
     "@lit-labs/react": "^1.0.4",
-    "@mui/icons-material": "^5.8.4",
```
teenoh, line 42: Removed mui/icons. See [internal slack comment](https://hashintel.slack.com/archives/C02LG39FJAU/p1661410388081759?thread_ts=1661346344.291039&cid=C02LG39FJAU)

packages/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -170,70 +175,60 @@ export const MockBlockDock: FunctionComponent<MockBlockDockProps> = ({
...
-      </Box>
-
-      <BottomView
-        graphProperties={propsToInject}
-        datastore={datastore}
-        readonly={debugReadonly}
-        setReadonly={setDebugReadonly}
-      />
-    </Layout>
+    <MockBlockDockProvider
```
teenoh, line 204: Wrapped both modes in a provider, so things like logs doesn't get clear when we switch on/off debug mode

packages/mock-block-dock/package.json
kachkaev, line 36: Optional name change: 

```suggestion
    "codegen": "node -p \"'export const MOCK_BLOCK_DOCK_VERSION = ' + JSON.stringiy(require('./package.json').version) + ';'\" > src/version.ts",
```

We already have `codegen` script in `site` and it is launched via turbo. If you look at `turbo.json`, you'll see that `build` already depends on `codegen`. This means that you won't have to include `yarn generate-version-file` in `build`. We can create the same  relationship between `codegen` and `dev` too.

kachkaev (commented):
Maybe I’m doing something really stupid but I’m getting this error when running ` yarn workspace mock-block-dock dev`:

```
[webpack] asset main.js 662 bytes [emitted] (name: main)
[webpack] 
[webpack] ERROR in main
[webpack] Module not found: Error: Can't resolve './src' in '/Users/ak/-/projects/hash/github/blockprotocol/packages/mock-block-dock'
[webpack] resolve './src' in '/Users/ak/-/projects/hash/github/blockprotocol/packages/mock-block-dock'
 ```
 
 Happy to try again in the morning.

File: packages/mock-block-dock/package.json (hunk 3 above)
Comment (teenoh) [reply packages/mock-block-dock/package.json:36]: Nice!

Comment (luisbettencourt) [reply packages/mock-block-dock/package.json:36]: Tested `dev`, `build` and `generate-version-file` on cmd.exe and they all seem to work.
`dev` printed an error (but worked anyway): 
```
[webpack] Module not found: Error: Can't resolve './src' in 'C:\Users\lmnbe\Desktop\blockprotocol\packages\mock-block-dock'
[webpack] resolve './src' in 'C:\Users\lmnbe\Desktop\blockprotocol\packages\mock-block-dock'
[webpack]   using description file: C:\Users\lmnbe\Desktop\blockprotocol\packages\mock-block-dock\package.json (relative path: .)
[webpack]     Field 'browser' doesn't contain a valid alias configuration
```

File: packages/mock-block-dock/dev/dev-app.tsx (hunk 2 above)
Comment (teenoh) [reply packages/mock-block-dock/dev/dev-app.tsx:98]: Updated MockBlockDock to accept a blockInfo prop in https://github.com/blockprotocol/blockprotocol/pull/555. Will update block templates in  #555

File: packages/mock-block-dock/package.json (hunk 3 above)
Comment (teenoh) [reply packages/mock-block-dock/package.json:36]: Thanks. @kachkaev also pointed out this issue yesterday. I will try to address in #555 or create a follow up task

Comment (teenoh) [reply packages/mock-block-dock/package.json:36]: I will address in #555

**Comment by teenoh** [comment]
> Thanks @teenoh! This looks great and has some awesome new features.
> 
> I had a play and have a couple of requests on the editing properties – I don't know which are possible in the library we're using, so don't worry if it's a big pain:
> 
> 1. Can we add tooltips to the icons to tell users what they do? Even a native HTML 'title' is fine
> 2. Can we not show delete icons if they're not allowed?
> 3. Is it possible to stop the icons moving about when the 'copy' icon changes to success?
> 
> Could we also center align these vertically? Screenshot 2022-08-31 at 09 21 33

Thanks @CiaranMn, I will address these in #555

File: .gitignore [hunk 6]
```diff
@@ -50,6 +50,8 @@ site/public/blocks
 site/public/schemas
 site/site-map.json
 site/test-results
+# this file is auto-generated when build or dev command is run
+packages/mock-block-dock/src/version.ts
```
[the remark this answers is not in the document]
teenoh: Done here => https://github.com/blockprotocol/blockprotocol/pull/539/commits/5f513df16fbd7a5e7c4018ef805f4e546ac18f00

packages/mock-block-dock/package.json
teenoh: Renamed to codegen in 5f513df16fbd7a5e7c4018ef805

teenoh:
> Maybe I’m doing something really stupid but I’m getting this error when running ` yarn workspace mock-block-dock dev`:
> 
> ```
> [webpack] asset main.js 662 bytes [emitted] (name: main)
> [webpack] 
> [webpack] ERROR in main
> [webpack] Module not found: Error: Can't resolve './src' in '/Users/ak/-/projects/hash/github/blockprotocol/packages/mock-block-dock'
> [webpack] resolve './src' in '/Users/ak/-/projects/hash/github/blockprotocol/packages/mock-block-dock'
> ```
> 
> Happy to try again in the morning.

Thanks, I can also reproduce the bug, I will look at addressing this in #555

.gitignore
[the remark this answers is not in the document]
kachkaev: > We should copy this line to other .*ignore files to keep them aligned.

Let's do that as well. We can also remove a double new line that got introduced in 5f513df16fbd7a5e7c4018ef805f4e546ac18f00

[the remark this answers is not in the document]
teenoh: Thanks for catching. Done here => [571c303](https://github.com/blockprotocol/blockprotocol/pull/539/commits/571c303a96e9b4216a840c27fb3fbecee13da3ab)

CiaranMn (approved):

packages/mock-block-dock/package.json
teenoh: Thanks!

PR #543, opened 2022-08-29
Implement `useSnackbar` hook

yusufkinatas:

This adds `useSnackbar` hook, which allows us to easily show snackbars without rendering anything in the component's `jsx`. 

_Originally, this was implemented & used at https://github.com/blockprotocol/blockprotocol/pull/526, but then design changed for that PR, and removed usage of `useSnackbar`. That's why I opened a seperate PR for this. We can consider merging this to keep `useSnackbar` in codebase, or keep this PR open until we need it._

## 🔍 What does this change?

- Adds a custom `useSnackbar` hook. 
- Global snackbar configuration / management is handled by [notistack library](https://github.com/iamhosseindhv/notistack), ([suggested by MUI](https://mui.com/material-ui/react-snackbar/#notistack))
- It adds another layer on top of the `notistack` hook for make it even easier to use. 

### With `notistack`'s hook
```js
import { useSnackbar } from 'notistack';

const Component = () => {
  const { enqueueSnackbar } = useSnackbar();

  const onError = () => enqueueSnackbar("Oops!", { variant: "error" });

  // rest of the component
};
```
### This the custom hook
```js
import { useSnackbar } from "components/hooks/use-snackbar";

const Component = () => {
  const snackbar = useSnackbar();

  const onError = () => snackbar.error("Oops!");

  // rest of the component
};
```
 
## 📹 Demo (taken from [notistack](https://github.com/iamhosseindhv/notistack))
68747470733a2f2f692e696d6775722e636f6d2f4d74696a76414b2e676966


kachkaev (commented):
If there is no immediate need to use snackbar, we can close this PR and recycle the diff later if needed. Even if we delete the branch. GitHub will still preserve the PR diff.
Adding something we don't plan to use immediately adds maintenance costs, which are probably higher than the cost of rebasing, if needed 🙂

yusufkinatas:
@kachkaev Makes sense, closing it, we can take the code from diff if we need to in future, when we decide to introduce `snackbar`s to the project

yusufkinatas:
@nonparibus What is the expected outcome from this PR now? Should we merge it, keep it open until we want to use a snackbar somewhere, or sth else?

nonparibus:
> @nonparibus What is the expected outcome from this PR now? Should we merge it, keep it open until we want to use a snackbar somewhere, or sth else?

There's a slack message about this somewhere w/ the rationale (just had a quick look but couldn't find it), but the TL;DR is we can merge it.

kachkaev (approved):
We don’t have immediate plans to use this feature so I’d personally keep this diff out of `main` for now.  Since @nonparibus has reopened the PR, we can merge it as is too. We can potentially improve `useSnackbar` later when (if?) the new hook is needed in practice.

PR #546, opened 2022-08-30
Improved Errors in Package System Package

Alfred-Mountfield:
This PR improves error handling and reporting for the Type System packages.

We previously used `try_from` implementations _within_ serde serialization/deserialization methods. This reduced a large amount of code duplication, but unfortunately `serde_json` turns your custom error types into a plain string. It then also manipulates this string by appending other information and such depending on the context. This makes it near impossible (certainly not in a safe and maintainable fashion) to extract your original error _type_ out. Some of the earlier commits on this PR show some attempts.

Due to this, the bulk of this PR is separating our representations of the elements of the ontology into their `repr` and non-`repr` forms. All `repr` types are re-exported at the top-level under `crate::repr`. 

These `repr` objects are simply representations of the JSON objects, `VersionedUri`s are represented as strings for example. A `TryFrom` is implemented to convert the `repr` into the non-`repr` forms, and an `Into` is implemented to convert back. These implementations are then used within custom `FromStr` implementations on the non-`repr` types, which allows us to keep our custom error types.

After separating all of the types out, this PR expanded the majority of our previously unspecified error types, as well as introducing a number of new ones. These types have `this_error::Error` implementations to handle their display implementations.

These errors are then exposed through WASM using a custom `Result` type. This type is used to bring over some Rust philosophy to TypeScript, allowing `validate*` functions to _return_ error messages (strongly typed) rather than _throwing_ them. This custom `Result` type required us to fork `tsify` due [to a bug](https://github.com/madonoharu/tsify/issues/6).

This PR then updates all of the existing `validate*` methods to return the new error messages, and updates all the tests to reflect the new mechanisms, as well as creating new tests.

**Disclaimer**
The Git Commit history is properly messed up on this one. I wanted to rebase to clean it up, but due to a large amount of refactoring, moving, and renaming, such a rebase would have been too much work.


crates/type-system/src/ontology/data_type/mod.rs
```diff
@@ -94,11 +74,38 @@ impl DataType {
...
 
-#[cfg_attr(target_arch = "wasm32", derive(Tsify))]
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
-#[serde(deny_unknown_fields)]
+impl FromStr for DataType {
+    type Err = ParseDataTypeError;
+
+    fn from_str(data_type_str: &str) -> Result<Self, Self::Err> {
+        let data_type_repr: repr::DataType = serde_json::from_str(data_type_str)
+            .map_err(|err| ParseDataTypeError::InvalidJson(err.to_string()))?;
```
TimDiekmann, line 82: Do we want to classify the JSON error further? This might be useful for the user as it's a difference, if the JSON is malformed or if e.g. a list was expected but an object was passed.

crates/type-system/src/ontology/data_type/wasm.rs
```diff
@@ -1,25 +1,40 @@
-#![allow(
+#![expect(
     non_snake_case,
-    reason = "We want camelCase names (variables, functions, args) in TS, and this module is WASM \
-              only"
+    reason = "We want camelCase names (variables, functions, args) in TS, and this module is WASM
+               only"
```
TimDiekmann, line 4:
```suggestion
              only"
```

crates/type-system/src/ontology/mod.rs
```diff
@@ -6,10 +6,43 @@
...
+    object::{error::ParsePropertyTypeObjectError, Object},
+    one_of::{error::ParseOneOfError, OneOf},
+    validate::{ValidateUri, ValidationError},
+};
+
+// Re-export the repr contents so they're nicely grouped and so that they're easier to import in
+// a non-ambiguous way where they don't get confused with their non repr counterparts.
+// For example, `import crate::ontology::repr` lets you then use `repr::DataType`
+#[expect(
+    clippy::redundant_pub_crate,
```
TimDiekmann, line 35: We replaced this lint in `error-stack` with `unreachable_pub`

crates/type-system/src/ontology/shared/one_of/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// # Errors
+    ///
+    /// - [`ValidationError`] if the object is not in a valid state.
+    pub fn new<U: Into<Vec<T>>>(one_of: U) -> Result<Self, ParseOneOfError> {
+        let one_of = Self::new_unchecked(one_of);
+        one_of
+            .validate()
+            .map_err(ParseOneOfError::ValidationError)?;
+        Ok(one_of)
+    }
```
TimDiekmann, line 30: This can only return `ValidationError`, not `ParseOneOfError`

crates/type-system/src/lib.rs
```diff
@@ -32,25 +32,22 @@
     clippy::use_debug,
     clippy::verbose_file_reads
 )]
-#![allow(clippy::use_self, reason = "Too many false positives")]
-#![allow(
+#![expect(clippy::use_self, reason = "Too many false positives")]
+#![expect(
```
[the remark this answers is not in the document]
Alfred-Mountfield: I have not noticed that, no. Do you have a config like this?
image

crates/type-system/src/ontology/data_type/mod.rs
Alfred-Mountfield: **At the moment** I don't think it's worth the time investment. JSON errors shouldn't really be possible if people are using the TypeScript types (or Rust types, or any other bindings we may produce)

crates/type-system/src/ontology/shared/one_of/mod.rs
Alfred-Mountfield: I don't understand your comment. It returns `ValidationError` which is wrapped inside `ParseOneOfError`, as it is a variant of a problem with the `OneOf` (this is the `OneOf::new` function after all).

crates/type-system/src/lib.rs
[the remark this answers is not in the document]
TimDiekmann: Yes, and as additional parameters I set `--tests --workspace --all-features`.

image
image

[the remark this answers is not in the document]
Alfred-Mountfield: Perhaps its to do with the workspace or all-features args.

crates/type-system/src/ontology/mod.rs
Alfred-Mountfield: Have done so, although it behaved a little weirdly.

crates/type-system/src/ontology/shared/one_of/mod.rs
TimDiekmann: OneOf::new will only construct a single variant of `ParseOneOfError`, so the code could be changed to
```rust
    pub fn new<U: Into<Vec<T>>>(one_of: U) -> Result<Self, ValidationError> {
        let one_of = Self::new_unchecked(one_of);
        one_of.validate()?;
        Ok(one_of)
    }
```

TimDiekmann (approved):
I can't say I like the workaround with `repr` (and I know that this was my suggestion), but this results in a very clear public API interface. I'd like to see some more documentation on the types on how it's expected to look as JSON without looking at the `repr` implementation, but this is not blocking the PR. E.g. when looking at `DataType`, it's not clear, how the JSON in `DataType::try_from(json!{ ... })` should look like.

Thanks for the immense work that went into this PR. The error messages on the TS page are much better now!

crates/type-system/src/ontology/data_type/wasm.rs
```diff
@@ -1,25 +1,40 @@
...
+ *
+ * @param {DataType} dataType - The Data Type object to validate.
+ * @returns {(Result.Ok|Result.Err<ParseDataTypeError>)} - an Ok with null inner if valid, or an Err with an inner ParseDataTypeError  
+ */
+export function validateDataType(dataType: DataType): Result<undefined, ParseDataTypeError>;
+"#;
+#[wasm_bindgen(skip_typescript, js_name = validateDataType)]
+pub fn validate_data_type(data_type_obj: &JsValue) -> JsValue {
+    #[cfg(debug_assertions)]
+    set_panic_hook();
```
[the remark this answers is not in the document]
Alfred-Mountfield: This is mostly an experiment, I think we'll want to see if we even want to keep it. You can see the description here:
https://rustwasm.github.io/docs/wasm-pack/tutorials/npm-browser-packages/template-deep-dive/src-utils-rs.html

We could also expose our own initialize function which the user has to call, but in the nodejs target you don't currently have to call `init` to use the library so I'd like to avoid that if possible.

packages/@blockprotocol/type-system-node/dist/index.d.ts
```diff
@@ -1,64 +1,124 @@
...
-    export type Value<T> = T;
-    export type Array<T> = __ValueOrArrayArray<T>;
+type __ValidationErrorBaseUri = BaseUri;
+type __ValidationErrorVersionedUri = VersionedUri;
+declare namespace ValidationError {
+    export type MissingRequiredProperty = { type: "MissingRequiredProperty"; inner: __ValidationErrorBaseUri };
+    export type BaseUriMismatch = { type: "BaseUriMismatch"; inner: { base_uri: __ValidationErrorBaseUri; versioned_uri: __ValidationErrorVersionedUri } };
+    export type MissingRequiredLink = { type: "MissingRequiredLink"; inner: __ValidationErrorVersionedUri };
+    export type MismatchedPropertyCount = { type: "MismatchedPropertyCount"; inner: { actual: number; expected: number } };
+    export type EmptyOneOf = { type: "EmptyOneOf"; inner?: null };
```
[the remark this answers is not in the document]
Alfred-Mountfield: See the issue linked in the PR description: https://github.com/madonoharu/tsify/issues/6

packages/@blockprotocol/type-system-node/test/data-type.test.ts
```diff
@@ -48,45 +48,74 @@ const primitiveDataTypes: DataType[] = [
...
+  ],
+  [
+    {
+      kind: "dataType",
+      $id: "https://blockprotocol.org/@blockprotocol/types/data-type/text/v/1",
+      title: "Text",
+    },
+    {
+      reason: "InvalidJson",
+      inner: "missing field `type` at line 1 column 108",
```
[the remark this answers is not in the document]
Alfred-Mountfield: That's just serde_json's display implementation. https://docs.rs/serde_json/latest/src/serde_json/error.rs.html#339-342

PR #548, opened 2022-08-30
Allow users to publish blocks directly to API

CiaranMn:
# What does this

Allows users to publish blocks by supplying a name and a tarball of source, as another publishing route in addition to linking an existing npm package.

This paves the way for introducing publishing via a CLI – a CLI can call the `create` and `update` endpoints from a block folder and provide a tarball of its files.

# What it changes

1. Expands the `/api/blocks/create` endpoint introduced in #522 to also accept a tarball of block source
2. Adds an `/api/blocks/update` endpoint to update either an npm-linked block (simply triggers a retrieval from npm) or by providing another tarball for directly-uploaded blocks
3. Extracts some metadata validation / rewriting from the npm logic so it can be re-used elsewhere
4. Moves `wrangler` code to a top-level `infra` folder and renames to `cloudflare-r2-wrangler` (follow up from #522)
5. Calls for regeneration of pages using blocks when blocks are added / updated

# How to test it

### 1. Creation with tarball

Get a tarball of some block:
```sh
tar -czf package.tar.gz *
```
make a multipart form POST request to `/api/blocks/create` with a `blockName` string and the file provided under `tarball.

### 2. Update with tarball

Change something in your files, create a new tarball and repeat the above except to the `/api/blocks/update` endpoint


site/src/lib/api/blocks/npm.ts
```diff
@@ -90,19 +82,6 @@ const mirrorNpmPackageToR2 = async (
     });
   }
 
-  let metadataJson;
```
CiaranMn, line 93: A bunch of metadata checking stuff has been moved inside `r2` where it can be shared across all publishing methods, but otherwise unchanged

site/src/lib/api/blocks/r2.ts
```diff
@@ -18,13 +24,105 @@ const s3 = new S3({
   signatureVersion: "v4",
 });
 
+const stripLeadingAt = (pathWithNamespace: string) =>
+  pathWithNamespace.replace(/^@/, "");
+
+const validateBlockFiles = async (
```
CiaranMn, line 30: here is where the validation code that was previously in `npm` has been moved to

CiaranMn:
Putting this back into draft to make changes following the merge of #526 and probably tweaks as a result of implementing the CLI using it

site/src/pages/api/blocks/shared.ts
```diff
@@ -0,0 +1,20 @@
+import { NextApiResponse } from "next";
```
kachkaev, line 1: We have parts of the codebase where a folder named `shared` is used to store resources re-used locally. Each file under `shared` is like a mini-library with a clearly defined scoped (= file name). Creating `shared.ts` means that _all_ shared resources will go to one file and it can become pretty messy. In this particular case, our resources are scattered between `src/pages/api/blocks/shared.ts` and `src/lib/api/blocks`. It’s not very clear why.

I suggest avoiding `shared.ts`, even if we have only few resources to put in there. It’s better to go for `shared/some-topic.ts` or at least stick with `src/lib/api/blocks` for consistency. The problem with the latter is that it’s too far from where the resources are actually used. It’s tolerable short-term though.

site/src/pages/api/blocks/update.api.ts
```diff
@@ -0,0 +1,118 @@
+import { body as bodyValidator } from "express-validator/src/middlewares/validation-chain-builders";
+
+import { getDbBlock } from "../../../lib/api/blocks/db";
+import { publishBlockFromNpm } from "../../../lib/api/blocks/npm";
+import { createAuthenticatedHandler } from "../../../lib/api/handler/authenticated-handler";
+import { ExpandedBlockMetadata } from "../../../lib/blocks";
+import { shouldAllowBlockPublishing } from "../../../lib/config";
```
kachkaev, line 7: Now that we have two publishing methods, we should probably rename `shouldAllowBlockPublishing` into something like or `shouldAllowBlockNpmPublishing`, `shouldAllowNpmPublishing` or `npmPublishing`. We can also replace the condition:

```
export const shouldAllowBlockPublishing =
  process.env.NEXT_PUBLIC_NPM_PUBLISHING === "1";
```

It can be `!!`, `"true"` or `envalid.cleanEnv` + `envalid.bool`

kachkaev: In addition, we can create another feature flag to allow direct publishing. I would not mix up two feature flag.s

CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/548/commits/3cad45ddfd72c25139aea7b48a00d987b36eeab7

I don't think we need a feature flag for CLI publishing. We will use that very soon

site/src/pages/api/blocks/shared.ts
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/548/commits/02150ef4a20a4b506f30f99dd8682fb97b18bd59

site/src/pages/api/blocks/create.api.ts
```diff
@@ -26,20 +37,61 @@ export default createAuthenticatedHandler<
...
+        formatErrors({
+          code: "SIGNUP_INCOMPLETE",
+          msg: "You must be signed up to complete this request",
+        }),
+      );
+    }
+
+    const { blockName: untransformedBlockName, npmPackageName } = req.body;
+
+    const slugifiedBlockName = generateSlug(untransformedBlockName);
```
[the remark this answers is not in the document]
CiaranMn: `blockName` is required to be specified and is supposed to be a slug, this is just making sure it is.

I've updated it to throw an error if it isn't instead of silently changing it, in https://github.com/blockprotocol/blockprotocol/pull/548/commits/5baef1df09ca34a3db9a18e381e29a0e0ae88600

site/src/pages/api/blocks/publish.api.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export default createApiKeyRequiredHandler<
+  ApiBlockPublishRequest,
+  ApiTypePublishResponse
+>()
+  .use(
+    multipartUploads({
+      fieldsLimit: 1,
+      filesLimit: 1,
+      maxFileSize: 10 * 1024 * 1024, // bytes = 10MB
```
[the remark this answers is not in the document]
CiaranMn: No particular reason, I've upped it in https://github.com/blockprotocol/blockprotocol/pull/548/commits/ec6796370c8b20e6eb197b7b8cd28c1a9bda9d0b

site/src/lib/api/blocks/from-tarball.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  const tarballPath = path.resolve(tarballFolder, "tarball.tar.gz");
+
+  await fs.writeFile(tarballPath, tarball);
+
+  const extractionFolderPath = path.resolve(tarballFolder, "package");
+  await fs.ensureDir(extractionFolderPath);
+
+  try {
+    // tar is not available on deployed lambdas
```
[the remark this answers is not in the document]
CiaranMn: The former. Clarified in https://github.com/blockprotocol/blockprotocol/pull/548/commits#:~:text=update%20slug%20check%2C%20file%20size%20limit%2C%20comments

site/src/pages/api/blocks/publish.api.ts
```diff
@@ -0,0 +1,132 @@
...
+ */
+export default createApiKeyRequiredHandler<
+  ApiBlockPublishRequest,
+  ApiTypePublishResponse
+>()
+  .use(
+    multipartUploads({
+      fieldsLimit: 1,
+      filesLimit: 1,
+      maxFileSize: 100 * 1024 * 1024, // bytes = 10MB
```
kachkaev, line 35:
```suggestion
      maxFileSize: 100 * 1024 * 1024, // bytes = 100MB
```

kachkaev (approved):

kachkaev (approved):

PR #551, opened 2022-08-30
Call ESLint from workspaces instead of monorepo root

kachkaev:
This PR migrates top-level `eslint:lint` and `eslint:fix` scripts to Turborepo. Yarn workspaces are configured with their own `.eslintrc.cjs`. This lets developers configure workspace-specific rules on top of monorepo-wide rules in a shared `@local/eslint-config` package. This makes ESLint setup more flexible and scaleable, saving us from maintaining a huge and complex configuration file. Because ESLint is called via Turborepo, `yarn lint:eslint` becomes faster on average thanks to caching.

We cannot publish block templates with references to `@local/eslint-config`. Running `yarn install` in a new block would fail because `@local/eslint-config` cannot be found on npm. We may publish our ESLint config in the future, but until then, we remove ESLint from block templates before publishing. See details in `packages/@local/package-chores/scripts/remove-eslint-from-templates.ts`.

---
  - https://app.asana.com/0/0/1202882101840487


packages/@local/eslint-config/index.js
```diff
@@ -250,184 +239,23 @@ module.exports = {
...
+    //       { devDependencies: true },
+    //     ],
+    //   },
+    // },
+    // {
+    //   files: ["packages/block-template-custom-element/src/app.ts"],
+    //   rules: {
+    //     "no-console": "off",
+    //   },
+    // },
```
kachkaev, line 256: TODO: figure out what to do with ESLint in block templates. We can't use our local rules there, unless we want to magically remove `.eslintrc.cjs` file and mentions of `@local/eslint-config` and `@local/eslint-config` in `package.json`.

packages/@local/eslint-config/index.js
```diff
@@ -464,6 +265,21 @@ module.exports = {
...
+      //       { devDependencies: true },
+      //     ],
+      //   },
+      // },
+      // {
+      //   files: ["packages/block-template-custom-element/src/app.ts"],
+      //   rules: {
+      //     "no-console": "off",
+      //   },
+      // },
```
kachkaev, line 282: TODO: figure out what to do with ESLint in block templates. We can't use our local rules there, unless we want to magically remove `.eslintrc.cjs` file and mentions of `@local/eslint-config` and `@local/eslint-config` in `package.json`.

package.json
```diff
@@ -38,18 +38,15 @@
   },
   "lint-staged": {
     "**/*": [
-      "suppress-exit-code eslint --fix",
```
kachkaev, line 41: Not sure how to run workspace-specific ESLint in a pre-commit hook, but I hope we can live without autofixes. ESLint should apply on save in a correctly configured IDE.

packages/@local/eslint-config/index.js
kachkaev: We've decided to remove ESLint from templates before publishing. On it.

.gitignore
```diff
@@ -1,49 +1,63 @@
...
+.sentryclirc
+
+## Rust / WASM
+*.rs.bk
+bin
+Cargo.lock
+pkg
+target
+wasm-pack.log
+
```
kachkaev, line 59: Moved these out of `crates/**` for simplicity

apps/site/.eslintrc.cjs
```diff
@@ -0,0 +1,108 @@
+const noRestrictedImportsConfig = require("@local/eslint-config").rules[
+  "no-restricted-imports"
+][1];
+
+/** @type {import("eslint").Linter.Config} */
+module.exports = {
+  ...require("@local/eslint-config/generate-workspace-config.cjs")(__dirname),
+  rules: {
```
kachkaev, line 8: These rules have been pulled out of `packages/@local/eslint-config/index.js`

package.json
```diff
@@ -37,18 +37,15 @@
...
     ]
   },
   "devDependencies": {
     "@changesets/changelog-github": "^0.4.6",
     "@changesets/cli": "^2.24.4",
-    "@local/eslint-config": "0.0.0-private",
     "check-dependency-version-consistency": "3.0.1",
     "dotenv-flow": "3.2.0",
-    "eslint": "8.24.0",
     "husky": "7.0.4",
```
kachkaev, line 49: No more top-level ESLint 🎉

.gitignore
```diff
@@ -1,49 +1,63 @@
-######################
-## Shared with linters
-######################
+#################################
+## Shared between git and linters
+#################################
 
-# Logs
+## Logs
```
kachkaev, line 5: `##` is for comments, `#` is for commands that can be commented out to run. e.g.:

```
## Hope you know what you’re doing!
# rm -rf /my-project
```

crates/type-system/.eslintrc.cjs
```diff
@@ -0,0 +1,4 @@
+/** @type {import("eslint").Linter.Config} */
+module.exports = {
+  ...require("@local/eslint-config/generate-workspace-config.cjs")(__dirname),
```
kachkaev, line 3: Using spread operation instead of `module.exports = require(...)` for future-proofing. This way it is easier to add custom rules while keeping diff small. Besides, `require` fits one line so is not chunked by Prettier.

.github/workflows/release.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Create Release Pull Request or Publish to npm
         uses: changesets/action@v1.3.0
         with:
-          publish: yarn changeset publish
+          publish: yarn workspace @local/package-chores exe scripts/remove-eslint-from-templates.ts && yarn changeset publish
```
CiaranMn, line 23: Can't we call this script via [prepublishOnly](https://docs.npmjs.com/cli/v8/using-npm/scripts#prepare-and-prepublish) in package.json instead? Then it will happen consistsently whether publishing from CI or not.

CiaranMn (approved):

.github/workflows/release.yml
kachkaev: Seems that `yarn workspace @local/package-chores exe scripts/remove-eslint-from-templates.ts` works from `packages/block-template-xyz`. That's good. We can probably tweak the script so that it removes ESLint from `cwd` rather than for a predefined list of packages. 

[Asana task](https://app.asana.com/0/1203007126736604/1203132574912424/f) (internal)

PR #553, opened 2022-08-31
Fix middleware redirecting guest users to `/dashboard` issue

yusufkinatas:
## Reason

The middleware at [middleware.page.ts](https://github.com/blockprotocol/blockprotocol/blob/966c79d6cbdc1cfb1a681f3ea69098294cca9629/site/src/middleware.page.ts) kept redirecting users to `/login` even if they are guests.

The bug could be reproduced by following these steps.

1. Run the app and login locally, navigate to /dashboard
2. Run yarn dev:seed-db to kill session cache
3. Refresh the page
4. Find yourself on the /login page
5. Click on the logo

**expected:** home page
**observed:** redirect back to /login

## Solution

Deleted the session cookie on `/me` api route to make session cookie is not present when there is no user session. 

This PR also comments-out some disabled tests temporarily.

---
_Thanks @kachkaev for noticing this issue._


yusufkinatas:
Code coverage decrease is because of commented-out integration tests. We can ignore it, because we'll enable those tests back soon.

site/tests/integration/block-publish-npm.test.ts
```diff
@@ -1,118 +1,121 @@
-import { Page } from "playwright";
-import { expect, test } from "playwright-test-coverage";
+import { test } from "playwright-test-coverage";
 
-import { publishBlock } from "../shared/blocks";
-import { resetDb } from "../shared/fixtures";
-import { login } from "../shared/nav";
+test("TODO: uncomment this file when NPM_PUBLISHING feature is released", () => {});
```
CiaranMn, line 3: Couldn't we set the environment variable in the test to make sure we don't break the feature?

yusufkinatas: Do you mean conditionally skipping every test depending on the env variable, instead of commenting out tests? @CiaranMn

CiaranMn: I meant just [set the environment variable](https://nextjs.org/docs/basic-features/environment-variables#test-environment-variables) so that the feature appears when running tests so that we keep testing as if the feature was released – that way we don't make any changes that break it without realising.

kachkaev, line 3: We can use [`test.skip`](https://playwright.dev/docs/test-annotations#skip-a-test) instead.

Lines we’ve commented out can still be linted. Leaving them as comment will produce inaccurate git blame when it's time to restore the test.

yusufkinatas: I thought something like test.skip, based on the env variable, so when we enable the env variable flag, tests would start running again right? @CiaranMn

CiaranMn: If we stop running the tests there is a chance we make some changes that break the feature without knowing, and then when we want to actually enable it we have to figure out what they were.

If the tests don't take long to run I don't see the advantage in keeping them running (by setting the environment variable for the tests)

CiaranMn: Possibly it's the 'can't publish the same npm-linked package twice', since this is only prevented if running in production. You could also make that check apply in test mode (or only skip it in development mode)

yusufkinatas: Hmm, I think it could be sth else, since it's calling `resetDb` in `beforeEach` on every test. Btw, I'm not sure if it was `can't publish the same npm-linked package twice`, need to check

site/tests/integration/block-publish-npm.test.ts
```diff
@@ -80,6 +82,8 @@ test("user should not be able to publish an invalid npm-package", async ({
 test("user should not be able to publish an already-used npm-package", async ({
   page,
 }) => {
+  test.skip();
+  /** @todo skip this test if `isProduction === false` */
```
CiaranMn, line 86: `isProduction` will never be false in these tests – we need some other approach, so I suggest updating the todo. We don't have to figure it out now though.
```suggestion
  /** @todo update 'already-taken' name check so that it can be checked in these tests (e.g. by having the check activated by custom environment variable, not `isProduction`) */
```

yusufkinatas: I suggest removing the todo, let me update it

CiaranMn (approved):

PR #555, opened 2022-08-31
Add block entity switching functionality to `mock-block-dock` debug mode

teenoh:
This PR
- introduces entity switching functionality via `mock-block-dock`'s debug UI and 
- adds a `block Info` tab for displaying information about the block

https://user-images.githubusercontent.com/16928688/187622061-2759247d-5d7d-4443-8fef-8f90c5ec729c.mov

Screenshot 2022-08-31 at 08 47 12


packages/mock-block-dock/dev/dev-app.tsx
```diff
@@ -15,16 +15,34 @@ type TestBlockType =
...
+  },
+  "html-at-url": {
+    entityTypeId: "test-type-1",
+    properties: { name: "World" },
+  },
+  "html-as-string": {
+    entityId: "entity-html-at-url",
+    entityTypeId: "test-type-1",
+    properties: { name: "World" },
+  },
```
teenoh, line 36: Added these to test scenarios where any or both of entityId and entityTypeId are absent to be sure it doesn't break

packages/mock-block-dock/src/use-mock-block-props.ts
```diff
@@ -75,81 +74,54 @@ export const useMockBlockProps = ({
...
-      entityTypeId,
-      schema: {
-        title: "BlockType",
-        type: "object",
-        $schema: "https://json-schema.org/draft/2019-09/schema",
-        $id: "http://localhost/blockType1",
-        ...(blockSchema ?? {}),
-      },
-    };
-
```
teenoh, line 94: Moved all this logic into useMockDatastore()

packages/mock-block-dock/src/use-mock-block-props.ts
```diff
@@ -158,23 +111,9 @@ export const useMockBlockProps = ({
...
-        (entityType) =>
-          entityType.entityTypeId === latestBlockEntity.entityTypeId,
-      ),
-    [entityTypes, latestBlockEntity.entityTypeId],
-  );
-
-  if (!latestBlockEntityType) {
-    throw new Error("Cannot find block entity type. Has it been deleted?");
-  }
-
```
teenoh, line 177: This is no longer needed. There are checks in `use-mock-datastore` that ensures the block always has an `entityId` and `entityTypeId`, even when they aren't present in the passed in `blockEntity`

PR #556, opened 2022-08-31
Prevent direct access to sandbox host

kachkaev:
This PR makes sure that https://www.blocksandbox.org redirects to https://blockprotocol.org unless we visit `/:shortname/blocks/:blockslug/sandboxed-demo`. Opening that specific path does the reverse: https://blockprotocol.org → https://www.blocksandbox.org.

I don’t know how to test the new code before merging to prod, so please review it carefully.

---
  - https://app.asana.com/0/0/1202882436362522


site/src/lib/api-client.ts
```diff
@@ -40,7 +40,7 @@ import {
 } from "../pages/api/verify-email.api";
 import { FRONTEND_URL } from "./config";
 
-const BASE_URL = `${FRONTEND_URL}/api/`;
+const BASE_URL = `${typeof window === "undefined" ? "" : FRONTEND_URL}/api/`;
```
kachkaev, line 43: This small change is not related to sandbox directly, but should resolve one issue with hostnames.

When visiting a production deployment by its permalink, e.g. (https://blockprotocol-j5e8ppem1.stage.hash.ai), whoami request goes to `https://blockprotocol.org/api/me` and fails CORS:

```
Response body is not available to scripts (Reason: CORS Allow Origin Not Matching Origin)
```

Using relative path on the client should fix this problem and reduce noise on Sentry. We visit permanent deployment URLs when investigating bugs in production.

site/src/middleware.page.ts
```diff
@@ -5,18 +5,42 @@ import { NextResponse } from "next/server";
...
+      url.pathname.match(/^\/@[\w_-]+\/blocks\/[\w_-]+\/sandboxed-demo$/),
+    );
 
-  // destroying session at `logout.api.ts` removes this cookie
-  const isLoggedIn = !!cookies.get(SESSION_COOKIE_NAME);
+    if (url.host === productionFrontendHost && openingBlockSandboxPage) {
+      const newUrl = new URL(url);
+      newUrl.host = productionSandboxHost;
+      return NextResponse.redirect(newUrl, 308);
+    }
```
yusufkinatas, line 27: I feel like defining a function like this
```ts
const changeHostAndRedirect = (host: string) => {
  const newUrl = new URL(url);
  newUrl.host = host;
  return NextResponse.redirect(newUrl, 308);
};
```

and reusing it in both if's looks a bit easier-to-read

```ts
if (url.host === productionFrontendHost && openingBlockSandboxPage) {
  return changeHostAndRedirect(productionSandboxHost);
}

if (url.host === productionSandboxHost && !openingBlockSandboxPage) {
  return changeHostAndRedirect(productionFrontendHost);
}
```

yusufkinatas (approved):
Tested on local, redirecting to sandbox works as expected. Couldn't find a way to try `redirect to localhost from sandbox` on my local. But code looks ok to me.

yusufkinatas (approved):

PR #557, opened 2022-08-31
Add block entity switching functionality to `mock-block-dock` debug mode

teenoh:
This PR
- introduces entity switching functionality via `mock-block-dock`'s debug UI and 
- adds a `block Info` tab for displaying information about the block

https://user-images.githubusercontent.com/16928688/187622061-2759247d-5d7d-4443-8fef-8f90c5ec729c.mov

Screenshot 2022-08-31 at 08 47 12

---
  - https://app.asana.com/0/0/1202882280310592
  - https://app.asana.com/0/0/1202736533402962


packages/mock-block-dock/dev/dev-app.tsx
```diff
@@ -15,16 +15,34 @@ type TestBlockType =
...
+  "html-at-url": {
+    entityTypeId: "test-type-1",
+    properties: { name: "World" },
+  },
+  "html-as-string": {
+    entityId: "entity-html-at-url",
+    entityTypeId: "test-type-1",
+    properties: { name: "World" },
+  },
+};
```
teenoh, line 37: Added these to test

scenarios where any or both of entityId and entityTypeId are absent to be sure it doesn't break
updating block entity externally (e.g from block properties tab in BP Hub) still works

packages/mock-block-dock/src/use-mock-block-props.ts
```diff
@@ -158,23 +111,9 @@ export const useMockBlockProps = ({
...
-  // @todo we don't do anything with this type except check it exists - do we need to do this?
-  const latestBlockEntityType = useMemo(
-    () =>
-      entityTypes.find(
-        (entityType) =>
-          entityType.entityTypeId === latestBlockEntity.entityTypeId,
-      ),
-    [entityTypes, latestBlockEntity.entityTypeId],
-  );
-
```
teenoh, line 173: This is no longer needed. There are checks in use-mock-datastore that ensures the block always has an entityId and entityTypeId, even when they aren't present in the passed in blockEntity

packages/mock-block-dock/src/use-mock-block-props/use-mock-datastore.ts
```diff
@@ -818,9 +840,111 @@ export const useMockDatastore = (
...
+    const entityId = newExternalBlockEntity?.entityId ?? defaultBlockEntityId;
+
+    const entityTypeInStore = newEntityTypes.find(
+      (entityType) => entityType.entityTypeId === entityTypeId,
+    );
+
+    if (entityTypeInStore) {
+      // If block schema changed, update entity type
+      if (JSON.stringify(blockSchema) !== JSON.stringify(prevBlockSchema)) {
+        // test this
```
teenoh, line 856:
```suggestion
```

packages/block-template-react/src/dev.tsx
```diff
@@ -17,6 +18,7 @@ const DevApp = () => {
         entityId: "test-block-1",
         properties: { name: "World" },
       }}
+      blockInfo={packageJSON.blockprotocol}
```
teenoh, line 21: This updates `block-template-react` to pull blockInfo from the block's `package.json#protocol` field.

packages/block-template-react/tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "lib": ["dom", "dom.iterable", "esnext"],
     "module": "commonjs",
     "noEmit": true,
-    "target": "es6"
+    "target": "es6",
+    "resolveJsonModule": true
```
teenoh, line 11: had to add this in-order to import the block's package.json. See https://github.com/blockprotocol/blockprotocol/pull/557/files#r959809732

packages/mock-block-dock/src/debug-view/dev-tools.tsx
```diff
@@ -132,16 +118,8 @@ export const DevTools = () => {
...
             </Tabs>
-
-            {blockType && (
-              <Chip
-                variant="outlined"
-                size="small"
-                label={chipInfo[blockType].label}
-                color={chipInfo[blockType].color}
-              />
-            )}
```
teenoh, line 144: Removed this from here, since the info is now displayed in the `block Info` tab

packages/mock-block-dock/src/use-mock-block-props/use-mock-datastore.ts
```diff
@@ -55,6 +72,11 @@ export const useMockDatastore = (
   const [linkedAggregations, setLinkedAggregations] = useDefaultArrayState<
     MockDataStore["linkedAggregationDefinitions"]
   >(initialData.linkedAggregationDefinitions);
+  const [blockEntity, setBlockEntity] = useState<Entity>(
+    externalBlockEntity ?? defaultBlockEntity,
```
teenoh, line 76: added a fallback entity so that blockEntity is not undefined

packages/mock-block-dock/src/use-mock-block-props/use-mock-datastore.ts
```diff
@@ -818,9 +840,110 @@ export const useMockDatastore = (
...
+    // This handles scenarios where the passed in block entity doesn't have an
+    // entityId or entityTypeId
+    if (!entity.entityId) {
+      entity.entityId = defaultBlockEntityId;
+    }
+    if (!entity.entityTypeId) {
+      entity.entityTypeId = defaultBlockEntityTypeId;
+    }
+    setBlockEntity(entity);
+  }
```
teenoh, line 940: Block Entity can be updated either
1. externally (e.g via block properties tab in BP hub) or
2. internally (e.g when user switches block entity)

This takes account of the different sources of changes and adds a default entity(Id/typeId) if it's not present

packages/mock-block-dock/src/use-mock-block-props/use-mock-datastore.ts
```diff
@@ -818,9 +840,110 @@ export const useMockDatastore = (
...
+  if (
+    !entities.find(
+      (entity) =>
+        entity.entityId === blockEntity.entityId &&
+        entity.entityTypeId === blockEntity.entityTypeId,
+    ) &&
+    [blockEntity.entityId, blockEntity.entityTypeId].every((key) => !!key)
+  ) {
+    updateStore(blockEntity);
+  }
```
teenoh, line 924: This ensures we update datastore to reflect accordingly if the block entity isn't present there

CiaranMn:
Need to fix a couple of things which I'll come back to after looking at the publishing CLI
1. State updates do not work (e.g. you cannot edit the name in a React block) - this is because the block entity is now stored in separate state. It should be derived from the mock datastore instead, otherwise updates to the data store from the block itself will not be reflected in the block entity.
2. There is provision for omitting the `entityTypeId` in the passed `blockEntity` but this does not work - if the `entityTypeId` is omitted, the block entity does not appear in the datastore (try removing `entityTypeId` from a type in `dev-app`) – we don't need to provide for `entityTypeId` being omitted, it should always be present in an `Entity`.

CiaranMn:
Closing in favour of #599

PR #560, opened 2022-08-31
Fix Entity Type in Type System Package

Alfred-Mountfield:
#559 was sloppy. This actually fixes the problem, adds integration tests to check it, and then also fixes other problems noticed as a result of trying to do it properly this time.


crates/type-system/src/ontology/entity_type/repr.rs
```diff
@@ -31,10 +31,10 @@ pub struct EntityType {
     #[serde(skip_serializing_if = "Option::is_none")]
     description: Option<String>,
     // TODO - Improve the typing of the values
-    #[cfg_attr(target_arch = "wasm32", tsify(type = "Record<VersionedUri, any>"))]
+    #[cfg_attr(target_arch = "wasm32", tsify(type = "Record<BaseUri, any>"))]
```
TimDiekmann, line 34: I probably would have made the same mistake 😄

TimDiekmann (approved):
LGTM 🎉

PR #562, opened 2022-09-02
Implement `next-seo` & improve block page meta tags

yusufkinatas:
## Reason

Block page had irrelevant `description` and `image` meta tags. Updated them to show a preview of the block. Also used `next-seo` for typescript support & consistency.

## What does this change?

- Add `next-seo` library to manager meta tags
- Improve meta tags used in block page

## Demo
Screenshot taken from https://metatags.io/ for `Code` block's page
demo


site/src/pages/_app.page.tsx
```diff
@@ -114,16 +115,7 @@ const MyApp = ({
...
-                <title>
-                  Block Protocol – an open standard for data-driven blocks
-                </title>
-                <meta itemProp="name" content="Block Protocol" />
-                <meta
-                  itemProp="description"
-                  content="An open standard for data-driven blocks"
-                />
-              </Head>
+              <DefaultSeo {...defaultSeoConfig} />
```
yusufkinatas, line 118: Removed the meta tags from `_document.page.tsx` and moved them here as [next-seo documentation suggests](https://github.com/garmeeh/next-seo#default-seo-configuration)

site/src/pages/hub.page.tsx
```diff
@@ -27,14 +30,14 @@ export const getStaticProps: GetStaticProps<PageProps> = async () => {
...
-        />
-      </Head>
+      <NextSeo
+        title={seoTitle}
+        description={seoDesc}
+        openGraph={{
+          description: seoDesc,
+          title: seoTitle,
+        }}
+      />
```
yusufkinatas, line 40: I searched for a reason to keep `itemProp="name"` and couldn't find any. Let me know if I missed sth and openGraph tags are not enough here.

site/src/pages/_app.page.tsx
```diff
@@ -20,6 +20,7 @@ import {
   UserContextValue,
   UserState,
 } from "../context/user-context";
+import { defaultSeoConfig } from "../default-seo-config";
```
kachkaev, line 23: It's helpful to keep things that are used together close to each other. Extracting a separate file makes sense if we plan to use `defaultSeoConfig` outside `_app.page.tsx`. I suggest defining this constant right in the file, but I may be missing something. What benefits of a separate file to you see?

kachkaev, line 23: It's helpful to keep things that are used together close to each other. Extracting a separate file makes sense if we plan to use `defaultSeoConfig` outside `_app.page.tsx`. I suggest defining this constant right in the file or even ditch the constant and directly set `<DefaultSeo ... />` props.

I may be missing something though. What benefits of a separate file to you see?

site/src/pages/hub.page.tsx
```diff
@@ -27,14 +30,14 @@ export const getStaticProps: GetStaticProps<PageProps> = async () => {
...
-          content="The Block Protocol's registry of open-source blocks"
-        />
-      </Head>
+      <NextSeo
+        title={seoTitle}
+        description={seoDesc}
+        openGraph={{
+          description: seoDesc,
+          title: seoTitle,
+        }}
```
kachkaev, line 39: Do we need to define `openGraph` title and description if they match `title` and `description`? Current diff only sets `title` in docs, but this gets propagated to `og:title` in page source. Seems like `NextSeo` resolves these fields correctly.

I suggest either removing `openGraph` in this file or setting it elsewhere too.

kachkaev (commented):
I haven’t used `next-seo` before – nice library! The diff looks good, just have a couple of small comments.

site/src/pages/_app.page.tsx
yusufkinatas: I wanted to keep `_app.page.tsx` clean & easy-to-read. Thought keeping this SEO content on somewhere else.

site/src/pages/hub.page.tsx
yusufkinatas: Tested the behavior of `NextSeo`, and you're right. only `title` / `description` is enough, no need to pass `openGraph` if we don't want to define sth else for openGraph.

Fixed here https://github.com/blockprotocol/blockprotocol/pull/562/commits/58f5c7ffb40d1cd0525e600d492933b31cb8bab8

site/src/pages/_app.page.tsx
yusufkinatas: Moved the props into the top of `_app.page.tsx` here https://github.com/blockprotocol/blockprotocol/pull/562/commits/1c2fe0d3010dacbd8f846be8aa1514d109383e5a

kachkaev (approved):

site/src/pages/[shortname]/blocks/[block-slug].page.tsx
```diff
@@ -251,11 +251,22 @@ const BlockPage: NextPage<BlockPageProps> = ({
...
-      <Head>
-        <title>
-          {`Block Protocol – ${blockMetadata.displayName} Block by ${shortname}`}
-        </title>
-      </Head>
+      <NextSeo
+        title={`Block Protocol – ${blockMetadata.displayName} Block by ${shortname}`}
+        description={
+          blockMetadata.description ||
+          "Check out this open-source block on Block Hub"
```
Comment (nonparibus) [inline site/src/pages/[shortname]/blocks/[block-slug].page.tsx:258]: This copy needs a pass

Comment (yusufkinatas) [reply site/src/pages/[shortname]/blocks/[block-slug].page.tsx:258]: Not sure if I understand. Should we change it to something else?

PR #563, opened 2022-09-02
Implement Error on `ParseBaseUriError` and `ParseVersionedUriError`

TimDiekmann:

`ParseBaseUriError` and `ParseVersionedUriError` does not implement `std::error::Error`, which is required in order to use them with `error-stack`.

## 🔍 What does this change?

- Implement `Error` for `ParseBaseUriError` and `ParseVersionedUriError`
- Drive-by: Fix a clippy lint


crates/type-system/src/ontology/entity_type/mod.rs
```diff
@@ -43,16 +43,7 @@ impl EntityType {
...
-            id,
-            title,
-            plural_title,
-            description,
-            default,
-            examples,
-            property_object,
-            links,
-        }
+        Self { id, title, plural_title, description, property_object, links, default, examples }
```
Alfred-Mountfield, line 46: Are you sure this formatting is correct? I've got the change locally to re-order and fix the lint but:
```rust
    ) -> Self {
        Self {
            id,
            title,
            plural_title,
            description,
            property_object,
            links,
            default,
            examples,
        }
    }
```

TimDiekmann: This is the clippy error:
image

crates/type-system/src/ontology/uri/error.rs
```diff
@@ -19,6 +20,8 @@ impl fmt::Display for ParseBaseUriError {
     }
 }
 
+impl Error for ParseBaseUriError {}
```
Alfred-Mountfield, line 23: Can you swap these to `thiserror` impls so they're consistent with other errors like `crates/type-system/src/ontology/entity_type/error.rs:14` and so that we have better format strings. (The current one is obviously not ideal 🙃)

crates/type-system/src/ontology/entity_type/mod.rs
Alfred-Mountfield: Yeah I know as I said I've fixed it locally
> I've got the change locally to re-order and fix the lint but:

My point is your commit puts them all on the same line. That's why I'm asking if you're sure the **formatting** is correct

crates/type-system/src/ontology/uri/error.rs
TimDiekmann: I replaced the `Display` implementation with `#[derive(Error])`, I used error messages according to the [API guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#error-types-are-meaningful-and-well-behaved-c-good-err). 298ed7710c0f31fa83e9360ab56233f3d37974c0

`ParseIntError` does not implement `Serialize`, so I converted it into a `String` 😞

crates/type-system/src/ontology/uri/mod.rs
```diff
@@ -153,9 +153,8 @@ impl FromStr for VersionedUri {
 
         Ok(Self::new(
             BaseUri::new(base_uri.to_owned()).map_err(ParseVersionedUriError::InvalidBaseUri)?,
-            version
-                .parse()
-                .map_err(|_| ParseVersionedUriError::InvalidVersion)?,
+            u32::from_str(version)
+                .map_err(|error| ParseVersionedUriError::InvalidVersion(error.to_string()))?,
```
Alfred-Mountfield, line 157: How come you're going with a string here? Does `parse` not return a useful type that we can store inside?

Alfred-Mountfield: Ah you replied

crates/type-system/src/ontology/uri/error.rs
Alfred-Mountfield: Hm, I wonder if we should get rid of the Serialize bound. Anyway looks good as is, thank you

Alfred-Mountfield (approved):

crates/type-system/src/ontology/uri/mod.rs
TimDiekmann: No worries 🙂

PR #571, opened 2022-09-05
Mock-block-dock fixes and UI tweaks

CiaranMn:
# What this does

Fixes a few things in `mock-block-dock`:

1. Makes it respond to changes in the `blockEntity` and `blockSchema` props (fixing the current bug in the hub where changes from the data source are not reflected)
2. Moves the readonly toggle in the debug UI, since it's currently out of sight below block properties
3. Rename 'Data Source' to 'Block Properties' on mobile hub view for consistency with the desktop view
4. Add an option to remove the debug toggle entirely, for use in the hub (where we don't want it until/unless we make the debug view look good there)

Internal task links:
1. [Fixing block properties update in hub](https://app.asana.com/0/1202839302145209/1202922733837187/f)
2. [Rename data source -> block properties](https://app.asana.com/0/1200211978612931/1202737317045344/f)


packages/mock-block-dock/dev/dev-app.tsx
```diff
@@ -92,10 +92,11 @@ const DevApp: FunctionComponent = () => {
       </div>
 
       <MockBlockDock
-        debug
         blockDefinition={blockDefinition}
         blockEntity={blockEntity}
         blockName="Test Block"
+        debug
+        key={testBlockType} // completely reset the state when block type has changed
```
CiaranMn, line 99: there are a few bugs when running MBD in dev mode (i.e. its own dev mode, which doesn't end up in templates) and switching between block types

kachkaev (approved):

PR #574, opened 2022-09-05
Remove react-block-loader

CiaranMn:
The `react-block-loader` package has been broken for a while pending an update to BP 0.2.

Its presence is causing some issues with dependencies, particularly now as it relies on an older version of `blockprotocol`, and we want to remove the deprecated code in `blockprotocol` and replace it with a CLI.

The presence of `react-block-loader` in the codebase complicates this.

We can restore the code when we need it, but for now not having to work around code we are not using is simpler.

---
  - https://app.asana.com/0/0/1202925802678759


kachkaev (changes requested):
Instead of removing

```
"fix:dependency-version-consistency": "check-dependency-version-consistency --fix . --ignore-package=react-block-loader",
...
"lint:dependency-version-consistency": "check-dependency-version-consistency . --ignore-package=react-block-loader",
```

we should only remove this bit: `--ignore-package=react-block-loader`

CiaranMn:
> Instead of removing
> 
> ```
> "fix:dependency-version-consistency": "check-dependency-version-consistency --fix . --ignore-package=react-block-loader",
> ...
> "lint:dependency-version-consistency": "check-dependency-version-consistency . --ignore-package=react-block-loader",
> ```
> 
> we should only remove this bit: `--ignore-package=react-block-loader`

oops 🤦 thanks!

packages/block-template-react/.bprc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "apiKey": "b10ck5.5551820619c3b59abcff6d07325cc6e6.e86fc0f4-2228-4d4d-96b5-6240bfee0745"
+}
```
kachkaev, line 3: 🚨

kachkaev (changes requested):

packages/block-template-react/.bprc.json
CiaranMn: Oops - thanks! I've been so careful to avoid `git add .` when jumping around but forgot it when sorting out merge conflicts on this one

The key is for my local db so fine to remain in history – have removed

kachkaev (approved):

PR #575, opened 2022-09-06
Introduce blockprotocol CLI

CiaranMn:
# What this does

This PR introduces a CLI for interacting with the Block Protocol API. The first command introduced is `blockprotocol publish`.

It repurposes the deprecated `blockprotocol` package to do so.

# How to test it

Go into a block folder in the repo, e.g. `packages/block-template-react`, and with the site running run 
```sh
BLOCK_PROTOCOL_SITE_HOST=http://localhost:3000 npx blockprotocol publish
```

# Next steps

1. Update the 'publish a block' page to describe this process instead of the previous one
2. Move our own blocks to be published via CLI instead of maintaining JSON files in the repo
3. Remove block building code from this repo

# Demo

Screenshot 2022-09-06 at 08 42 04

Screenshot 2022-09-06 at 08 49 28

## The happy path

Screenshot 2022-09-06 at 08 57 58

## Error messages from the CLI

Screenshot 2022-09-06 at 08 57 10
Screenshot 2022-09-06 at 08 42 45

Screenshot 2022-09-06 at 08 59 56
Screenshot 2022-09-06 at 09 00 00

Screenshot 2022-09-06 at 09 12 49

## Error states from the API

Screenshot 2022-09-06 at 09 06 16
Screenshot 2022-09-06 at 09 06 20

# Known issues

Needs #574 to merge for CI to pass, to remove the final package that depends on the deprecated `blockprotocol` types this PR removes.

---
  - https://app.asana.com/0/0/1202882583267696


.changeset/make-blockprotocol-a-cli.md
```diff
@@ -0,0 +1,8 @@
+---
+"blockprotocol": minor
+"block-template-custom-element": patch
+"block-template-html": patch
+"block-template-react": patch
+---
+
+make blockprotocol package into a cli, and remove deprecated types
```
kachkaev, line 8: Current description does not reflect what we’ve done in templates and may be confusing. I think we can create two md files in this PR.

packages/block-template-custom-element/.gitignore.dist
```diff
@@ -1,2 +1,3 @@
 dist
 node_modules
+.bprc.json
```
kachkaev, line 3: New lines in all three `.gitignore.dist` 🙏

packages/blockprotocol/cli/commands.js
```diff
@@ -0,0 +1,7 @@
+import * as help from "./commands/help.js";
+import * as publish from "./commands/publish.js";
+
+export const commands = {
```
kachkaev, line 4: optional:

```suggestion
export const commandLookup = {
```

Plural suggests we are dealing with an array here, which is not true.

packages/blockprotocol/cli/commands/help.js
```diff
@@ -0,0 +1,31 @@
...
+    ],
+  },
+];
+
+// ********************* SCRIPT ********************* //
+
+/**
+ * Print the main help guide
+ */
+const script = () => {
```
kachkaev, line 25: Unless `script` is some established term you have borrowed upstream, I would try to avoid it. Existing scripts follow [specific rules](https://github.com/blockprotocol/blockprotocol/pull/189), which cannot be used in commands – these are different. Not sure what exact term to use, but something like `run` would not be too much off.

packages/blockprotocol/cli/commands/publish.js
```diff
@@ -0,0 +1,185 @@
...
+
+import { printSpacer } from "../print-spacer.js";
+import { doesUserAgree } from "./publish/does-user-agree.js";
+import { findApiKey } from "./publish/find-api-key.js";
+import { findBlockFolder } from "./publish/find-block-folder.js";
+import { postPublishForm } from "./publish/post-form.js";
+
+// ********************* MANUAL ********************* //
+
+const manual = [
```
kachkaev, line 16: Can we add typings via `/** comments */`?

packages/blockprotocol/cli/commands/publish/find-api-key.js
```diff
@@ -0,0 +1,99 @@
...
+
+  // see if there's a config file in the current working directory
+  const configFilePathInCwd = path.resolve("./", configFileName);
+  const fileExistsInCwd = await fs.pathExists(configFilePathInCwd);
+  if (fileExistsInCwd) {
+    return extractKeyFromJsonFile(configFilePathInCwd);
+  }
+
+  // if not, find the project root and check for the file there
+  const projectRootPath = await findProjectRoot();
```
kachkaev, line 60: Not sure we need to search for project root in this case. AFAIU, config files are searched upwards till `/` is reached. For example, `~/.bprc.json` can be a valid location for the config.

packages/blockprotocol/cli/commands/publish/shared/find-project-root.js
```diff
@@ -0,0 +1,13 @@
+import { findUp } from "find-up";
+import path from "node:path";
+
+/**
+ * Find the path to the project root, by searching upwards for the nearest package.json
+ * @returns {Promise<string>}
```
kachkaev, line 6:
```suggestion
 * @returns {Promise<string|undefined>}
```

If we switch to search for `.bprc.json` via `find-up`, this function will be used only once, so won't be needed.

packages/blockprotocol/cli/print-spacer.js
```diff
@@ -0,0 +1,9 @@
+import chalk from "chalk";
+
+export const printSpacer = () => {
+  console.log(
+    chalk.bold(
+      "––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––",
```
kachkaev, line 6: I guess we can use `process.stdout.rows` here as a cap (if defined).

site/README.md
```diff
@@ -297,6 +297,32 @@ Uploads a user avatar and apply it to logged in profile page.
 - Response Body:
   - `avatarUrl`: Url pointing to the newly uploaded user avatar.
 
+#### `POST /api/blocks/create`
+
+Creates and mirrors a block published to npm. Only permitted if `process.env.NEXT_PUBLIC_NPM_PUBLISHING === "1"`
```
kachkaev, line 302:
```suggestion
Creates and mirrors a block published to npm. Only permitted if `process.env.NEXT_PUBLIC_NPM_PUBLISHING` is set.
```

```suggestion
Creates and mirrors a block published to npm. Only permitted if `process.env.NEXT_PUBLIC_NPM_PUBLISHING` is truthy.
```

(same below)

site/src/middleware.page.ts
```diff
@@ -46,3 +46,11 @@ export async function middleware(request: NextRequest) {
     }
   }
 }
+
+/**
+ * Exclude API routes from the middleware as it stops them accessing the raw body (needed for file uploads)
+ * @see https://github.com/vercel/next.js/issues/39262
+ */
+export const config = {
+  matcher: ["/", "/((?!api/).*)"],
```
kachkaev, line 55: Will this still match `/hub` etc.?

kachkaev:
Optional suggestion: add `version` command

packages/block-template-custom-element/.gitignore.dist
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/575/commits/bd8dd3436a6d588d763e6211c473151178f9d134

packages/blockprotocol/cli/commands.js
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/575/commits/bd8dd3436a6d588d763e6211c473151178f9d134

packages/blockprotocol/cli/commands/help.js
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/575/commits/bd8dd3436a6d588d763e6211c473151178f9d134

site/README.md
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/575/commits/bd8dd3436a6d588d763e6211c473151178f9d134

packages/blockprotocol/cli/commands/publish.js
```diff
@@ -0,0 +1,185 @@
...
+  {
+    header: "blockprotocol publish",
+    content:
+      "Publish a block to the Block Protocol hub. See https://blockprotocol.org/docs",
+  },
+  {
+    header: "Options",
+    optionList: [
+      {
+        name: "dry",
```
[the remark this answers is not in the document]
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/575/commits/bd8dd3436a6d588d763e6211c473151178f9d134

packages/blockprotocol/cli/commands/publish/find-api-key.js
CiaranMn: Fair enough – simplified in https://github.com/blockprotocol/blockprotocol/pull/575/commits/c9d54518a8f1b58b023e7d4c72df8529552fbe96

packages/blockprotocol/cli/commands/publish/shared/find-project-root.js
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/575/commits/6ca50aaf85446fc13c248004439d1929ff9261a0

It is still needed in the API key file in order to find the location in which to create one, if the user wants

CiaranMn: We could alternatively create it in their user directory. Having thought about it, we probably want to add it to their `.gitignore` as well if we create it in the project...

packages/blockprotocol/cli/print-spacer.js
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/575/commits/d7c8886fa34d7792f4861a587fe2e137ecb69aa1

site/src/middleware.page.ts
CiaranMn: Yes, it will.

packages/block-template-react/.gitignore.dist
```diff
@@ -1,2 +1,3 @@
 dist
 node_modules
+.bprc
```
kachkaev, line 3: `EOL` worth fixing given that it's a template

kachkaev: You can avoid this in the future by adding `"files.insertFinalNewline": true,` to VSCode config

packages/blockprotocol/cli/commands/publish.js
kachkaev: Hmmm can we use a typing from `@types/command-line-usage` instead?

packages/block-template-react/.gitignore.dist
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/575/files/fdc585c6d6e709fbce49ca0060eb44baa4972bad..7666a5ab623fb4777c2ffe3a26276dbbc157eeaa

CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/575/commits/2b6e49cd069970b0ae07b039aa9c09724750e0d0 now?

CiaranMn: can we do this automatically as part of formatting?

packages/blockprotocol/cli/commands/publish.js
CiaranMn: Let's just convert this to TS later if we want to have proper type security, rather than a halfway house.

kachkaev: Hmm not sure we need to convert CLI code to TS — JS can be covered by TSC perfectly, with typings living in JSDoc. Not a blocker for MVP though.

kachkaev (approved):

PR #576, opened 2022-09-06
Expose registerCallbacks to end users. Fix related bug in mock-block-dock

CiaranMn:
# What this does

This PR allows users to update multiple callbacks for the graph service at once at any point via exposing `graphService.registerCallbacks`, making it easier to update callbacks beyond the initial construction of the service (the constructor already accepts multiple callbacks to register).

It uses this new functionality to fix a bug in `mock-block-dock`, whereby callbacks were not updated in the service if they changed during the session. For example, they would be fixed in whatever `readonly` state was applied when the service was first initialized.

## Test case
1. Add `readonly` to the props passed to `MockBlockDock in `packages/mock-block-dock/dev/dev-app.tsx`
2. `cd packages/mock-block-dock && yarn dev`
4. See the block load in readonly mode (no text input)
5. Click the read only toggle to turn readonly mode off, and try and update the block's name in the text input field – it should work
6. In a separate window `git checkout main`
7. Repeat step 4-5 - it should not work

- [Internal Asana task link](https://app.asana.com/0/1200211978612931/1202923029147031/f)


packages/@blockprotocol/graph/src/graph-block-handler.ts
```diff
@@ -68,6 +68,20 @@ export class GraphBlockHandler
     return {};
   }
 
+  /**
+   * Registers multiple callbacks at once.
+   * Useful for bulk updates to callbacks after the service is first initialised.
+   */
+  registerCallbacks(callbacks: Partial<BlockGraphMessageCallbacks>) {
+    super.registerCallbacks(callbacks);
```
nathggns, line 76: What's the point of overriding a parent function when all you do is call it?

nathggns (approved):

packages/@blockprotocol/graph/src/graph-block-handler.ts
CiaranMn: For narrower typing. This could be achieved with generics, and I just started to do it, but I think it's going to take time disproportionate to its benefit to do properly

PR #577, opened 2022-09-06
Extract `@local/tsconfig` workspace

kachkaev:
_similar to https://github.com/hashintel/hash/pull/470; enables further work in #551_

This PR introduces a new workspace named `@local/tsconfig`. This new private package contains three TS configs to extend from:
- `node.json` – for CLIs or script-only packages
- `dom+jsx.json` – for packages that target browsers (being built)
- `next.json` – for Next.js apps (no build step)

These three config files save us from repeating compiler options in each workspace. This helps us maintain high strictness level without worrying about things going out of sync.

Note that we cannot extend from `@local/tsconfig` in block templates, so their `tsconfig.json` files are still standalone.  I've removed a couple of unused compiler options and have enabled strict mode in them.

Importantly, root `tsconfig.json` and `tsconfig.base.json` files have been removed. This means that each workspace (package) needs to be configured individually. This is good because it reminds us to pick the right config to extend from. IDEs should be fine with this change (at least VSCode has no problem picking the closest `tsconfig.json` file).

---

**Caveat:** I've switched to `"target": "ES2020"`, which is higher than what we used (it was a mix). This change should reduce the size of `dist` folders, but might also introduce syntax issues in some runtimes. I believe that this should not happen because:
- Our min engine requirement is Node 16.14, which is pretty modern
- When packages are used in apps, they should be bundled / polyfilled by a build tool if needed

I suggest trying this target and lowering it only if we experience real problems.

---

See more details in PR comments.

---
  - https://app.asana.com/0/0/1202210072243248


packages/@blockprotocol/type-system-node/tsconfig.json
```diff
@@ -1,31 +1,5 @@
...
-    "declaration": true,
-    "esModuleInterop": true,
-    "forceConsistentCasingInFileNames": true,
-    "isolatedModules": true,
-    "jsx": "react-jsx",
-    "lib": ["dom", "dom.iterable", "esnext"],
-    "module": "esnext",
-    "moduleResolution": "node",
-    "outDir": "dist/esm",
-    // @todo: Change this and check as part of CI, we want to make sure that the generated bindings files are correct
```
kachkaev, line 14: [Asana task](https://app.asana.com/0/1202839302145209/1202922733837195/f) (internal)

.changeset/config.json
```diff
@@ -9,7 +9,7 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@blockprotocol/site", "@local/*"],
+  "ignore": ["@blockprotocol/site", "@crates/*", "@local/*"],
```
kachkaev, line 12: Without root `tsconfig.json`, the only way for us to lint `crates/type-system/scripts/build-wasm.ts` is to create a new workspace at `crates/type-system`. We don't want to publish it, hence this change in Changeset config

packages/@blockprotocol/core/package.json
```diff
@@ -29,8 +29,8 @@
   ],
   "scripts": {
     "build": "yarn clean && yarn build:cjs && yarn build:esm",
-    "build:cjs": "tsc --module commonjs --outDir dist/cjs ",
-    "build:esm": "tsc",
+    "build:cjs": "tsc --project tsconfig.build.cjs.json",
+    "build:esm": "tsc --project tsconfig.build.esm.json",
```
kachkaev, line 33: Sticking with `--project` file instead of on-the-fly args

packages/@blockprotocol/core/src/html.ts
```diff
@@ -167,7 +167,7 @@ export const markBlockScripts = (container: HTMLElement, url: URL | string) => {
         let nextSource = prevSource;
 
         nextSource += html.substring(
-          idx === 0 ? 0 : relevantImports[idx - 1].se,
+          idx === 0 ? 0 : relevantImports[idx - 1]!.se,
```
kachkaev, line 170: Looks like we did not have strict null checks in this package.

packages/@blockprotocol/type-system-web/package.json
```diff
@@ -18,6 +18,10 @@
   "scripts": {
     "lint:tsc": "tsc --noEmit"
   },
+  "devDependencies": {
+    "@local/tsconfig": "0.0.0-private",
+    "typescript": "4.8.2"
```
kachkaev, line 23: Both `@local/tsconfig` and `typescript` need to be present in all TS workspaces.

packages/@local/eslint-config/index.js
```diff
@@ -253,6 +250,54 @@ module.exports = {
         "global-require": "off",
       },
     },
+    {
```
kachkaev, line 253: This long list of overrides will be removed in #551. We need for now after removing root `tsconfig.json`.

packages/@local/script-resources/suppress-experimental-warnings.cjs
```diff
@@ -4,10 +4,11 @@
 
 const originalEmit = process.emit;
 
+// @ts-expect-error -- TS is unable to untangle all emit overloads
```
kachkaev, line 7: `checkJs` was not enabled for this package

packages/@local/tsconfig/dom+jsx.json
```diff
@@ -0,0 +1,8 @@
+{
+  "$schema": "http://json.schemastore.org/tsconfig",
+  "extends": "./node.json",
+  "compilerOptions": {
+    "jsx": "react-jsx",
+    "lib": ["DOM", "DOM.Iterable", "ESNext"]
```
kachkaev, line 6: Using letter casing from schema 🤷

packages/block-scripts/bin.js
```diff
@@ -8,7 +8,7 @@ import { listAvailableScriptNames } from "./shared/list-available-script-names.j
 const scriptConfig = parser(process.argv.slice(2));
 const scriptName = scriptConfig.help
   ? "help"
-  : (scriptConfig._.shift() ?? "help").toLowerCase();
+  : `${scriptConfig._.shift() ?? "help"}`.toLowerCase();
```
kachkaev, line 11: This file was not checked previously

packages/block-scripts/package.json
```diff
@@ -48,13 +48,19 @@
     "typescript": "4.8.2",
     "typescript-json-schema": "^0.53.0",
     "uuid": "^8.3.2",
-    "webpack": "^5.72.0",
+    "webpack": "^5.74.0",
```
kachkaev, line 51: I added `@types/webpack-assets-manifest` and `@types/webpack-bundle-analyzer` to fix missing typings and TSC reported type mismatches. Upgrading webpack by two minor versions fixed this (webpack comes with built-in typings).

packages/block-scripts/scripts/build.js
```diff
@@ -15,8 +15,8 @@ const script = async () => {
     await generateBuildWebpackConfig("production"),
   ]);
 
-  if (stats.hasErrors()) {
-    console.log(stats.toString());
+  if (!stats || stats.hasErrors()) {
```
kachkaev, line 18: File was not checked previously

packages/block-scripts/shared/webpack-config.js
```diff
@@ -53,7 +53,7 @@ const generateBaseWebpackConfig = async (mode) => {
       filename: mode === "production" ? "[name].[contenthash].js" : "[name].js",
     },
     externals: Object.fromEntries(
-      Object.keys(peerDependencies ?? {}).map((key) => [key, key]),
+      Object.keys(peerDependencies || {}).map((key) => [key, key]),
```
kachkaev, line 56: We don't compile block-scripts, so using || instead of ?? just in case for lower Node versions.

CiaranMn (approved):

PR #583, opened 2022-09-07
Patch `@changesets/get-dependents-graph`

kachkaev:
Context: https://github.com/changesets/changesets/issues/906#issuecomment-1239468046

Not sure which of the three dist files is used, so I’ve patched all of them.

The change mirrors a potential permanent fix: https://github.com/changesets/changesets/pull/907/files#diff-9816585d4f6aa370adc5fbc0d534aa7e558282bba32232a63c487952972c7a4cR23


kachkaev:
Release workflow turned green again! 🕺
https://github.com/blockprotocol/blockprotocol/runs/8232022738?check_suite_focus=true#step:4:1

PR #584, opened 2022-09-07
Enable TSC in `blockprotocol` package

kachkaev:
`blockprotocol` package was re-introduced in #581 as a CLI. When merging #577, I kept this package unlinted to keep the diff focused. The aim of this PR is to make sure that `blockprotocol` has `lint:tsc` and passes this check.

---
  - https://app.asana.com/0/0/1202922734858538


packages/blockprotocol/cli/commands/publish/post-form.js
```diff
@@ -35,33 +37,46 @@ export const postPublishForm = async ({
...
+              { msg: `Could not connect to host ${blockProtocolSiteHost}` },
+            ],
+          };
+        }
+        if ("errors" in error) {
+          // should be a meaningful error provided by the API about why the request failed, e.g. bad API key
+          return error;
+        }
       }
-      return err; // should be a meaningful error provided by the API about why the request failed, e.g. bad API key
```
kachkaev, line 65: Am I right that we expect something like this here? `{ errors: [{ msg: "hello" }] }`

CiaranMn: yes, exactly

CiaranMn (approved):
Thanks Alex - tested and all working as expected

PR #588, opened 2022-09-08
Fix opening modals resetting page scroll position issue

yusufkinatas:

This PR fixes the issue that causes opening modals to reset page scroll position. Losing the scroll position is pretty annoying, especially when you're reading through the documentation, and lose where you were few seconds ago.

## 🔍 What does this change?

- Remove custom `useScrollLock` solution that we implemented over MUI's default scroolLock [in this PR](https://github.com/blockprotocol/blockprotocol/pull/115), and started using MUI's scroll-lock again.
- The reason for custom scrollLock implementation was the concern of MUI's scroll-lock breaking the layout. Because MUI adds `padding-right` to `body` element. 
- I investigated this, and found out that using MUI's solution breaks the header layout, because it's adding the padding to the body, but the header does not respect that padding. Since our header positioned `fixed` on some cases, which does not respects the padding of the body.
- And the custom implementation was adding `position:fixed` to `html` element itself. Which resets the scroll position of the content, and I found no easy way of fixing that.
- My solution to `mui's scroll-lock breaking the fixed header` issue is basically the following:
  - Apply the same logic thta MUI applies to the `body` element when the scroll-lock is enabled to our custom header wrapper.
  - More detailed explanation: Make header wrapper always `100vw`, and add the scrollbar width as `padding-right` when scrollbar is hidden. So we can achieve the same logic as MUI. (Check [MUI's implementation](https://github.com/mui/material-ui/blob/master/packages/mui-base/src/ModalUnstyled/ModalManager.ts#L100-L103))
  - Actually MUI is aware of fixed items jumping around when modals are opened. So they suggest (See [MUI's `mui-fixed` suggestion](https://mui.com/material-ui/getting-started/faq/#why-do-the-fixed-positioned-elements-move-when-a-modal-is-opened)) using a helper class called `.mui-fixed`, and they provide the calculated padding-right to the fixed elements. But unfortunately our `mobile slide navigation` is causing the same jumping issue. And it's a `Slide`, not a `Modal`. So using `mui-fixed` className does not prevent the header jumping issue on mobile. So I implemented the same behavior, but not inside `modal` component, rather than our only fixed element that's having this issue, `header`. 

## ❓ How to test this?

1.  Scroll down in a page
2. Open any modal (e.g. press `shift + /` to open search modal)
3. Close the modal, and check if scroll position is resetted or not. It should not reset.
4. Also check opening/closing modals and see header is not jumping because of scrollbar showing/hiding

## 📹 Demo

Detailed comparison of modal scroll-lock behavior between `this branch` and `main`
https://youtu.be/5usAFMBa3QM

_Uploaded on youtube because of file size limit here._

---
  - https://app.asana.com/0/0/1202765662321717


nathggns:
Looking at this now

site/src/theme/components/utils/mui-css-baseline-theme-options.ts
```diff
@@ -151,9 +151,7 @@ export const MuiCssBaselineThemeOptions: Components["MuiCssBaseline"] = {
           }
 
           body {
-            overflow: auto;${
-              "" /** @todo: find a pernament solution for preventing the navbar from shifting to the right when a modal is opened on an overflowing page */
-            }
+            overflow-x: hidden !important;
```
nathggns, line 154: Why is this still necessary?

nathggns:
@yusufkinatas can we create a ticket around the page jumping around when opening the mobile nav? I don't know what we're doing wrong that's causing that (but don't think you caused it in this PR)

nathggns (approved):

site/src/theme/components/utils/mui-css-baseline-theme-options.ts
yusufkinatas: Since header is `100vw` and gets a padding same as scrollbar, is overflows on x axis as much as scrollbar width if scrollbar is visible. But this prevents the jumping on header.

PR #589, opened 2022-09-08
Introduce 'labelProperty' to mock-block-dock and EntityType

yusufkinatas:
## Purpose of this Pull Request

The Block Protocol specifies that entity type schemas may define a '[labelProperty](https://blockprotocol.org/docs/spec/graph-service#json-schema-extensions)' which indicates the field on an entity that should be used as its display property.

This PR adds field to the `EntityType` definition in `@blockprotocol/core`, and adds values for the field to the dummy data in `mock-block-dock`

Implements changes originally created as patches in this PR: https://github.com/hashintel/hash/pull/1033

---
  - https://app.asana.com/0/0/1202944443009560


.changeset/calm-lies-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blockprotocol/graph": patch
+---
+
+create a patch release for /graph
```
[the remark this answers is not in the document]
yusufkinatas: Hmm, I see. Due to the [task description here](https://app.asana.com/0/1201095311341924/1202944443009560/f) _(internal)_, I thought this is a common practice to write this if we're creating a changeset for a patch-package. Let me update the descriptions.

packages/@blockprotocol/graph/src/types.ts
```diff
@@ -173,6 +173,7 @@ export type EntityType = {
     $schema: string;
     title: string;
     type: string;
+    labelProperty: string;
```
CiaranMn, line 176:
```suggestion
    labelProperty?: string;
```
I think we made it required in HASH to remind us to enforce it on predefined schemas, but it should be optional here because it's optional in the spec – apologies, I should've noticed / said this before.

CiaranMn (dismissed):

kachkaev (approved):

packages/@blockprotocol/graph/src/types.ts
yusufkinatas: Updated

PR #598, opened 2022-09-12
Improve clarity and explanation in block templates, and base custom element class

CiaranMn:
# What this does

This PR introduces a few things to improve the usability of the block templates:
1. Comments explaining what is going on
2. Simplifies templates where appropriate to make it clearer the underlying graph features being used (e.g. remove unnecessary memoization and abstractions from the React template, and the custom element base class)
3. Adds a `format` command to each template so that users can easily format (or their editors autoformat) when blocks are created outside of an existing repo

As part of this it also adds formatting of CSS files and HTML files to Prettier for the entire repo.

---
  - https://app.asana.com/0/0/1202952622546732


packages/block-template-react/src/app.tsx
```diff
@@ -2,41 +2,101 @@ import {
...
+ * To change the structure of the entity your block expects, change this type.
+ */
 type BlockEntityProperties = {
   name: string;
 };
 
+/**
+ * This function is to help illustrate a property being change when the button is pressed.
+ * You can remove it when you remove the button below.
+ */
```
yusufkinatas, line 29: Not sure if we need to explain what this function does (it looks too obvious to me). Or maybe we can remove the second line. Imagine default React project explains us "you can remove this image import when you remove the `img` below" on the default React project. I felt like we're explaining "you should remove unused functions from you code" to developers. 

Maybe I'm wrong, just explained my gut feeling when I first saw this comment

packages/block-template-react/src/app.tsx
```diff
@@ -2,41 +2,101 @@ import {
...
 
-  const updateSelf = useCallback(
-    (newProperties: Partial<BlockEntityProperties>) =>
-      graphService?.updateEntity({
-        data: { properties: newProperties, entityId },
-      }),
-    [entityId, graphService],
-  );
-
+  /** Here we extract the 'name' property from the blockEntity's properties ("object destructuring") */
```
yusufkinatas, line 65: Hmm, not sure if we should explain ""object destructuring".

yusufkinatas (dismissed):
Great explanation & comment lines! Dropped few minor questions. Feel free to address them.

packages/block-template-react/src/app.tsx
CiaranMn: Thanks! Removed the second line in https://github.com/blockprotocol/blockprotocol/pull/598/commits/3c41770f16d43573e718be49b1e379a9d3444454

packages/block-template-react/src/app.tsx
CiaranMn: Removed in https://github.com/blockprotocol/blockprotocol/pull/598/commits/3c41770f16d43573e718be49b1e379a9d3444454

PR #599, opened 2022-09-12
Add entity switching to mock block dock

CiaranMn:
Recreation of #557 

- introduces entity switching functionality via mock-block-dock's debug UI (choose any entity from the datastore to load into the block)
- adds a block Info tab for displaying information about the block

Plus
- removes the 'upload schema' feature (the schema doesn't affect anything at the moment so unclear why users would want to do it - I think it's a distraction until we make it do something)
- adds a `name` field to the `blockprotocol` object in package.json in the templates (and as part of `create-block-app`), and uses this to populate the `name` field in `block-metadata.json` – this is to allow blocks to have a different `name` field to the one in `package.json`, e.g. in case they need to have a local workspace or npm-unique name in package.json, but something shorter for their block name in the Hub
- updates construction of the graph service so that it responds correctly to toggling debug / not debug mode
- UI improvements
- makes debug styles a module (this was an attempt to mollify NextJS, but it doesn't allow importing CSS modules either – styles to be in-lined in a followup PR)


packages/mock-block-dock/src/debug-view/dev-tools.tsx
```diff
@@ -89,18 +88,6 @@ const ResizeHandle = forwardRef<HTMLDivElement, any>((props, ref) => {
   );
 });
 
-const chipInfo = {
```
CiaranMn, line 92: moved into `block-info-view.tsx`

packages/mock-block-dock/src/assets/debug-view-styles.module.css
```diff
@@ -27,17 +28,23 @@
   src: url("./jet-brains-mono-regular.woff2") format("woff2");
 }
 
-.mbd-non-debug-mode-wrapper {
-  padding-top: 60px;
+body {
+  margin: 0;
+}
```
kachkaev, line 33: Is this allowed downstream? E.g. in Next.js

kachkaev (approved):
👍 Works on my machine™

Screenshot 2022-09-14 at 14 37 03

packages/mock-block-dock/src/assets/debug-view-styles.module.css
CiaranMn: No – it seems we will have to inline these styles. I will do it in a follow up since the problem exists already

PR #600, opened 2022-09-13
Update `@changesets/get-dependents-graph` patch

kachkaev:
Replaces https://github.com/blockprotocol/blockprotocol/pull/583

The original patch was inspired by https://github.com/changesets/changesets/pull/907. It ignored `devDependencies` when calculating new versions to release, which worked generally well. The solution, however, did not cover block templates, which had `block-scripts` in `devDependencies`.

This PR updates the way we list dependencies in a topological graph. A package is excluded if it is `private`. This should cover existing cases and also work for `block-scripts`.

[Slack thread](https://hashintel.slack.com/archives/C02LG39FJAU/p1662969660652819) (internal)

### DIFF
--- a/patches/@changesets+get-dependents-graph+1.3.3.patch
+++ b/patches/@changesets+get-dependents-graph+1.3.3.patch
@@ -1,42 +1,42 @@
 diff --git a/node_modules/@changesets/get-dependents-graph/dist/get-dependents-graph.cjs.dev.js b/node_modules/@changesets/get-dependents-graph/dist/get-dependents-graph.cjs.dev.js
-index 9bd1b9d..9e1a33d 100644
+index 9bd1b9d..cbacc42 100644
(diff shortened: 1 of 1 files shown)


patches/@changesets+get-dependents-graph+1.3.3.patch
```diff
@@ -1,42 +1,42 @@
...
-+    
-     const deps = config[type];
-     if (!deps) continue;
+@@ -71,6 +71,9 @@ function getDependencyGraph(packages, opts) {
+     for (let [depName, depRange] of allDependencies) {
+       const match = packagesByName[depName];
+       if (!match) continue;
++      if (match.packageJson.private) {
++        continue;
++      }
```
kachkaev, line 11: 👆 this

yusufkinatas (approved):

PR #602, opened 2022-09-13
Bump package version when `devDependencies` update

kachkaev:
Follows https://github.com/blockprotocol/blockprotocol/pull/600. With this change, we should hopefully see new versions in `block-template-*` when `mock-block-dock` and `block-scripts` are updated.

There is a chance that the PR will break _Version Packages_ completely. If this happens, I’ll revert the PR.


CiaranMn (approved):
🤞

PR #603, opened 2022-09-13
Add cd step to post-install instructions

CiaranMn:
After creating a block, we tell users they can run 'yarn install && yarn dev' to get started.

Actually they'll need to change into the block dir first. This PR adds that step to the instructions, and a bit of colour.

Screenshot 2022-09-13 at 17 39 01

### DIFF
--- a/packages/create-block-app/create-block-app.js
+++ b/packages/create-block-app/create-block-app.js
@@ -157,7 +157,18 @@ const usage = commandLineUsage(helpSections);
     await fs.writeJson(blockMetadataPath, blockMetadata, { spaces: 2 });
   }
 
+  const chalk = (await import("chalk")).default;
+
   console.log(
-    `Your ${blockName} block is ready to code in ${resolvedBlockPath}.\nRun 'yarn install && yarn dev' to get started`,
+    chalk.bold("-".repeat(Math.min(process.stdout.columns ?? 48, 48))),
+  );
+  console.log(
+    `Your ${chalk.bold(blockName)} block is ready to code in ${chalk.bold(
+      resolvedBlockPath,
+    )}.\n` +
+      `Run ${chalk.blue(`cd ${resolvedBlockPath}`)} and then \n` +
+      `${chalk.blue("yarn install && yarn dev")} or ${chalk.blue(
+        "npm install && npm run dev",
+      )} to get started`,
   );
 })();
--- a/packages/create-block-app/create-block-app.js
+++ b/packages/create-block-app/create-block-app.js
@@ -171,7 +171,7 @@ const usage = commandLineUsage(helpSections);
   );
   console.log(
     `Your ${chalk.bold(blockName)} block is ready to code in ${chalk.bold(
-      escapedPath,
+      resolvedBlockPath,
     )}\n` +
       `Run ${chalk.blue(`cd ${escapedPath}`)} and then \n` +
       `${chalk.blue("yarn install && yarn dev")} or ${chalk.blue(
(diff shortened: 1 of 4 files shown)


packages/create-block-app/create-block-app.js
```diff
@@ -157,7 +157,18 @@ const usage = commandLineUsage(helpSections);
...
+
+  console.log(
+    chalk.bold("-".repeat(Math.min(process.stdout.columns ?? 48, 48))),
+  );
   console.log(
-    `Your ${blockName} block is ready to code in ${resolvedBlockPath}.\nRun 'yarn install && yarn dev' to get started`,
+    `Your ${chalk.bold(blockName)} block is ready to code in ${chalk.bold(
+      resolvedBlockPath,
+    )}.\n` +
+      `Run ${chalk.blue(`cd ${resolvedBlockPath}`)} and then \n` +
```
kachkaev, line 169: If `resolvedBlockPath` contains spaces, the command will end up broken. Let’s escape this value: https://github.com/nodejs/node/issues/34840#issuecomment-676398171

packages/create-block-app/create-block-app.js
```diff
@@ -157,7 +157,18 @@ const usage = commandLineUsage(helpSections);
...
+      `${chalk.blue("yarn install && yarn dev")} or ${chalk.blue(
+        "npm install && npm run dev",
```
kachkaev, line 171: Optional: if we mention `npm install` here, we could do the same [in docs](https://blockprotocol.org/docs/developing-blocks#creating-a-block).

kachkaev (changes requested):
See comments above

packages/create-block-app/create-block-app.js
kachkaev: Optional (not sure in this). We can also use [`path.relative`](https://nodejs.org/api/path.html#pathrelativefrom-to) in `cd` command to make it shorter. It will be just `./block-subpath` in most cases.

packages/create-block-app/create-block-app.js
CiaranMn: Thanks - done in https://github.com/blockprotocol/blockprotocol/pull/603/commits/cd1b786aa46d7cd98341c96b12de6a9d4ace44d8

packages/create-block-app/create-block-app.js
CiaranMn: Both done in https://github.com/blockprotocol/blockprotocol/pull/603/commits/cd1b786aa46d7cd98341c96b12de6a9d4ace44d8

packages/create-block-app/create-block-app.js
```diff
@@ -158,7 +158,24 @@ const usage = commandLineUsage(helpSections);
...
+    ? relativeBlockPath
+    : `'${relativeBlockPath.replace(/'/g, `'"'`)}'`;
+
+  console.log(
+    chalk.bold("-".repeat(Math.min(process.stdout.columns ?? 48, 48))),
+  );
   console.log(
-    `Your ${blockName} block is ready to code in ${resolvedBlockPath}.\nRun 'yarn install && yarn dev' to get started`,
+    `Your ${chalk.bold(blockName)} block is ready to code in ${chalk.bold(
+      escapedPath,
```
kachkaev, line 174:
```suggestion
      resolvedBlockPath,
```

kachkaev (commented):
Nice! Just one small suggestion

kachkaev (approved):

PR #585, opened 2022-09-07
Remove `skipLibCheck` from most workspaces

kachkaev:
Enabled by #577, [Asana task](https://app.asana.com/0/0/1202922733837195) (internal)

This PR removes [`skipLibCheck`](https://www.typescriptlang.org/tsconfig#skipLibCheck) from `@local/tsconfig`. This makes `yarn lint:tsc` more rigorous, which helps us spot issues in `.d.ts` files (mostly in `node_modules`). That can guard us from runtime bugs, because with `skipLibCheck` we may get `any` from some library typings withour realising that.

TSC without `skipLibCheck` are a bit slower, so I’ve introduced `--skipLibCheck` argument in build scripts, where applicable. This makes cold `yarn install` a bit faster (≈25s → ≈15s). Type integrity is still validated via `lint:tsc`.

I had to fix several TypeScript issues for CI to pass – you can find details of these changes in PR comments. The only workspace I have left with `skipLibCheck` for now is `site`. It still have quite a few issues:

<details>
<summary><code>Found 38 errors in 24 files.</code></summary>
<pre>
../node_modules/mdast-util-mdx-expression/complex-types.d.ts:35:5 - error TS2300: Duplicate identifier 'mdxFlowExpression'.

35     mdxFlowExpression: MdxFlowExpression
       ~~~~~~~~~~~~~~~~~

../node_modules/mdast-util-mdx-expression/complex-types.d.ts:36:5 - error TS2300: Duplicate identifier 'mdxFlowExpression'.

36     mdxFlowExpression: MdxFlowExpression
       ~~~~~~~~~~~~~~~~~

../node_modules/mongodb/mongodb.d.ts:1744:97 - error TS2344: Type 'TLocal' does not satisfy the constraint 'Document'.

1744     watch<TLocal = TSchema>(pipeline?: Document[], options?: ChangeStreamOptions): ChangeStream<TLocal>;
                                                                                                     ~~~~~~

  ../node_modules/mongodb/mongodb.d.ts:1744:11
    1744     watch<TLocal = TSchema>(pipeline?: Document[], options?: ChangeStreamOptions): ChangeStream<TLocal>;
                   ~~~~~~~~~~~~~~~~
    This type parameter might need an `extends Document` constraint.

../node_modules/mongodb/mongodb.d.ts:2536:99 - error TS2344: Type 'TSchema' does not satisfy the constraint 'Document'.

[...]

PR #605, opened 2022-09-13
MBD graph visualisation of local datastore

benwerner01:
This PR adds a graph visualisation to the "datastore" tab in the mock-block-dock developer environment. It uses the `echarts` library to create a force directed graph of the entities and links in the datastore.

## Shortcomings

Due to unexpected limitations encountered with the ECharts library, there are several shortcomings of the visualisation implementation introduced in this PR. A [follow-up task](https://app.asana.com/0/1200211978612931/1202974955902879/f) has been created for addressing these.

### Shortcoming 1 - Buggy behaviour when setting both `draggable: true` and `roam: true`

When setting `draggable: true` (so that nodes can be dragged around) and `roam: true` (so that the user can pan around the visualisation using their mouse) buggy behaviour occurs in the ECharts visualisation:

https://user-images.githubusercontent.com/42802102/190125968-8eaae7aa-432c-4135-8b3c-6cef5d9da78b.mov

### Shortcoming 2 - Cannot set existing node in simulation to a fixed position on click when it is clicked

To set the position of a node in a force directed layout, you can toggle the node to [`fixed: true`](https://echarts.apache.org/en/option.html#series-graph.data.fixed) and give it an `x` and `y` position. This works well when initialising the visualisation with a fixed node (as in [this official echarts example](https://echarts.apache.org/examples/en/editor.html?c=graph-force-dynamic)). However, when updating an existing node in the visualisation to go from `fixed: false` to `fixed: true` using `setOptions(..., { notMerge: false })` the change is not reflected in the graph (i.e. the node is not moved to the fixed position).

Interestingly when using `setOptions(..., { notMerge: false })` the change does get get reflected in the graph (i.e. the node that was clicked moves to the specified position). However there’s no animation between the positions of the node, and the position of all other nodes is reset in the layout - so it results in an undesirable transition. I think this is because `notMerge: true` essentially replaces the entire state of the chart, which is why we should be using `notMerge: false` instead.

https://user-images.githubusercontent.com/42802102/190134356-1eaac574-2c31-484e-9b33-555b06375430.mp4

### Shortcoming 3 - Cannot display link labels on hover without moving the position of nodes

To display a hidden label of an edge I need to send an update to the ECharts state (setting `show: true` for the label of the edge). However on every update to the ECharts state, the force simulation moves all the node, resulting in an undesirable effect:

https://user-images.githubusercontent.com/42802102/190136088-77817e25-2409-4990-acb3-799c438010bd.mov

I have an idea of why this may be happening based on some prior experience in D3. In D3 you are able to control [an alpha variable](https://github.com/d3/d3-force#simulation_alpha) in force directed graphs, which starts at some value and gradually decreases to 0, which controls whether or not the nodes are allowed to be moved by the forces in the simulation. When adding/removing nodes typically you’d reset the alpha variable to some number to give nodes an opportunity to adjust to a “natural” position in the force layout, and then come to rest when `alpha` reaches `0`. However when performing other updates (such as displaying a label), you typically would not reset the alpha variable as you don’t want nodes to move unnecessarily.

What I think is happening is that on every call to `setOptions` the the equivalent of the `alpha` value in echarts gets reset, so that if any nodes were added to the graph they move to a "natural" position in the force layout. I couldn't find a way to stop this behaviour from happening for updates which do not change the number of nodes/edges in the graph. I also couldn't find an alternative way of displaying/hiding the labels, short of manually overlaying them over the visualisation.

---
  - https://app.asana.com/0/0/1202765662321747


packages/mock-block-dock/src/debug-view/dev-tools/datastore-graph-visualisation.tsx
```diff
@@ -0,0 +1,196 @@
...
+  initialEdges?: GraphSeriesOption["edges"];
+}): EChartOption => ({
+  color: ["#6F59EB"],
+  series: [
+    {
+      type: "graph",
+      layout: "force",
+      draggable: true,
+      /** @todo: find way to get this working, currently interferes with `draggable` */
+      // roam: true,
```
benwerner01, line 32: When setting `draggable: true` (so that nodes can be dragged around) and `roam: true` (so that the user can pan around the visualisation using their mouse) buggy behaviour occurs in the ECharts visualisation:

https://user-images.githubusercontent.com/42802102/190125968-8eaae7aa-432c-4135-8b3c-6cef5d9da78b.mov

What I've opted to do instead is to not set `roam: true` for now,

benwerner01, line 32: @CiaranMn we have a choice of either:
1. `draggable: false` and `roam: true` (user cannot rearrange nodes, but can pan the visualisation) or
2. `draggable: true` and `roam: false` (user can rearrange nodes, but can't pan the visualisation).

I've opted for 2 for now. This works fine for small graphs, but will be problematic for graphs that exceed the size of the wrapping container (either because they're very large, or because there is a small screen width). Let me know if you'd like me to change this.

CiaranMn: I think we're better off with option (1) – I think it'll be more important for users to be able to explore the graph than to rearrange it. Let me know if there's an advantage to rearranging that would change this (is it hard to see without rearranging in some circumstances?)

benwerner01: > is it hard to see without rearranging in some circumstances?

Sometimes yes, when labels overlap for instance.

benwerner01: I don't think so, based on [their docs](https://apache.github.io/echarts-handbook/en/concepts/chart-size/) and examples it seems you have to tell echarts the dimensions of the container before it even starts creating the graph. I haven't come across some way of extracting how big the produced graph is if it exceeds the bounds of the container.

It might be possible to give it a temporary dimension, let it produce the graph and then manually query the SVG in the HTML tree to figure out how big the graph is and then readjust the container - but I think that would be overkill

packages/mock-block-dock/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "@local/tsconfig/dom+jsx.json",
+  "compilerOptions": {
+    "skipLibCheck": true
+  },
```
benwerner01, line 5: Keen to get thoughts on this, without this I was getting type errors when linting (in `node_modules/echarts/types/dist/shared.d.ts` )

cc @kachkaev

kachkaev: What error is it? Can you fix it via `patch-package`? That’d be a preferred option IMO. When `skipLibCheck` is `true`, TypeScript replaces conflicting types with `any` and this increases the chances of runtime errors.

I’m not against `skipLibCheck`, but we should treat it as a last resort, if all other solutions are infeasible. ATM the only package without this flag is `site` – it was excluded from #585 to get the PR sooner than later.

packages/mock-block-dock/src/debug-view/dev-tools/datastore-graph-visualisation.tsx
CiaranMn: Fair enough - let's leave it how it is for now

packages/mock-block-dock/src/debug-view/dev-tools/datastore-graph-visualisation.tsx
```diff
@@ -0,0 +1,211 @@
+import { Entity, Link } from "@blockprotocol/graph/.";
+import { Box } from "@mui/material";
+import { GraphChart, GraphSeriesOption } from "echarts/charts";
+import * as echarts from "echarts/core";
+import { SVGRenderer } from "echarts/renderers";
+import { useEffect, useRef, useState } from "react";
+
+import { useMockBlockDockContext } from "../../mock-block-dock-context";
+
+const parseNameFromEntity = (entity: Entity): string =>
```
CiaranMn, line 10: We have a newish schema property in the BP, `labelProperty`, which allows users to specify which field on an entity should be used as its display name. We have this set on the sample entities in MBD so we should use it here, with some sensible fallbacks.

I've more or less lifted the approach in HASH in [this commit](https://github.com/blockprotocol/blockprotocol/commit/1538378979a2ac693fa07f1c0d5dda0f0202890e) as an example

CiaranMn (commented):
Thanks @benwerner01 - I think we can make one improvement to entity label generation. Looks good otherwise, will help understanding / debugging a lot!

packages/mock-block-dock/src/debug-view/dev-tools/datastore-graph-visualisation.tsx
benwerner01: Thanks Ciaran, missed that this made it into MBD already - have cherry-picked your commit in https://github.com/blockprotocol/blockprotocol/pull/605/commits/d01e699d3b7fa7c6add939d174f7078569831065 👍

packages/mock-block-dock/tsconfig.json
benwerner01: Have taken another look and think have fixed it with a patch in https://github.com/blockprotocol/blockprotocol/pull/605/commits/bd8d5a901ed5ac12faec925ae2fe41f4b6cfc82a. If this is preferable, I'll merge-in the patch instead of `skipLibCheck: true`?

CiaranMn: This is definitely preferable – this way we're not skipping the check of all libraries (safer), and we can easily tell where the issue is (because there's a patch) – vs `skipLibCheck: true` where we'll have no idea in 4 weeks why it's on without turning it off and seeing what breaks.

CiaranMn (approved):

PR #610, opened 2022-09-15
type-system overview through diagram

indietyp:
This PR adds a visualization of the type-system via a diagram in the `README.md`, because there isn't any documentation just yet, I have created a new heading with the diagram.


Alfred-Mountfield (approved):
Awesome! Always appreciate architectural diagrams.

PR #611, opened 2022-09-15
Use `React.lazy` in `mock-block-dock` to optimize MUI loading

kachkaev:
The goal of this PR is to optimize sandbox performance. The changes cannot be tested in full until we've released the package on `esm.sh`. Manual smoke-testing testing can be done locally by following these steps:

1. Build MBD  
   ```sh
   cd packages/mock-block-dock
   yarn build
   ```

2. Launch React template   
   ```sh
   cd packages/block-template-react
   yarn dev
   ```

3. Replace `debug` with `hideDebugToggle` in `packages/block-template-react/src/dev.tsx` 

4. Confirm that MBD at http://localhost:63212/ still works

---

  - https://app.asana.com/0/0/1202449143008136

### DIFF
--- a/packages/mock-block-dock/src/mock-block-dock.tsx
+++ b/packages/mock-block-dock/src/mock-block-dock.tsx
@@ -186,17 +186,12 @@ export const MockBlockDock: FunctionComponent<MockBlockDockProps> = ({
     }
   }, [graphService, graphServiceCallbacks]);
 
-  const wrappedBlockRenderer = (
+  const blockRendererWithBorder = (
     <div
-      ref={wrapperRef}
-      style={
-        debugMode
-          ? {
-              border: "1px dashed rgb(0,0,0,0.1)",
-              marginTop: 30,
-            }
-          : {}
-      }
+      style={{
+        border: "1px dashed rgb(0,0,0,0.1)",
+        marginTop: 30,
+      }}
     >
       {graphService ? (
         <BlockRenderer
@@ -236,13 +231,15 @@ export const MockBlockDock: FunctionComponent<MockBlockDockProps> = ({
       setEntityIdOfEntityForBlock={setEntityIdOfEntityForBlock}
       updateEntity={graphServiceCallbacks.updateEntity}
     >
-      <Suspense>
-        {hideDebugToggle && !debugMode ? (
-          wrappedBlockRenderer
-        ) : (
-          <MockBlockDockUi>{wrappedBlockRenderer}</MockBlockDockUi>
-        )}
-      </Suspense>
+      <div ref={wrapperRef}>
+        <Suspense>
+          {hideDebugToggle && !debugMode ? (
+            blockRendererWithBorder
+          ) : (
+            <MockBlockDockUi>{blockRendererWithBorder}</MockBlockDockUi>
+          )}
+        </Suspense>
+      </div>
     </MockBlockDockProvider>
   );
 };
--- a/packages/mock-block-dock/src/mock-block-dock.tsx
+++ b/packages/mock-block-dock/src/mock-block-dock.tsx
@@ -240,12 +240,7 @@ export const MockBlockDock: FunctionComponent<MockBlockDockProps> = ({
         {hideDebugToggle && !debugMode ? (
           wrappedBlockRenderer
         ) : (
-          <MockBlockDockUi
-            debugMode={debugMode}
-            onDebugModeChange={() => setDebugMode(true)}
-          >
-            {wrappedBlockRenderer}
-          </MockBlockDockUi>
+          <MockBlockDockUi>{wrappedBlockRenderer}</MockBlockDockUi>
         )}
       </Suspense>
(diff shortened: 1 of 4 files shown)


packages/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -59,16 +66,16 @@ type MockBlockDockProps = {
...
+ * @param blockDefinition.blockEntity the starting properties for the block entity
+ * @param blockDefinition.blockInfo metadata about the block
+ * @param blockDefinition.blockSchema the schema for the block entity
+ * @param blockDefinition.debug `= false` display debugging information
+ * @param blockDefinition.hideDebugToggle `= false` hide the ability to toggle the debug UI
+ * @param blockDefinition.initialEntities the entities to include in the data store (NOT the block entity, which is always provided)
+ * @param blockDefinition.initialEntityTypes the entity types to include in the data store (NOT the block's type, which is always provided)
+ * @param blockDefinition.initialLinks the links to include in the data store
+ * @param blockDefinition.initialLinkedAggregations The linkedAggregation DEFINITIONS to include in the data store (results will be resolved automatically)
+ * @param blockDefinition.readonly `= false` whether the block should display in readonly mode or not
```
kachkaev, line 78: Sidecar fix in jsdoc

packages/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -179,7 +186,7 @@ export const MockBlockDock: FunctionComponent<MockBlockDockProps> = ({
     }
   }, [graphService, graphServiceCallbacks]);
 
-  const Component = (
+  const wrappedBlockRenderer = (
```
kachkaev, line 189: `Component` (with capital letter) is confusing because it implies usage of `<Component />`. Renaming the local var reduces confusion.

packages/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -229,25 +236,13 @@ export const MockBlockDock: FunctionComponent<MockBlockDockProps> = ({
...
-              </button>
-            </Box>
-          )}
-          {Component}
-        </Box>
-      ) : (
-        <DebugView>{Component}</DebugView>
-      )}
+      <Suspense>
+        {hideDebugToggle && !debugMode ? (
```
kachkaev, line 240: Optional suggestion (out of scope): we can rename `hideDebugToggle` to `pure` or `plain` to communicate the intent a bit better.

packages/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -7,18 +7,25 @@ import {
...
 import { BlockRenderer } from "./block-renderer";
-import { DebugView } from "./debug-view";
-import { OffSwitch } from "./debug-view/icons";
 import { MockBlockDockProvider } from "./mock-block-dock-context";
 import { useDefaultState } from "./use-default-state";
 import { useMockBlockProps } from "./use-mock-block-props";
 import { useSendGraphValue } from "./use-send-graph-value";
 
+const MockBlockDockUi = lazy(async () => ({
+  default: (await import("./mock-block-dock-ui")).MockBlockDockUi,
```
kachkaev, line 26: Not quite like in docs, but it should work too:
https://reactjs.org/docs/code-splitting.html#named-exports

packages/mock-block-dock/src/mock-block-dock.tsx
CiaranMn: `blockDefinition` is one value on the props object - it is not the only one. This implies that there's an object called `blockDefinition` which all these things exist on, which is not the case.

This has also removed the [optional](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values) indicators.

What problem were you trying to fix? I don't think this does it

kachkaev: Ah yeah you are right I've made a typo. It should be fixed in https://github.com/blockprotocol/blockprotocol/pull/611/commits/e476456e59252526724dfbbb492f75743d0e1378.

I’ve been trying to resolve this notice:

Screenshot 2022-09-15 at 18 57 47

kachkaev: [JSDoc ref](https://jsdoc.app/tags-param.html#optional-parameters-and-default-values)

CiaranMn: different editor implementations of JSDoc, doh 😅 
Screenshot 2022-09-16 at 10 47 58

kachkaev: Hmmm how about with this change? https://github.com/blockprotocol/blockprotocol/pull/611/commits/82c038de24245f15ee809c57bedcb275d061cc5f

Conceptually, the function has only one param, which is `props`

CiaranMn (approved):

PR #609, opened 2022-09-14
Rename and move `site` workspace

kachkaev:
The goal of this PR is to improve alignment with common monorepo standards. We will eventually apply the same structure in [HASH repo](https://github.com/hashintel/hash) too.

### TODO

- ~~Merge https://github.com/blockprotocol/blockprotocol/pull/565~~
- ~~Update Vercel config once PR is reviewed, redeploy for testing~~

### Commands to run locally after pulling `main`

```
mv site/playwright-report apps/site/playwright-report
mv site/public/blocks apps/site/public/blocks
mv site/public/schemas apps/site/public/schemas
mv site/blocks-data.json apps/site/blocks-data.json
mv site/site-map.json apps/site/site-map.json
mv site/.env.local apps/site/.env.local
```

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202540836241707

PR #613, opened 2022-09-15
Update GitHub labels

kachkaev:
TODO after approval but before merging: sync [GitHub labels](https://github.com/blockprotocol/blockprotocol/labels) with labels in Yaml

---
  - https://app.asana.com/0/0/1202922734858534


.github/workflows/labeler.yml
```diff
@@ -6,6 +6,9 @@ jobs:
   triage:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/labeler@v3
+      - uses: kachkaev/labeler@d89797c51d07680aec17049cc6790e9d323d9a93
```
kachkaev, line 9: https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-a-commits-sha-for-release-management

.github/labeler.yml
```diff
@@ -1,104 +1,69 @@
+"area: crates":
+  - "crates/**/*"
```
nonparibus, line 2: We want to rename `crates` to `libs` in line with our agreed [monorepo structuring principles](https://www.notion.so/hashintel/Monorepo-Structure-216b3150cfdf48bb9f1bc149dd25861b), and additionally move the contents of `packages` in here (except where things may in fact be `apps`).

.github/labeler.yml
```diff
@@ -1,104 +1,69 @@
...
   - "packages/**/*"
-  - "packages/**/.*"
+
+"area: rfcs":
+  - "rfcs/**/*"
 
 "area: site":
-  - "site/*"
-  - "site/.*"
   - "site/**/*"
```
nonparibus, line 26: We should make `site` a sub-folder within `apps`

.github/labeler.yml
```diff
@@ -1,104 +1,69 @@
...
-  - "**/patches/*"
-  - "**/patches/.*"
   - "**/patches/**/*"
-  - "**/patches/**/.*"
   - "**/yarn.lock"
 
 "area: hub":
-  - "hub/*"
-  - "hub/.*"
   - "hub/**/*"
```
nonparibus, line 9: Please liaise with @CiaranMn to remove the `hub` folder in its entirety now we support `npm`-based publishing. 

Side-note: designs for a Git-based publication flow are also complete (but the corresponding dev work not yet planned). In any case, we don't want to be maintaining inbound PRs from users containing their blocks if we can avoid it. The `hub` folder and its contents should be removed once their contents have been mirrored into our block-serving solution.

.github/labeler.yml
kachkaev: The goal of this PR is to sync existing labels with existing repo structure. I suggest syncing labels again ones (if?) `crates` and `packages` are put under `lib`. Might be worth checking this idea again with @CiaranMn and @Alfred-Mountfield in early Oct.

.github/labeler.yml
kachkaev: That's in scope for https://github.com/blockprotocol/blockprotocol/pull/609 (see TODO in PR description)

.github/labeler.yml
nonparibus: There's no 'if', as our monorepo structure/principles were agreed. Unless there's new information we didn't have then, let's not re-open that finished discussion. Happy for this change to be made separately, but it does need making.

nonparibus (approved):
Thanks, Alex

PR #614, opened 2022-09-16
Notify Slack upon block publication/update on the Hub

yusufkinatas:
## What is the purpose of this PR?

This PR sends notifications to an internal slack channel when blocks on Block Hub gets updated, or new ones gets created.

## What does this change?

- Adds `SLACK_BLOCK_NOTIFICATION_WEBHOOK_URL` env variable to vercel/blockprotocol project
- Calls new `notifySlackAboutBlock` function when `api/blocks/create` or `api/blocks/publish` endpoints succeeds

## How to test this?

1. Check an [example slack notification](https://hashintel.slack.com/archives/C042NKNA2EN/p1663301969710919) _(internal)_
**OR**
1. Checkout to the branch
2. Uncomment this line: https://github.com/blockprotocol/blockprotocol/blob/24c985acfd83b6528d259783ff4b56526e6faeb3/site/src/lib/api/blocks/slack.ts#L11
3. Try publishing via local BP CLI, or http://localhost:3000/blocks/publish/npm

---
  - https://app.asana.com/0/0/1202961851742601


site/src/lib/api/blocks/slack.ts
```diff
@@ -0,0 +1,60 @@
...
+import { ExpandedBlockMetadata } from "../../blocks";
+import { isProduction } from "../../config";
+
+export const notifySlackAboutBlock = async (
+  block: ExpandedBlockMetadata,
+  changeType: "publish" | "update",
+) => {
+  if (!isProduction) return;
+
+  const webhookURL = mustGetEnvVar("SLACK_BLOCK_NOTIFICATION_WEBHOOK_URL");
```
kachkaev, line 13: Hmm not sure why we need two checks here. We can configure `SLACK_BLOCK_NOTIFICATION_WEBHOOK_URL` in prod only, which will work as both. This will allow us to test the web hook locally if we want and also disable notifications in prod, if necessary.

yusufkinatas: Hmm, so do you mean only enabling `production` for this specific env variable on Vercel?

kachkaev: The app can only care about the presence of `process.env.SLACK_BLOCK_NOTIFICATION_WEBHOOK_URL`. If the variable is not set, notifications are silent. This gives us freedom to toggle the feature in prod only via this UI:
https://vercel.com/docs/concepts/projects/environment-variables#declare-an-environment-variable

yusufkinatas: Updated in this commit, and also updated variable on Vercel https://github.com/blockprotocol/blockprotocol/pull/614/commits/68220b46969512447f72371c55d56a59e703e60c

kachkaev: Cool! I guess we’ll also need to drop `mustGetEnvVar` as it will fail locally and in preview releases. You can use `envalid` instead with `default: ""` (you'll find an example in BP or HASH repo)

site/src/lib/api/blocks/slack.ts
```diff
@@ -0,0 +1,57 @@
+import axios from "axios";
+
+import { mustGetEnvVar } from "../../../util/api";
+import { ExpandedBlockMetadata } from "../../blocks";
+
+export const notifySlackAboutBlock = async (
+  block: ExpandedBlockMetadata,
+  changeType: "publish" | "update",
+) => {
+  const webhookURL = mustGetEnvVar("SLACK_BLOCK_NOTIFICATION_WEBHOOK_URL");
```
nathggns, line 10: What happens if the env variable isn't there because it's being run locally, for example?

nathggns (changes requested):

site/src/lib/api/blocks/slack.ts
yusufkinatas: Ciaran updated this here https://github.com/blockprotocol/blockprotocol/pull/614/commits/bfa10af1fefe2710d15f5bdd92fe5faa2db996ab

Now it just returns.

nathggns (dismissed):

site/src/pages/workshop/workshop.md
```diff
@@ -375,9 +375,9 @@ export const App: BlockComponent<BlockEntityProperties> = ({
 }) => {
   ...
   // find only the entities in linkedEntities which appear as the destination of one of the projectLinks
-  const 
-  
-  
+  const
```
kachkaev, line 378: This file has been fixed in https://github.com/blockprotocol/blockprotocol/pull/641. I'm going to merge `main` and accept remote changes to make this PR ready to go 🙂

kachkaev (approved):

PR #615, opened 2022-09-16
Handle block element changing

CiaranMn:
# What it does

This PR fixes an issue in the spec and in the implementation whereby the embedding application was instructed to listen for an event from the block, and then _listen for_ and dispatch all subsequent message events from the _same_ element the block dispatched from.

The application needs to _dispatch_ from the same element so that the block can catch the events, but _listening_ on the same element means that if the block removes/replaces the element, the embedder has lost its listener and can no longer receive messages from the block.

This PR fixes that by updating the spec and implementation so that the application only dispatches from the same element the block does – it keeps its listener on an element outside the block, so that it'll always receive messages from within it. The block is required to re-send its `init` message if it changes the element it is dispatching from.

[Task](https://app.asana.com/0/1200211978612931/1202808830477307/f) (internal)

## Drive-by

- Fix a nesting issue in `service-meta.json`


nathggns:
Looking at this now

nathggns (approved):

PR #618, opened 2022-09-16
Hotfix mock-block-dock

kachkaev:
Follows #611 

- Load ESM instead of CJS via esm.sh
- Hide dashed border when `hideDebugToggle`is true

[Asana task](https://app.asana.com/0/1200211978612931/1202449143008136/f) (internal)

## Before

Screenshot 2022-09-16 at 14 28 05

## After

Screenshot 2022-09-16 at 14 28 53

### DIFF
--- a/site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
+++ b/site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
@@ -87,7 +87,7 @@ const handler: NextApiHandler = async (req, res) => {
       import React from "https://esm.sh/react@${reactVersion}"
       import ReactDOM from "https://esm.sh/react-dom@${reactVersion}"
       import { jsx as _jsx } from "https://esm.sh/react@${reactVersion}/jsx-runtime.js";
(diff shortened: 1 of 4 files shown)


site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
```diff
@@ -87,7 +87,7 @@ const handler: NextApiHandler = async (req, res) => {
       import React from "https://esm.sh/react@${reactVersion}"
       import ReactDOM from "https://esm.sh/react-dom@${reactVersion}"
       import { jsx as _jsx } from "https://esm.sh/react@${reactVersion}/jsx-runtime.js";
-      import { MockBlockDock } from "https://esm.sh/mock-block-dock@${mockBlockDockVersion}?alias=lodash:lodash-es&deps=react@${reactVersion}"
+      import { MockBlockDock } from "https://esm.sh/mock-block-dock@${mockBlockDockVersion}/dist/esm/index.js?deps=react@${reactVersion}";
```
kachkaev, line 90: We no longer use `lodash` so don't need `alias` here. I think we

CiaranMn (approved):

PR #622, opened 2022-09-16
Various block author DX improvements

CiaranMn:
# What this PR does

A collection of small things to improve block DX:
- when generating `block-metadata.json`, copy the `repository` field from `package.json` as well as other info in there
- improve the debug UI layout – previously, it was possible for the block content to be hidden without any ability to scroll, because the debug window was fixed to the bottom of the screen. Rather than giving the block content window a fixed height, I've opted to let it increase the total content beyond the view port.
- persist the user's selected debug tab to localStorage so that they don't lose it on every hot reload
- flip the 'Preview Mode' toggle, which previously was 'on' if the debug UI was visible. 'Preview Mode' makes more sense as 'hide the debug UI and show me the block as it would be in an app'
- hide the dotted border indicating the extent of the block, in preview mode (this should only appear in debug mode)
- hide the dark mode toggle since it turns the block background black and raises questions about styling etc which we haven't fully concluded the answers to yet
- add a missing import to `r2.ts` - `fetch` was called without an import, which would've crashed when this code path was hit, since it's server-side 
- copy and README tweaks / clarifications


packages/mock-block-dock/src/debug-view/header.tsx
```diff
@@ -134,18 +134,19 @@ export const Header = ({ darkMode, setDarkMode }: Props) => {
...
+      {/*  {darkMode ? ( */}
+      {/*    <OnSwitch sx={{ height: 20, width: 40, ml: 1.25 }} /> */}
+      {/*  ) : ( */}
+      {/*    <OffSwitch sx={{ height: 20, width: 40, ml: 1.25 }} /> */}
+      {/*  )} */}
+      {/* </Button> */}
 
       <Button onClick={() => setDebugMode(false)}>
         Preview Mode
-        <OnSwitch sx={{ height: 20, width: 40, ml: 1.25 }} />
```
benwerner01, line 148: Likely unrelated to this PR, but for my understanding why does the block currently lose its font styling in "Preview Mode"?
image
image

packages/mock-block-dock/src/debug-view/header.tsx
```diff
@@ -134,18 +134,19 @@ export const Header = ({ darkMode, setDarkMode }: Props) => {
...
+        <OffSwitch sx={{ height: 20, width: 40, ml: 1.25 }} />
```
benwerner01, line 149: If we stick with this UI for longer, might be nice to replace this with the MUI toggle component (which [can be customised](https://mui.com/material-ui/react-switch/#customization) to match the styling here, but comes with animations out of the box which would help make this feel less static). But we should probably address this down the line, may be worth adding a todo here though (if you agree).

benwerner01 (approved):
Thanks Ciaran, these changes look good to me and are working locally as expected 👍

Have left two thoughts, but neither block merging

packages/mock-block-dock/src/debug-view/header.tsx
CiaranMn: Done in https://github.com/blockprotocol/blockprotocol/pull/622/commits/3315ddd5dfe7948d739ca4a67ffa90e933e7d1da

packages/mock-block-dock/src/debug-view/header.tsx
CiaranMn: No MUI components are added in this view, so I'm assuming there is some styling / theming data that is also not loadd

benwerner01: Ah that makes sense, thanks for explaining

benwerner01 (approved):

PR #631, opened 2022-09-20
Spec: update entities and types by overwrite, not merge

CiaranMn:
In introducing BP `0.2` we specified that updates to entity properties and entity type schemas should be by _merge_, not overwrite, i.e. any properties not mentioned in the update already on the entity/schema would be retained.

This requires more being specified to work, e.g. to have a way of deleting properties, and so I am reverting it to 'by overwrite' for now until we fully specify merge updates.

---
Unrelated change I'm sneaking in here: get rid of an old section in RFCs on block types, since we have a GitHub discussion on it now.

---
  - https://app.asana.com/0/0/1202721046738726


nathggns:
Looking now

nathggns (approved):

PR #640, opened 2022-09-26
Allow multiple entity types as constraints on links

Alfred-Mountfield:
The Type System currently only allows one `EntityTypeReference` as a constraint on a link's destination. This PR wraps that within a `OneOf`, this means that Entity Types can say a `Foo` link points to `Bar` _or_ `Baz`.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1200211978612931/1202953520344825/f) _(internal)_


TimDiekmann (approved):
LGTM 🎉

PR #642, opened 2022-09-27
Fix several markup issues

kachkaev:
See PR comments for details.

I was also trying to fix hydration issues on the `/hub` page [as reported by Sentry recently](https://sentry.io/organizations/hashintel/issues/3476303370/?project=6603021&query=is%3Aunresolved) (internal). However, I was unable to reproduce this issue both locally and in production. I suggest merging existing fixes and monitoring Sentry logs again.

---
  - https://app.asana.com/0/0/1202923366972165


site/src/components/button.tsx
```diff
@@ -22,19 +22,19 @@ const loadingAnimation = (
...
           left: 8,
           animation: "lds-ellipsis2 0.6s infinite",
         },
-        "&:nth-child(3)": {
+        "&:nth-of-type(3)": {
           left: 32,
           animation: "lds-ellipsis2 0.6s infinite",
         },
-        "&:nth-child(4)": {
+        "&:nth-of-type(4)": {
```
kachkaev, line 37:
```
The pseudo class ":nth-child" is potentially unsafe when doing server-side rendering. Try changing it to ":nth-of-type". 
```

site/src/components/pages/blocks/block-form-styles.tsx
```diff
@@ -4,7 +4,9 @@ export const BlockFormContainer = styled(Paper)(
   sx({ flex: 1, width: "100%", p: 6 }),
 );
 
-export const BlockListContainer = styled(BlockFormContainer)<{
+export const BlockListContainer = styled(BlockFormContainer, {
+  shouldForwardProp: (propName) => propName !== "hasBlocks",
+})<{
```
kachkaev, line 9:
```
Warning: React does not recognize the `hasBlocks` prop on a DOM element.
If you intentionally want it to appear in the DOM as a custom attribute, spell it
as lowercase `hasblocks` instead. If you accidentally passed it from a parent
component, remove it from the DOM element.
```
https://mui.com/system/styled/#what-problems-does-it-solve

site/src/components/pages/hub/block-data-tabs.tsx
```diff
@@ -21,6 +21,9 @@ const DesktopTabs = styled(
       {children}
     </Tabs>
   ),
+  {
+    shouldForwardProp: (propName) => propName !== "modalOpen",
+  },
```
kachkaev, line 26:
```
Warning: React does not recognize the `modalOpen` prop on a DOM element.
If you intentionally want it to appear in the DOM as a custom attribute, spell it
as lowercase `modalopen` instead. If you accidentally passed it from a parent
component, remove it from the DOM element.
```

https://mui.com/system/styled/#what-problems-does-it-solve

site/src/components/pages/user/block-list-item.tsx
```diff
@@ -18,10 +18,10 @@ export const BlockListItem = ({ block }: BlockListItemProps) => {
       title={block.displayName!}
       description={block.description}
       extraContent={
-        <Box display="flex" gap={1.5}>
-          <Link href={`/@${block.author}`}>@{block.author}</Link>
+        <Box display="flex" gap={1.5} component="span">
+          @{block.author}
           <span>{`V${block.version}`}</span>
-          <span>{formatUpdatedAt(block.lastUpdated)}</span>
+          {formattedUpdatedAt ? <span>{formattedUpdatedAt}</span> : null}
```
kachkaev, line 24: Minor: avoids empty span

site/src/components/pages/user/block-list-item.tsx
```diff
@@ -18,10 +18,10 @@ export const BlockListItem = ({ block }: BlockListItemProps) => {
       title={block.displayName!}
       description={block.description}
       extraContent={
-        <Box display="flex" gap={1.5}>
-          <Link href={`/@${block.author}`}>@{block.author}</Link>
+        <Box display="flex" gap={1.5} component="span">
```
kachkaev, line 21: Each list item on the `/@username/blocks` page is a link (`<a>`):

Screenshot 2022-09-27 at 19 31 57

This means that we cannot use `<div>`, `<p>` or `<a>` components inside list items. I switched to `span` in all cases to fix hydration issues.

site/src/theme/components/navigation/mui-tabs-theme-options.ts
```diff
@@ -1,10 +1,6 @@
 import { Components } from "@mui/material";
 
 export const MuiTabsThemeOptions: Components["MuiTabs"] = {
-  defaultProps: {
-    disableRipple: true,
-    disableTouchRipple: true,
-  },
```
kachkaev, line 7:
```
Warning: React does not recognize the `disableRipple` prop on a DOM element.
If you intentionally want it to appear in the DOM as a custom attribute, spell it
as lowercase `disableripple` instead. If you accidentally passed it from a parent
component, remove it from the DOM element.
```

https://stackoverflow.com/q/66570397/1818285#comment122941778_66570397

site/src/components/pages/user/block-list-item.tsx
```diff
@@ -18,10 +18,10 @@ export const BlockListItem = ({ block }: BlockListItemProps) => {
...
+          @{block.author}
```
kachkaev, line 22: Fixes

```
Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.
```

Note that we no longer link to user inside a link to a block. I _guess_ it makes sense product-wise.

site/src/components/pages/user/block-list-item.tsx
yusufkinatas: I'm not sure if we should get rid of the link, or render it without an `<a>` tag to keep the same design. I feel like this is a design choice.

yusufkinatas (commented):
Looks OK, I only have one concern. Let me know what do you think.

site/src/components/pages/user/block-list-item.tsx
kachkaev: Not sure if a link inside a link is a good UX and whether it is accessible. @nonparibus WDYT?

kachkaev: For context, we’re talking about nested links in this list: https://blockprotocol.org/@hash/blocks

Screenshot 2022-09-28 at 14 05 47

kachkaev: I removed author from `ListViewCard`. This component is only used at `/@user/blocks` or `/blocks` (i.e. _my blocks_).

https://blockprotocol-jcempba1y.stage.hash.ai/@hash/blocks

Screenshot 2022-09-28 at 21 35 08

---

I’m still not convinced that having `<a>` inside `<a>` produces good UX / a11y. If it was, Next.js would not warn about it:

```
Warning: validateDOMNesting(...): <a> cannot appear as a descendant of <a>.
```

It is possible to fool Next with `span` + `onClick`, but this won’t improve UI a11y. Happy to change this opinion if we can find examples in high-quality web apps.

nonparibus: I defer on best implementation, but there are going to be times we want clickable islands within otherwise clickable elements that don't point to the same place.

nonparibus:
Functionally seems fine, but leaving code review/approval to others. Cheers

yusufkinatas (approved):

PR #650, opened 2022-10-04
Inline mock-block-dock styles

CiaranMn:
`mock-block-dock` uses some styles in its debug UI. 

It was loading some global styles from a stylesheet, but NextJS doesn't allow this, preventing its use in NextJS projects (e.g. [this blog post ](https://hash.dev/blog/build-blocks-in-any-language)– not the Hub here because MBD is loaded from an ESM module in an iFrame and is not part of the NextJS app).

This PR loads styles via `styled-jsx`, which NextJS seems fine with (`hash.dev` builds using the output of this branch's version of MBD).

## Known issues

Although the hash.dev builds with this version of MBD where it didn't before, the blog post crashes for some other reason to do with the calculation block <> MBD. This bug is already present in the published version of MBD so needn't block this PR merging, which at least removes the earlier error.

---
  - https://app.asana.com/0/0/1203004971448412


packages/mock-block-dock/src/mock-block-dock-ui.tsx
```diff
@@ -26,17 +25,89 @@ export const MockBlockDockUi: FunctionComponent<{ children: ReactNode }> = ({
...
+            align-content: baseline;
+            display: flex;
+            align-items: center;
+          }
+
+          :global(.mbd-debug-mode-toggle > svg) {
+            height: 20px;
+            width: 40px;
+            margin-left: 10px;
+          }
```
kachkaev, line 104: How about passing `className` to `OnSwitch` instead? Seems like `:global` is avoidable (unless I miss something)

CiaranMn: Yep fair enough - used `sx` instead, in https://github.com/blockprotocol/blockprotocol/pull/650/commits/bc95c5ba8569ec3252d38b488ad3fe256e12f062

PR #651, opened 2022-10-04
Simplify globs in package.json scripts

kachkaev:
Continues #649 (same idea)


package.json
```diff
@@ -17,17 +17,17 @@
   "scripts": {
     "fix": "npm-run-all --continue-on-error \"fix:*\"",
     "fix:dependency-version-consistency": "check-dependency-version-consistency --fix .",
-    "fix:eslint": "eslint \"**/*\" --fix",
-    "fix:markdownlint": "markdownlint --fix \"**/*\"",
-    "fix:prettier": "prettier --cache --write --ignore-unknown \"**/*\"",
+    "fix:eslint": "eslint --fix \"**\"",
+    "fix:markdownlint": "markdownlint --dot --fix \"**\"",
```
kachkaev, line 21: `--dot` includes files/folders with a dot (for example `.github`) — see [docs](https://www.npmjs.com/package/markdownlint-cli). We don't have markdowns in `.` folders but covering these files can be handy in the future.

package.json
```diff
@@ -17,17 +17,17 @@
   "scripts": {
     "fix": "npm-run-all --continue-on-error \"fix:*\"",
     "fix:dependency-version-consistency": "check-dependency-version-consistency --fix .",
-    "fix:eslint": "eslint \"**/*\" --fix",
-    "fix:markdownlint": "markdownlint --fix \"**/*\"",
-    "fix:prettier": "prettier --cache --write --ignore-unknown \"**/*\"",
+    "fix:eslint": "eslint --fix \"**\"",
```
CiaranMn, line 20: since we're simplifying, can't we use `eslint . --fix`, `prettier . --write`?

kachkaev: Nice idea! Done in https://github.com/blockprotocol/blockprotocol/pull/651/commits/2df1b0876bf088eb526b2ddf46cc48fe94161b39 (kept options before `.`)

PR #655, opened 2022-10-11
Configure ESLint `curly` rule

kachkaev:
https://eslint.org/docs/latest/rules/curly

Using `all` makes code rhythm more consistent which improves cognition.

- [Slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1664887623783489) (internal)
- [Asana task](https://app.asana.com/0/0/1203089470814745) (internal)
- https://github.com/blockprotocol/blockprotocol/pull/655

### DIFF
--- a/packages/@local/eslint-config/index.js
+++ b/packages/@local/eslint-config/index.js
@@ -27,6 +27,7 @@ module.exports = {
     "no-nested-ternary": "off",
     "no-restricted-syntax": "off",
     camelcase: "off",
+    curly: ["error", "all"],
     "no-unused-vars": [
       "error",
       {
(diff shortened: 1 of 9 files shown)


packages/@local/eslint-config/index.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
     "no-nested-ternary": "off",
     "no-restricted-syntax": "off",
     camelcase: "off",
+    curly: ["error", "all"],
```
kachkaev, line 30: 👀

kachkaev: `all` is default, but the rule works as `multi` if the argument is not passed. This could be to do with rule inheritance: AirBnB config must be setting `multi` for us.

kachkaev:
CI fails because of codecov: https://github.com/codecov/codecov-action/issues/598#issuecomment-1267261226

PR #656, opened 2022-10-13
Implement text hook service in mock-block-dock

CiaranMn:
# What this does

Implement a handler for requests for a 'text' hook under the Hook Service (formally introduced into the spec in https://github.com/blockprotocol/blockprotocol/pull/676), in `mock-block-dock` – i.e. the embedding application takes over rendering/editing of text at a specific node at the block's request.

This uses the Slate rich text editor, as it is one of the lighter (in bundle size) off-the-shelf editors.

Will need to see how much this slows down loading of blocks in the Block Hub and possibly adjust.

---
  - https://app.asana.com/0/0/1202668694585879


packages/mock-block-dock/dev/dev-app.tsx
```diff
@@ -136,4 +136,8 @@ const DevApp: FunctionComponent = () => {
   );
 };
 
-render(<DevApp />, node);
+if (node) {
+  createRoot(node).render(<DevApp />);
+} else {
+  throw new Error("Unable to find DOM element with id 'app'");
+}
```
CiaranMn, line 143: Leftover React 18 migration

packages/mock-block-dock/dev/test-react-block.tsx
```diff
@@ -2,6 +2,7 @@ import {
   BlockComponent,
   useGraphBlockService,
 } from "@blockprotocol/graph/react";
+import { useHook, useHookBlockService } from "@blockprotocol/hook/react";
```
CiaranMn, line 5: Added use of the hook service to the React test block - MBD dev environment only

packages/mock-block-dock/src/util.ts
```diff
@@ -39,6 +39,31 @@
...
+
+  let i;
+
+  for (i = 0; i < keys.length - 1; i++) {
+    // @ts-expect-error -- expected ‘No index signature with a parameter of type 'string' was found on type '{}'’
+    currentObj = currentObj[keys[i]!];
+  }
+
+  // @ts-expect-error -- expected ‘No index signature with a parameter of type 'string' was found on type '{}'’
+  currentObj[keys[i]!] = value;
```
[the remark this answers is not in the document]
CiaranMn: Now guarded against

packages/@blockprotocol/hook/src/hook-service.json
```diff
@@ -56,40 +56,7 @@
...
-              },
-              "hookId": {
-                "type": ["string", "null"],
-                "description": "The ID of the hook as provided in the response when first sending a hook message. This should be null in the first 'hook' message for a given node, and use the value provided by the embedding application in 'hookResponse' on subsequent messages."
-              }
-            }
-          }
-        }
-      ]
+      "errorCodes": ["INVALID_INPUT", "NOT_FOUND", "NOT_IMPLEMENTED"]
```
CiaranMn, line 59: changeset for this added in #676

packages/mock-block-dock/dev/dev-app.tsx
```diff
@@ -1,6 +1,6 @@
 import { Entity } from "@blockprotocol/graph";
 import { ChangeEvent, FunctionComponent, useState } from "react";
-import { render } from "react-dom";
+import { createRoot } from "react-dom/client";
```
nathggns, line 3: Why're we importing from `react-dom/client` and not `react-dom`?

packages/mock-block-dock/dev/dev-app.tsx
nathggns: Is this necessary for this PR or you just doing it while you're here? Have we verified there are no issues?

packages/mock-block-dock/src/hook-portals/text.tsx
```diff
@@ -0,0 +1,154 @@
...
+              children: [{ text: text ?? "" }],
+            },
+          ]
+        : text,
+    [text],
+  );
+
+  const textHasChanged = useDidTextChange(text);
+  if (textHasChanged) {
+    resetNodes(editor, nodesFromProps);
```
nathggns, line 108: This looks like a side effect – shouldn't it be in `useEffect`?

packages/mock-block-dock/src/hook-portals/text/use-did-text-prop-change.tsx
```diff
@@ -0,0 +1,19 @@
+import { useEffect, useRef } from "react";
+
+import { MaybePlainOrRichText } from "./shared";
+
+const generateComparableString = (text: MaybePlainOrRichText) =>
+  typeof text === "string" ? text : JSON.stringify(text);
+
+export const useDidTextChange = (text: MaybePlainOrRichText) => {
```
nathggns, line 8: I think the result of this is only used the side effect bit that should be an effect, right? It's probably better having this return a function that tells you if its changed – reading from refs during render is a no-no. 

image

https://beta.reactjs.org/apis/react/useRef#useref

packages/mock-block-dock/src/util.ts
```diff
@@ -39,6 +39,34 @@ const get = (
...
+};
+
+export const debounce = (func: Function, delayMs: number) => {
+  let timerId: NodeJS.Timeout;
+
+  return (...args: unknown[]) => {
+    clearTimeout(timerId);
+    timerId = setTimeout(() => func(...args), delayMs);
+  };
+};
```
nathggns, line 68:
```suggestion
export const debounce = <T extends (...args: any[]) => any>(func: T, delayMs: number) => {
  let timerId: NodeJS.Timeout;

  return (...args: Parameters<T>) => {
    clearTimeout(timerId);
    timerId = setTimeout(() => func(...args), delayMs);
  };
};
```

packages/mock-block-dock/dev/dev-app.tsx
CiaranMn: Just doing it while I'm here. There are no issues – this is not packaged with MBD, it's just the thing to test it out when developing it.

packages/mock-block-dock/dev/dev-app.tsx
CiaranMn: That's what the code example I found did, e.g. [here](https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html). Happy to change it if it matters

nathggns: Oh I misread something elsewhere. Ignore me!

packages/mock-block-dock/src/hook-portals/text/use-did-text-prop-change.tsx
CiaranMn: How about this? https://github.com/blockprotocol/blockprotocol/pull/656/commits/2896b994ecfbec767fe2f990d9c7e3f74e95005d

packages/mock-block-dock/src/hook-portals/text.tsx
```diff
@@ -90,6 +99,10 @@ export const TextHookView = ({
     );
   }
 
+  const previousTextRef = useRef<MaybePlainOrRichText>(
+    generateComparableString(text),
+  );
```
nathggns, line 104: You don't want to call `generateComparableString` on every render – but also this ref will always contain a string not `MaybePlainOrRichText`. 

```suggestion
  const previousTextRef = useRef<string | null>(null);
  if (previousTextRef.current === null) {
    previousTextRef.current = generateComparableString(text)
  }
```

packages/mock-block-dock/src/hook-portals/text.tsx
```diff
@@ -103,10 +116,16 @@ export const TextHookView = ({
...
-  }
+  useEffect(() => {
+    if (
+      generateComparableString(text) !==
+      generateComparableString(previousTextRef.current)
+    ) {
+      resetNodes(editor, nodesFromProps);
+
+      previousTextRef.current = generateComparableString(text);
+    }
```
nathggns, line 127: You're calling `generateComparableString` on `previousTextRef.current` even though it already stores a comparable string, and you're doing the work twice. 

```suggestion
   const comparableString = generateComparableString(text);
    if (comparableString !== previousTextRef.current) {
      previousTextRef.current = comparableString;
      resetNodes(editor, nodesFromProps);
    }
```

nathggns (changes requested):

nathggns (approved):

PR #681, opened 2022-10-18
Use `prepublishOnly` script in block templates

kachkaev:
#551 and #658 introduced package pre-processing for block templates. The original implementation relied on running a script before publishing all packages. This script removed ESLint from block templates to avoid referencing `@local/eslint-config` in them.

This PR refactors the script to call it from `prepublishOnly`, which is more robust. Related discussion: https://github.com/blockprotocol/blockprotocol/pull/551#discussion_r990869983

---
  - https://app.asana.com/0/0/1203132574912424


packages/block-template-custom-element/package.json
```diff
@@ -24,6 +24,7 @@
     "format": "prettier --write --ignore-unknown .",
     "lint:eslint": "eslint .",
     "lint:tsc": "tsc --noEmit",
+    "prepublishOnly": "PACKAGE_DIR=$(pwd) yarn workspace @local/package-chores exe scripts/cleanup-before-publishing.ts",
```
kachkaev, line 27: I assume that we won't need to publish from Windows. Running `yarn workspace @local/package-chores exe scripts/cleanup-before-publishing.ts` without `PACKAGE_DIR` gives us `packages/@local/package-chores` as `process.cwd`. This value is set by `yarn workspace`.

CiaranMn (approved):

PR #682, opened 2022-10-18
Configure Sentry Replay

kachkaev:
- [Asana task](https://app.asana.com/0/1200211978612931/1203176157000686/f) (internal)
- [`@sentry/replay` on npm](https://www.npmjs.com/package/@sentry/replay)
- [`@sentry/replay` on GitHub](https://github.com/getsentry/sentry-replay)
- https://github.com/getsentry/sentry-replay/issues/246

The original diff in this PR also included an update to `@sentry/nextjs` (from 7.13.0 to 7.15.0 / 7.16.0). I discovered an upstream incompatibility of newer versions with `@vercel/nft` so reverted the update after a pretty long investigation. Bug report: https://github.com/getsentry/sentry-javascript/issues/5998.

Replay is configured via `NEXT_PUBLIC_SENTRY_REPLAYS_SAMPLING_RATE` on Vercel. If unset or set to 0, `Replay` is disabled. Thus, if we encounter errors in production, we can quickly turn this feature off on Vercel and redeploy.


apps/site/sentry.server.config.ts
```diff
@@ -1,7 +1 @@
-// Context: https://docs.sentry.io/platforms/javascript/guides/nextjs/
-
-import * as Sentry from "@sentry/nextjs";
-
-import { sentryConfig } from "./sentry.client.config";
-
-Sentry.init(sentryConfig);
```
kachkaev, line 7: I guess we don't need `Sentry.init` because `sentry.client.config` has this side effect too.

yusufkinatas (approved):

apps/site/sentry.client.config.ts
```diff
@@ -1,14 +1,30 @@
 // Context: https://docs.sentry.io/platforms/javascript/guides/nextjs/
 
 import * as Sentry from "@sentry/nextjs";
-import type { NextjsOptions } from "@sentry/nextjs/types/utils/nextjsOptions";
+import { Replay } from "@sentry/replay";
 
 const dsn = process.env.NEXT_PUBLIC_SENTRY_DSN;
 
-export const sentryConfig: NextjsOptions = {
+const replaysSamplingRate = Number.parseInt(
```
nonparibus, line 8: @kachkaev Do we want to update this to `parseFloat`, too, as flagged in https://github.com/hashintel/hash/pull/1232#discussion_r1001141982?

apps/site/sentry.client.config.ts
```diff
@@ -1,14 +1,30 @@
...
+
+Sentry.init({
   dsn,
   enabled: !!dsn,
   environment: process.env.NEXT_PUBLIC_VERCEL_ENV ?? "unset",
-};
-
-Sentry.init(sentryConfig);
+  integrations:
+    replaysSamplingRate > 0 &&
```
nonparibus, line 18: @kachkaev Ditto comments from https://github.com/hashintel/hash/pull/1232#discussion_r1001140022 here

apps/site/sentry.client.config.ts
kachkaev: This is being addressed in https://github.com/blockprotocol/blockprotocol/pull/698

apps/site/sentry.client.config.ts
kachkaev: This is being addressed in https://github.com/blockprotocol/blockprotocol/pull/698

PR #697, opened 2022-10-21
Restore text blocks to hub

CiaranMn:
This PR updates the commit for blocks to account for updates to text blocks, and restores text blocks to the hub.

Text blocks on the hub now work as we introduced the hook service to mock-block-dock in #656 (mock-block-dock is what renders blocks in the hub, and acts as an embedding application)

How to test:
1. https://blockprotocol-hsvoiwjjr.stage.hash.ai/@hash/blocks/paragraph
2. https://blockprotocol-hsvoiwjjr.stage.hash.ai/@hash/blocks/callout
3. https://blockprotocol-hsvoiwjjr.stage.hash.ai/@hash/blocks/header 

### Known issues

The formatting toolbar can be hidden on the top line as it is outside the iFrame's boundary. We could add arbitrary padding to the top of the iFrame to give blocks a bit more breathing room to fix this in follow up.

---

  - https://app.asana.com/0/0/1202510130282422

PR #699, opened 2022-10-23
Fix bug in dock text comparison

CiaranMn:
This PR fixes a few bugs in mock-block-dock's hook service which only appear when updates to the property the block has called the hook service on are sent from outside the block, either very quickly or as plain text. This was noticed when trying out text blocks in the Hub and typing in the external editor.

Fixes as follows:
1. when a plain string is sent as an update to a property that is currently rich text, the string was not being properly compared to rich text in the block. This PR fixes comparing strings to rich text
2. when resetting the slate editor content in response to external updates, the `onChange` is triggered. This PR stops this sending an update back to the embedding application if the content is the same as the value from props – otherwise, if updates are coming fast, this can cause a loop when an external update comes in, and is then sent back out again while another external update is coming in.
3. fixes an issue in the Toolbar whereby Slate crashes if the user selects it and hits backspace

### Known issues

Sending external text updates very quickly can still cause some jankiness in the text updating – this will only happen in the Hub if someone types quickly into the JSON editor, and not when developing a block, so it's good enough for now. T[ask to fix this](https://app.asana.com/0/1201095311341924/1203225015178902/f) (internal).

### Next steps

Once this is merged and new mock-block-dock released, add back text blocks to the Hub – #697 

---
  - https://app.asana.com/0/0/1203225015178904

### DIFF
--- a/.changeset/chilly-parents-kneel.md
+++ b/.changeset/chilly-parents-kneel.md
@@ -1,0 +1,5 @@
+---
+"mock-block-dock": patch
+---
+
+fix bug in comparing text
--- a/packages/mock-block-dock/src/hook-portals/text.tsx
+++ b/packages/mock-block-dock/src/hook-portals/text.tsx
@@ -4,7 +4,6 @@ import {
   useEffect,
   useMemo,
   useRef,
-  useState,
 } from "react";
 import {
   BaseEditor,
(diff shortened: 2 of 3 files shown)


kachkaev (approved):
The diff makes sense. Happy for you to wait for @nathggns for a more in-depth review.

PR #676, opened 2022-10-17
Add 'Hook Service' spec and FAQ alongside small fixes in hook package

CiaranMn:
## What this PR does

Following on from https://github.com/blockprotocol/blockprotocol/pull/479, which introduced the `@blockprotocol/hook` package, this PR:
1. Adds the formal spec to the website
2. Adds a FAQ mentioning the hook service
3. Makes a couple of tweaks to the `@blockprotocol/hook` package to reflect the following:
  - we now allow blocks to request a hook is updated with any property is changed, not just the node, so that they do not have to send separate teardown and recreate messages when changing e.g. the path or entity
  - properly export the service JSON which had a file listed in package.json's `exports` that didn't exist
  - bug fix to make sure an id is re-used when it exists

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202510130282424
  - https://app.asana.com/0/0/1202783036300604

PR #707, opened 2022-10-25
Hook service: useHook and mock-block-dock improvements

CiaranMn:
This PR makes the following improvements following testing of the hook service in the Hub:

**mock-block-dock**

Make toolbar appear on hover, not omnipresent, and improve formatting + styling.

Kapture 2022-10-25 at 09 31 24

Also now doesn't show the toolbar in readonly mode

**@blockprotocol/hook**

Previously, `useHook` was sending a request to tear down the hook before requesting any updates. This meant that the embedding application would tear down the hook and then be asked to update it. 

MBD's implementation of the hook service was returning an error when the hook to update was not found, causing the Header block to crash in the Hub when switching DOM nodes.

This PR stops `useHook` from sending any tear down requests before updating a hook, since #676 made it so that any hook property can be changed in an update request (making it unnecessary to tear down the hook unless the component is being unmounted).

## Next steps

1. Update text blocks to use the new packages
2. Update and make #697 ready to review


packages/@blockprotocol/hook/src/react.ts
```diff
@@ -243,50 +240,45 @@ export const useHook = <T extends HTMLElement>(
 
       existingHookRef.current = hook;
 
-      teardownPromise
```
CiaranMn, line 246: enable hide whitespace to see what's changed other than removing this teardown

packages/@blockprotocol/hook/src/react.ts
```diff
@@ -243,50 +240,45 @@ export const useHook = <T extends HTMLElement>(
...
+                // eslint-disable-next-line no-console
+                console.error(response.errors);
+                throw new Error("Unknown error in hook");
               }
-            });
-        })
-        .catch((err) => {
-          catchError(() => {
-            throw err;
-          });
```
nathggns, line 289: Blocking: this needs to be here so React catches the errors thrown above – unsure why it's been removed.

packages/mock-block-dock/src/hook-portals/text/toolbar.tsx
```diff
@@ -1,24 +1,77 @@
+import { ReactElement, useEffect, useRef } from "react";
+import { createPortal } from "react-dom";
+import { Editor, Range } from "slate";
+import { useFocused, useSlate } from "slate-react";
+
 import { MarkButton } from "./mark-button";
 
+const Portal = ({ children }: { children: ReactElement }) => {
+  return createPortal(children, document.body);
+};
```
nathggns, line 10: Could you just call `createPortal` in `Toolbar`?

nathggns (dismissed):

packages/@blockprotocol/hook/src/react.ts
CiaranMn: Thank you, restored in https://github.com/blockprotocol/blockprotocol/pull/707/commits/8cbb862915ec99adce6ca230d209ec5a07b5a8af

packages/@blockprotocol/hook/src/react.ts
```diff
@@ -190,9 +190,6 @@ export const useHook = <T extends HTMLElement>(
 
     const existingHookId = existingHookRef.current?.id;
 
-    const teardownPromise =
-      existingHookRef.current?.teardown?.().catch() ?? Promise.resolve();
-
```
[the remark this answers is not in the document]
CiaranMn: Something like this? https://github.com/blockprotocol/blockprotocol/pull/707/commits/8cbb862915ec99adce6ca230d209ec5a07b5a8af

packages/mock-block-dock/src/hook-portals/text/toolbar.tsx
CiaranMn: Yes - https://github.com/blockprotocol/blockprotocol/pull/707/commits/8cbb862915ec99adce6ca230d209ec5a07b5a8af

packages/@blockprotocol/hook/src/react.ts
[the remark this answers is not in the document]
CiaranMn: @nathggns, I've asked @kachkaev to check that I've addressed your feedback and approve on that basis (if he agrees that I have) so we can close out the text tasks. 

If you think that what has been done here needs improvement (and this PR has been merged) I am happy to make further improvements in follow up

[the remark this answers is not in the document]
CiaranMn: @kachkaev, the concern here was that we've removed the teardown that cancelled any pending request for the hook, which might mean that a response to an earlier hook mesage was received out of order. I've therefore added in a `cancel` back in this place which cancels any in-flight requests and bails out of the `.then` part of the promise - see check for `(!controller.signal.aborted)`

kachkaev (approved):

PR #708, opened 2022-10-25
Switch Playwright tests to ESM

kachkaev:
Removes `apps/site/tests/package.json` (with `"type": "commonjs"`), unblocks dependency updates such as [`execa@6`](https://github.com/sindresorhus/execa/releases/tag/v6.0.0) (which is needed in https://github.com/blockprotocol/blockprotocol/pull/703 / [internal Asana task](https://app.asana.com/0/1203007126736610/1203225227202716/f))

[Playwright docs on ESM](https://playwright.dev/docs/test-typescript#typescript-with-esm)

CJS to ESM conversion can be expanded to the entire `@apps/site` package in #709.

Also fixes `import/no-unresolved` for ESM imports by adding `"eslint-import-resolver-typescript"`. This can improve linting in other ESM-only packages in the future (`block-scripts` etc.).

### DIFF
--- a/apps/site/tests/package.json
+++ b/apps/site/tests/package.json
@@ -1,3 +1,0 @@
-{
-  "type": "commonjs"
-}
--- a/apps/site/tests/integration/accountpage.test.ts
+++ b/apps/site/tests/integration/accountpage.test.ts
@@ -1,9 +1,7 @@
 import { expect, test } from "playwright-test-coverage";
 
-import blocksData from "../../blocks-data.json";
-import type { ExpandedBlockMetadata } from "../../src/lib/blocks";
-import { resetDb } from "../shared/fixtures";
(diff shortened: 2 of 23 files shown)


apps/site/tests/package.json
```diff
@@ -1,3 +0,0 @@
-{
-  "type": "commonjs"
-}
```
kachkaev, line 3: 🧹

CiaranMn (commented):
If `import/no-unresolved` is now fixed for ESM packages, can we remove it being disabled in:
- `@blockprotocol/graph/.eslintrc.cjs`
- `@blockprotocol/hook/.eslintrc.cjs`
- `@local/eslint-config/index.js`

CiaranMn (approved):

PR #709, opened 2022-10-25
Switch site workspace to strict ESM path resolution

kachkaev:
This PR follows #708. It sets `moduleResolution` to `Node16` in `apps/site/tsconfig.json` and removes `--es-module-specifier-resolution=node` from `apps/site/package.json` → `exe` script.

This turns `app/site` into a workspace that strictly follows ESModule spec. File extensions have been added to import paths and imports from folders had to be replaced with `folder-name/index.js`. Until this PR, import styles were different between `app/site/src` (no file extension) and `apps/site/tests` (file extension present).

When importing from `.ts{,x}`, `.js{,x}` extension is used as per [TypeScript docs](https://www.typescriptlang.org/docs/handbook/esm-node.html).

---

I had to do this in a few places:

```diff
-import x from "x";
+import _x from "x";
+const x = _x as unknown as typeof _x.default;
```

This must be to do with ESM-like typings for CJS packages. If typings for a CJS module contain `export = x`, all works. If it’s `export default x` instead, the above hack is necessary. Local package patching does not work if there are more exports in addition to the default one. Upstream discussion: https://github.com/ajv-validator/ajv/issues/2132. **Update:** there is now a fix for this issue, albeit not merged to the project's `main` or as yet officially released: https://github.com/ajv-validator/ajv/issues/2132#issuecomment-1845811081


apps/site/scripts/prepare-blocks.ts
```diff
@@ -30,8 +30,10 @@ import {
   expandBlockMetadata,
   ExpandedBlockMetadata,
   StoredBlockInfo,
-} from "../src/lib/blocks";
-import { FRONTEND_URL } from "../src/lib/config";
+} from "../src/lib/blocks.js";
+import { FRONTEND_URL } from "../src/lib/config.js";
+
+const Ajv = _Ajv as unknown as typeof _Ajv.default;
```
kachkaev, line 36: Here and below: see https://github.com/ajv-validator/ajv/issues/2132

I had to do this in a few places:

```diff
-import x from "x";
+import _x from "x";
+const x = _x as unknown as typeof _x.default;
```

This must be to do with ESM-like typings for CJS packages. If a CJS module says `export = x`, all works. If it says `export default x`, the above hack is necessary. Package patching does not work if there are more exports in addition to the default one.

apps/site/src/components/navbar/account-dropdown.tsx
```diff
@@ -2,16 +2,17 @@ import {
   Box,
   Divider,
   ListItemButton,
+  // eslint-disable-next-line no-restricted-imports
```
kachkaev, line 5: Result of changing ESLint rules. Importing Popover was restricted, but only from `"@mui/material/Popover`

patches/next-mdx-remote+4.0.3.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/node_modules/next-mdx-remote/index.d.ts b/node_modules/next-mdx-remote/index.d.ts
+index 9247c2a..09e9c9b 100644
+--- a/node_modules/next-mdx-remote/index.d.ts
++++ b/node_modules/next-mdx-remote/index.d.ts
+@@ -1 +1 @@
+-export * from './dist/index'
++export * from './dist/index.js'
```
kachkaev, line 7: Here and in a couple of other patches: I had to add extensions for TS imports to resolve

packages/@local/eslint-config/index.js
```diff
@@ -234,7 +234,13 @@ module.exports = {
       },
     },
     {
-      files: ["*.config.{c,m,}{j,t}s", "*.d.ts", "*rc.{c,m,}js"],
+      files: [
+        "*.config.{c,m,}{j,t}s",
+        "*.d.ts",
+        "*rc.{c,m,}js",
+        "*.test.{j,t}s{,x}",
+        "tests/**",
```
kachkaev, line 242: Compensates removal of `overrides` from `apps/site/.eslintrc.cjs`

apps/site/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "extends": "@local/tsconfig/next.json",
   "include": ["next-env.d.ts", "theme-override.d.ts", "**/*.ts", "**/*.tsx"],
   "compilerOptions": {
+    "moduleResolution": "Node16",
```
kachkaev, line 5: ### 💥

apps/site/.eslintrc.cjs
```diff
@@ -30,69 +30,39 @@ module.exports = {
...
           },
           {
             name: "@mui/material",
-            importNames: ["Button", "TextField", "Alert"],
+            importNames: ["Alert", "Button", "Popover", "TextField"],
             message:
               "Please use the custom wrapper component in src/component instead.",
           },
-          {
-            name: "@mui/material/Button",
```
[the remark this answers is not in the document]
CiaranMn: Can we keep these specific ones still, banning `@mui/material/Button*` etc? 

If someone imports `@mui/material/Button.js` and doesn't notice the linting error in their editor (or the editor isn't configured properly), they are going to get a linting error in CI which tells them to use `@mui/material`, and once they fix that get _another_ one that tells them to use the custom component, which feels like it could waste time.

With the current change we are hiding an error behind another error, when we have the ability to surface the correct fix immediately.

apps/site/.eslintrc.cjs
```diff
@@ -30,69 +30,39 @@ module.exports = {
...
           {
-            name: "next/link",
+            name: "next/link.js",
             message:
               "Please use the custom src/components/Link component instead to ensure Next.js and MUI compatibility.",
           },
           {
             name: "@mui/material",
-            importNames: ["Button", "TextField", "Alert"],
+            importNames: ["Alert", "Button", "Popover", "TextField"],
```
CiaranMn, line 46: There is no `src/components/popover` 
```suggestion
            importNames: ["Alert", "Button", "TextField"],
```

apps/site/src/components/navbar/account-dropdown.tsx
CiaranMn, line 5: This rule should not exist, there is no custom `Popover` component to use 
```suggestion
```

nonparibus:
No longer planned.

PR #713, opened 2022-10-27
Update publishing docs and add pull request template 

CiaranMn:

This PR updates the publishing docs with the new process, since publishing is now available via CLI rather than making a pull request to the repo.

As a consequence we can now add a PR template, so I've added one. We previously couldn't because a change template would have been overkill/confusing for people adding blocks via PR.

- It makes one.

## ❓ How to test this?

1.  Visit `/docs/developing-blocks#updating-your-block` in the preview deployment
2. We'll have to check the template once it's merged


apps/site/src/_pages/docs/1_developing-blocks.mdx
```diff
@@ -190,48 +190,33 @@ You now have a block package that you can provide to apps to use, by **publishin
...
+Once you've built a block, you can add it to the [Hub](/hub), so that your block will have an instant online demo playground,
+and will be searchable via our block API.
+
+To publish a block on the Hub take the following steps.
+
+### In your browser
 
-### My block source is in a subfolder
+- Sign up for an account on this site if you haven’t already – click 'sign up' in the top right
+- [Create an API key](https://blockprotocol.org/settings/api-keys and copy it
```
kachkaev, line 201:
```suggestion
- [Create an API key](https://blockprotocol.org/settings/api-keys) and copy it
```

```suggestion
- [Create an API key](/settings/api-keys) and copy it
```

The second option aligns the link with `[Hub](/hub)` above. You may want to use absolute links everywhere instead.

apps/site/src/_pages/docs/1_developing-blocks.mdx
```diff
@@ -190,48 +190,33 @@ You now have a block package that you can provide to apps to use, by **publishin
...
 
-- If it’s in a yarn workspace, please provide `workspace`
-- If it’s not in a yarn workspace, please provide its `folder` path
+### In your terminal
 
-### My block requires building from its source
+- Run `npm run build` or `yarn build` to create a production build of your block (it will appear in the `dist/` folder)
```
kachkaev, line 205:
```suggestion
- Run `npm run build` or `yarn build` to create a production build of your block (it will appear in the `dist` folder)
```

(to align with _Build_ section)

apps/site/src/_pages/docs/1_developing-blocks.mdx
```diff
@@ -190,48 +190,33 @@ You now have a block package that you can provide to apps to use, by **publishin
...
+`npm run build && npx blockprotocol@latest publish` or `yarn build && npx blockprotocol@latest publish`
 
-We assume that running `yarn install && yarn build` will build your block (after switching to the relevant folder/workspace, if provided).
-The specified `distDir` will be relative to the workspace directory or folder, if provided.
+### Changing the Hub preview
 
-</InfoCard>
+- Take a screenshot of your block in action and replace `public/block-preview.png`
+- update the `description` in package.json
+- add an icon to `public/` and update `blockprotocol.icon` in `package.json`
```
kachkaev, line 220:
```suggestion
- add an icon to the `public` folder and update `blockprotocol.icon` in `package.json`
```

kachkaev (changes requested):
Looks good, but there is a missing `)` next to one link.

A bit unrelated: do we still need the `/workshop` page?

apps/site/src/_pages/docs/1_developing-blocks.mdx
CiaranMn: I've gone for absolute links everywhere in https://github.com/blockprotocol/blockprotocol/pull/713/commits/d22ea1720e6eeacd4b164b3a5df46e233d478c51

I have in my head that this is better because it allows the links to work wherever the markdown is being read. It has no effect in development since links starting with `https://blockprotocol.org` are [treated as internal](https://github.com/blockprotocol/blockprotocol/blob/4bf770dd3c2f854f2109f59b0043137475c63d91/apps/site/src/components/base-link.tsx#L18) and will be routed accordingly.

CiaranMn:
> A bit unrelated: do we still need the /workshop page?

@kachkaev no, but we should repurpose the material as a tutorial since it's basically that.

I've created[ a task to do this](https://app.asana.com/0/1200211978612931/1203254239427249/f) (internal)

apps/site/src/_pages/docs/1_developing-blocks.mdx
kachkaev: Thanks, makes total sense. If we need to fix URLs in preview deployments, I guess we can do that in a separate PR by tweaking markdown AST.

kachkaev (approved):

PR #428, opened 2022-07-11
RFC - Type extension and duplication

thehabbos007:
This RFC proposes a way for the type system (given by #352) to extend (inherit) types and duplicate (fork) types.

---
[Rendered view](https://github.com/blockprotocol/blockprotocol/blob/3ac880de5e5905dd51449bc84056280be487bfeb/rfcs/text/0428-extending-and-duplicating-types.md)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202517364041306

PR #714, opened 2022-10-27
Add foundation for Inheritance in Entity Types

Alfred-Mountfield:
This adds an `AllOf` field to the `EntityType` meta-schema. This will likely be used in the future for allowing extending of types (see #428).

This is being added to the experimental Type System package for now, ahead of RFC approval, in order to continue to experiment and iterate to find the appropriate solutions.

[**Asana Task**](https://app.asana.com/0/0/1203250001255245)  _(internal)_


crates/type-system/src/ontology/entity_type/mod.rs
```diff
@@ -75,6 +78,11 @@ impl EntityType {
         self.description.as_deref()
     }
 
+    #[must_use]
+    pub const fn inherits_from(&self) -> &AllOf<EntityTypeReference> {
```
TimDiekmann, line 82: Do we want to return a slice instead?

crates/type-system/src/ontology/shared/all_of/repr.rs
```diff
@@ -0,0 +1,94 @@
...
+
+use crate::{repr, EntityTypeReference, ParseAllOfError};
+
+#[cfg_attr(target_arch = "wasm32", derive(Tsify))]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct AllOf<T> {
+    #[serde(
+        rename = "allOf",
+        default = "Vec::new",
```
TimDiekmann, line 13:
```suggestion
        default,
```

Alfred-Mountfield: This is needed because otherwise it expects `T` to implement default... I have no idea why

packages/@blockprotocol/type-system-web/dist/index.js
```diff
@@ -119,38 +119,38 @@ function takeObject(idx) {
     return ret;
 }
 /**
-* @param {any} link_type_obj
+* @param {any} data_type_obj
```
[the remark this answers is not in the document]
Alfred-Mountfield: It's non-deterministic 🙃

crates/type-system/src/ontology/entity_type/mod.rs
Alfred-Mountfield: I wondered that as well, but I think there's _some_ value in having the `AllOf` rather than `&[EntityTypeReference]` because we may add more information/implementations to `AllOf`. When we look at inheritance we're going to want some logic about composing the types involved. 

_**Probably**_ it won't happen in the `AllOf` object, but it's hard to tell, so might as well keep the object. I had previously actually just had `inherits_from: Vec<EntityTypeReference>` but we're doing something else for `OneOf` so I'd prefer to be consistent for now.

crates/type-system/src/ontology/shared/all_of/repr.rs
TimDiekmann: Annoying, didn't expected that. 😞

crates/type-system/src/ontology/entity_type/mod.rs
TimDiekmann: fair 🙂

crates/type-system/src/ontology/shared/all_of/repr.rs
Alfred-Mountfield: Me neither, I lose about 10 minutes trying to figure out why it wanted `Default` when I just want an empty vec 🙃 

It's not like we're even saying `default = "Vec<T>::new` so this is just arcane macro blackbox funtimes

TimDiekmann: I thought this was because of https://github.com/rust-lang/rust/issues/26925 but the implementation is
```rust
impl<T> const Default for Vec<T> {
    /// Creates an empty `Vec<T>`.
    fn default() -> Vec<T> {
        Vec::new()
    }
}
```

TimDiekmann (approved):

PR #715, opened 2022-10-27
Reduce timeout if workflows using `actions/upload-artifact@v3`

kachkaev:
Workaround for https://github.com/actions/upload-artifact/issues/358

Reducing job timeout saves ≈300 Workflow minutes each time the bug occurs. This helps unblock pending jobs faster.


.github/actions/warm-up-repo/action.yml
```diff
@@ -9,6 +9,7 @@ inputs:
 
 runs:
   using: composite
+
```
kachkaev, line 12: Just added some whitespace to sync formatting between workflows and actions.

CiaranMn (approved):

PR #716, opened 2022-10-27
Use entity type references in `links` object rather than link type references 

Alfred-Mountfield:
This replaces references of `LinkType`s in `links` in Entity Types with references of entity types instead. This is part of an experimental change whereby links _are_ a special kind of entity.

This is being added to the experimental Type System package for now, ahead of RFC approval, in order to continue to experiment and iterate to find the appropriate solutions.

[**Asana Task**](https://app.asana.com/0/1200211978612931/1203250001255247/f)  _(internal)_


crates/type-system/src/ontology/entity_type/mod.rs
```diff
@@ -183,6 +189,13 @@ impl EntityTypeReference {
     }
 }
 
+impl From<&VersionedUri> for &EntityTypeReference {
+    fn from(uri: &VersionedUri) -> Self {
+        // SAFETY: Self is `repr(transparent)`
+        unsafe { &*(uri as *const VersionedUri).cast::<EntityTypeReference>() }
+    }
+}
+
```
TimDiekmann, line 198: With this only being used once (in `link_mapping`) do you think it's worth adding `unsafe` to the crate?

TimDiekmann (approved):
Looks good, thanks!

I have one comment, feel free to merge if you don't agree 🙂

crates/type-system/src/ontology/entity_type/mod.rs
Alfred-Mountfield: I think so, we hadn't disabled unsafe anyway, just didn't have any. And we both agree that really there should be a safe form of this operation if it's `repr(transparent)`

PR #720, opened 2022-10-28
Show updates to `devDependencies` in `block-template-*` CHANGELOGs

kachkaev:

The goal of this PR is to avoid empty changelog entries in `block-template-*` packages ([example](https://github.com/blockprotocol/blockprotocol/blob/5b4ec346ed887e1a68ef5134c2050e214a842f94/packages/block-template-react/CHANGELOG.md)).

````markdown
# block-template-react

## 0.1.18

## 0.1.17

## 0.1.16

## 0.1.15
````

This is done by checking `devDependencies` in addition to `dependencies` and `peerDependencies` when generating changelog lines. The exception is made only for `block-template-*` because other packages are agnostic of updates to `devDependencies`.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1203207083851550/f) _(internal)_

## ⚠️ Known issues

The rule for including `devDependencies` is hardcoded. We might want to move the rule to `.changesets/config.json` or propose a change upstream. Alternatively, we can stop publishing block templates as individual npm packages.

## 🐾 Next steps

Check https://github.com/blockprotocol/blockprotocol/pull/702, create a new PR if the result is not as expected.

## 🛡 What tests cover this?

- None 🙈

## ❓ How to test this?

Merge and check Check https://github.com/blockprotocol/blockprotocol/pull/702

### CODE REVIEW

File: patches/@changesets+apply-release-plan+6.1.1.patch [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+       depVersionRange: versionRange,
+-      depType: dependencyVersionRange ? "dependencies" : "peerDependencies"
++      depType: dependencyVersionRange ? "dependencies" : devDependencyVersionRange ? "devDependencies" : "peerDependencies"
+     }, {
+       minReleaseType: updateInternalDependencies,
+       onlyUpdatePeerDependentsWhenOutOfRange
+diff --git a/node_modules/@changesets/apply-release-plan/src/get-changelog-entry.ts b/node_modules/@changesets/apply-release-plan/src/get-changelog-entry.ts
+index 1f07982..cb3c43f 100644
+--- a/node_modules/@changesets/apply-release-plan/src/get-changelog-entry.ts
++++ b/node_modules/@changesets/apply-release-plan/src/get-changelog-entry.ts
```
kachkaev, line 82: ↑ autogenerated using `yarn build` in [Changesets monorepo](https://github.com/changesets/changesets)
↓ written by hand

CiaranMn: Should we add a comment into the patch which explains this, in case someone else other than you has to recreate it when updating `changesets`?

CiaranMn (approved):

PR #724, opened 2022-10-28
Improve TS Types of the Type System Package

Alfred-Mountfield:

We had a number of spots in the Type System packages where non-ideal types were being used, strings instead of `VersionedUri` for example, and a few other things.

This PR generally improves the usabilities of the interfaces. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1201095311341924/1202981263631815/f) _(internal)_

## 🔍 What does this change?

- `DataTypeReference`, `PropertyTypeReference`, and `EntityTypeReference` now has `VersionedUri` instead of `string`
- The `links` object on `EntityType` has been much improved
- A lot of fields have been marked as optional when they were previously required
- `$id` in schemas are correctly typed as `VersionedUri`s
- `BaseUri` is used in more places where it should be
- Some other improvements, see the generated binding file

- The TypeScript bindings act as the doc at the moment

## 🛡 What tests cover this?

- The Rust unit tests
- Node-variant of the Type System package integration tests

## ❓ How to test this?

1.  Run `cargo make test` from inside the crate
1.  Run `yarn test` from inside the node variant of the type-system package.

### DIFF
--- a/crates/type-system/src/ontology/data_type/repr.rs
+++ b/crates/type-system/src/ontology/data_type/repr.rs
@@ -22,6 +22,7 @@ enum DataTypeTag {
 pub struct DataType {
     #[cfg_attr(target_arch = "wasm32", tsify(type = "'dataType'"))]
     kind: DataTypeTag,
+    #[cfg_attr(target_arch = "wasm32", tsify(type = "VersionedUri"))]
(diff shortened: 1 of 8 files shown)


TimDiekmann (approved):
I like that, thanks! 🎉

PR #726, opened 2022-10-28
mock-block-dock: fix stale hook callback and dom warnings

CiaranMn:

Fixes an oversight in MBD's implementation of the hook service, where the hook callback was not updated when the state it included changes, meaning that the state was fixed at the time the service was constructed. This PR re-registers the callback when it is reconstructed, as we already do in an effect for `graphServiceCallbacks`. We should probably just do this inside the `useXService` hooks to stop whoever is using them having to remember to do so, and I've created a [task](https://app.asana.com/0/1201095311341924/1203257689029614/f) to do so (internal).

As a drive-by I've also fixed a couple of DOM warnings that were in the console whenever someone uses MBD, and which make it hard to see actual errors from the block. These are only visible when MBD is compiled, not when running `yarn dev` in the package itself:
1. Pass a string instead of boolean for `styled jsx` tags
2. Fix an issue with `react-json-view` where they haven't updated for React changes and were passing an `inputRef` instead of `ref` to a component


packages/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -151,71 +154,82 @@ export const MockBlockDock: FunctionComponent<MockBlockDockProps> = ({
     readonly,
   });
 
-  useHookEmbedderService(wrapperRef, {
-    callbacks: {
-      hook: async ({ data }) => {
-        if (!data) {
+  const hookCallback = useCallback<EmbedderHookMessageCallbacks["hook"]>(
+    async ({ data }) => {
```
CiaranMn, line 158: no changes to the function itself - hide whitespace

kachkaev (approved):

kachkaev (approved):

PR #761, opened 2022-11-14
Add Semgrep GitHub Action

nonparibus:

Adds a CI job for Semgrep scanning the repo on proposed merges into `main`, similar to that in use in [`@hashintel/hash`](https://github.com/hashintel/hash).

## Next steps

Once this has been reviewed and we're happy with it, we should update the [corresponding file](https://github.com/hashintel/hash/blob/main/.github/workflows/semgrep.yml) in `@hashintel/hash` to reflect our new understanding of best-practice here (minor modifications only expected).

---
  - https://app.asana.com/0/0/1203364310192415

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,22 @@
+name: Semgrep
+on:
+  pull_request: {}
+  push:
+    branches:
+      - main
+    paths:
+      - .github/workflows/semgrep.yml
+  schedule:
+    - cron: '0 0 * * 0'
+jobs:
+  semgrep:
+    name: Scan
+    runs-on: ubuntu-20.04
+    env:
+      SEMGREP_APP_TOKEN: ${{ secrets.SEMGREP_APP_TOKEN }}
+    container:
+      image: returntocorp/semgrep
+    if: (github.actor != 'dependabot[bot]')
+    steps:
+      - uses: actions/checkout@v3
+      - run: semgrep ci
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,5 +1,6 @@
 name: Semgrep
 on:
+  # Scan changed files in PRs (diff-aware scanning):
   pull_request: {}
   push:
     branches:
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -9,25 +9,25 @@ on:
       - main
   # Schedule the CI job (this method uses cron syntax):
   schedule:
-    - cron: '30 0 1,15 * *' # scheduled for 00:30 UTC on both the 1st and 15th of the month
-    
+    - cron: "30 0 1,15 * *" # scheduled for 00:30 UTC on both the 1st and 15th of the month
+
 jobs:
   semgrep:
     name: Scan
-    
+
     # Change this in the event of future self-hosting of Action runner:
     runs-on: ubuntu-latest
-    
+
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,22 @@
+name: Semgrep
+on:
+  pull_request: {}
+  push:
+    branches:
+      - main
+    paths:
+      - .github/workflows/semgrep.yml
```
nonparibus, line 8:
```suggestion
```

Doesn't seem to feature in the [Semgrep sample CI config](https://semgrep.dev/docs/semgrep-ci/sample-ci-configs/#github-actions) in their docs, even though present in their auto-generated code from the app.

.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,22 @@
...
+  schedule:
+    - cron: '0 0 * * 0'
```
nonparibus, line 10:
```suggestion
    - cron: '30 0 1,15 * *' # scheduled for 00:30 UTC on both the 1st and 15th of the month
```

Have Semgrep scans run periodically as well as reactively.

.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,22 @@
+name: Semgrep
+on:
+  pull_request: {}
```
nonparibus, line 3:
```suggestion
  # Scan changed files in PRs (diff-aware scanning):
  pull_request: {}
```

.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,22 @@
...
+  push:
```
nonparibus, line 4:
```suggestion
  # Scan mainline branches and report all findings:
  push:
```

.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,22 @@
...
+  schedule:
+    - cron: '0 0 * * 0'
+jobs:
+  semgrep:
+    name: Scan
+    runs-on: ubuntu-20.04
+    env:
+      SEMGREP_APP_TOKEN: ${{ secrets.SEMGREP_APP_TOKEN }}
+    container:
+      image: returntocorp/semgrep
```
nonparibus, line 18:
```suggestion
      # A Docker image with Semgrep installed:
      image: returntocorp/semgrep
```

.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,22 @@
...
+    branches:
+      - main
+    paths:
+      - .github/workflows/semgrep.yml
+  schedule:
+    - cron: '0 0 * * 0'
+jobs:
+  semgrep:
+    name: Scan
+    runs-on: ubuntu-20.04
```
nonparibus, line 14:
```suggestion
    # Change this in the event of future self-hosting of Action runner:
    runs-on: ubuntu-latest
```

kachkaev (commented):
Not sure I see benefits in most of the added comments. Who are they for?

I also wonder if we should stick with a a fixed Ubuntu version, e.g. `ubuntu-22.04`. It's useful to control the timing of such updates.

nonparibus:
> Not sure I see benefits in most of the added comments. Who are they for?
> 
> I also wonder if we should stick with a a fixed Ubuntu version, e.g. `ubuntu-22.04`. It's useful to control the timing of such updates.

They and the use of `latest` reflects Semgrep's current docs/recommended approach.

PR #754, opened 2022-11-10
Update Sentry client and Replay

kachkaev:
## 🌟 What is the purpose of this PR?

Extracted from #692, was needed to investigate the cause of Playwright failures. After splitting that PR, I can conclude that Playwright tests fail because of Next 13 and not updates to Sentry. We can merge this PR to start using the latest Sentry libs.

This PR is a part of attempts to fix [Asana task](https://app.asana.com/0/1203150072758509/1203311682285013/f) (internal)

PR #768, opened 2022-11-17
Revert "Update Sentry client and Replay"

CiaranMn:
Reverts blockprotocol/blockprotocol#754

This update appears to be the cause of the Block Hub crashing when session storage access is prevented, e.g. because third-party cookie access in Chrome is disabled, which also prevents storage access in iFrames. 

To recreate: visit https://blockprotocol.org/@hash/blocks/code in Chrome incognito mode (assuming you have third-party access in incognito mode enabled – if not, enable it).

We can re-revert with a new version once https://github.com/getsentry/sentry-replay/issues/316 is fixed


kachkaev (approved):

apps/site/sentry.client.config.ts
```diff
@@ -9,13 +9,12 @@ Sentry.init({
   dsn,
   enabled: !!dsn,
   environment: process.env.NEXT_PUBLIC_VERCEL_ENV ?? "unset",
-  integrations: process.env.NEXT_PUBLIC_SENTRY_REPLAY_SESSION_SAMPLING_RATE
+  integrations: process.env.NEXT_PUBLIC_SENTRY_REPLAYS_SAMPLING_RATE
```
kachkaev, line 12: When we revert this revert, we can change

`NEXT_PUBLIC_SENTRY_REPLAY_SESSION_SAMPLING_RATE`
to
`NEXT_PUBLIC_SENTRY_REPLAY_SESSION_SAMPLE_RATE`

This will better align with `sessionSampleRate`. I've already used this name in https://github.com/hashintel/hash/pull/1377

PR #779, opened 2022-11-21
Check console messages in Playwright

kachkaev:

This PR adds tracking of console messages in Playwright tests. It is inspired by https://github.com/hashintel/hash/pull/436 and https://github.com/microsoft/playwright/discussions/11690#discussioncomment-2060397. With the new check for console messages, it is less likely that we’ll miss new warnings or errors. They can pop up when updating dependencies, adding new features or refactoring code.

## 🔗 Related links

- [Asana task](link) _(internal)_
- https://github.com/hashintel/hash/pull/436

## 🔍 What does this change?

- Tests import Playwright resoutces from a different location:

   ```diff
   - import type { Page } from "@playwright/test";
   - import { expect, test } from "playwright-test-coverage";
   + import { type Page, expect, test } from "../shared/wrapped-playwright.js";
   ```

- `no-restricted-imports` ESLint rule is updated.

- New helper function `tolerateCustomConsoleMessages` can be used to define additional messages per test. It is useful when testing edge cases and allows us to keep the ‘global’ list of tolerable messages small.

## ⚠️ Known issues

- There are quite a few console messages to triage. Some are dev-only warnings while others point to real issues in the app. These issues are small but are at least worth knowing about.

- Not new, but worth noting: when importing types from `wrapped-playwright`, [`type` prefix is needed](https://www.typescriptlang.org/docs/handbook/2/modules.html#typescript-specific-es-module-syntax) due to ESM resolution / transpiling:

   ```ts
   import { expect, type Page, test } from "../shared/wrapped-playwright.js";
   ```
   
   Node fails otherwise:

   ```ts
   import { expect, Page, test } from "../shared/wrapped-playwright.js";
   ```
   
   ```
   SyntaxError: The requested module '../shared/wrapped-playwright.js' does not provide an export named 'Page'
   ```

## 🐾 Next steps

- [Triage detected console messages in Block Protocol tests](https://app.asana.com/0/1203312852763953/1203414492513784/f) (internal Asana task)

## 🛡 What tests cover this?

- CI
- Local integration tests with `yarn dev` (which contains additional warnings)


apps/site/tests/integration/docs.test.ts
```diff
@@ -159,7 +158,7 @@ test("Docs page should contain key elements and interactions should work", async
   ).toHaveAttribute("href", "/docs/developing-blocks");
 });
 
-test("invalid docs page should redirect to 4040", async ({ page }) => {
+test("invalid docs page should redirect to 404", async ({ page }) => {
```
kachkaev, line 161: Unrelated typo

apps/site/tests/integration/block-publish-npm.test.ts
```diff
@@ -23,6 +26,11 @@ test.beforeEach(async ({ page }) => {
 
   await page.goto("/");
 
+  // @todo triage: https://app.asana.com/0/1203312852763953/1203414492513784/f
+  tolerateCustomConsoleMessages([
+    /Failed to load resource: the server responded with a status of 500 \(Internal Server Error\)/,
+  ]);
```
kachkaev, line 32: Not sure why we see 500 in this test, but I leave investigation out of scope. This PR does not touch app code.

CiaranMn (approved):

PR #794, opened 2022-11-28
Cleanup `blockprotocol` and `block-scripts`

kachkaev:

This small maintenance PR improves the structure and ESLint config for two CLI packages.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203301853897838/1203400309200172/f) _(internal)_

## 🔍 What does this change?

See PR comments


packages/block-scripts/.eslintrc.cjs
```diff
@@ -3,6 +3,8 @@ module.exports = {
   ...require("@local/eslint-config/generate-workspace-config.cjs")(__dirname),
   rules: {
     "no-console": "off",
-    "import/extensions": ["error", "always"],
+    "import/extensions": ["error", "ignorePackages"],
```
kachkaev, line 6: Allows us to remove `// eslint-disable-next-line` before `import { fileFromPath } from "formdata-node/file-from-path"`. This still works because `formdata-node` is CJS and is therefore subject to a ‘classic’ file resolution mechanism.

packages/block-scripts/.eslintrc.cjs
```diff
@@ -3,6 +3,8 @@ module.exports = {
...
+    "import/named": "off", // Need to investigate this (Maximum call stack size exceeded Occurred while linting ... Rule: "import/named")
+    "import/no-cycle": "off", // Need to investigate this (Maximum call stack size exceeded Occurred while linting ... Rule: "import/no-cycle")
```
kachkaev, line 8: These two rules were hanging VSCode, so I disabled them for now. This may be to do with some recent update in `eslint-plugin-import`. We can try the rules again in a few weeks and see what happens.

packages/blockprotocol/cli.js
```diff
@@ -1,20 +1,10 @@
...
-import { printSpacer } from "./cli/print-spacer.js";
-
-/**
- * @param {string} errorMessage
- */
-const printErrorMessage = (errorMessage) => {
-  console.log(chalk.red(`${errorMessage}. Please check usage.`));
-  printSpacer();
-};
+import { printErrorMessage } from "./cli/print-error-message.js";
```
kachkaev, line 7: This fn was extracted to `cli/print-error-message.js` to be able to move `print-spacer` into `cli/shared` for clarity.

CiaranMn (approved):

PR #799, opened 2022-11-30
Reconfigure Renovate (cover all dependencies)

kachkaev:

This PR extends the scope of Renovate, which now covers all dependencies in the repo. As before, new PRs are not created automatically and we need to approve them via Dependency dashboard (#788) first.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203312852763953/1203464390652052/f) _(internal)_
- https://github.com/hashintel/hash/pull/1535

## 🔍 What does this change?

I’ve removed some package rules as they are no longer needed. We extend from a preset named [`config:base`](https://docs.renovatebot.com/presets-config/#configbase) which in turn includes [monorepo presets](https://docs.renovatebot.com/presets-monorepo/). This means that dependencies within well-known monorepos are grouped together without us having to configure anything. For example, `next` is upgraded with `@next/bundle-analyzer` in a single PR. I’ve kept a few rules that span across multiple monorepos (e.g. everything related to ESLint or Jest), but we can revise this later.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,55 +1,14 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:base"],
+
   "automerge": true,
   "dependencyDashboard": true,
   "dependencyDashboardApproval": true,
-  "enabledManagers": ["github-actions", "npm"],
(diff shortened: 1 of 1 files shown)


TimDiekmann (approved):
PR looks good to me, thank you!

PR #815, opened 2022-12-07
Use Type System Types in Graph Service

Alfred-Mountfield:

As part of a new iteration of the Block Protocol, we will need to start using the type system. This PR updates the graph service module to use the `EntityType` definition from the type system package, and disables mutation methods (which will be re-enabled in time as we learn their new APIs).

It also applies some drivebys to improve structuring of the package in preparation for bigger additions, and adds a drive-by alias for `EntityId` which makes the code more readable. Note: this is exported, so perhaps we want to revert the commit, it adds a dependency for `@blockprotocol/graph` to the hook service, technically the service is dependent on that, so this somewhat nicely encapsulates it in code.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203509182568505/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

Please review commit by commit, or the diff may be confusing.

- The docs will need updating in follow-ups.

## ⚠️ Known issues

This breaks mock-block-dock.

## 🐾 Next steps

- Implement the `Subgraph`
- Use the `Subgraph` in various graph service responses
- Fix `mock-block-dock`

## 🛡 What tests cover this?

None to my knowledge

## ❓ How to test this?

Rely on TSC passing for the packages changed

## 📹 Demo

N/A


CiaranMn (approved):
Looks like a good starting point.

I assume the shape of the `graph` object we pass to blocks will change to be closer to the `subgraph` (it's currently `blockEntity`, `linkedEntities` and `linkGroups` which are like (a) a helper to get the root, (b) vertices and (c) edges in slightly different shapes)

Alfred-Mountfield:
Yep I think we'll completely rework the block-graph

PR #817, opened 2022-12-08
Add README.md files to local workspaces

kachkaev:

This PR explains the purpose of three local workspace we’ve got.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203494829477059/f) _(internal)_
- [Slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1670499445582139) _(internal)_
- #818 _(related)_

### DIFF
--- a/packages/@local/eslint-config/README.md
+++ b/packages/@local/eslint-config/README.md
@@ -1,0 +1,3 @@
+## `@local/eslint-config`
+
+This workspace contains shared preferences for `.eslintrc.cjs` files.
--- a/packages/@local/package-chores/README.md
+++ b/packages/@local/package-chores/README.md
@@ -1,0 +1,4 @@
+## `@local/package-chores`
+
+This workspace contains CLI scripts which are needed for package maintenance.
+These scripts are used in GitHub Workflows for CI checks and publishing.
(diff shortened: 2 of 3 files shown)


nonparibus (approved):
Thanks @kachkaev! Any more info about how the shared preference setup for `eslint` & `tsconfig` would be great, but this starts the process (very minimally) of improving our public documentation of these packages.

kachkaev:
I expect that devs will be able to figure out the setup for a new space by looking at what we have already. It is possible to  search for `@local/tsconfig` and `@local/esint-config` and then copy found files to a new folder. If we document this, we’ll need to maintain this doc but there is not guarantee that anyone would ever read it. If someone experiences real problems when creating new workspaces, we can discuss them with that person or document them specifically if needed.

nonparibus:
@kachkaev Even if it's just a case of saying "search for @local/tsconfig and @local/esint-config and then copy found files to a new folder" I think this would be preferable. We are underweight this, and it helps non-native developers (who may not be as experienced as our team) settle in and understand. I am not convinced this level of direction/detail really creates a maintenance burden, either.

PR #819, opened 2022-12-08
Add the `Subgraph` types

Alfred-Mountfield:

The Block Protocol Graph service is centered around a data abstraction which is represented as a Graph. With the introduction of the Type System there are an increasing variety of elements within this graph, and the previous ways of representing their relationships weren't sufficient in the face of this.

This PR introduces the `Subgraph` type (and its constituent types). The `Subgraph` is likely to replace a lot of the return types for various `get` and `aggregate` queries within the Graph Service, which should allow users to configure how much contextual information they want about the elements they're querying (for example, they should be able to say how deep link entities should be resolved and returned).

It is being designed for the future to support complex EAs which may have temporal versioning schemes (a somewhat new introduction to the Block Protocol), but the intention is to abstract a lot of this complexity behind a standard library of functions which provide the `Subgraph` API. These will be created in follow-ups

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203516598340492/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

Please see commits, effort has been made to keep them clean, and reviewing commit-by-commit should help mitigate some of the diff caused by restructuring of the package.

- Docs will be updated with the merge of the 0.3 branch

## ⚠️ Known issues

- `mock-block-dock` has not been updated, and is therefore broken, this will be fixed in follow-ups
- Some of the type guards require usage of the Type System package due to the use of `validateBaseUri`, this means that usages of this library also becomes dependent on the WASM being initialized. 

## 🐾 Next steps

- Add standard library for the `Subgraph`
- Add tests for the standard library

## 🛡 What tests cover this?

- Typescript compilation

## ❓ How to test this?

N/A at this time

## 📹 Demo

N/A

### DIFF
--- a/packages/@blockprotocol/graph/src/types/subgraph/edges.ts
+++ b/packages/@blockprotocol/graph/src/types/subgraph/edges.ts
@@ -1,9 +1,17 @@
+import { BaseUri } from "@blockprotocol/type-system/slim";
+
 import { EntityId } from "../entity";
-import { KnowledgeGraphOutwardEdge } from "./edges/outward-edge";
+import {
+  KnowledgeGraphOutwardEdge,
+  OntologyOutwardEdge,
+} from "./edges/outward-edge";
 import { Timestamp } from "./time";
 
-/** @todo - Add ontology edges here */
 export type Edges = {
+  [_: BaseUri]: {
+    [_: number]: OntologyOutwardEdge[];
+  };
+} & {
   [_: EntityId]: {
     [_: Timestamp]: KnowledgeGraphOutwardEdge[];
   };
--- a/packages/@blockprotocol/graph/src/types/subgraph/edges.ts
+++ b/packages/@blockprotocol/graph/src/types/subgraph/edges.ts
@@ -1,0 +1,10 @@
+import { EntityId } from "../entity";
+import { KnowledgeGraphOutwardEdge } from "./edges/outward-edge";
+import { Timestamp } from "./time";
+
+/** @todo - Add ontology edges here */
+export type Edges = {
+  [_: EntityId]: {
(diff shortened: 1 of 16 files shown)


packages/@blockprotocol/graph/src/types/subgraph/edges.ts
```diff
@@ -0,0 +1,18 @@
...
+import { Timestamp } from "./time";
+
+export type Edges = {
+  [_: BaseUri]: {
+    [_: number]: OntologyOutwardEdge[];
+  };
+} & {
+  [_: EntityId]: {
+    [_: Timestamp]: KnowledgeGraphOutwardEdge[];
+  };
```
CiaranMn, line 17: We could include further documentation in the key name (not suggesting these exact values, but making them all `_` is missing an opportunity to help understand what's going on)
```suggestion
  [typeBaseUri: BaseUri]: {
    [typeVersion: number]: OntologyOutwardEdge[];
  };
} & {
  [entityBaseId: EntityId]: {
    [edgeFirstCreatedAt: Timestamp]: KnowledgeGraphOutwardEdge[];
  };
```

CiaranMn (approved):
One suggestion, otherwise looks good!

packages/@blockprotocol/graph/src/types/subgraph/edges.ts
Alfred-Mountfield: That's a good point! I'm reworking these types a little bit to make them more `tsc` friendly in a follow-up PR so I'll do it then to avoid more rebasing work for the stacked PRs

PR #820, opened 2022-12-08
Introduce the start of the `Subgraph` standard library

Alfred-Mountfield:

As expressed in #819, the `Subgraph` type is being designed for hopeful future support of temporal versioning, however the intention is to abstract a lot of this complexity behind a standard library of functions which provide the Subgraph API. 

We would like to support embedding applications (EAs) that have temporal versioning schemes, but we would also like to not force most blocks to deal with such things if they don't need to. (In the vast amount of cases, blocks will be interested in the latest version of something, and in a lot of cases, EAs may not even support historical data).

The `Subgraph` object is being designed to support complex temporal versioning, but the hope is that the API to interact with it can be compatible with non-temporal versioned use-cases.

This PR introduces stdlib methods for retrieving elements from the `Subgraph`, and introduces overloaded methods that abstract away temporal versioning to only expose dealing with the latest editions of things.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203516598340493/f) _(internal)_

## 🚫 Blocked by

- [x] #819 

## 🔍 What does this change?

Again see commits please, they _should_ be somewhat self explanatory, although a lot of rebasing was involved so they may be incorrect in places.

- The docs will be updated when the 0.3 branch is more mature.

## ⚠️ Known issues

- `mock-block-dock` is broken at the moment as it has not been updated
- The graph service has not been updated

## 🐾 Next steps

- Update graph service to use `Subgraph`
- Update `mock-block-dock`
- Improve documentation

## 🛡 What tests cover this?

- Only `tsc` at the moment, although it's still somewhat broken until the next PR

## ❓ How to test this?

`tsc` and manual code-review I'm afraid

## 📹 Demo

N/A


thehabbos007 (commented):
Thanks, Alfie! Looks good, just have a couple comments

packages/@blockprotocol/graph/src/types/ontology.ts
```diff
@@ -1,7 +1,10 @@
-import { BaseUri } from "@blockprotocol/type-system/dist/cjs-slim/index-slim";
-import { validateBaseUri } from "@blockprotocol/type-system/wasm/type-system";
+import { BaseUri, validateBaseUri } from "@blockprotocol/type-system/slim";
+
+export * from "./ontology/data-type.js";
```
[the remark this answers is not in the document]
Alfred-Mountfield: I had opened a comment about this but force-pushed over it or something I think 😅 

I'm not sure why this is required, I haven't seen it in other packages, I think it might be because of how the `dist` is generated and the module is ESM. I encountered an issue when _using_ this package (not building it or through `tsc` which is frustrating) in `mock-block-dock`, I don't remember the error unfortunately, but I think it was about ESM and it explicitly told me I needed file endings. There were also file endings in some of the code already, but inconsistently, which confused me even more.

Perhaps @kachkaev has an idea.

packages/@blockprotocol/graph/src/stdlib.ts
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @param subgraph
+ * @param {EntityId} entityId - The ID of the link entity
+ */
+export const getLeftEntityForLinkEntity = (
+  subgraph: Subgraph,
+  entityId: EntityId,
+): Entity => getLeftEntityForLinkEntityTemporal(subgraph, entityId);
+/**
+ * Gets the "right entity" (by default this is the "destination") of a given link entity.
```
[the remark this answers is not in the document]
Alfred-Mountfield: I'm not sure anything is "used" as such yet, we need to decide on terminology generally. I think we should be consistent here so I'll go with `target` but I don't particularly like it either 🙃

packages/@blockprotocol/graph/src/stdlib.ts
```diff
@@ -0,0 +1,102 @@
...
+ * Gets all incoming link entities from a given entity.
+ *
+ * @param subgraph
+ * @param {EntityId} entityId - The ID of the source entity to search for outgoing links from
+ */
+export const getIncomingLinksForEntity = (
+  subgraph: Subgraph,
+  entityId: EntityId,
+): Entity[] => getIncomingLinksForEntityTemporal(subgraph, entityId);
+/**
```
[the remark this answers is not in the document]
Alfred-Mountfield: `eslint` and `prettier` seems happy for me locally. I've added spaces anyway because this wasn't intended and I don't like the lack of them 😅

packages/@blockprotocol/graph/src/stdlib.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+export const getOutgoingLinksForEntity = (
+  subgraph: Subgraph,
+  entityId: EntityId,
+): Entity[] => getOutgoingLinksForEntityTemporal(subgraph, entityId);
+/**
+ * Gets all incoming link entities from a given entity.
+ *
+ * @param subgraph
+ * @param {EntityId} entityId - The ID of the source entity to search for outgoing links from
```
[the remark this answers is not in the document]
Alfred-Mountfield: Good spot, updated other occurrences, thanks!

thehabbos007 (approved):

packages/@blockprotocol/graph/src/types/subgraph/edges/outward-edge.ts
```diff
@@ -20,18 +23,17 @@ type GenericOutwardEdge<K, E> = {
...
 };
 
+export type EntityIdAndTimestamp = {
+  baseId: EntityId;
+  timestamp: Timestamp;
+};
+
 export type OntologyOutwardEdge =
   | GenericOutwardEdge<OntologyEdgeKind, OntologyTypeEditionId>
   | GenericOutwardEdge<SharedEdgeKind, EntityEditionId>;
```
CiaranMn, line 33: Not introduced by this PR, but type implies you can have:
```ts
{
  kind: "IS_OF_TYPE",
  reversed: false,
  rightEndpoint: <some entity's id>
}
```
– is that right / known / fine?

packages/@blockprotocol/graph/src/stdlib/subgraph/element/data-type.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @throws if the vertex isn't a `DataTypeVertex`
+ */
+export const getDataTypeById = (
+  subgraph: Subgraph,
+  dataTypeId: VersionedUri,
+): DataTypeWithMetadata | undefined => {
+  const [baseUri, version] = [
+    extractBaseUri(dataTypeId),
+    extractVersion(dataTypeId),
+  ];
```
CiaranMn, line 43: Presumably at some point we can have a
```ts
const { baseUri, version } = destructureVersionedUrI(dataTypeId);
```

packages/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  const editionVersions = Object.keys(entityEditions).sort();
+
+  let entityEdition: Entity | undefined;
+
+  // Short circuit for efficiency, just take the latest
+  if (targetEditionInformation === undefined) {
+    [entityEdition] = editionVersions
+      .slice(-1)
+      .map((latestVersion) => entityEditions[latestVersion]!.inner);
```
CiaranMn, line 63: Alternative if mutating the original array isn't a concern, which I find a bit easier to parse
```suggestion
    entityEdition = entityEditions[edititionVersions.pop()]!.inner;
```

packages/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
```diff
@@ -0,0 +1,124 @@
...
+    let targetVersion: EntityVersion | undefined;
+    for (let idx = 0; idx < editionVersions.length; idx++) {
+      /** @todo - If we expose endTimes we can do an interval check here per edition, rather than needing to infer it */
+      // Rolling window: we've sorted, so for each edition's version check if the given timestamp is in between the
+      // start of this edition and the next one (lower-bound-inclusive)
+      const [editionVersion, nextEditionVersion] = editionVersions.slice(
+        idx,
+        idx + 1,
+      ) as [EntityVersion, EntityVersion | undefined];
+
```
CiaranMn, line 79: Presumably an easy optimization would be to bail out of the loop with `undefined` if the very first `editionVersion > targetTime`, because there's no point checking the remainder(?)

packages/@blockprotocol/graph/src/stdlib/subgraph/roots.ts
```diff
@@ -0,0 +1,151 @@
...
+import { mustBeDefined } from "../must-be-defined.js";
+import { getDataTypeByEditionId } from "./element/data-type.js";
+import { getEntity } from "./element/entity.js";
+import { getEntityTypeByEditionId } from "./element/entity-type.js";
+import { getPropertyTypeByEditionId } from "./element/property-type.js";
+
+/**
+ * Returns all root elements.
+ *
+ * The type of this can be constrained by using some of the helper type-guards:
```
CiaranMn, line 18: I think this is a bit clearer, if I've understood the intent
```suggestion
 * For a narrower return type, first narrow the type of `subgraph` by using one of the helper type-guards:
```

packages/@blockprotocol/graph/src/types/ontology.ts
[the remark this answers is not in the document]
CiaranMn: You need file extensions on relative import statements in ESM packages, i.e. those with `type: "module"` - [docs](https://nodejs.org/api/esm.html#import-specifiers)

packages/@blockprotocol/graph/src/types/subgraph/edges/outward-edge-alias.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * A collection of 'aliases' which describe various variants of outward edges in more accessible-forms
+ */
+
+import { EntityIdAndTimestamp, OutwardEdge } from "./outward-edge";
```
CiaranMn, line 5:
```suggestion
import { EntityIdAndTimestamp, OutwardEdge } from "./outward-edge.js";
```

packages/@blockprotocol/graph/src/types/subgraph/edges/outward-edge-alias.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/** @todo - is there a way to have TS force us to make this always satisfy `KnowledgeGraphOutwardEdge`? */
```
CiaranMn, line 7: Can you give me an example of one where it doesn't?

CiaranMn (commented):
A few comments, mostly minor suggestions on code, nothing blocking

packages/@blockprotocol/graph/src/types/subgraph/edges/outward-edge-alias.ts
Alfred-Mountfield: What I mean is you can change it to be whatever you want, you can make the string "foo" if you want. Which is unfortunate if we were to refactor `KnowledgeGraphOutwardEdge`, as this wouldn't be picked up, and wouldn't cause type errors (except in usage).

Again I want to be able to say `OutgoingLinkEdge satisfies KnowledgeGraphOutwardEdge`. It should be possible to actually define a subtype.

packages/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -0,0 +1,225 @@
...
+} from "../../../types/subgraph/edges/outward-edge-alias.js";
+import { mustBeDefined } from "../../must-be-defined.js";
+import { getEntity } from "../element/entity.js";
+
+const convertTimeToStringWithDefault = (
+  timestamp?: Date | string,
+  defaultConstructor = () => new Date().toISOString(),
+) => {
+  return timestamp === undefined
+    ? defaultConstructor()
```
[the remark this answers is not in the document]
Alfred-Mountfield: I can't remember my intentions around this, but I agree it's not used. Have removed, thanks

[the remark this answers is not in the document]
Alfred-Mountfield: [Remove unnecessary default function](https://github.com/blockprotocol/blockprotocol/pull/820/commits/ada39514998a50be48b8444c649539ed4609330e)

packages/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -0,0 +1,225 @@
...
+    : typeof timestamp === "string"
+    ? timestamp
```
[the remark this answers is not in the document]
Alfred-Mountfield: I'm not sure we should do this, it could cause strange behaviour with timezones and other parsing things. I think we should just require in documentation that the string is ISO formatted

packages/@blockprotocol/graph/src/stdlib-temporal.ts
```diff
@@ -0,0 +1,24 @@
+export {
```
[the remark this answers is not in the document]
Alfred-Mountfield: Have added tiny docs at the tops of the files, appreciate it's not comprehensive but as you say, can do it later
[Add basic top-level docs](https://github.com/blockprotocol/blockprotocol/pull/820/commits/fba43ee96c2bf5a0387e3229b73982200a7c73f7)

packages/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -0,0 +1,225 @@
...
+  if (!entityEdges) {
+    return [];
+  }
+
+  const uniqueEntitiesFilter = getUniqueEntitiesFilter();
+
+  return (
+    Object.entries(entityEdges)
+      // Only look at outgoing edges that were created before or at the timestamp
+      .filter(([edgeTimestamp, _]) => edgeTimestamp <= timestampString)
```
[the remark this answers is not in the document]
Alfred-Mountfield: Is there a reason you prefer this? I like it being explicit that we're destructuring the tuple here

packages/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -0,0 +1,225 @@
...
+      // Extract the link `EntityEditionId`s from the endpoints of the link edges
+      .flatMap(([_, outwardEdges]) => {
+        return (outwardEdges as OutwardEdge[])
+          .filter(isOutgoingLinkEdge)
+          .map((edge) => {
+            return edge.rightEndpoint;
+          });
+      })
+      .map(({ baseId: linkEntityId, timestamp: _firstEditionTimestamp }) => {
```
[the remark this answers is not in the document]
Alfred-Mountfield: Yeah this was added on a previous PR where _not_ having it was actually causing some confusion when @thehabbos007 was trying to understand the choice of timestamp. 

I also like that the entry is explicitly ignored here, as it would be quite easy to accidentally assume the wrong timestamp had been used

packages/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -0,0 +1,225 @@
...
+        return mustBeDefined(
+          getEntity(
+            subgraph,
+            linkEntityId,
+            // Find the edition of the link at the given moment (not at `_firstEditionTimestamp`, the start of its history)
+            timestampString,
+          ),
+        );
+      })
+      .filter((x): x is Entity => x !== undefined)
```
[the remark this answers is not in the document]
Alfred-Mountfield: It would, I noticed that after reading your previous comment, it's removed 
@Alfred-Mountfield
[Remove unnecessary undefined checks](https://github.com/blockprotocol/blockprotocol/pull/820/commits/ebab17cf139e04667baa4705fb72afa07e6858fa)

packages/@blockprotocol/graph/src/types/subgraph/edges/outward-edge.ts
Alfred-Mountfield: I did try looking at one that constrains the type properly but without explicit ways to define subtypes it's not really ergonomic :( 

(I want `satisfies` for types)

packages/@blockprotocol/graph/src/stdlib/subgraph/element/data-type.ts
Alfred-Mountfield: Yeah we've already got that in HASH, we could look adding it here later on as an extension too

packages/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
Alfred-Mountfield: This behaves differently though, the pop will throw an error whereas the map will return `undefined`

Alfred-Mountfield: Oh wait it doesn't throw, let me check if I'm happy with it being mutated

Alfred-Mountfield: That syntax doesn't work anyway because pop can be undefined so we'd need a temporary variable and such. Going to leave it as is
image

packages/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
Alfred-Mountfield: I think that makes sense too, good spot!

packages/@blockprotocol/graph/src/stdlib/subgraph/roots.ts
Alfred-Mountfield: Good call, thanks

[Improve doc](https://github.com/blockprotocol/blockprotocol/pull/820/commits/bf123b4e52437d7ec340ef5b0211175417983ae1)

packages/@blockprotocol/graph/src/types/subgraph/edges/outward-edge-alias.ts
Alfred-Mountfield: I'm confused, I swear I checked all of them 😭 

[Missing .js import](https://github.com/blockprotocol/blockprotocol/pull/820/commits/836126f168db91e4f2f6f46a7d46d0dae03ac11d)

packages/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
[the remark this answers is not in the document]
CiaranMn: Just seems a bit pointless to me – destructuring in JS tends to be 'take what you need, and no more'.

The fact that we're destructuring a tuple should be clear from the value implied by `Object.entries` and the syntax. If people don't know what's produced by that, then I'm not sure introducing a mysterious `_` is going to help.

It's not a big deal to leave it in but it'll probably look weird to the next JS developer who'll remove it (which also won't be a big deal, so let's not sweat it)

packages/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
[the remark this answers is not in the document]
CiaranMn: Fair enough!

packages/@blockprotocol/graph/src/types/subgraph/edges/outward-edge.ts
CiaranMn: Simple tweak that prevents someone constructing the incorrect example above
```ts
type GenericOutwardEdge<K, E, Reversed = boolean> = {
  kind: K;
  reversed: Reversed;
  rightEndpoint: E;
};

...

export type OntologyOutwardEdge =
  | GenericOutwardEdge<OntologyEdgeKind, OntologyTypeEditionId>
  | GenericOutwardEdge<SharedEdgeKind, EntityEditionId, true>;

export type KnowledgeGraphOutwardEdge =
  | GenericOutwardEdge<KnowledgeGraphEdgeKind, EntityIdAndTimestamp>
  | GenericOutwardEdge<SharedEdgeKind, OntologyTypeEditionId, false>;
```

CiaranMn: Obviously if `SharedEdgeKind` is expanded to include things that can go ontology<>knowledge then it'd have to be amended further. 

Probably at that point it would be worth introducing an separate record of which edges can go from ontology<>ontology, knowledge<>ontology (I think you just have to know this at the moment?)

packages/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
CiaranMn: Fair enough

packages/@blockprotocol/graph/src/types/subgraph/edges/outward-edge-alias.ts
CiaranMn: Got it – this is maybe as close as you can get
```ts
type SubType<Type extends Record<string, unknown>, Sub extends Type> = {
  [Key in keyof Type]: Sub[Key];
};

export type OutgoingLinkEdge = SubType<
  KnowledgeGraphOutwardEdge,
  {
    reversed: true;
    kind: "asdad"; // error, 
    rightEndpoint: EntityIdAndTimestamp;
  }
>;
```
This will not prevent additional properties being assigned, which is hard to eliminate everywhere in TS (see [here](https://stackoverflow.com/a/62099607/17217717)).

Alfred-Mountfield: Hm, I wish the IDE understood this a little better :( 

You don't get any autocomplete for example.

image

And refactoring (e.g. changing `reversed` to `test`):
image

Doesn't get picked up:
image

At least it does throw an error so I guess it's better than before. I'll add it, I just wish there was native support :(

Thanks again

packages/@blockprotocol/graph/src/types/subgraph/edges/outward-edge.ts
Alfred-Mountfield: Bit concerned about adding loads of generics and decreasing the readability of the types, but I think generally this is a great suggestion, so I've added it. Thanks! (also added a few more generics 😅)

[Make GenericOutwardEdge even more generic](https://github.com/blockprotocol/blockprotocol/pull/820/commits/e5ed3aa96c043daec71842c3a0c6d1717d701456)

packages/@blockprotocol/graph/src/types/subgraph/edges/outward-edge-alias.ts
Alfred-Mountfield: [Add Subtype util and constrain link aliases](https://github.com/blockprotocol/blockprotocol/pull/820/commits/e3e6394788fa9574b8ae5eede0c5a8311bd6d2d5)

packages/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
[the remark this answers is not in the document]
Alfred-Mountfield: I would agree if it wasn't a `filter`. As in if this was a `map` or something and you were destructuring to just select one part of it, that'd make more sense to me, because you don't need to care about the next thing, but because the second element is genuinely used in the next function, I'd prefer it to be explicit, otherwise it takes that extra second to understand what's going on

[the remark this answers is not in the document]
CiaranMn: Fair enough. If it's for clarity I would consider using variable names like `_edge` because it makes it even easier to see what's going on, but isn't important to do in this PR

CiaranMn (approved):

packages/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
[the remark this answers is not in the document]
Alfred-Mountfield: Tiny change to do so I'll driveby and then merge. Cheers

PR #825, opened 2022-12-09
Add start of non-versioned `Subgraph` stdlib entry-point

Alfred-Mountfield:

We would like to support EAs that have temporal versioning schemes, but we would also like to not force most blocks to deal with such things if they don't need to. (In the vast amount of cases, blocks will be interested in the latest version of something, and in a lot of cases, EAs may not even support historical data).

The `Subgraph` object is being designed to support complex temporal versioning, but the hope is that the API to interact with it can be compatible with non-temporal versioned use-cases.

This PR introduces stdlib methods for retrieving entities from the `Subgraph`, and introduces overloaded methods that abstract away versioning to only dealing with the latest editions of things. The more complex methods are currently not exposed through the `exports` of the package.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203516598340493/f) _(internal)_

## 🚫 Blocked by

- [ ] #820 

## 🔍 What does this change?

See commits

- The docs will need updating when the 0.3 branch is ready for consumption, this will be done in follow-ups

## ⚠️ Known issues

- Standard library functions around link entities aren't provided yet.
- `mock-block-dock` has not been updated

## 🐾 Next steps

- Introduce more stdlib functions for link entities
- Add tests

## 🛡 What tests cover this?

- None besides `tsc` at the moment

## ❓ How to test this?

Check `tsc` and read the code I'm afraid

## 📹 Demo

N/A


Alfred-Mountfield:
Closed as I rebased and joined this with #820

PR #826, opened 2022-12-09
Update and Fix the Graph Service

Alfred-Mountfield:

#815 , #819, and #820 have progressively been updating parts of the Graph Service to encode the `Subgraph` type and some of the new shapes of `Entity`s and other graph elements.

This PR updates the rest of the Graph Service, using `Subgraph` on most `get*` and `aggregate*` methods, comments out linkedAggregations for reimplementation later, and removes outdated definitions (such as the older links definitions).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/home/1200969824692742/1203509182568508) _(internal)_

## 🚫 Blocked by

- [ ] #820 

## 🔍 What does this change?

See commits

The docs will need updating as the 0.3 branch matures

## ⚠️ Known issues

- There is no way to create links at the same time as an entity now, this will be a problem for required links
- The JSON schema for the Graph Service has not been updated yet

## 🐾 Next steps

- Reimplement some of the Type System library in TS to avoid the need for asynchronous initialization on simple functions like `validateBaseUri`
- Fix Mock Block Dock
- Write tests for the subgraph stdlib

## 🛡 What tests cover this?

- `tsc` only to my knowledge

## ❓ How to test this?

- Run `tsc`? The next PR for fixing `mock-block-dock` will be better evidence.

## 📹 Demo

See follow-up PR


packages/@blockprotocol/graph/src/custom-element.ts
```diff
@@ -5,19 +5,15 @@ import {
   EntityPropertiesObject,
   GraphBlockHandler,
 } from "./index.js";
+import { getRoots } from "./stdlib/subgraph/roots";
 
-export interface BlockElementBase<
-  BlockEntityProperties extends EntityPropertiesObject | null,
```
Alfred-Mountfield, line 10: I must say I didn't fully understand the purpose of all of these generic parameters (I think it was to let blocks tell TypeScript what the type of the Block Entity was). With my removal this isn't going to be possible, so perhaps we want to rework the `Subgraph` object to be able to even further constrain the typings of the `roots` with another generic parameter.

Probably by making `Entity` generic.

It's a little hard to figure this out prior to us knowing the shape of TS type generation from the type system.

CiaranMn: To note that the block templates won't compile now (because they are still passing a generic), but I suggest we just leave them because we're going to want to plug in a generic somehow.

Agree we can figure out what happens here later

packages/@blockprotocol/graph/src/custom-element.ts
```diff
@@ -57,28 +53,34 @@ export abstract class BlockElementBase<
...
     }
 
     return this.graphService.updateEntity({
       data: {
-        entityId: this.graph.blockEntity.metadata.editionId.baseId,
+        entityId: blockEntity.metadata.editionId.baseId,
+        entityTypeId: blockEntity.metadata.entityTypeId,
         properties,
+        leftToRightOrder: blockEntity.linkData?.leftToRightOrder,
+        rightToLeftOrder: blockEntity.linkData?.rightToLeftOrder,
```
CiaranMn, line 83: As it stands it's impossible to update the link order via this method if the root entity is a link entity. We could add in additional arguments for order to this

We could also consider removing this method as it's just a helper to make calling `updateEntity` for the root entity a bit easier

CiaranMn (approved):

packages/@blockprotocol/graph/src/custom-element.ts
CiaranMn: No, it's fine, we can clean up later and will need to revisit this in light of whatever we do with generics in `Subgraph` anyway

Alfred-Mountfield: 😅  I jumped the gun and did it beforehand, I'm happy to drop the commit but I think it's better to have it be consistent in the interim anyway

CiaranMn: No worries, sounds good!

PR #830, opened 2022-12-11
Fix Mock-Block-Dock

Alfred-Mountfield:

#826 updated the Graph Service for the new changes we're experimenting with in 0.3 These broke `MockBlockDock` (as expected). This PR updates `MockBlockDock` for the new definitions, giving it an understanding of the Type System, as well as implementing `Subgraph` resolution, creation, and exploration.

It implements the new format of the Graph Service hooks, and removes outdated code. 

There are a few driveby's for cleanliness too.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432081/1203528958301159/f) _(internal)_

## 🚫 Blocked by

- [x] #826
- [x] #829

## 🔍 What does this change?

See commits, effort has been made to keep them self-explanatory

- A full docs review will be needed when the 0.3 branch is more mature

## ⚠️ Known issues

I have tried to capture issues in code comments, or comments directly on the PR

## 🐾 Next steps

- Update MBD to support hooks for the ontology elements
- Update MBD to support temporal versioning

## 🛡 What tests cover this?

- Unknown, `tsc` is passing, and I have tested locally

## ❓ How to test this?

1.  Checkout the branch 
2.  Merge #829 in locally if it hasn't been merged upstream already (and this branch updated to include it)
3.  Run and test mock-block-dock as usual

**If you want to see the demo data** then change the following line in `useMockBlockProps`
```ts
entities: [...(initialEntities ?? initialMockData.entities)],
```
to this
```ts
entities: [...initialEntities ?? [], ...initialMockData.entities],
```

## 📹 Demo

Uploading Screen Recording 2022-12-11 at 03.50.14.mov…


packages/mock-block-dock/dev/test-react-block.tsx
```diff
@@ -2,39 +2,63 @@ import {
...
   if (readonly) {
     return (
       <div ref={blockRef}>
         <h1>
-          Hello {properties.name}! The id of this block is {entityId}
+          Hello{" "}
+          {
+            blockEntity?.properties[extractBaseUri(propertyTypes.name.$id)] as
+              | string
+              | undefined
```
Alfred-Mountfield, line 50: This is the sort of thing that should become a lot more ergonomic with type generation

packages/mock-block-dock/dev/test-react-block.tsx
```diff
@@ -44,19 +68,36 @@ export const TestReactBlock: BlockComponent<AppProps> = ({ graph }) => {
...
-                entityId,
-                properties: { ...properties, name: event.target.value },
+                entityId: blockEntity.metadata.editionId.baseId,
+                entityTypeId: blockEntity.metadata.entityTypeId,
+                properties: {
+                  ...blockEntity.properties,
+                  [extractBaseUri(propertyTypes.name.$id)]: event.target.value,
+                },
+                leftToRightOrder: blockEntity.linkData?.leftToRightOrder,
+                rightToLeftOrder: blockEntity.linkData?.rightToLeftOrder,
```
Alfred-Mountfield, line 100: Until we implement support for JSON patch or something (to be able to express removal of properties), we're requiring the full contents to be passed on update

packages/mock-block-dock/package.json
```diff
@@ -40,12 +40,14 @@
...
     "@blockprotocol/graph": "0.0.18",
     "@blockprotocol/hook": "0.0.8",
+    "@blockprotocol/type-system": "0.0.2",
     "@emotion/react": "^11.10.0",
     "@emotion/styled": "^11.10.0",
     "@lit-labs/react": "^1.0.4",
     "@mui/material": "^5.10.4",
     "ajv": "^8.9.0",
     "echarts": "^5.3.3",
+    "lodash": "4.17.21",
```
Alfred-Mountfield, line 50: See TODO about lodash

packages/mock-block-dock/src/datastore/mutate-subgraph.ts
```diff
@@ -0,0 +1,123 @@
...
+      [atTime]: [outwardEdge],
+    };
+  } else if (!subgraph.edges[sourceEntityId]![atTime]) {
+    subgraph.edges[sourceEntityId]![atTime] = [outwardEdge];
+  } else {
+    const outwardEdgesAtTime = subgraph.edges[sourceEntityId]![atTime]!;
+    /**
+     * @todo - Q for PR review: Added a lodash dependency for this, equality of the outward edge is actually complicated
+     *    fine to keep?
+     */
```
Alfred-Mountfield, line 33: I saw our `util.ts` file had some patched utilities to avoid a heavy weight lodash dependency or something. I didn't see any for `isEqual`. I'd experimented with some other deep equality strategies but the types (lack of branding and such) makes it quite difficult to write it nicely.

packages/mock-block-dock/src/data/entities.ts
```diff
@@ -1,49 +1,130 @@
...
+};
+
+const createWorksForLink = (
+  sourceEntityId: string,
+  destinationEntityId: string,
+): Entity => {
+  return {
+    metadata: {
+      editionId: {
+        baseId: `${sourceEntityId}-works-for-${destinationEntityId}`,
```
Alfred-Mountfield, line 53: These look horrible in the visualiser, probably shouldn't do this 🙃

packages/mock-block-dock/src/data/entity-types.ts
```diff
@@ -1,61 +1,119 @@
...
+};
+
+export const entityTypes = {
+  company,
+  person,
+  worksFor,
+  founderOf,
+  testType,
+};
+// satisfies Record<string, EntityType> };
```
Alfred-Mountfield, line 119: I ran into problems with using `satisfies` (can't remember if it was Webpack or something else).

If I could have used it then we could have inlined the `entityTypes` definitions and had them `as const`, which means their type would be preserved. Unfortunately if I type this as `Record<string, EntityType>` then their usages become buggered because TypeScript doesn't agree it's const.

Adding `as const` here also breaks it because the type isn't compatible (`EntityType` isn't `readonly` or something)

packages/mock-block-dock/src/datastore/mutate-subgraph.ts
```diff
@@ -0,0 +1,123 @@
...
+  EntityVertex,
+  KnowledgeGraphOutwardEdge,
+  KnowledgeGraphRootedEdges,
+  KnowledgeGraphVertices,
+  Subgraph,
+  Timestamp,
+} from "@blockprotocol/graph";
+import isEqual from "lodash/isEqual";
+
+/** @todo - clean up the assertions here */
```
Alfred-Mountfield, line 13: I'm actually not sure why there are so many.

I think a lot of code like this could be fixed up with a `addOrSetWithDefault` utility method or something

packages/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -0,0 +1,252 @@
...
+    (traversalSubgraph as any).vertices[elementIdentifier.baseId] = {
+      [elementIdentifier.versionId]: element,
+    };
+  }
+
+  const toResolve: ResolveMap = new ResolveMap({});
+
+  for (const [edgeKind, depths] of typedEntries(unresolvedDepths)) {
+    // Little hack for typescript, this is wrapped in a function with a return value to get type safety to ensure the
+    // switch statement is exhaustive. Try removing a case to see.
```
Alfred-Mountfield, line 70: Unsure if I want to keep this. It makes it harder to read due to the increased indentation but it would be nice if we add support for the other edge kinds this would complain (and did complain when I messed up some stuff while developing).

Extracting it into a standalone function makes it a lot less ergonomic, you have to figure out all the input params, and it'd really need documentation I think.

packages/mock-block-dock/src/datastore/traverse/traversal-context.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+}
+
+/**
+ * Tracks the depths at which each element within the graph has been explored. This means that if the same element is
+ * encountered multiple times during traversal, at different respective depths, the context can determine if its
+ * neighbours have already been explored to a deeper depth than the current branch of the traversal, or if further
+ * exploration is needed.
+ */
+export class TraversalContext {
```
Alfred-Mountfield, line 65: This is very similar to our `DependencyContext` in the HASH graph implementation. Should probably try and take some lessons from the way I did it here and backport it back. Already discussing with @TimDiekmann about some of them

packages/mock-block-dock/src/datastore/use-mock-datastore.ts
```diff
@@ -0,0 +1,618 @@
...
+};
+
+export const useMockDatastore = (
+  initialData: MockData = {
+    entities: [],
+    // linkedAggregationDefinitions: [],
+  },
+  readonly?: boolean,
+): MockDatastore => {
+  const [graph, setGraph] = useState(mockDataToSubgraph(initialData));
```
Alfred-Mountfield, line 43: I used `useDefaultState` here and it cost me about an hour because the state would seemingly get reset on every render. I don't really get the point of the hook (`useDefaultState`)

packages/mock-block-dock/src/datastore/use-mock-datastore.ts
```diff
@@ -0,0 +1,618 @@
...
+                  },
+                ],
+              });
+              return currentGraph;
+            }
+
+            /**
+             * @todo - This assumes that once an entity is a link, it's always a link, and its endpoints can't be changed
+             *   however we don't enforce this currently. We should probably check the given entityType here
+             */
```
Alfred-Mountfield, line 186: This has some wider implications about the endpoints being mutable, which we were thinking they shouldn't be

packages/mock-block-dock/src/datastore/use-mock-datastore.ts
```diff
@@ -0,0 +1,618 @@
...
+      return {
+        errors: [
+          {
+            code: "NOT_IMPLEMENTED",
+            message: `Uploading files is not currently supported`,
+          },
+        ],
+      };
+
+      /** @todo - create the file entity-type and re-enable file uploading */
```
Alfred-Mountfield, line 499: This seemed out of scope for this PR, I think it'd have to be a Block Protocol managed type and that's a whole can of worms

packages/mock-block-dock/src/debug-view/dev-tools/datastore-graph-visualization.tsx
```diff
@@ -0,0 +1,368 @@
...
+import { Box } from "@mui/material";
+import { GraphChart, GraphSeriesOption } from "echarts/charts";
+import * as echarts from "echarts/core";
+import { SVGRenderer } from "echarts/renderers";
+import { useEffect, useRef, useState } from "react";
+
+import { useMockBlockDockContext } from "../../mock-block-dock-context";
+import { typedEntries } from "../../util";
+
+const parseLabelFromEntity = (entityToLabel: Entity, subgraph: Subgraph) => {
```
Alfred-Mountfield, line 24: Took heavy inspiration from our implementation in HASH of a similar function

packages/mock-block-dock/src/debug-view/dev-tools/datastore-graph-visualization.tsx
```diff
@@ -0,0 +1,368 @@
...
+const getSubgraphEntitiesAsEChartNodes = (subgraph: Subgraph): EChartNode[] => {
+  const allEntities = getEntities(subgraph);
+
+  /** @todo - Render link entities differently */
+  // const [linkEntities, nonLinkEntities] = partitionArrayByCondition(
+  //   allEntities,
+  //   (entity) =>
+  //     entity.linkData?.leftEntityId !== undefined &&
+  //     entity.linkData?.rightEntityId !== undefined,
+  // );
```
Alfred-Mountfield, line 181: I had an implementation which had split them up, but I didn't want to find all of the code around `EChartNode` and change if they have a border or whatever.

I do think we should play around with this a bit in the near future though for the sake of knowing how ergonomic the links are entities change even is.

Alfred-Mountfield: Would also be nice to have ontology nodes in the graph (potentially different colours, filterable, etc.)

packages/mock-block-dock/src/debug-view/dev-tools/properties.tsx
```diff
@@ -7,23 +8,37 @@ import {
...
+  const blockEntityType = useMemo(
+    () => getEntityTypeById(graph, blockEntity.metadata.entityTypeId),
+    [graph, blockEntity],
+  );
+
+  /** @todo - reimplement validation */
+
+  // const validate = ajv.compile(blockSchema ?? {});
+  // validate(blockEntity.properties);
+  // const errors = validate.errors?.map((error) => error.message);
```
Alfred-Mountfield, line 39: @CiaranMn related to your current work I think

packages/mock-block-dock/src/data/entity-types.ts
CiaranMn: NextJS [does not yet support](https://github.com/vercel/next.js/discussions/40895) `satisfies`

packages/mock-block-dock/src/datastore/mutate-subgraph.ts
CiaranMn: These are just objects, right? Simplest is this

```ts
const isObjectEqual = (
  first: Record<string, unknown>, 
  second: Record<string, unknown>
) => 
  JSON.stringify(first) === JSON.stringify(second)
```

packages/mock-block-dock/src/datastore/use-mock-datastore.ts
CiaranMn: The point of `useDefaultState` is to initialise a local state from the value passed to `useDefaultState`, and then re-initialiaze it if that value changes on susbequent renders (so that callers of this function can affect the local state by passing a new value for the 'default state' argument).

e.g. if I have MBD loaded in some app, and set `initialData` through its props, that initial datastore will get initialized in this local state. If I then want to change the initial datastore mid-session, I cannot do that via MBD props, because this normal `useState` is going to ignore the value passed to it after the first initialization.

The reason it kept re-rendering is because `mockDataToSubgraph(initialData)` returns a new value on every render (it creates a new object in memory), so as far as `useDefaultState` is concerned, it's getting a new value to reinitialize with on every render.

The fix would be to memoize the return of `mockDataToSubgraph`, e.g.

```ts
export const useMockDataToSubgraph = (mockData: MockData) => {
  const subgraph = useMemo(() => {
    const { entities } = mockData;

    const subgraph: Subgraph = {
      roots: [],
      vertices: {},
      edges: {},
      depths: {
        hasLeftEntity: { incoming: 255, outgoing: 255 },
        hasRightEntity: { incoming: 255, outgoing: 255 },
      },
    };

    addEntitiesToSubgraph(subgraph, entities);
  }, [mockData]);

  return subgraph;
};
```

CiaranMn (commented):
Haven't read all the code but a few comments on questions I saw on a first skim

packages/mock-block-dock/src/datastore/use-mock-datastore.ts
Alfred-Mountfield: Ah that makes sense, thank you for the great explanation!

Glad I opened a comment about this because I sensed something was off.

Fixed in [Turn mockDataToSubgraph into a hook](https://github.com/blockprotocol/blockprotocol/pull/830/commits/221b45e3773e968578f92ed2c6ed499f4af4391f)

packages/mock-block-dock/dev/dev-app.tsx
```diff
@@ -1,42 +1,61 @@
...
-
 const blockEntityMap = {
   react: {
-    entityId: "entity-react",
-    entityTypeId: "test-type-1",
-    properties: { name: "World" },
+    metadata: {
+      editionId: {
+        baseId: "entity-react",
+        versionId: new Date().toISOString(),
```
CiaranMn, line 19: Feels a bit odd to have `versionId` part of this when it is unused .3

CiaranMn, line 19: Feels a bit odd to have `versionId` part of this when it is unused .3

CiaranMn, line 19: Feels a bit odd to have `versionId` part of this when it is unused .3

CiaranMn, line 19: Feels a bit odd to have `versionId` part of this when it is unused .3

CiaranMn, line 19: Feels a bit odd to have `versionId` part of this when it is unused .3

CiaranMn, line 19: Feels a bit odd to have `versionId` part of this when it is unused .3

Alfred-Mountfield: `baseId` is `EntityId`

So you have 
```ts
type EntityEditionId: {
  baseId: EntityId,
  recordId: string
}
```

Alfred-Mountfield: > and the difference between baseId and recordId
This: 

> No, because blocks will need the knowledge about what identifies an entity, and what identifies a specific edition of an entity. 

The `EntityId` establishes the _identity_ of an entity (e.g. "Bob"), it doesn't identify a specific edition of an entity.

We need two pieces of information, what is the thing, and what edition of that thing is it.

We previously were using version to identify that edition, but if we want to support multi-temporal-axis systems then a singular version string will not be good enough to identify the specific record we're talking about.

So we need 
- `EntityId` the identity of the thing we're talking about
- `X` the identity of the exact record that we're looking at here, a specific edition of the thing
- `firstAxisTime`, the version of this on the first axis 
- `secondAxisTime`, the version of this on the second axis

Where `firstAxisTime` and `secondAxisTime` are not part of the record's identifier, so they can live inside `entity.metadata.version` or something.

Alfred-Mountfield: X _opaquely_ defines a specific `(firstAxisTime, secondAxisTime)` combination of a record. You can think of it as the primary key in the data base.

CiaranMn: I would keep X as `versionId` then, because 
(a) it makes it more obvious that it is irrelevant to non-versioned systems (is it omitted in these cases? any old junk passed?)
(b) in non-versioned systems, the id of the record is possibly identical to the `baseId` (I'd also consider renaming this – why not `entityId`?)

CiaranMn: If we are saying that X is _only required to distinguish a specific version of a thing_

in non-versioned systems this description:
> just an opaque unique identifier for the specific record in the DB

will be `baseId` (as it is currently called)

Alfred-Mountfield: Let's jump on a call tomorrow or something about this then, I think I'm not conveying this well, we physically cannot combine them, and `versionId` is misleading because versions are distinguished by their timestamps, and the whole multi-axis thing makes that confusing. Hence recordId, the whole point is to not misleadingly convey meaning about temporal versioning within it, it's just a necessary piece of metadata in some systems, and in other systems they can just set it to any string they like

packages/mock-block-dock/dev/dev-app.tsx
```diff
@@ -130,6 +149,7 @@ const DevApp: FunctionComponent = () => {
         }}
         debug
         key={testBlockType} // completely reset the state when block type has changed
+        initialEntities={Object.values(blockEntityMap)}
```
CiaranMn, line 152: Will need to have a bit more of a think about this because it means the dummy company data is never going to be used so long as someone wants to provide a starting block entity... maybe we have a prop to either add or don't add the bundled dummy data on top of whatever is provided here. Doesn't need to be figured out in this PR

CiaranMn (commented):
A few small comments – welcome your thoughts on those, but we can get this merged tomorrow.

Have played around all seems to work. The only surprise (which I should have expected) was the absence of the company etc dummy data in the mock datastore, but I've left a comment about that

We'll need to test blocks for 0.3 so it'll get more of a testing then

packages/mock-block-dock/src/datastore/mutate-subgraph.ts
Alfred-Mountfield: It's pretty messy but it'll almost certainly cover this case. I'll add it

packages/mock-block-dock/src/datastore/traverse.ts
Alfred-Mountfield: We could do, perhaps in a separate follow-up @kachkaev or someone could look at. 

I'd be slightly concerned at how well eslint might be able to introspect something like this if the union type is generated rather than being explicitly written.

packages/mock-block-dock/dev/dev-app.tsx
```diff
@@ -1,42 +1,61 @@
...
+        baseId: "entity-html-as-string",
+        versionId: new Date().toISOString(),
+      },
+      entityTypeId: entityTypes.testType.$id,
+    },
+    properties: { [extractBaseUri(propertyTypes.name.$id)]: "World" },
   },
-} as Record<string, Entity>;
+} as const;
+// } as const satisfies Record<string, Entity>;
```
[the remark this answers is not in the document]
Alfred-Mountfield: It would be nice to start using these appropriately, it's not really superfluous as it actually removes the need for casts and assertions in certain usages.

I can remove it if you'd prefer to not have commented out code (I tend to err on the side of removing it too), or I can put a todo about reintroducing this when satisfies is better supported?

packages/mock-block-dock/dev/test-react-block.tsx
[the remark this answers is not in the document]
Alfred-Mountfield: > Maybe we can make updateEntity take a generic which is either an Entity or whatever type for LinkEntity, and not take these link fields if it's known to be an Entity? feels weird to have to pass these fields to updateEntity for all entities.

You don't have to pass the fields if you know it's not a link entity. In this context we _probably_ could safely assume it's not, but because we haven't improved the block's parameter typing yet, I wouldn't iterate on this right now. When we start propagating the generic from `Entity` again with type generation then I think we can introduce a `LinkEntity` subtype which affirms there's `linkData`. 

> why are they at the root level here when they are nested in an Entity – i.e. why doesn't the data object passed to updateEntity have a linkData property instead of these two at the top level?

`LinkData` contains `leftEntityId` and `rightEntityId` which are currently not mutable (as I've highlighted above or on a previous PR about being a potential issue)

As such, I would prefer if we don't pass the fields in with `...(blockEntity.linkData ?? {})` just in case an implementation was badly implemented to do a plain replace.

packages/mock-block-dock/dev/dev-app.tsx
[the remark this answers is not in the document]
CiaranMn: I'm not sure why `satisfies` wouldn't work here. Leave it in and I'll look at it when I'm doing stuff in MBD

packages/mock-block-dock/dev/test-react-block.tsx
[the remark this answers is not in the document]
CiaranMn: Understood, thank you for the explanation

CiaranMn (approved):

PR #831, opened 2022-12-11
Expose ontology types of `repr` module

TimDiekmann:

In order to generalize over ontology types and entities in downstream applications (e.g. the HASH Graph API), it's required that the ontology types implements `Serialize` and `Deserialize` because of the orphan rule. 

## 🔗 Related links

- Part of [Asana task](https://app.asana.com/0/1203363157432084/1203449432963806/f) _(internal)_

## 🔍 What does this change?

Implements `Deserialize` and `Serialize` for `DataType`, `PropertyType`, and `EntityType`

## ⚠️ Known issues

The current design requires the struct to be cloned. This is unfortunate, but without changing the implementation it's hard to work around this. However, it's better to have those traits implemented in the crate rather than specifying the implementation by hand in downstream crates.

### DIFF
--- a/crates/type-system/src/ontology/mod.rs
+++ b/crates/type-system/src/ontology/mod.rs
@@ -36,21 +36,20 @@ pub use shared::{
     validate::{ValidateUri, ValidationError},
 };
 
-#[allow(
-    unused_imports,
-    reason = "We want to keep them here in case for the convenience of re-exporting"
-)]
 // Re-export the repr contents so they're nicely grouped and so that they're easier to import in
 // a non-ambiguous way where they don't get confused with their non repr counterparts.
 // For example, `import crate::repr` lets you then use `repr::DataType`
-pub(crate) mod repr {
+pub mod repr {
+    pub use super::{
+        data_type::repr::DataType, entity_type::repr::EntityType, property_type::repr::PropertyType,
+    };
     pub(crate) use super::{
-        data_type::repr::{DataType, DataTypeReference},
+        data_type::repr::DataTypeReference,
         entity_type::{
             links::repr::{Links, MaybeOneOfEntityTypeReference},
-            repr::{EntityType, EntityTypeReference},
+            repr::EntityTypeReference,
         },
-        property_type::repr::{PropertyType, PropertyTypeReference, PropertyValues},
+        property_type::repr::{PropertyTypeReference, PropertyValues},
         shared::{
             all_of::repr::AllOf,
             array::repr::{Array, ValueOrArray},
--- a/crates/type-system/src/ontology/mod.rs
+++ b/crates/type-system/src/ontology/mod.rs
@@ -41,15 +41,12 @@ pub use shared::{
 // For example, `import crate::repr` lets you then use `repr::DataType`
 pub mod repr {
     pub use super::{
-        data_type::repr::DataType, entity_type::repr::EntityType, property_type::repr::PropertyType,
-    };
-    pub(crate) use super::{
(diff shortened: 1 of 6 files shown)


Alfred-Mountfield (changes requested):
Thanks for the PR, and thanks for identifying pain-points of using this downstream. 

Back when we originally implemented this, we had discussed the possibility of implementing (de)serialize for the types and decided against it. Members of the community emphasises that `serde` is not a parsing library, and we shouldn't be trying to encapsulate the validation steps into deserialization, that's the reasoning for the `repr` boundary.

I think the main pain-point is probably that we don't provide easy access to the `repr` shapes within the crate, as we had wanted to keep them an implementation detail for the most part. It would seem that it'd be useful to be able to include them in other things though, benefiting from `deserialize` macros, etc.

I think we have two considerations, perhaps we should do both.
- Implement a `from_unchecked` for the types from their `repr`'s, for scenarios where users know the thing they have is valid (e.g. something they've serialized into a DB). We could potentially have a `deserialize_unchecked` method or something too.
- Expose the `repr` internal types, this would let users manually construct unvalidated forms of the types.. I'm not sure if that's particularly helpful, but it would at least let them have access to something with `(de)serialize` implemented on it, if they need to use it in something that requires those bounds.

TimDiekmann:
We agreed offline to expose the `repr` module and use them as opaque structures.

crates/type-system/src/ontology/mod.rs
```diff
@@ -36,21 +36,20 @@ pub use shared::{
...
         entity_type::{
             links::repr::{Links, MaybeOneOfEntityTypeReference},
-            repr::{EntityType, EntityTypeReference},
+            repr::EntityTypeReference,
         },
-        property_type::repr::{PropertyType, PropertyTypeReference, PropertyValues},
+        property_type::repr::{PropertyTypeReference, PropertyValues},
         shared::{
             all_of::repr::AllOf,
             array::repr::{Array, ValueOrArray},
```
Alfred-Mountfield, line 55: I think we might want to just expose the whole `repr` module publicly. This means people won't easily be able to do things with the `repr` like destructuring nicely (won't have access to the type), manually create it, etc.

crates/type-system/src/utils.rs
```diff
@@ -59,20 +60,24 @@ pub(crate) mod tests {
...
+        R: From<T> + PartialEq + Debug + Serialize + DeserializeOwned,
     {
-        let deserialized: T = T::from_str(input).expect("failed to deserialize");
-        let _reserialized =
-            serde_json::to_value(deserialized.clone().into()).expect("failed to serialize");
+        let deserialized_repr: R = serde_json::from_str(input).expect("failed to deserialize");
+        let value: T = deserialized_repr.try_into().expect("failed to convert");
+        let re_serialized_repr: R = value.clone().into();
 
         if let Some(repr) = expected_native_repr {
```
Alfred-Mountfield, line 73: Can we duplicate this check to still also check `deserialized_repr`?

TimDiekmann: Done in 5fb72f0e3a62774aa79aab26fcfe5b6a7bbccef1

crates/type-system/src/ontology/mod.rs
TimDiekmann: Sure: 00b073bd2c153e6f45eb7cfeaf13a36af841c567

Alfred-Mountfield (approved):

PR #848, opened 2022-12-16
Fix-up LICENSE files to improve GitHub display

nonparibus:

https://github.com/blockprotocol/blockprotocol/pull/839 was meant to improve the identification and display of licenses in our repo. Instead it resulted in this...

Capture 2022-12-16 at 17 03 53@2x

This PR fixes that. @kachkaev Have used absolute paths here, not sure if relative links would have been better. Lmk what you think.

---
  - https://app.asana.com/0/0/1203568198115142


kachkaev (approved):
Absolute paths should be fine here I guess.

kachkaev (approved):

PR #855, opened 2022-12-20
Voting section UX improvements

luisbettencourt:

This PR makes several improvements to the voting section on the WP landing page.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432087/1203568198115158/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203363157432087/1203568198115176/f) _(internal)_

## 🔍 What does this change?

- Allow users to suggest multiple applications (they will be separated by a comma, ex: "application1, application2");
- Removed the voted indicator on the `Other` button (used to be displayed after the user suggested an application);
- Added autofocus on inputs to make it easier to go through screens;
- Fix the voting section card changing height when going through the voting screens;
- Split the various voting screens into their own components.


luisbettencourt:
@nonparibus made the below change to the divider between sections, do you agree?

Before:

image

image


After:

image

image

kachkaev:
🤔

```
Error: @apps/site:lint:tsc: src/lib/api-client.ts(5,10): error TS2459: Module
'"../components/pages/wordpress/request-another-application"' declares 'ApplicationIds' locally, but it is not exported.
```

apps/site/src/components/pages/wordpress/voting/applications.tsx
```diff
@@ -0,0 +1,63 @@
+import { ReactElement } from "react";
+import {
+  CommentDotsIcon,
+  ContentfulIcon,
+  GithubIcon,
+  SanityIcon,
+  StrapiIcon,
+} from "../../../icons";
+
+export enum ApplicationIds {
```
kachkaev, line 10: It'd be great to avoid enums if possible. We can use discriminated unions instead. [Internal Slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1667303607693499).

kachkaev: Here, you can use  

```
export interface ApplicationInfo {}

const applicationInfoById = {
 "ECO_SANITY": {...}
} satisfies Record<string, ApplicationInfo>

Object.entries(applicationInfoById).forEach([applicationId, applicationInfo])
```

kachkaev (commented):
How feasible would it be to add integration tests for this new feature?

apps/site/src/components/pages/wordpress/voting/applications.tsx
nonparibus: I see some value in this. We are likely to use the voting component again in the future, in other contexts.

nonparibus:
To @kachkaev's suggestion above, I see some value in this. We are likely to use the voting component again in the future, in other contexts, so at least basic integration tests would be good.

luisbettencourt:
I'll convert this PR to draft until I get integration tests for this

apps/site/src/components/pages/wordpress/voting/applications.tsx
luisbettencourt: Addressed in eef141d6028484a0c3393645292c18a6bced3fc3. Thank you

luisbettencourt:
Integration tests will come in a follow-up, as decided in [slack](https://hashintel.slack.com/archives/C02LG39FJAU/p1671573560103359?thread_ts=1671573423.569379&cid=C02LG39FJAU)

nonparibus (approved):
We're going to get this in for the timebeing and will rollback if we need to.

PR #861, opened 2022-12-22
Generate TS types from schema URI, and use in block template

CiaranMn:
## 🌟 What is the purpose of this PR?

In summary, to add improved automatic type generation in block templates to prepare for version 0.3 of the spec.

The new type system being introduced in 0.3 changes how types are defined (see https://github.com/blockprotocol/blockprotocol/discussions/418), which means we can no longer require on the previous, approach to typing the "block entity" – the entity the graph sent to the block is rooted at.

We also want to improve how the entities _linked to_ the block entity are typed.

Previously:
- an entity's shape was defined by a self-contained JSON Schema that could be shipped with the block code
- there was no unambiguous defined way of expressing the valid links from an entity
- the `blockEntity` was sent separately to the things linked to it, which were under `blockGraph`

New context for 0.3:
- an entity type schema must be composed of individual "property type" schemas, hosted at their own URIs, which in turn must reference either other property type schemas, or "data type" schemas
- links are formalised in these schemas as entity types which inherit from a specific `Link` entity type (more docs to follow)
- all data is sent in a single graph, referred to as a "subgraph" as a hint to the fact that it may not be the entire graph available in the embedding application's datastore (although it may be)

We were previously generating TypeScript types for the block entity from the self-contained JSON schema shipped with a block's source, and none at all for anything linked to it.

This PR makes the initial necessary corrections and further improvements in preparation for introducing the new type system:
1. Introduces a utility to `@blockprotocol/graph` which, given a URI to a valid entity type schema
  - generates the necessary types for the entity type and its dependencies (property types and data types)
  - (if requested) generates types representing the valid links from the entity type, including their `linkEntity` (the special entity which stores metadata about the link) and `rightEntity` (the entity on the other side of the link), and possibly the valid links from _those_ entities if further depth is requested, and so on
2. Includes aliases for the `RootEntity` and a few other `RootX` types representing things linked from it
3. Starts to add a few generic slots in types in the helper functions in `@blockprotocol/graph/subgraph` which these types can be plugged in to
4. Adds generation script to `block-scripts` and uses it in `block-template-react` as a starting point, to generate the types and use them in `app.tsx`

This is a starting point we can iterate on based on usage – see next steps.

## 🔗 Related links

- This follows other prep for DX in the new type system in e.g. #820, #826, #830

## 📜 Does this require a change to the docs?

- Not yet – we'll be writing some.

## 🐾 Next steps

To make this usable:
1. Add generics to more types in `subgraph`
[...]

PR #866, opened 2023-01-03
Fix + update block templates to use new generated types

CiaranMn:

Makes further changes to the block templates to use the new generated types, building on #861. Fixes stuff I encountered when attempting to make a custom element block.

1. Move the type generator to its own export path in `@blockprotocol/graph` so that it doesn't need to be evaluated by everything using that package, and rename it to `codegen` to reflect its intended use as a CLI tool
2. Fixes the custom element base class
3. Updates `block-template-custom-element` and `block-template-react` to use the new generated types
4. removes the old block-schema generating code from `block-scripts`
5. adds `addSimpleAccessors` function to the `stdlib` (since I was messing with it – not sure we want it, to discuss)
6. Adds generics to another function in `stdlib`– I stopped as it was becoming difficult to construct the types – probably fine as it is for now(?)

## 🐾 Next steps

- Publish a collection of canary versions of packages that can be used to build 0.3 blocks with outside of this repo 

---
  - https://app.asana.com/0/0/1203533305475324


packages/@blockprotocol/graph/src/stdlib.ts
```diff
@@ -109,12 +109,36 @@ type GenericLinkEntityAndRightEntity = {
...
+>(
+  entity: Entity<Properties>,
+) => {
+  for (const [key, value] of Object.entries(entity.properties)) {
+    const [simpleAccessor, trailingSpace] = key.split("/").slice(-2);
+    if (trailingSpace !== "" || !simpleAccessor) {
+      throw new Error(`Property key ${key} is not a valid base URI.`);
+    }
+    // eslint-disable-next-line no-param-reassign -- intentional mutation
+    entity.properties[simpleAccessor] = value;
```
CiaranMn, line 140: If we do want a function such as this, this is the wrong way to do it, I've belatedly realise – it will have surprising results if people are spreading out the object.

The better way would be to create a `Proxy` with a `get` trap. I didn't do that first time round because it would have worse performance, because you'd need to check the key on every access rather than just throwing the new keys on at the start. 

I'm not going to update this yet until we've discussed if this is worth having at all.

packages/@blockprotocol/graph/src/codegen.ts
```diff
@@ -1,10 +1,15 @@
+// This file is has its own import path "@blockprotocol/graph/codegen"
```
Alfred-Mountfield, line 1:
```suggestion
// This file has its own import path "@blockprotocol/graph/codegen"
```

packages/@blockprotocol/graph/src/custom-element.ts
```diff
@@ -35,10 +49,36 @@ export abstract class BlockElementBase extends LitElement {
...
   };
 
+  private updateDerivedProperties() {
+    const blockEntitySubgraph = this.graph?.blockEntitySubgraph;
+
+    if (blockEntitySubgraph) {
+      const rootEntity = getRoots<{
+        editionId: EntityEditionId;
+        element: RootEntity;
+      }>(blockEntitySubgraph)[0];
```
Alfred-Mountfield, line 59: It feels like you shouldn't be passing in this generic argument. Is it not encapsulated within `blockEntitySubgraph`?

packages/@blockprotocol/graph/src/stdlib.ts
Alfred-Mountfield: I think we definitely need simple accessors, but as I mentioned in passing a while ago, we should _**not**_ be relying on the URL format. If we generate simple accessors they should be based on the `Title` property of the schemas.

I agree about your `Proxy` approach, and I also do not think we should be putting this method here alongside the `Subgraph` library functions. (Not suggesting we have to limit the `stdlib` to the `Subgraph` type but we should restructure the package otherwise.)

Alfred-Mountfield: I think for now we should probably remove this from the PR, log it as a potential approach, and create a follow-up to investigate the correct solution.

CiaranMn: I've removed the function and renamed the type in https://github.com/blockprotocol/blockprotocol/pull/866/commits/f1d4077f10d4c09a41baf5455f030d81d8d2026d– I'd like to keep the type around in case to experiment with it in block buiding to see if I can figure out some better pattern. We can remove it if it turns out to be not helpful

packages/@blockprotocol/graph/src/custom-element.ts
CiaranMn: Yes, correct, thank you – removed this and another instance in https://github.com/blockprotocol/blockprotocol/pull/866/commits/8c9cb91791d702baafd54164ac2bea189b3b230d

Alfred-Mountfield (approved):

PR #869, opened 2023-01-06
Export Subgraph building tools from @blockprotocol/graph

CiaranMn:

The new 0.3 spec will introduce the concept of a `Subgraph`, the types for which were introduced in #819.

A `Subgraph` represents the result of a query over a graph, whereby:
1. the graph is explored starting from `roots` (identifiers for `vertices`, which can be an `Entity`, `EntityType`, `PropertyType` or `DataType`)
2. connections from these vertices are followed to the next vertices, and so on according to specified `depths`
3. the resulting `Subgraph` is then all the `vertices` encountered in this exploration, and the `edges` that connect them, along with the `roots` and `depths` of the query which produced the data

In the BP data model, the data stored in vertices (`Entity`, `EntityType` etc) are sufficient to know every connection to or from a given vertex, but _only if every vertex is checked_. 

The `Subgraph` object therefore includes `edges`: a map providing quick lookups of all edges from a given edition of a vertex.

Because it is possible to derive `edges` from the data stored in `vertices`, embedding applications may not persist separate records of `edges` in a datastore, and instead produce the `Subgraph` as needed in response to queries.

This PR introduces a `buildSubgraph` function which completes a `Subgraph` with the necessary `edges`, extracting an implementation which did so internally in `mock-block-dock`, making it consumable by anyone via the `@blockprotocol/graph` package.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203363157432094/1203665143398379/f) _(internal)_

## 🔍 What does this change?

- Move some code out of `mock-block-dock` into `@blockprotocol/graph`
  -  includes a new `internal` export path for things that people shouldn't typically need to use (needed to be consumed by MBD)
- Introduce `buildSubgraph` – see docstring for its behavior

- Not yet, we will need to write some explaining subgraphs etc.

## ⚠️ Known issues

- See `@todo`s in `buildSubgraph`

## 🐾 Next steps

- Use this in the WordPress plugin.

## ❓ How to test this?

1.  Currently difficult as MBD is not building due to an unrelated issue which will be fixed by #866


packages/@blockprotocol/graph/src/internal/mutate-subgraph/is-equal.ts
```diff
@@ -0,0 +1,67 @@
+export const isEqual = (first: any, second: any): boolean => {
```
Alfred-Mountfield, line 1: Is this from somewhere? Do we want to attribute it? (In case there are updates there or someone discovers limitations)

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
```diff
@@ -0,0 +1,68 @@
...
+import {
+  Entity,
+  EntityEditionId,
+  GraphResolveDepths,
+  Subgraph,
+} from "../../types.js";
+
+/**
+ * Builds a Subgraph from a given set of entities, which may include 'link entities' –
+ * i.e. entities that represent relationships between other entities – or other entities.
```
Alfred-Mountfield, line 11:
```suggestion
 * Builds a Subgraph from a given set of entities, which may include 'link entities' –
 * i.e. entities that represent relationships between other entities 
```

I think that just confuses the sentence

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * The set of entities should represent the result of a query on a graph,
+ * and the 'roots' and 'depths' for that query provided along with the data.
```
Alfred-Mountfield, line 14: I don't really understand this sentence. The set of entities represent the roots and depths for ... ?

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Unbounded queries can currently only be represented via very high values for depths.
```
Alfred-Mountfield, line 16: I think we should just say that query depths are bounded by a maximum of 255. (We should encode this in the spec, or change our backend implementation to allow for greater depths)

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @returns a Subgraph containing:
+ *   - 'vertices' containing the provided entities
+ *   - 'edges' calculated by the function, representing connections between vertices
+ *   - 'depths' as provided by the caller
+ *   - 'roots' as provided by the caller
+ *
+ * @throws if the provided roots are not present in the data
+ *
+ * @todo 1. add support for ontology vertices (e.g. entity types)
+ * @todo 2. support null values for depths, for the result of queries with unlimited depths.
```
Alfred-Mountfield, line 36: Why do we want to support this? I can't really imagine situations where this would be a reasonable expectation of an EA, and except in _**very very very**_ specific edge-cases you should be expecting exponential growth in the size of a query (where honestly even a depth of > 10 might be too much to handle in memory).

I think we should get blocks to make follow-up requests for data if they really do need more than a depth of 255.

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
```diff
@@ -0,0 +1,68 @@
...
+ */
+export const buildSubgraph = (
+  data: { entities: Entity[] },
+  roots: EntityEditionId[],
+  depths: GraphResolveDepths,
+) => {
+  const missingRoot = roots.find(
```
Alfred-Mountfield, line 43: The `find` will only short-circuit on the unhappy path, so I think we should go with a `filter` here to provide more information to the user (i.e. the full list of missing roots rather than one)

packages/@blockprotocol/graph/src/internal/mutate-subgraph/is-equal.ts
Alfred-Mountfield: I see this was actually moved from an existing file, did I add it or did you? 😅 

My question still stands if we do know where it's from

CiaranMn: You added it – I think it was probably from a gist that I sent you. I can check if you adopted it and link it

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
CiaranMn: I previously had it as you have suggested, but then thought it might be misintrepreted as being "it may include link entities [and that's all it may include]"

how about
```suggestion
 * Builds a Subgraph from a given set of entities, which may include either or both of 'link entities' 
 * – i.e. entities that represent relationships between other entities – or non-link entities.
```

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
CiaranMn: any clearer?
```suggestion
 * The set of entities should represent the result of a query on a graph.
 * The 'roots' and 'depths' used for that query should be provided along with the data.
```

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
CiaranMn: To be honest when I added this I was thinking about `mock-block-dock`, which uses this to represent its entire datastore as a subgraph. Happy to just leave it as `255` max and we can revisit if there's any ever reason to change it

packages/@blockprotocol/graph/src/internal/mutate-subgraph/is-equal.ts
Alfred-Mountfield: Oops, please do

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
Alfred-Mountfield:
```suggestion
 * Builds a Subgraph from a given set of entities, some (or all) of which may be 'link entities' –
 * i.e. entities that represent relationships between other entities – or other entities.
```

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
Alfred-Mountfield: Much, thank you

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
Alfred-Mountfield: I think lets go with 255 until reason to not do so. (If only the 255 thing was _**easily**_ captured within JS/TS like it is in languages with differently sized number types)

packages/@blockprotocol/graph/src/internal/mutate-subgraph/is-equal.ts
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/869/commits/d049a78aa2e9a55bc48ca7d28a5817ca0130e765

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/869/commits/74ed5c6af3aa1339a460d6af1b62c6b49435ef1c

packages/@blockprotocol/graph/src/internal/mutate-subgraph.ts
```diff
@@ -7,12 +7,19 @@ import {
...
+import { isEqual } from "./mutate-subgraph/is-equal.js";
 
-import { isEqual } from "../util";
-
-/** @todo - clean up the assertions here */
-export const addKnowledgeGraphEdge = (
+/**
+ * Looking to build a subgraph? You probably want `import { buildSubgraph } from @blockprotocol/graph/stdlib`
+ *
+ * This MUTATES the given `Subgraph` by adding the given outwardEdge to the entity at the specified atTime.
```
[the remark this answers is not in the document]
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/869/commits/ed8370b832faee6439cd340cba6f0aeb89107a51

packages/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/869/commits/74ed5c6af3aa1339a460d6af1b62c6b49435ef1c

Alfred-Mountfield (approved):

PR #878, opened 2023-01-10
Add 'publish canary packages' manual GitHub action

CiaranMn:

Adds a GitHub workflow to trigger canary releases for any changesets captured in `.changeset/`

This workflow is manually triggered via the GitHub actions UI.

Assuming there are any changeset files in `.changeset/` on a branch, triggering the workflow for a branch will do the following:
1. Add a commit to the branch which converts the changesets into snapshot release data (CHANGELOG and package.json changes)
2. Reverts that commit so that the branch is left unchanged
3. Pushes the two commits so that they are in the remote (Changesets will access the commit being publsihed from)
4. Checkout the commit with the snapshot release data
5. Publishes canary versions from it

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543021352032/1203673362090847/f) _(internal)_

- I will update internal publishing docs.

## 🐾 Next steps

- Use this to publish canary versions on a branch off 0.3.

## ❓ How to test this?

1. You can check the git commands locally
2. The action must be in `main` before it is usable via GitHub

## Demo

You can see an example of the commits generated in this branch.


kachkaev (approved):
The diff makes sense, let’s see if it works!

When we have an opportunity to improve the pipeline we can potentially:

- split one bash task into several named ones to improve feedback (it will be easier to track progress and investigate failures)

- potentially rename the yaml to `release-canary.yml` to place two pipelines together.

PR #879, opened 2023-01-11
Add changesets for 0.3 package updates

CiaranMn:

Adds changeset files required for the introduction of the new specification(s) and type system.

Principles behind the version increments:
- Given that packages have either received breaking changes versus their `latest` release, or are receiving their first public advertisement (`@blockprotocol/type-system`, `@blockprotoocl/service`), every package will receive a minor bump
  - the one exception is `blockprotocol` which has only had a minor dependency bump unrelated to the Block Protocol as part of the `0.3` branch, and as a consequence of repo version consistency rather than any intended changes to it
- There is no attempted correlation between the version of a module (e.g. `graph`, which is now at `0.3`) and the package that implements it (e.g. `@blockprotocol/graph`). Any correlation is coincidental.

We should adopt some rule for future increases to avoid mental overhead in picking version numbers. While [semver](https://semver.org/) has no rules for versioning prior to major version 1 (`1.0.0`), we could for example adopt a policy that minor versions are breaking, patch versions are not.

## 🐾 Next steps

- Merge `0.3` into `main`
- Test packages using a `canary` release
- `latest` release


.changeset/small-dingos-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@blockprotocol/type-system": patch
+---
+
+add stdlib functions for interacting with subgraph. reorganize types
```
Alfred-Mountfield, line 5: Shouldn't this be the changelog for the graph package?

I think the changes we made in the type-system were a variety of updates to the schema representation such as "additionalProperties" or removing redundant fields on entities such as "requiredLinks"

Alfred-Mountfield:
```suggestion
update schema representations of ontology types to remove redundant fields, and add further validation
```

Alfred-Mountfield (approved):

PR #893, opened 2023-01-14
Add lib links to root README

nonparibus:

Adds convenience links from given paths to their actual locations.


kachkaev (approved):
These links will become invalid in old README versions if we change repo structure. That's a minor problem though, so I guess we stick with absolute links and not turn them to relative ones.

nonparibus:
> These links will become invalid in old README versions if we change repo structure. That's a minor problem though, so I guess we stick with absolute links and not turn them to relative ones.

Would their relative counterparts not also? AFAIK GitHub offers no kind of redirect.

kachkaev:
Relative links point to a file in the same commit as the markdown. Absolute links we’ve added point to tip of the future `main` branch.

PR #897, opened 2023-01-17
Fix workspace dirs in knip config

webpro:

Just saw that the knip config contains odd workspace keys. They should be the actual workspaces as defined in `package.json`. Not sure what happened, why it was like this. Anyway, here's a fix.

### DIFF
--- a/.knip.jsonc
+++ b/.knip.jsonc
@@ -28,18 +28,23 @@
     // thinking every workspace uses all ESLint plugins. That's why Knip's ESLint plugin is disabled everywhere else.
     // This resulted in two false positives: eslint-import-resolver-typescript for this workspace, and
     // @local/eslint-config for the others.
-    "packages/@local/eslint-config": {
+    "libs/@local/eslint-config": {
       "entry": [],
       "project": [],
       "eslint": "index.js"
     },
-    "packages/@local/tsconfig": {
+    "libs/@local/tsconfig": {
       "typescript": ["*.json", "!package.json"]
     },
-    "packages/@blockprotocol/core": {
+    "libs/@local/package-chores": {
+      "entry": "scripts/*.ts",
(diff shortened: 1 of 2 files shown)


kachkaev (dismissed):
Thanks for spotting! 👍

kachkaev (approved):
Dropping `-c .knip.jsonc` is great 👏 🚀

webpro:
> Dropping `-c .knip.jsonc` is great 👏 🚀

That was actually based on the discussion of my first PR here, great suggestion :)

kachkaev (commented):
Let’s merge this PR after https://github.com/blockprotocol/blockprotocol/pull/895

kachkaev:
#895 is merged, so we are on 1.6.1 now. @webpro does the output in https://github.com/blockprotocol/blockprotocol/actions/runs/3943196165/jobs/6747753419#step:8:39 make sense to you? Shall we merge?

webpro:
> #895 is merged, so we are on 1.6.1 now. @webpro does the output in https://github.com/blockprotocol/blockprotocol/actions/runs/3943196165/jobs/6747753419#step:8:39 look correct to you?

Short answer: Yes, that looks correct.

Long answer: there's a few things to decipher there. One example: the `playwright` dependency is unused from `apps/site/package.json`, but missing from `.github/actions/warm-up-repo/action.yml` and `.github/workflows/site-deployment.yml`. Even though `playwright` might really belong to the `apps/site` workspace, Knip has no way to understand this. Two options:

- Move the `playwright` dependency to the root workspace.
- Move the GA workflow files to `github-actions.config` of the `apps/site` workspace, and remove them from the root workspace config. Not sure if dependencies can be separated properly this way, though.

> Shall we merge?

Sure :)

PR #899, opened 2023-01-20
Rename various elements of the Graph Service

Alfred-Mountfield:

Over the course of developing this new release of the BP spec (and accompanying code) we've incrementally added quite a few new concepts. This PR will shortly be followed up by a number of PRs that greatly expand on the concept of "temporal versioning" within the Block Protocol. With the introduction of that code, we'd like to center on a common terminology that we can use consistently to encourage improved intuition.

This PR lays the foundation for that by renaming various fields to try and establish a consistent reasoned high-level picture of things. It also adds a few QoL improvements as a driveby (such as adding aliases for various Subgraph types rooted at different elements)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203756383156843/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- See commits for a discrete list, effort has been made to keep them self-contained, however here is the basic gist:
  - Elements of the abstract "datastore" will be associated with `RecordId`s. These should be viewed as immutable records, where `update` operations result in the creation of a new "record". This is made up of:
    - A base identifier (either `entityId` or `baseUri`), the identity of the element
    - An edition qualifier (either `editionId` or `number`), the discriminating identifier to find the specific row _for_ that element
  - We introduce the term "_revision_", which refers to a specific instance of some _thing_ within a specific `Subgraph`. In a `Subgraph` with temporal support (more information in a future PR), there may be multiple instances of a _thing_ within the `Subgraph`, these instances are associated with some versioning scheme, and each instance is referred to as a _revision_.
  - Elements contained within the `vertices` of a specific `Subgraph` will be identified by a `VertexId`. This is made up of:
    - a `baseId`, which is the _identity_ of the element, e.g. a specific ontology type or entity, and corresponds to the first level of nesting inside `vertices` and `edges`
    - a `revisionId`, which identifies the specific _revision_ of that element, e.g. the instance of the ontology type associated with a specific version, or the instance of the entity associated with a specific timestamp on that `Subgraph`'s variable temporal axis.

- The model described above will need to be propagated to users in some form, that is not done as part of this PR.

## ⚠️ Known issues

- Using the `EntityRootedSubgraph` alias in `useMockBlockDockContext` seems to breaks TS generic resolution in `getRoots` (see [this commit](https://github.com/blockprotocol/blockprotocol/pull/899/commits/e202f31fa363b991defa9d121cf051d751575bd9))
- The HTML block in `mock-block-dock` wasn't updated for 0.3. [Asana task](https://app.asana.com/0/1203358502199087/1203866892869243) _(internal)_

## 🐾 Next steps

- Introduce multi-axis temporal versioning to the Graph service
- Update MBD and accompanying packages to handle temporal versioning

## 🛡 What tests cover this?

- I'm unsure, at the very least `tsc` and `eslint`. It's also possible to check, through running it, that MBD has not been broken incidentally

## ❓ How to test this?

- Confirm that `lint:tsc` and `lint:eslint` pass for all packages _except_:
  - `site`
  - `block-scripts`
- Confirm that MBD still behaves as expected

## 📹 Demo

N/A

### DIFF
--- a/libs/mock-block-dock/src/mock-block-dock-context.tsx
+++ b/libs/mock-block-dock/src/mock-block-dock-context.tsx
@@ -1,7 +1,7 @@
 import { Message } from "@blockprotocol/core";
 import {
   EmbedderGraphMessageCallbacks,
-  EntityEditionId,
+  EntityRecordId,
   Subgraph,
   SubgraphRootTypes,
 } from "@blockprotocol/graph";
@@ -33,7 +33,7 @@ type MockBlockDockInfo = {
   logs: Message[];
   readonly: boolean;
   setDebugMode: Dispatch<SetStateAction<boolean>>;
-  setEntityEditionIdOfEntityForBlock: Dispatch<SetStateAction<EntityEditionId>>;
+  setEntityRecordIdOfEntityForBlock: Dispatch<SetStateAction<EntityRecordId>>;
   setLogs: Dispatch<SetStateAction<Message[]>>;
   setReadonly: Dispatch<SetStateAction<boolean>>;
   updateEntity: EmbedderGraphMessageCallbacks["updateEntity"];
--- a/libs/mock-block-dock/src/mock-block-dock-context.tsx
+++ b/libs/mock-block-dock/src/mock-block-dock-context.tsx
@@ -2,8 +2,8 @@ import { Message } from "@blockprotocol/core";
 import {
   EmbedderGraphMessageCallbacks,
   EntityRecordId,
-  EntityRootedSubgraph,
   Subgraph,
+  SubgraphRootTypes,
 } from "@blockprotocol/graph";
 import {
   createContext,
@@ -17,7 +17,7 @@ import {
 } from "react";
 
 type MockBlockDockInfo = {
-  blockEntitySubgraph: EntityRootedSubgraph;
+  blockEntitySubgraph: Subgraph<SubgraphRootTypes["entity"]>;
   blockInfo: {
     blockType: {
       entryPoint?: "react" | "html" | "custom-element" | string;
--- a/libs/mock-block-dock/src/mock-block-dock-context.tsx
+++ b/libs/mock-block-dock/src/mock-block-dock-context.tsx
@@ -2,8 +2,8 @@ import { Message } from "@blockprotocol/core";
 import {
   EmbedderGraphMessageCallbacks,
   EntityRecordId,
+  EntityRootType,
   Subgraph,
-  SubgraphRootTypes,
 } from "@blockprotocol/graph";
 import {
   createContext,
@@ -17,7 +17,7 @@ import {
 } from "react";
 
 type MockBlockDockInfo = {
-  blockEntitySubgraph: Subgraph<SubgraphRootTypes["entity"]>;
+  blockEntitySubgraph: Subgraph<EntityRootType>;
   blockInfo: {
     blockType: {
       entryPoint?: "react" | "html" | "custom-element" | string;
--- a/libs/mock-block-dock/src/mock-block-dock-context.tsx
+++ b/libs/mock-block-dock/src/mock-block-dock-context.tsx
@@ -2,8 +2,8 @@ import { Message } from "@blockprotocol/core";
 import {
   EmbedderGraphMessageCallbacks,
   EntityRecordId,
+  EntityRootedSubgraph,
   Subgraph,
-  SubgraphRootTypes,
 } from "@blockprotocol/graph";
 import {
   createContext,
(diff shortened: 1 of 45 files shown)


libs/mock-block-dock/src/mock-block-dock-context.tsx
```diff
@@ -17,7 +17,7 @@ import {
 } from "react";
 
 type MockBlockDockInfo = {
-  blockEntitySubgraph: EntityRootedSubgraph;
+  blockEntitySubgraph: Subgraph<SubgraphRootTypes["entity"]>;
```
Alfred-Mountfield, line 20: This really confuses me.. Any ideas @CiaranMn

CiaranMn: It looks like TypeScript does not or cannot infer the generic `RootType` from the `subgraph` argument when the declared type of the argument passed hides the generic behind an alias (as the `EntityRootedSubgraph` does). I had a try to google existing issues to do with this but didn't get anywhere.

Screenshot 2023-01-31 at 15 31 08

Because of that, I wonder if it is better not having these aliases (sorry) so that we don't cause ourselves or others this pain. A slight improvement over the previous position would be to split out the root types so that we could just do `Subgraph<EntityRootType>` rather than `Subgraph<SubgraphRootTypes["entity"]>`

```ts

export type EntityRootType = {
  vertexId: EntityVertexId;
  element: Entity;
};

export type EntityTypeRootType = {
  vertexId: OntologyTypeVertexId;
  element: EntityTypeWithMetadata;
};

export type DataTypeRootType = {
  vertexId: OntologyTypeVertexId;
  element: DataTypeWithMetadata;
};

export type PropertyTypeRootType = {
  vertexId: OntologyTypeVertexId;
  element: PropertyTypeWithMetadata;
};

export type SubgraphRootType =
  | EntityRootType
  | EntityTypeRootType
  | DataTypeRootType
  | PropertyTypeRootType;
```

Alfred-Mountfield: I was able to recreate this in a _somewhat_ simple form on a [TS playground](https://www.typescriptlang.org/play?#code/PTAEEsHMDsHsCcCmoAuALZ5rUfVBPAB0QGdQMkAaVDfAciVBIFdi8BjWaFecAN0QATUADMEoAIZM0CFKEQAPCQFtCAG0QAoFEWQBRDcsTcAKrrIBeUAG9NoVIpQAuG3fsRBLgEaxYGidDUbvZILGrOTDxYkG4AvpRuWCiIkLgutu4eLrBeAFaI7CgJmaHM4S7QzMpeuHGasZqaioSyBMSgBohGprqgVp3dKGbEJADaANaI+LAiHYbGQ+YAuo067QBivgA8JvIKydCCZAMLw4gAfH2u9uCeoCajAES3jyv2IKAAdN-1jR8mGBIyHYEmYQNAhHgOUMZCwoAABl4JPB4dpeiZHJtYFcsVsTj0Rk9kvtXuc0e0AJLcFK4LE47b4xaE57U1LwUmNTjQEhyJF4Kw7PYHI5zLqnXTnAAUYlgLlxJnOAEoXA9HqVwq8+pdrKAkChmPBoKAAAySMgBfAAblAsVAH3Aqnm3AkKHAXFAsGU4BQyWEnK4PIc+wAysR2OAROAhHKGU6maQiY5SVcdbcXCIJGpwQ0uYHiSgAIJqcASEgqzG+FNZUSZ7OaXNyJI0+ChgoRqN3XGMs5jFnJNnJqypu7WW05gON1m4IslsugKn92mVofV0c2-3cuQkMPtoQYnkARiufMl+db4cjQkV1o+nHgSEKanwbWQPN40BiDYcPJnpaPVhPfNfxIa87TALA7wfFAnxfSJ30gUAAB9QEqapai-bc20vQR9xQAAmY9kUlJs2XPXdBFA28ECgmC1mQVCangesJ2-Qti1LAiAKIkjp3YkCb3A6BIIKaDnzouDoiQlCqkYoA)

I'm going to ask on the TS discord and see if anyone has ideas. Your alternative posed here does seem like an improvement on what we had before though, I'll probably go with that.

Alfred-Mountfield: Seems that this is to do with how TS evaluates types and its reliance on structural typing ([Discord thread](https://discord.com/channels/508357248330760243/1070295766930817024/1070295766930817024)).

I've opted to go with the suggestion you gave.

CiaranMn (approved):

Alfred-Mountfield:
Shouldn't use trunk merge on PRs targetting branches it seems 😅

PR #906, opened 2023-01-24
Fix canary block templates. Commit auto-generated types

CiaranMn:

1. Fix some bugs in the block templates (see comments on diff)

2. Commit the auto-generated types for blocks, rather than ignoring them – this removes type generation from an external URI as a dependency in the build process. Types are immutable and should not change. I've set `linguist-generated=true` to collapse them in the diff by default

## 🔗 Related links

- [Asana task for committing types](https://app.asana.com/0/1203358502199087/1203774997009906/f) _(internal)_
- [Bug report on block template](https://hashintel.slack.com/archives/C02LG39FJAU/p1674582285578809) _(internal)_

## 🐾 Next steps

– Update #879 and publish new canary versions

### DIFF
--- a/libs/@blockprotocol/graph/src/codegen/entity-type-to-typescript.ts
+++ b/libs/@blockprotocol/graph/src/codegen/entity-type-to-typescript.ts
@@ -17,8 +17,7 @@ import {
 import { hardcodedBpTypes } from "./hardcoded-bp-types.js";
 import { fetchTypeAsJson, typedEntries } from "./shared.js";
 
-const bannerComment = (uri: string, depth: number) => `/* eslint-disable */
-/**
+const bannerComment = (uri: string, depth: number) => `/**
  * This file was automatically generated – do not edit it.
  * @see ${uri} for the root JSON Schema these types were generated from
  * Types for link entities and their destination were generated to a depth of ${depth} from the root
@@ -218,8 +217,9 @@ const _jsonSchemaToTypeScript = async (
       generateLinkEntityAndRightEntityDefinition({
         sourceEntityTypeName: typeName,
         linkEntityTypeName: linkEntityType.typeName,
-        rightEntityTypeNames: (
-          rightEntityTypes ?? [{ typeName: "Entity" }]
+        rightEntityTypeNames: (rightEntityTypes.length > 0
+          ? rightEntityTypes
+          : [{ typeName: "Entity" }]
         ).map((type) => type.typeName),
       });
 
--- a/libs/@blockprotocol/graph/src/custom-element.ts
+++ b/libs/@blockprotocol/graph/src/custom-element.ts
@@ -3,7 +3,6 @@ import { LitElement } from "lit";
 import {
   BlockGraphProperties,
   Entity,
-  EntityEditionId,
   EntityPropertiesObject,
   GraphBlockHandler,
   LinkEntityAndRightEntity,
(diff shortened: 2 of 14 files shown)


libs/@blockprotocol/graph/src/codegen/entity-type-to-typescript.ts
```diff
@@ -218,8 +217,9 @@ const _jsonSchemaToTypeScript = async (
       generateLinkEntityAndRightEntityDefinition({
         sourceEntityTypeName: typeName,
         linkEntityTypeName: linkEntityType.typeName,
-        rightEntityTypeNames: (
-          rightEntityTypes ?? [{ typeName: "Entity" }]
+        rightEntityTypeNames: (rightEntityTypes.length > 0
+          ? rightEntityTypes
+          : [{ typeName: "Entity" }]
```
CiaranMn, line 222: Bug fix: `rightEntityTypes` is an empty array if there is no constraint on the expected type

libs/@blockprotocol/graph/src/custom-element.ts
```diff
@@ -43,10 +42,17 @@ export abstract class BlockElementBase<
...
      * @see https://blockprotocol.org/docs/spec/graph-service#message-definitions for a full list
      */
     graph: { type: Object },
+
+    /**
+     * These are properties derived from the block subgraph – the root entity, and those linked from it
+     * // @see https://lit.dev/docs/components/properties/#internal-reactive-state
+     */
+    blockEntity: { state: true },
+    linkedEntities: { state: true },
```
CiaranMn, line 55: Bug fix: HTML template uses derived properties, and this is required for Lit to know the element must be rendered

libs/@blockprotocol/graph/src/graph-block-handler.ts
```diff
@@ -97,7 +98,9 @@ export class GraphBlockHandler
 
   // @todo automate creation of these methods from graph-service.json and types.ts
 
-  createEntity({ data }: { data?: CreateEntityData }) {
+  createEntity<
+    ValidProperties extends EntityPropertiesObject = EntityPropertiesObject,
+  >({ data }: { data?: CreateEntityData & { properties: ValidProperties } }) {
```
CiaranMn, line 103: Improvement: allow uses of the library to pass a generic to enforce a certain properties object (useful to get autocomplete and type safety when passing things in for an update, especially given keys are long URIs now)

libs/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
```diff
@@ -48,7 +48,7 @@ export const buildSubgraph = (
       ),
   );
 
-  if (missingRoots) {
+  if (missingRoots.length > 0) {
```
CiaranMn, line 51: Bug fix: `missingRoots` is the result of a `filter` (this was left over from when it was a `find`)

libs/block-template-react/src/app.tsx
```diff
@@ -66,9 +65,10 @@ export const App: BlockComponent<RootEntity> = ({
 
   const entityId = blockEntity.metadata.editionId.baseId;
 
-  const simpleBlockEntity = addSimpleAccessors(blockEntity);
+  const titleKey: keyof RootEntity["properties"] =
+    "https://alpha.hash.ai/@hash/types/property-type/title/";
 
-  const { title } = simpleBlockEntity.properties;
+  const title = blockEntity.properties[titleKey];
```
CiaranMn, line 71: `addSimpleAccessors` has been removed

libs/create-block-app/create-block-app.js
```diff
@@ -131,7 +131,6 @@ const usage = commandLineUsage(helpSections);
   } else {
     packageJson.blockprotocol.displayName = blockName;
     packageJson.blockprotocol.name = slugifiedBlockName;
-    packageJson.postinstall = "yarn codegen";
```
CiaranMn, line 134: we don't need this now the types are committed

kachkaev (approved):

PR #919, opened 2023-01-30
Better 'Object' type. Rename codegen script in templates

CiaranMn:

1. Overwrites the default generated type for an opaque object with our own `JsonObject` type
  - this involves re-exporting `JsonObject` in `@blockprotocol/graph` to avoid users having to add `@blockprotocol/core` as a separate dependency (and therefore potentially becoming out of sync)
2. Renames `codegen` in templates to `schema` as a temporary measure. The generated type output is non-deterministic (types may appear in a different order), meaning that any run of Turborepo can result in a diff if the template's `codegen` scripts are run (internal Asana [task](https://app.asana.com/0/1203179076056209/1203843256013361/f) to figure out a permanent fix for this)

### DIFF
--- a/libs/@blockprotocol/graph/src/codegen/entity-type-to-typescript.ts
+++ b/libs/@blockprotocol/graph/src/codegen/entity-type-to-typescript.ts
@@ -179,6 +179,12 @@ const _jsonSchemaToTypeScript = async (
     "export type Link =",
   );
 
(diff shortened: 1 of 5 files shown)


yusufkinatas (approved):
Didn't tested it on my local but code looks good to me!

PR #922, opened 2023-01-30
Make Þ services available before element is available

nathggns:

Allow constructing `ServiceHandler` implementations before the element is available (but not core service). Messages will then be queued. 

## 🔗 Related links

- [Ensure that services are always defined](https://app.asana.com/0/1203179076056186/1203474494377437/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Allow creating a `ServiceHandler` without an element. This means coreHandler is not always available now, making `coreHandler` no longer readonly, but reimplementing that behaviour with a private field and a getter
- New `coreQueue` field and `processCoreCallbacks` method on `ServiceHandler` to defer commands (sendMessage, registerCallback) until the core handler is available
- We don't register services when handlers are constructed if there is no element – there is a new `initialize` method to do this 
- `sendMessage` in `ServiceHandler` is now deferred until the core handler is available
- All handlers for graph/hook service classes updated to handle not immediately having an element (as well as type signatures for hooks), and to immediately create the service instead of waiting until a useEffect 

I don't *think* these functions are documented yet.

## ⚠️ Known issues

I haven't changed the core service yet. It's much more complicated, and is not in any case accessible via React, which is the main motivation for this change. We can only have one core service per block, so we have to have an element to store it against, and there is a dispatching/listening element. Meanwhile graph/hook services mostly use the element to get the core service, and they can just queue messages themselves. 

## 🐾 Next steps

- Consider whether there is a need to do this in the core service (@CiaranMn) 

## 🛡 What tests cover this?

N/A


libs/@blockprotocol/core/src/service-handler.ts
```diff
@@ -10,9 +15,19 @@ import { GenericMessageCallback, MessageContents, MessageData } from "./types";
  */
 export abstract class ServiceHandler {
   /** the CoreHandler this service is registered with, for passing messages via */
-  readonly coreHandler: CoreHandler;
+  private _coreHandler: CoreHandler | null = null;
+
+  public get coreHandler(): CoreHandler | null {
+    return this._coreHandler;
+  }
```
CiaranMn, line 22: Why is this necessary? It looks like `coreHandler` is only used in this class, and could remain `private`

libs/@blockprotocol/core/src/service-handler.ts
```diff
@@ -48,27 +63,56 @@ export abstract class ServiceHandler {
...
+    if (!coreHandler) {
+      throw new Error("Could not initialize – missing core handler");
+    }
+
+    coreHandler.initialize();
+
+    this.processCoreCallbacks();
+  }
+
+  registerService(element: HTMLElement) {
```
CiaranMn, line 94: I think this can be private
```suggestion
  private registerService(element: HTMLElement) {
```

libs/@blockprotocol/core/src/service-handler.ts
```diff
@@ -112,11 +156,27 @@ export abstract class ServiceHandler {
...
-    });
+    this.coreQueue.push((coreHandler) =>
+      coreHandler.registerCallback({
+        callback,
+        messageName,
+        serviceName: this.serviceName,
+      }),
+    );
+
+    this.processCoreCallbacks();
```
CiaranMn, line 167: I think it's a little confusing that we have `registerCallback` and `processCoreCallbacks` and they refer to completely different things – maybe this could be `processCoreQueue` or `processCoreActions` or something else?

libs/@blockprotocol/graph/src/graph-block-handler.ts
```diff
@@ -54,13 +54,17 @@ export class GraphBlockHandler
...
     super({ element, serviceName: "graph", sourceType: "block" });
+
     if (callbacks) {
       this.registerCallbacks(callbacks);
     }
-    this.coreHandler.initialize();
+
+    if (element) {
+      this.initialize(element);
+    }
```
CiaranMn, line 67: I wonder if we could move these steps into the parent `ServiceHandler` class's constructor – if it accepted callbacks in the constructor, we wouldn't need to add these same steps in all its implementations (which we'll have to do for every new service if we keep it here). We could keep this constructor for better typing of the callbacks.

libs/@blockprotocol/core/src/service-handler.ts
```diff
@@ -48,27 +63,56 @@ export abstract class ServiceHandler {
...
-    this.element = element ?? null;
     this.serviceName = serviceName;
     this.sourceType = sourceType;
 
+    if (element) {
+      this.registerService(element);
+    }
+  }
+
+  initialize(element: HTMLElement) {
```
CiaranMn, line 78: As this is a method that is exposed to end-users, I think we should document this – here's a starting point
```suggestion
  /**
   * You only need to use this if you are constructing a service directly.
   * You do not need to use it if you're using a React hook or block template.
   *
   * This initializes a service with the element it will listen for messages on,
   * and must be called if the service was constructed without an element.
   */
  initialize(element: HTMLElement) {
```

libs/@blockprotocol/core/src/service-handler.ts
```diff
@@ -48,27 +63,56 @@ export abstract class ServiceHandler {
...
+    if (!this.element) {
+      this.registerService(element);
+    }
```
CiaranMn, line 81: It is possible that this is called with an element different to that provided in the constructor, which we might want to at least throw an error on here.

libs/@blockprotocol/core/src/types.ts
```diff
@@ -237,3 +237,4 @@ export type HtmlBlockDefinition = {
    */
   source?: string;
 };
+export type CoreHandlerCallback = (handler: CoreHandler) => void;
```
CiaranMn, line 240:
```suggestion

export type CoreHandlerCallback = (handler: CoreHandler) => void;
```

libs/@blockprotocol/graph/src/graph-embedder-handler.ts
```diff
@@ -38,7 +38,7 @@ export class GraphEmbedderHandler
     blockEntity?: Entity;
     blockGraph?: BlockGraph;
     callbacks?: Partial<EmbedderGraphMessageCallbacks>;
-    element: HTMLElement;
+    element?: HTMLElement | null;
     entityTypes?: EntityType[];
     linkedAggregations?: LinkedAggregations;
     readonly?: boolean;
```
CiaranMn, line 44: `initialize` isn't called in the constructor here, regardless of whether there's an element or not, which forces people to call `initialize` even if they originally provided an element.

I believe it originally was not present because the core embedder handler doesn't do anything on initialization, but it's doing more things now so probably worth calling – or moving the `initialize` calls into the super class so that we don't have to worry about doing it in all its implementations.

CiaranMn (changes requested):
Thanks @nathggns – this is a very exciting change! 

I've tested this and it works well. I have a few suggestions which I think will make our life easier in future when introducing new services and features.

libs/@blockprotocol/core/src/service-handler.ts
nathggns: It was public readonly before and I didn't want to make a breaking change if it wasn't necessary. Can remove if you think it being private is fine.

libs/@blockprotocol/core/src/types.ts
nathggns: WebStorm has started messing with spacing between lines 😅 thank you!

libs/@blockprotocol/graph/src/graph-block-handler.ts
nathggns: I'm now calling initialize and registerCallbacks in the super constructor

CiaranMn (approved):
Excellent! 🥳

PR #923, opened 2023-01-30
Track block downloads, expose last week's count via API

CiaranMn:

1. Record download events for block source files when they go via the Cloudflare R2 worker
2. Expose a count of the last week's downloads via the API

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543026679380/1203624210273373/f) _(internal)_

## 🔍 What does this change?

- Update the Cloudflare worker to record an event when a block's source entry point is requested
- Expose the last week's download count on the API return

## ⚠️ Known issues

- This only works for blocks which have been published via the CLI and are therefore (a) in the db and (b) served from R2, not those in the `hub/` folder which are served via Vercel (and whose metadata is in the repo, not the db)
- Even for blocks which are stored in R2, this will not count downloads which are cached in the browser (now) or at Cloudflare's edge (once we add caching in Cloudflare)

## 🐾 Next steps

- Deploy the production wrangler (only the development one deployed so far)
- Expose the weekly download count in the frontend

## ❓ How to test this?

1. Cumbersome as it involves publishing blocks to R2, hitting their source endpoint and then checking if the weekly download count has been incremented

## 📹 Demo

Screenshot 2023-01-30 at 23 02 41

### DIFF
--- a/apps/site/src/lib/api/blocks/db.ts
+++ b/apps/site/src/lib/api/blocks/db.ts
@@ -1,15 +1,44 @@
 import { ExpandedBlockMetadata } from "../../blocks";
 import { connectToDatabase } from "../mongodb";
-import { blocksDbCollectionName } from "./shared";
+import { blockDownloadsCollectionName, blocksDbCollectionName } from "./shared";
 
-const queryOptions = { projection: { _id: 0 } };
+const defaultProjection = { _id: 0 };
+
+const weeklyDownloadCountAggregationStage = [
+  {
+    $lookup: {
+      from: blockDownloadsCollectionName,
+      let: { id: "$_id" },
+      pipeline: [
+        {
+          $match: {
+            $expr: { $eq: ["$$id", "$blockId"] },
+            downloadedAt: {
+              $gte: new Date(
+                new Date().valueOf() - 7 * 60 * 60 * 24 * 1000,
+              ).toISOString(),
+            },
+          },
+        },
+      ],
+      as: "weeklyDownloads",
+    },
+  },
+  { $addFields: { downloads: { weekly: { $size: "$weeklyDownloads" } } } },
+  { $project: { weeklyDownloads: 0, ...defaultProjection } },
+];
 
 export const getDbBlocks = async (filter: { shortname?: string }) => {
   const { db } = await connectToDatabase();
 
   return db
-    .collection<ExpandedBlockMetadata>(blocksDbCollectionName)
-    .find(filter.shortname ? { author: filter.shortname } : {}, queryOptions)
+    .collection(blocksDbCollectionName)
+    .aggregate<ExpandedBlockMetadata>([
+      {
+        $match: filter.shortname ? { author: filter.shortname } : {},
+      },
+      ...weeklyDownloadCountAggregationStage,
+    ])
     .toArray();
 };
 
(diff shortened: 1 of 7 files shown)


apps/site/src/components/pages/hub/block-data-container.tsx
```diff
@@ -197,7 +197,7 @@ export const BlockDataContainer: FunctionComponent<BlockDataContainerProps> = ({
           }}
         >
           <Tab label="Preview" />
-          <Tab label="Source Code" />
+          <Tab label="Block Data" />
```
CiaranMn, line 200: Drive-by: this is the name for the tab in mobile which contains the block example properties and block schema. I think 'Source Code' is an inappropriate name for these things so I've renamed it.

apps/site/src/lib/blocks.ts
```diff
@@ -256,7 +262,15 @@ export const retrieveBlockFileContent = async ({
...
-    : await fetch(metadataSchemaUrl).then((response) => response.json());
+    : await fetch(metadataSchemaUrl)
+        .then((response) => response.json())
+        .catch((err) => {
+          // eslint-disable-next-line no-console -- intentional log to flag problem without tanking site
+          console.error(
+            `Could not fetch and parse schema at ${metadataSchemaUrl} for block ${pathWithNamespace}: ${err}`,
+          );
+          return { title: "Unparseable schema" };
+        });
```
CiaranMn, line 273: Drive-by: I had tests failing first time round and it was difficult to debug because there was a `500` error and the app crashing when a schema was unreachable. This handles the error a little better.

infra/cloudflare-r2-wrangler/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "cloudflare-r2-wrangler",
```
kachkaev, line 2: WDYT of using

```suggestion
  "name": "@infra/cloudflare-r2-wrangler",
```
?

This will reduce the chance of clashes with existing npm packages.

package.json
```diff
@@ -8,7 +8,8 @@
       "libs/@blockprotocol/*",
       "libs/@local/*",
       "libs/@local/internal-api-client-generator/typescript",
-      "libs/*"
+      "libs/*",
+      "infra/*"
```
kachkaev, line 12: Alphabetical sorting 🙏

apps/site/src/lib/blocks.ts
kachkaev: Optional: It'd be great to avoid a mix of `async` / `await` syntax with `then` / `catch`.

apps/site/src/lib/api/blocks/db.ts
```diff
@@ -1,15 +1,44 @@
...
+            },
+          },
+        },
+      ],
+      as: "weeklyDownloads",
+    },
+  },
+  { $addFields: { downloads: { weekly: { $size: "$weeklyDownloads" } } } },
+  { $project: { weeklyDownloads: 0, ...defaultProjection } },
+];
```
kachkaev, line 29: Does there exist a type in Mongo package which we can use here with `satisfies`?

apps/site/src/lib/blocks.ts
CiaranMn: Sure – https://github.com/blockprotocol/blockprotocol/pull/923/commits/111d3bebf83071ffdab96ccfbcf9814c049d9760

infra/cloudflare-r2-wrangler/package.json
CiaranMn: Yep – https://github.com/blockprotocol/blockprotocol/pull/923/commits/62b1d290aca43f7fd7ade313a0538a5d7046009c

package.json
CiaranMn: https://github.com/blockprotocol/blockprotocol/pull/923/commits/62b1d290aca43f7fd7ade313a0538a5d7046009c

kachkaev (approved):

PR #924, opened 2023-01-31
Introduce primitives for bi-temporal versioning

Alfred-Mountfield:

As expressed in previous PRs, we intend to add the foundations of support for bitemporal datastores within the BP as part of the 0.3 release. This PR introduces relevant primitives that will be used to expand the `Subgraph` type and associated queries.

Notably, performing queries across a temporal datastore requires a lot of interfacing with `Interval`s of time. That is the focus of this PR.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203765721937276/f) _(internal)_

## 🚫 Blocked by

- [x] #899 

## 🔍 What does this change?

See commits and in-code documentation, effort has been made to keep them self-contained and explanatory.

- This PR does not expose these types on the package's public API, that will need to be done, and we will need to think about how to appropriately write user documentation to introduce them to these concepts. The stdlib on the intervals has (hopefully) comprehensive documentation however.

## ⚠️ Known issues

N/A

## 🐾 Next steps

- Extend `Entity` and the `Subgraph` type to support temporal axes
- Migrate the stdlib and mock-block-dock to handle temporal versioning

## 🛡 What tests cover this?

- I'm unsure, at the very least `tsc` and `eslint`. It's also possible to check, through running it, that MBD has not been broken incidentally
- I wanted to write a proper suite of tests to check this, but made the uncomfortable compromise at the moment to defer that until later on, to unblock some further work. We intend to write a test suite for the `Subgraph` library, and I'd encourage us to write tests for temporal operations at the same time. 
 - It's worth noting that these temporal operations were ported from some logic we had implemented [for the HASH graph query layer ](https://github.com/hashintel/hash/blob/795bcf1e4f73c2c186da576b314d0a3ffc5bfa8d/apps/hash-graph/lib/interval-ops/src/interval.rs#L11), which are heavily tested, so confidence is fairly high that these will be correct.

## ❓ How to test this?

- Confirm that `lint:tsc` and `lint:eslint` pass for all packages _except_:
  - `site`
  - `block-scripts`
- Confirm that MBD still behaves as expected

## 📹 Demo

N/A, code is unused at present

### DIFF
--- a/libs/@blockprotocol/graph/src/shared.ts
+++ b/libs/@blockprotocol/graph/src/shared.ts
@@ -1,0 +1,28 @@
+type TupleEntry<
+  T extends readonly unknown[],
+  I extends unknown[] = [],
+  R = never,
+> = T extends readonly [infer Head, ...infer Tail]
+  ? TupleEntry<Tail, [...I, unknown], R | [`${I["length"]}`, Head]>
+  : R;
+type ObjectEntry<T extends {}> = T extends object
+  ? { [K in keyof T]: [K, Required<T>[K]] }[keyof T] extends infer E
+    ? E extends [infer K, infer V]
+      ? K extends string | number
+        ? [`${K}`, V]
+        : never
+      : never
+    : never
+  : never;
+// Source: https://dev.to/harry0000/a-bit-convenient-typescript-type-definitions-for-objectentries-d6g
+type Entry<T extends {}> = T extends readonly [unknown, ...unknown[]]
+  ? TupleEntry<T>
+  : T extends ReadonlyArray<infer U>
+  ? [`${number}`, U]
+  : ObjectEntry<T>;
+
+// @todo deduplicate this and packages/mock-block-dock/src/util.ts
+/** `Object.entries` analogue which returns a well-typed array */
+export function typedEntries<T extends {}>(object: T): ReadonlyArray<Entry<T>> {
+  return Object.entries(object) as unknown as ReadonlyArray<Entry<T>>;
+}
--- a/libs/@blockprotocol/graph/src/shared.ts
+++ b/libs/@blockprotocol/graph/src/shared.ts
@@ -5,6 +5,7 @@ type TupleEntry<
 > = T extends readonly [infer Head, ...infer Tail]
   ? TupleEntry<Tail, [...I, unknown], R | [`${I["length"]}`, Head]>
   : R;
+
 type ObjectEntry<T extends {}> = T extends object
   ? { [K in keyof T]: [K, Required<T>[K]] }[keyof T] extends infer E
     ? E extends [infer K, infer V]
@@ -14,7 +15,7 @@ type ObjectEntry<T extends {}> = T extends object
       : never
     : never
   : never;
-// Source: https://dev.to/harry0000/a-bit-convenient-typescript-type-definitions-for-objectentries-d6g
+
 type Entry<T extends {}> = T extends readonly [unknown, ...unknown[]]
   ? TupleEntry<T>
   : T extends ReadonlyArray<infer U>
@@ -22,7 +23,10 @@ type Entry<T extends {}> = T extends readonly [unknown, ...unknown[]]
   : ObjectEntry<T>;
 
 // @todo deduplicate this and packages/mock-block-dock/src/util.ts
-/** `Object.entries` analogue which returns a well-typed array */
+/** `Object.entries` analogue which returns a well-typed array
+ *
+ * Source: https://dev.to/harry0000/a-bit-convenient-typescript-type-definitions-for-objectentries-d6g
+ */
 export function typedEntries<T extends {}>(object: T): ReadonlyArray<Entry<T>> {
   return Object.entries(object) as unknown as ReadonlyArray<Entry<T>>;
 }
--- a/libs/@blockprotocol/graph/src/stdlib/bound.ts
+++ b/libs/@blockprotocol/graph/src/stdlib/bound.ts
@@ -4,58 +4,60 @@ import {
 } from "../types/temporal-versioning.js";
 
 export const compareBounds = (
-  lhs: TemporalBound,
-  rhs: TemporalBound,
-  lhsType: keyof NonNullTimeInterval,
-  rhsType: keyof NonNullTimeInterval,
+  left: TemporalBound,
+  right: TemporalBound,
+  leftType: keyof NonNullTimeInterval,
+  rightType: keyof NonNullTimeInterval,
 ): number => {
   // If the bound values are not equal, then the bound with the lower value is less than the bound with the higher
   // value.
-  if (lhs.kind !== "unbounded" && rhs.kind !== "unbounded") {
-    if (lhs.limit !== rhs.limit) {
-      return lhs.limit.localeCompare(rhs.limit);
+  if (left.kind !== "unbounded" && right.kind !== "unbounded") {
+    if (left.limit !== right.limit) {
+      return left.limit.localeCompare(right.limit);
     }
   }
 
-  if (lhs.kind === rhs.kind && lhsType === rhsType) {
+  if (left.kind === right.kind && leftType === rightType) {
     return 0;
   }
 
   if (
-    (lhs.kind === "unbounded" && lhsType === "start") ||
-    (rhs.kind === "unbounded" && rhsType === "end") ||
-    (lhs.kind === "exclusive" &&
-      rhs.kind === "exclusive" &&
-      lhsType === "end" &&
-      rhsType === "start") ||
-    (lhs.kind === "exclusive" &&
-      rhs.kind === "inclusive" &&
-      lhsType === "end") ||
-    (lhs.kind === "inclusive" &&
-      rhs.kind === "exclusive" &&
-      rhsType === "start")
+    (left.kind === "unbounded" && leftType === "start") ||
+    (right.kind === "unbounded" && rightType === "end") ||
+    (left.kind === "exclusive" &&
+      right.kind === "exclusive" &&
+      leftType === "end" &&
+      rightType === "start") ||
+    (left.kind === "exclusive" &&
+      right.kind === "inclusive" &&
+      leftType === "end") ||
+    (left.kind === "inclusive" &&
+      right.kind === "exclusive" &&
+      rightType === "start")
   ) {
     return -1;
   }
 
   if (
-    (lhs.kind === "unbounded" && lhsType === "end") ||
-    (rhs.kind === "unbounded" && rhsType === "start") ||
-    (lhs.kind === "exclusive" &&
-      rhs.kind === "exclusive" &&
-      lhsType === "start" &&
-      rhsType === "end") ||
-    (lhs.kind === "exclusive" &&
-      rhs.kind === "inclusive" &&
-      lhsType === "start") ||
-    (lhs.kind === "inclusive" && rhs.kind === "exclusive" && rhsType === "end")
+    (left.kind === "unbounded" && leftType === "end") ||
+    (right.kind === "unbounded" && rightType === "start") ||
+    (left.kind === "exclusive" &&
+      right.kind === "exclusive" &&
+      leftType === "start" &&
+      rightType === "end") ||
+    (left.kind === "exclusive" &&
+      right.kind === "inclusive" &&
+      leftType === "start") ||
+    (left.kind === "inclusive" &&
+      right.kind === "exclusive" &&
+      rightType === "end")
   ) {
     return 1;
   }
   throw new Error(
     `Implementation error, failed to compare bounds.\nLHS: ${JSON.stringify(
-      lhs,
-    )}\nRHS: ${JSON.stringify(rhs)}`,
+      left,
+    )}\nRHS: ${JSON.stringify(right)}`,
   );
 };
 
@@ -65,18 +67,18 @@ export const compareBounds = (
  *
  * As such, two inclusive (or two exclusive) bounds with the same limit are *not* adjacent.
  *
- * @param lhs - The first bound of the comparison (order is unimportant)
- * @param rhs - The second bound of the comparison
+ * @param left - The first bound of the comparison (order is unimportant)
+ * @param right - The second bound of the comparison
  */
 export const boundIsAdjacentTo = (
-  lhs: TemporalBound,
-  rhs: TemporalBound,
+  left: TemporalBound,
+  right: TemporalBound,
 ): boolean => {
   if (
-    (lhs.kind === "inclusive" && rhs.kind === "exclusive") ||
-    (lhs.kind === "exclusive" && rhs.kind === "inclusive")
+    (left.kind === "inclusive" && right.kind === "exclusive") ||
+    (left.kind === "exclusive" && right.kind === "inclusive")
   ) {
-    return lhs.limit === rhs.limit;
+    return left.limit === right.limit;
   } else {
     return false;
   }
--- a/libs/@blockprotocol/graph/src/stdlib/bound.ts
+++ b/libs/@blockprotocol/graph/src/stdlib/bound.ts
@@ -1,0 +1,83 @@
+import {
+  NonNullTimeInterval,
+  TemporalBound,
+} from "../types/temporal-versioning.js";
+
+export const compareBounds = (
+  lhs: TemporalBound,
+  rhs: TemporalBound,
+  lhsType: keyof NonNullTimeInterval,
+  rhsType: keyof NonNullTimeInterval,
+): number => {
+  // If the bound values are not equal, then the bound with the lower value is less than the bound with the higher
+  // value.
+  if (lhs.kind !== "unbounded" && rhs.kind !== "unbounded") {
+    if (lhs.limit !== rhs.limit) {
+      return lhs.limit.localeCompare(rhs.limit);
+    }
+  }
+
+  if (lhs.kind === rhs.kind && lhsType === rhsType) {
+    return 0;
+  }
+
+  if (
+    (lhs.kind === "unbounded" && lhsType === "start") ||
+    (rhs.kind === "unbounded" && rhsType === "end") ||
+    (lhs.kind === "exclusive" &&
+      rhs.kind === "exclusive" &&
+      lhsType === "end" &&
+      rhsType === "start") ||
+    (lhs.kind === "exclusive" &&
+      rhs.kind === "inclusive" &&
+      lhsType === "end") ||
+    (lhs.kind === "inclusive" &&
+      rhs.kind === "exclusive" &&
+      rhsType === "start")
+  ) {
+    return -1;
+  }
+
+  if (
+    (lhs.kind === "unbounded" && lhsType === "end") ||
+    (rhs.kind === "unbounded" && rhsType === "start") ||
+    (lhs.kind === "exclusive" &&
+      rhs.kind === "exclusive" &&
+      lhsType === "start" &&
+      rhsType === "end") ||
+    (lhs.kind === "exclusive" &&
+      rhs.kind === "inclusive" &&
+      lhsType === "start") ||
+    (lhs.kind === "inclusive" && rhs.kind === "exclusive" && rhsType === "end")
+  ) {
+    return 1;
+  }
+  throw new Error(
+    `Implementation error, failed to compare bounds.\nLHS: ${JSON.stringify(
+      lhs,
+    )}\nRHS: ${JSON.stringify(rhs)}`,
+  );
+};
+
+/**
+ * Checks whether two given temporal bounds are adjacent to one another, where adjacency is defined as being next to one
+ * another on the timeline, without any points between, *and where they are not overlapping*.
+ *
+ * As such, two inclusive (or two exclusive) bounds with the same limit are *not* adjacent.
+ *
+ * @param lhs - The first bound of the comparison (order is unimportant)
+ * @param rhs - The second bound of the comparison
+ */
+export const boundIsAdjacentTo = (
+  lhs: TemporalBound,
+  rhs: TemporalBound,
+): boolean => {
+  if (
+    (lhs.kind === "inclusive" && rhs.kind === "exclusive") ||
+    (lhs.kind === "exclusive" && rhs.kind === "inclusive")
+  ) {
+    return lhs.limit === rhs.limit;
+  } else {
+    return false;
+  }
+};
--- a/libs/@blockprotocol/graph/src/stdlib/bound.ts
+++ b/libs/@blockprotocol/graph/src/stdlib/bound.ts
@@ -70,7 +70,7 @@ export const compareBounds = (
  * @param left - The first bound of the comparison (order is unimportant)
  * @param right - The second bound of the comparison
  */
-export const boundIsAdjacentTo = (
+export const boundIsAdjacentToBound = (
   left: TemporalBound,
   right: TemporalBound,
 ): boolean => {
--- a/libs/@blockprotocol/graph/src/stdlib/interval.ts
+++ b/libs/@blockprotocol/graph/src/stdlib/interval.ts
@@ -11,12 +11,12 @@ import { boundIsAdjacentTo, compareBounds } from "./bound.js";
  * being next to one another on the timeline, without any points between, *and where they are not overlapping*. Thus,
  * if adjacent, the two intervals should span another given interval.
  *
- * @param lhs - The first interval of the comparison (order is unimportant)
- * @param rhs - The second interval of the comparison
+ * @param left - The first interval of the comparison (order is unimportant)
+ * @param right - The second interval of the comparison
  */
 export const intervalIsAdjacentToInterval = (
-  lhs: NonNullTimeInterval,
-  rhs: NonNullTimeInterval,
+  left: NonNullTimeInterval,
+  right: NonNullTimeInterval,
 ): boolean => {
   /*
    Examples           |     1     |     2     |     3     |     4     |     5
@@ -27,21 +27,21 @@ export const intervalIsAdjacentToInterval = (
    Contains Interval  |   true    |   true    |   false   |   false   |   false
    */
   return (
-    boundIsAdjacentTo(lhs.end, rhs.start) ||
-    boundIsAdjacentTo(lhs.start, rhs.end)
+    boundIsAdjacentTo(left.end, right.start) ||
+    boundIsAdjacentTo(left.start, right.end)
   );
 };
 
 /**
- * Returns whether or not the `rhs` {@link NonNullTimeInterval} is *completely contained* within the `lhs`
+ * Returns whether or not the `right` {@link NonNullTimeInterval} is *completely contained* within the `left`
  * {@link NonNullTimeInterval}.
  *
- * @param {NonNullTimeInterval} lhs - Checked if it contains the other
- * @param {NonNullTimeInterval} rhs - Checked if it's contained _within_ the other
+ * @param {NonNullTimeInterval} left - Checked if it contains the other
+ * @param {NonNullTimeInterval} right - Checked if it's contained _within_ the other
  */
 export const intervalContainsInterval = (
-  lhs: NonNullTimeInterval,
-  rhs: NonNullTimeInterval,
+  left: NonNullTimeInterval,
+  right: NonNullTimeInterval,
 ): boolean => {
   /*
    Examples           |     1     |     2     |     3     |     4     |     5     |     6
@@ -52,8 +52,8 @@ export const intervalContainsInterval = (
    Contains Interval  |   true    |   true    |   true    |   true    |   false   |   false
    */
   return (
-    compareBounds(lhs.start, rhs.start, "start", "start") <= 0 &&
-    compareBounds(lhs.end, rhs.end, "end", "end") >= 0
+    compareBounds(left.start, right.start, "start", "start") <= 0 &&
+    compareBounds(left.end, right.end, "end", "end") >= 0
   );
 };
 
@@ -88,12 +88,12 @@ export const intervalContainsTimestamp = (
 /**
  * Checks whether there is *any* overlap between two {@link NonNullTimeInterval}
  *
- * @param lhs
- * @param rhs
+ * @param left
+ * @param right
  */
 export const intervalOverlapsInterval = (
-  lhs: NonNullTimeInterval,
-  rhs: NonNullTimeInterval,
+  left: NonNullTimeInterval,
+  right: NonNullTimeInterval,
 ): boolean => {
   /*
    Examples |     1     |     2     |     3     |     4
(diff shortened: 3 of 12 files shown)


libs/@blockprotocol/graph/src/stdlib/interval.ts
```diff
@@ -0,0 +1,333 @@
...
+ * if adjacent, the two intervals should span another given interval.
+ *
+ * @param lhs - The first interval of the comparison (order is unimportant)
+ * @param rhs - The second interval of the comparison
+ */
+export const intervalIsAdjacentToInterval = (
+  lhs: NonNullTimeInterval,
+  rhs: NonNullTimeInterval,
+): boolean => {
+  /*
```
Alfred-Mountfield, line 21: I did try adding these comments to the actual doc-strings, but none of the formatting ended up being readable so I opted to put it back down in a normal in-code comment like we've done in HASH. The problems with the formatting included the table rendering removing whitespaces (which is essential), weirdly formatting the intervals, etc.

image
image
image

libs/@blockprotocol/graph/src/stdlib/bound.ts
```diff
@@ -0,0 +1,83 @@
...
+  lhsType: keyof NonNullTimeInterval,
+  rhsType: keyof NonNullTimeInterval,
+): number => {
+  // If the bound values are not equal, then the bound with the lower value is less than the bound with the higher
+  // value.
+  if (lhs.kind !== "unbounded" && rhs.kind !== "unbounded") {
+    if (lhs.limit !== rhs.limit) {
+      return lhs.limit.localeCompare(rhs.limit);
+    }
+  }
```
TimDiekmann, line 18: As you don't use a `match` pattern, this function can be written more efficiently: After checking for `"unbounded"` in line 14, you can do any `"unbounded"` comparison before invoking any comparison operator at all. This means,
- if both bounds are `"unbounded"`
    - and both `.kind`s are equal:  the bounds are equal.
    - if `lhs.kind` is `"start"`: `lhs` is smaller
    - else `rhs` is smaller
- if `lhs` is `"unbounded"`
    - if `lhsType` is `"start"`: `lhs` is smaller
    - else `rhs` is smaller
- if `rhs` is `"unbounded"`
    - if `rhsType` is `"start"`: `rhs` is smaller
    - else `lhs` is smaller

With that, all possibilities with `"unbounded"` are captured. Then you can continue with the logic when the limits are equal. 
I don't ask to change this here as the corresponding Rust code is very similar but very well tested and writing the same tests again is a little bit too much. Ideally, I'd favor to have this functionality exported from a Rust crate (this applies to all interval-related operations).

libs/@blockprotocol/graph/src/stdlib/bound.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Checks whether two given temporal bounds are adjacent to one another, where adjacency is defined as being next to one
+ * another on the timeline, without any points between, *and where they are not overlapping*.
+ *
+ * As such, two inclusive (or two exclusive) bounds with the same limit are *not* adjacent.
+ *
+ * @param lhs - The first bound of the comparison (order is unimportant)
+ * @param rhs - The second bound of the comparison
+ */
+export const boundIsAdjacentTo = (
```
TimDiekmann, line 71: The function is not called on a `bound` but is a free standing function. Maybe we should rename it to `boundsAreAdjacent`?

TimDiekmann, line 71: I'd also prefer if this is not exposed as this is an implementation detail of `intervalIsAdjacentToInterval` (which maybe should also be renamed to `IntervalsAreAdjacent`)

libs/@blockprotocol/graph/src/stdlib/interval.ts
```diff
@@ -0,0 +1,333 @@
...
+   ====================|===========|===========|===========|===========
+   Interval            |    [----] |   (--]    |  [--)     | [--]
+   Timestamp           |      .    |      .    |     .     |       .
+   --------------------|-----------|-----------|-----------|-----------
+   Contains Timestamp  |   true    |   true    |   false   |   false
+   */
+  return (
+    compareBounds(
+      interval.start,
+      { kind: "inclusive", limit: timestamp },
```
TimDiekmann, line 81: Could we move this out of the `return` statement? Then, we can reuse it and it's more readable.

libs/@blockprotocol/graph/src/stdlib/interval.ts
```diff
@@ -0,0 +1,333 @@
...
+ * Given a collection of {@link NonNullTimeInterval}s, this returns a list of non-adjacent, non-overlapping
+ * {@link NonNullTimeInterval}'s which span the space spanned by the input intervals.
+ *
+ * Conceptually this recursively calls {@link intervalUnionWithInterval} pairwise until all intervals have been unioned
+ * with one another. The space spanned by the result will not necessarily be contiguous (may contain gaps).
+ *
+ * @param {NonNullTimeInterval[]} intervals
+ */
+export const unionOfIntervals = <IntervalsType extends NonNullTimeInterval>(
+  ...intervals: IntervalsType[]
```
TimDiekmann, line 295: Instead of using an array, does TS has something like a BTreeSet we can utilize here instead?

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+export type TemporalAxes = "transactionTime" | "decisionTime";
+
+/**
+ * The bound of a time-interval that is either exclusively or inclusively limited by a `Timestamp`
+ */
+export type TimestampLimitedTemporalBound = {
+  kind: "inclusive" | "exclusive";
+  limit: Timestamp;
+};
```
TimDiekmann, line 23: Just for comparison: In Rust we use `"included"` and `"excluded"` as `bound` and `timestamp` instead of `limit`. I'm not requesting to change this, just want to flag it as this is subject to discussion. Personally, I like `kind` and `limit`, but prefer `"included"` and `"excluded"`. To deviate from those we could also use `"open"` or `"closed"`.

TimDiekmann (commented):
The code logic seems good to me, however, I'd prefer to have at least some test cases in place as the logic is not easy. I find the type definitions a bit too complex and don't know if they are needed at all, but I think @CiaranMn is the better person to ask here.

libs/@blockprotocol/graph/src/stdlib/interval.ts
Alfred-Mountfield: To my knowledge, nothing like you're hoping for with a Rust parallel where you can define custom hash or comparison operators. I also don't think it'd necessarily be faster, arrays are fairly fast and we're only sorting it once.

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
Alfred-Mountfield: I was going to open a comment to highlight this as well but had to jump off yesterday, thank you!

I guess the open question we'd like answered as part of merging this PR, is do we prefer a `kind` of:

1.`"included" | "excluded"` or,
2. `"inclusive" | "exclusive"`

I prefer "inclusive" and "exclusive", it aligns more with what I think I've heard more commonly in the past. "Inclusive bounds"

Alfred-Mountfield: cc @CiaranMn @thehabbos007

libs/@blockprotocol/graph/src/stdlib/bound.ts
Alfred-Mountfield: You are correct that it is not a method, but that is because we aren't using classes. I'd prefer the naming convention I've used here because for quite a few of them the ordering of the parameters matters, I admit for a few of these it does not (this for example), but for some of them like checking if an interval contains an interval.

libs/@blockprotocol/graph/src/stdlib/bound.ts
```diff
@@ -0,0 +1,83 @@
+import {
```
Alfred-Mountfield, line 1: @TimDiekmann making a comment so we can thread replies

> I'd prefer to have at least some test cases in place as the logic is not easy.

I'd prefer that too, I wanted to discuss the decision to defer tests, but had to jump off yesterday. I've updated the PR description with my thoughts about it:

> - I wanted to write a proper suite of tests to check this, but made the uncomfortable compromise at the moment to defer that until later on, to unblock some further work. We intend to write a test suite for the `Subgraph` library, and I'd encourage us to write tests for temporal operations at the same time. 
 > - It's worth noting that these temporal operations were ported from some logic we had implemented [for the HASH graph query layer ](https://github.com/hashintel/hash/blob/795bcf1e4f73c2c186da576b314d0a3ffc5bfa8d/apps/hash-graph/lib/interval-ops/src/interval.rs#L11), which are heavily tested, so confidence is fairly high that these will be correct.

libs/@blockprotocol/graph/src/stdlib/bound.ts
TimDiekmann: > but I think with the documentation this function is fine as standalone.

I disagree. The function is only well defined, if the two bounds don't have the same kind, e.g. when comparing a start-bound with an end-bound and this is not stated in the docs. I don't know what it means if a start bound is adjacent to another start bound.

libs/@blockprotocol/graph/src/stdlib/interval.ts
```diff
@@ -0,0 +1,333 @@
...
+  }
+};
+
+/**
+ * Advanced type to provide stronger type information when using {@link intervalMergeWithInterval}.
+ *
+ * If *both* of the `start` {@link TemporalBound}s are bounded, then the resultant `start` {@link TemporalBound} will be
+ * bounded, same goes for end respectively
+ */
+type MergeReturn<
```
Alfred-Mountfield, line 185: @TimDiekmann also starting a comment here so we can thread replies.

> I find the type definitions a bit too complex and don't know if they are needed at all, but I think @CiaranMn is the better person to ask here.

I also had this thought, I admit that there is a risk to maintainability here, but I think they really do help with usability of these functions and this code is probably very unlikely to be touched again much if at all. (And it helps that it's basically all pure TS/JS without any dependencies to introduce exterior reasons for maintenance).

Having these types means that in various places you won't need to check for things like `if (interval.start.kind !== "unbounded")...` and TS doesn't complain if you go directly for `interval.start.limit`.

I did discuss the complexity of these with @thehabbos007 as well, and if I remember correctly, neither of us were _convinced_ that they're worth keeping, but nor was it clear if we should delete them.

They are fairly big types, but I don't think they're that complicated, they're basically a few if statements.

libs/@blockprotocol/graph/src/stdlib/bound.ts
Alfred-Mountfield: I'm going to opt to not change it right now as from reading it, I'm not sure if your suggestion will end up more efficient, or clearer to read. And as you highlight, we'd like to be certain about correctness here first.

libs/@blockprotocol/graph/src/stdlib/interval.ts
TimDiekmann: > this code is probably very unlikely to be touched again much if at all.

I highly disagree with this reason.

In what situation do you expect to need some kind of `interval.start.kind !== "unbounded"`? The functions are currently only expected to be needed to calculate the intervals, which are used to query data from the graph, which could be any combination of bound types, and I don't see the need for those functions in other places. It feels like an overengineered approach, especially with the statement in mind, that you don't expect to change the code in the future.

libs/@blockprotocol/graph/src/stdlib/bound.ts
TimDiekmann: The approach is more efficient, but only by a very small margin, and in our use case, this will very very unlikely have any impact, so I agree to keep the logic as it is.

libs/@blockprotocol/graph/src/stdlib/interval.ts
Alfred-Mountfield: The _intersection_ of an entity interval with a resolved subgraph interval will necessarily be bounded on both sides for example. This happens throughout subgraph traversal, and subgraph querying.

libs/@blockprotocol/graph/src/stdlib/bound.ts
TimDiekmann: Adjacence requires that the bounds are directed, and the directions of the bounds need to be different. I appreciate, that this function does not take the bound kind, but only because for the logic used in this scope it has some invariants, which don't need to be checked. If both bounds are of the same kind, it should always return false, but we know, that we will always call this function with two different bound kinds.

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
CiaranMn: "inclusive" / "exclusive" is definitely more familiar to me

libs/@blockprotocol/graph/src/stdlib/bound.ts
Alfred-Mountfield: Can you point me to any material online that defines what it means for two _bounds_ to be adjacent?

If you cannot (and I did have a search), then I'd pose that we are free to define this function however we want. And in the definition I have given its behaviour, which I have described in the documentation, there are no hidden invariants, it should _not_ return false if they're of the same kind.

That is why the _interval_ adjacency function does additional logic (such as only checking a `start` bound against and `end` bound)

libs/@blockprotocol/graph/src/stdlib/interval.ts
Alfred-Mountfield: I have done so, although I don't think it really affected readability much 😅

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
Alfred-Mountfield: Also worth highlighting another change I've made from our other implementation. I describe the endpoints of the intervals here as the `start` and `end` bounds of the time intervals as I feel that this is more intuitive when talking about ranges of time.

We have opted for something more general in the other code, with `lower` and `upper` bounds.

Not sure if there are any other thoughts on that.

libs/@blockprotocol/graph/src/stdlib/bound.ts
TimDiekmann: There is no definition for adjacency for bounds, only for intervals. To check for the adjacence of an interval you need to compare bounds, and this is exactly, what this function is doing. This is why this function **is** an implementation detail. It's only a helper function to avoid writing the same functionality twice.

libs/@blockprotocol/graph/src/shared.ts
```diff
@@ -0,0 +1,28 @@
+type TupleEntry<
+  T extends readonly unknown[],
+  I extends unknown[] = [],
+  R = never,
+> = T extends readonly [infer Head, ...infer Tail]
+  ? TupleEntry<Tail, [...I, unknown], R | [`${I["length"]}`, Head]>
+  : R;
+type ObjectEntry<T extends {}> = T extends object
```
CiaranMn, line 8:
```suggestion

type ObjectEntry<T extends {}> = T extends object
```

libs/@blockprotocol/graph/src/shared.ts
```diff
@@ -0,0 +1,28 @@
...
+  ? { [K in keyof T]: [K, Required<T>[K]] }[keyof T] extends infer E
+    ? E extends [infer K, infer V]
+      ? K extends string | number
+        ? [`${K}`, V]
+        : never
+      : never
+    : never
+  : never;
+// Source: https://dev.to/harry0000/a-bit-convenient-typescript-type-definitions-for-objectentries-d6g
```
CiaranMn, line 17:
```suggestion

// Source: https://dev.to/harry0000/a-bit-convenient-typescript-type-definitions-for-objectentries-d6g
```

libs/@blockprotocol/graph/src/stdlib/bound.ts
Alfred-Mountfield: Right, which is why I'm suggesting we are free to define the operation as we wish, and _not_ encapsulate those additional invariants you describe

libs/@blockprotocol/graph/src/stdlib/bound.ts
```diff
@@ -0,0 +1,83 @@
...
+  NonNullTimeInterval,
+  TemporalBound,
+} from "../types/temporal-versioning.js";
+
+export const compareBounds = (
+  lhs: TemporalBound,
+  rhs: TemporalBound,
+  lhsType: keyof NonNullTimeInterval,
+  rhsType: keyof NonNullTimeInterval,
```
CiaranMn, line 10: I think all these functions would be more immediately parseable if we didn't use `lhs`/`rhs` to name the arguments for all of them . 

Where order doesn't matter here, I think `firstBound`/`secondBound`, `leftHandSide`/`rightHandSide` also fine (I'd skip acronyms where possible as they introduce additional processing time for people who don't have lhs/rhs internalised). Or even `left` and `right`

We should call them something more descriptive where order does matter (I'll suggest elsewhere)

Can we add a docstring to this explaining that order doesn't matter and what it's for, i.e. sorting?

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
thehabbos007: I also prefer `"inclusive" | "exclusive"` as it would describe the `kind` of `Bound`. 

`"open" | "closed"` also works, but I don't feel strongly about it.

libs/@blockprotocol/graph/src/stdlib/bound.ts
TimDiekmann: I'm not going to block this PR in this comment, but I want to emphasize that I'm not sold on exposing this function as is

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
TimDiekmann: >  I describe the endpoints of the intervals here as the start and end bounds of the time intervals as I feel that this is more intuitive when talking about ranges of time.
> 
> We have opted for something more general in the other code, with lower and upper bounds.

Personally, I find `start` and `end` more general than `lower` and `upper`. `lower` and `upper` encapsulate an ordering, `lower` is strictly less or equal to `upper`, while the `start` could greater than `end` - depending on your point of view. Maybe this is fine for time-related things as it's unusual, that a time interval starts in the present and ends in the past, but I think that `lower` and `upper` is more precise here.

libs/@blockprotocol/graph/src/stdlib/bound.ts
TimDiekmann: Why it has to be inlined? Why can't you just not export it?

Alfred-Mountfield: Because they are in different modules, and you need to `export` to make it available in another module.

TimDiekmann: Well, I consider this a language limitation, which could be solved by either moving the function next to the required function or at least adjusting the documentation to state, how to correctly use that function

CiaranMn: I also prefer `areBoundsAdjacent` (putting `are` first because it's more consistent with boolean fields that are called `isSomething`).

I would expect `boundIsAdjacentTo` to return the thing that a bound is adjacent to. So at minimum I think this could be called `boundIsAdjacentToBound`

Alfred-Mountfield: > could be solved by either moving the function next to the required function

Just bumping to highlight I edited my previous comment as I accidentally submitted it while typing.

> or at least adjusting the documentation to state, how to correctly use that function

As I said before, I am of the opinion that the documentation _does_ state what the function does and therefore how to correctly use it. The function's logic and return values make absolutely no presumptions about start/end bounds, etc. Those semantics are encapsulated by the _calling_ function.

It says:

> where adjacency is defined as being next to one another on the timeline, without any points between, *and where they are not overlapping*

That is what the function does. If in some strange world someone decided to check if two start bounds are adjacent (with this definition of adjacency), then that's their prerogative and the function would still be _correct_.

libs/@blockprotocol/graph/src/stdlib/interval.ts
```diff
@@ -0,0 +1,333 @@
...
+/**
+ * Returns whether or not the `rhs` {@link NonNullTimeInterval} is *completely contained* within the `lhs`
+ * {@link NonNullTimeInterval}.
+ *
+ * @param {NonNullTimeInterval} lhs - Checked if it contains the other
+ * @param {NonNullTimeInterval} rhs - Checked if it's contained _within_ the other
+ */
+export const intervalContainsInterval = (
+  lhs: NonNullTimeInterval,
+  rhs: NonNullTimeInterval,
```
CiaranMn, line 44: Because order does matter here, I would call them `outerInterval` and `innerInterval` or something else that makes it easier to track in the code – if we ever edit this it will make it easier to spot if they get mixed up.

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * @todo - doc
+ */
+export type TemporalAxes = "transactionTime" | "decisionTime";
+
+/**
+ * The bound of a time-interval that is either exclusively or inclusively limited by a `Timestamp`
+ */
+export type TimestampLimitedTemporalBound = {
```
CiaranMn, line 20: Do we need the word `Timestamp` in these types? Doesn't 'limited temporal bound' already imply it?

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
CiaranMn: Slightly prefer `start` and `end`, not a very strong feeling as anyone using them already has to make their way through the concept of multi-axis versioning and can probably deal with `upper` and `lower`

libs/@blockprotocol/graph/src/stdlib/bound.ts
Alfred-Mountfield: Correct, the _interval_ overlaps, but that's not what we're discussing here. The bounds as limit points do not overlap.

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
```diff
@@ -0,0 +1,91 @@
...
+  kind: "inclusive" | "exclusive";
+  limit: Timestamp;
+};
+
+export type InclusiveTimestampLimitedTemporalBound = Subtype<
```
CiaranMn, line 25: What are we going to use these for?

CiaranMn (commented):
A few comments on code clarity. I haven't looked at the implementation, assume Tim has got that but let me know if I should

libs/@blockprotocol/graph/src/stdlib/bound.ts
Alfred-Mountfield: > So at minimum I think this could be called boundIsAdjacentToBound

Alfred-Mountfield: I personally think starting the function with the object makes it easier to find what you're looking for

libs/@blockprotocol/graph/src/shared.ts
Alfred-Mountfield: I left these grouped together to try and encourage some intuition that they're all directly related, but most importantly they're implementation details for the thing below (which has a comment sharing the source).

I'm happy to add the newlines if you'd prefer just for general style consistency though.

libs/@blockprotocol/graph/src/stdlib/bound.ts
CiaranMn: Fine with either

If you go with `areXY` note that the first entry in the table should be `areBoundsAdjacent`

libs/@blockprotocol/graph/src/stdlib/bound.ts
Alfred-Mountfield: I feel like `first` and `second` might actually imply _more_ of an order than "lhs" or "rhs" which are fairly common parameter choices for operators.

I generally completely avoid acronyms too, so I'm happy to rename them here but I think it really will inhibit readability of the function below. It adds a lot text which makes it harder to quickly see the actual logic that's happening here. (And the identity of the things does not matter, so really I _almost_ just want to call it `x` and `y`).

I have tried to highlight where parameter order _does_ matter, do you think it would help that much to add lines in all the documentation about ones where parameter _doesn't_ matter?

That said, I can add a doc-comment to this function, I opted not to just because it was a bog-standard JS comparison operation.

```ts
export const compareBounds = (
  leftHandSide: TemporalBound,
  rightHandSide: TemporalBound,
  leftHandSideType: keyof NonNullTimeInterval,
  rightHandSideType: keyof NonNullTimeInterval,
): number => {
  // If the bound values are not equal, then the bound with the lower value is less than the bound with the higher
  // value.
  if (leftHandSide.kind !== "unbounded" && rightHandSide.kind !== "unbounded") {
    if (leftHandSide.limit !== rightHandSide.limit) {
      return leftHandSide.limit.localeCompare(rightHandSide.limit);
    }
  }

  if (
    leftHandSide.kind === rightHandSide.kind &&
    leftHandSideType === rightHandSideType
  ) {
    return 0;
  }

  if (
    (leftHandSide.kind === "unbounded" && leftHandSideType === "start") ||
    (rightHandSide.kind === "unbounded" && rightHandSideType === "end") ||
    (leftHandSide.kind === "exclusive" &&
      rightHandSide.kind === "exclusive" &&
      leftHandSideType === "end" &&
      rightHandSideType === "start") ||
    (leftHandSide.kind === "exclusive" &&
      rightHandSide.kind === "inclusive" &&
      leftHandSideType === "end") ||
    (leftHandSide.kind === "inclusive" &&
      rightHandSide.kind === "exclusive" &&
      rightHandSideType === "start")
  ) {
    return -1;
  }

  if (
    (leftHandSide.kind === "unbounded" && leftHandSideType === "end") ||
    (rightHandSide.kind === "unbounded" && rightHandSideType === "start") ||
    (leftHandSide.kind === "exclusive" &&
      rightHandSide.kind === "exclusive" &&
      leftHandSideType === "start" &&
      rightHandSideType === "end") ||
    (leftHandSide.kind === "exclusive" &&
      rightHandSide.kind === "inclusive" &&
      leftHandSideType === "start") ||
    (leftHandSide.kind === "inclusive" &&
      rightHandSide.kind === "exclusive" &&
      rightHandSideType === "end")
  ) {
    return 1;
  }
  throw new Error(
    `Implementation error, failed to compare bounds.\nLHS: ${JSON.stringify(
      leftHandSide,
    )}\nRHS: ${JSON.stringify(rightHandSide)}`,
  );
};
```

libs/@blockprotocol/graph/src/shared.ts
CiaranMn: I think it makes it easier to read, yes. We don't do this grouping elsewhere for types which are just utilities for types below them

libs/@blockprotocol/graph/src/stdlib/interval.ts
Alfred-Mountfield: Happy to change that, just flagging that I'm never sure about naming variables like that when they're being used to check it. 

For example take a type-guard: 
```ts
isFish(pet: Pet): pet is Fish
```

I wouldn't want to call it:
 ```ts
isFish(fish: Pet): fish is Fish
```

libs/@blockprotocol/graph/src/stdlib/bound.ts
CiaranMn: How about `left` and `right`? It's one more character. Or even `x` and `y`, sure.

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
Alfred-Mountfield: This is another one of those things I had intended to open a comment highlighting but forgot to 😅. I'm really uncomfortable with how long and unwieldy some of these type names are, they really are pretty rough. That said, I opted for correctness and explicitness over something being accidentally misleading.

For this specific case, I'm not against us going to `LimitedTemporalBound`, I was just trying to say it had a timestamp _at_ its limit, (because a potential interpretation of "unbounded" is that its limit is infinity, or negative infinity)

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
Alfred-Mountfield: These will end up used in my next PR
image

I wasn't even sure about exporting the intermediary types but it can be a bit annoying when a package doesn't export those types and you're trying to manipulate an object or create a type guard and you can't just reuse their intermediary ones and have to define your own or extract from the type somehow.

libs/@blockprotocol/graph/src/stdlib/bound.ts
Alfred-Mountfield: I don't despise `left` and `right` I guess 😅, does get rid of the acronym. I feel allergic to using `x` and `y` outside of code to do with co-ordinates or something tbh (I know I suggested it), so I've opted for `left` and `right`.

libs/@blockprotocol/graph/src/stdlib/bound.ts
Alfred-Mountfield: I've renamed `boundIsAdjacentTo` to `boundIsAdjacentToBound`.

I'm fine going with the other side of the table if someone feels stronger about this than me, but for now I'm leaning to leaving it as is

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
Alfred-Mountfield: Judging from the feelings in this thread I'm going to opt to leave it as is
- `start` and `end`
- `"inclusive"` and `"exclusive"`

TimDiekmann (approved):
Implementation logic looks good to me ✅ 
Probably it would be good if @CiaranMn judges code style and best practices in TS

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
CiaranMn: I think because we have an explicit separate type for `Unbounded`, we could get away with this being `LimitedTemporalBound`. Can think about it in follow up though

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
CiaranMn: I'm not really any the wiser from that screenshot but we can talk about it in that PR :)

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
Alfred-Mountfield: Not hard to do, have pushed

PR #928, opened 2023-02-01
Introduce support for bi-temporal versioning in the Graph Service

Alfred-Mountfield:

As expressed in previous PRs, we intend to add the foundations of support for bi-temporal datastores within the BP as part of the 0.3 release. This PR modifies the Graph Service to support such use-cases. It is necessarily a large PR as adding generics tends to propagate immensely. 

I recommend you review this commit by commit.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543026679380/1203765721937276/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

Please see commits :(

- As discussed in previous PRs, this absolutely will need changes to the docs. This is deferred to a further point, but effort has been made to have good in-code documentation.

## ⚠️ Known issues

- We ideally want users to be able to import different views of the package.
  - Blocks or EAs which don't care about (or do not support) temporal versioning should ideally be able to import types and functions as if there was no temporal versioning. It would be nice for us to be able to strip the generics, or prefill them in with defaults for these consumers
  - Blocks which do care about temporal versioning should ideally be able to import types and functions where it was if there were no generics and instead everything had _**required**_ temporal versioning. Similar to above, would be nice to be able to prefill or strip them or something.
  - Blocks or EAs that need to handle general cases (EAs with support for temporal versioning may need to interface with blocks that do not have it for example) need to be able to import the types which have the generics in place, without default values. An example of such a consumer is MBD which will be updated in the next PR.
  - **I have been unable to find a way to ergonomically do this in this PR. Because of that, I have consolidated the `stdlib` and `stdlib-temporal` endpoints for now. And am exposing the most general implementations of everything on the API**
- Some other issues have been highlighted by GitHub comments on the diff

## 🐾 Next steps

- Migrate MBD and fix other accompanying packages
- Figure out how to expose the various endpoints

## 🛡 What tests cover this?

- I'm unsure, at the very least `tsc` and `eslint`. 
- I wanted to write a proper suite of tests to check this, but made the uncomfortable compromise at the moment to defer that until later on, to unblock some further work.

## ❓ How to test this?

- Confirm that `lint:tsc` and `lint:eslint` pass for the `@blockprotocol/graph` package
- See #931 

## 📹 Demo

N/A


libs/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
```diff
@@ -65,11 +72,20 @@ export const buildSubgraph = (
     revisionId: rootRecordId.editionId,
   }));
 
-  const subgraph: Subgraph = {
+  // @ts-expect-error -- @todo, how do we convince TS that we're only setting this when the generic is satisfied
+  const subgraph: Subgraph<Temporal, EntityRootType<Temporal>> = {
```
Alfred-Mountfield, line 76: Didn't want to sink much time into this but completely ignoring `tsc` here is also not great.. I haven't found a good way to inform TSC that the generic has been narrowed at certain points within function bodies. This is a recurring inconvenience and in the following PR there will be many places where it's something like `if (subgraph.temporalAxes !== undefined)`... and it would be really great if there was a way to then tell TS "hey `Temporal` is actually `true` here"

libs/@blockprotocol/graph/src/stdlib/subgraph/element/map-revisions.ts
```diff
@@ -0,0 +1,43 @@
...
+
+type BaseIdToRevisions<
+  TemporalSupport extends boolean,
+  GraphElementType extends Vertex<TemporalSupport>["inner"],
+> = Record<
+  /*
+   * @todo - we _should_ be able to use `Extract<GraphElementForIdentifier<TemporalSupport, VertexId<any, any>>`
+   *   here to actually get a strong type (like `EntityId` or `BaseUri`). TypeScript seems to break on using it with a
+   *   generic though. So for now we write `string` because all of the baseId's of `VertexId` are string aliases anyway.
+   */
```
Alfred-Mountfield, line 11: This is actually quite frustrating and if we end up making the BaseIds template strings or branded types this will catch us out.

libs/@blockprotocol/graph/src/stdlib/subgraph/roots.ts
```diff
@@ -11,23 +11,26 @@ import {
...
 import { getPropertyTypeByVertexId } from "./element/property-type.js";
 
 /**
  * Returns all root elements.
  *
  * For a narrower return type, first narrow the type of `subgraph` by using one of the helper type-guards:
- * - isSubgraph<DataTypeRootType>
- * - isSubgraph<PropertyTypeRootType>
- * - isSubgraph<EntityTypeRootType>
- * - isSubgraph<EntityRootType>
```
Alfred-Mountfield, line 25: Oops, slipped through on the previous one

libs/@blockprotocol/graph/src/types/subgraph/edges.ts
```diff
@@ -2,14 +2,14 @@ import { BaseUri } from "@blockprotocol/type-system/slim";
...
-} from "./edges/outward-edge.js";
+import { KnowledgeGraphOutwardEdge } from "./edges/variants/knowledge.js";
+import { OntologyOutwardEdge } from "./edges/variants/ontology.js";
 
 export * from "./edges/kind.js";
 export * from "./edges/outward-edge.js";
-export * from "./edges/outward-edge-alias.js";
+export * from "./edges/variants.js";
+
+/** @todo - Re-express these and `Vertices` as `Record`s? */
```
Alfred-Mountfield, line 12: I'm unsure why I ended up expressing them like this originally, it seems that they'd be better as `Record`s, and perhaps they'd behave slightly nicer sometimes with narrowing.

I don't want to change it here though because I want to make sure it doesn't break any code in usage in MBD, so will probably look into this as part of the follow-up PR.

libs/@blockprotocol/graph/src/types/subgraph/edges/variants/knowledge.ts
```diff
@@ -71,19 +64,36 @@ export const isIncomingLinkEdge = (
...
+): outwardEdge is IsOfTypeEdge => {
+  return outwardEdge.kind === "IS_OF_TYPE" && !outwardEdge.reversed;
 };
+
+export type KnowledgeGraphOutwardEdge =
+  | OutgoingLinkEdge
+  | IncomingLinkEdge
+  | HasLeftEntityEdge
+  | HasRightEntityEdge
+  | IsOfTypeEdge;
```
Alfred-Mountfield, line 87: I did this because it turns out that expressing the things like below (`GenericOutwardEdge<KnowledgeGraphEdgeKind, boolean, EntityIdAndTimestamp>`) meant they just _would not_ play nicely with advanced types such as `Extract`.

Those types end up being pretty important in the ergonomics of using the `Subgraph` so I decided it'd be better to be fully explicit and stick to plain boring typescript here without generic stuff. 

This has had the side-effect benefit of having standardised names for the various edge permutations now, which we didn't have before.

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -62,46 +73,74 @@ export const getOutgoingLinksForEntity = (
...
-      // Only look at outgoing edges that were created before or at the timestamp
+      // Only look at outgoing edges that were created before or within the search interval
       .filter(
-        ([edgeTimestamp, _outwardEdges]) => edgeTimestamp <= timestampString,
+        ([edgeTimestamp, _outwardEdges]) =>
+          compareBounds(
+            { kind: "inclusive", limit: edgeTimestamp },
+            searchInterval.end,
+            /** @todo - should this be end */
+            "start",
```
TimDiekmann, line 83: Yes, I think so. As `searchInterval.end` could be `"exclusive"`, this would return `true` if `searchInterval.limit === edgeTimestamp` if I got it correctly.

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -113,90 +152,175 @@ export const getIncomingLinksForEntity = (
...
-      // Only look at edges that were created before or at the timestamp
+      // Only look at outgoing edges that were created before or within the search interval
       .filter(
-        ([edgeTimestamp, _outwardEdges]) => edgeTimestamp <= timestampString,
+        ([edgeTimestamp, _outwardEdges]) =>
+          compareBounds(
+            { kind: "inclusive", limit: edgeTimestamp },
+            searchInterval.end,
+            /** @todo - should this be end */
+            "start",
```
TimDiekmann, line 162: Same as above, I think this is `"end"`

TimDiekmann: Maybe we should add a function `boundIsStrictlyRightOfInterval` and use that instead.

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -205,26 +329,85 @@ export const getRightEntityForLinkEntity = (
...
+    timestamp !== undefined
+      ? ({
+          start: {
+            kind: "inclusive",
+            limit: convertTimeToStringWithDefault(timestamp),
+          },
+          end: {
+            kind: "inclusive",
+            limit: convertTimeToStringWithDefault(timestamp),
+          },
```
TimDiekmann, line 354: We have this pattern quite often, maybe we should add a function to convert a timestamp into an interval.

TimDiekmann (commented):
The logic largely seems to be correct, but a look from another person would be much appreciated. It's really hard to figure out the logic and I'm looking for the tests, which will be added in a follow-up.

libs/@blockprotocol/graph/src/types/subgraph/edges/outward-edge.ts
```diff
@@ -1,19 +1,29 @@
...
 
 export type EntityIdAndTimestamp = {
   baseId: EntityId;
   timestamp: Timestamp;
 };
 
+export type EntityValidInterval = {
+  entityId: EntityId;
+  validInterval: TimeInterval<LimitedTemporalBound, TemporalBound>;
+};
```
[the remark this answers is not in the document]
Alfred-Mountfield: It refers to a period in an entity's existence whereby an edge was active, yes. It could be more than that I guess. In _some_ ways this implies that this entity was valid for at least this range, and it also indirectly implies that the other endpoint (not captured in this object) was valid for at least this range too.

libs/@blockprotocol/graph/src/types/subgraph/element-mappings.ts
```diff
@@ -0,0 +1,120 @@
...
+  | {
+      identifier: BaseUri;
+      element:
+        | DataTypeWithMetadata[]
+        | PropertyTypeWithMetadata[]
+        | EntityTypeWithMetadata[];
+      vertex: DataTypeVertex[] | PropertyTypeVertex[] | EntityTypeVertex[];
+    }
+  | {
+      identifier: EntityIdAndTimestamp | EntityVertexId | EntityRecordId;
```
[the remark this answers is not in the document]
Alfred-Mountfield: That's how it's used, sure. But it's just what it says, it's an `EntityId` and a `Timestamp`. 

Assuming you were to query a `Subgraph` for that, you'd either get back a single revision (because revisions of an entity do not overlap), or you'd get back `undefined`.

libs/@blockprotocol/graph/src/types/subgraph/element-mappings.ts
```diff
@@ -0,0 +1,120 @@
...
+      element: Entity<Temporal>;
+      vertex: EntityVertex<Temporal>;
+    }
+  | {
+      identifier: EntityId | EntityValidInterval;
+      element: Entity<Temporal>[];
+      vertex: EntityVertex<Temporal>[];
+    };
```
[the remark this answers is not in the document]
Alfred-Mountfield: That's a very interesting question. I'm not sure if it's worth separating them out in this type (it could cause problems with other advanced types like `Extract`).

I don't think it'd necessarily be an error for a user to try and use these things on a non-temporal subgraph. _At the moment_ revisionId's in a non-temporal subgraph are still timestamps, they're just mostly meaningless, and there won't be multiple revisions.

Prior to the past two PRs for example, the stdlib relied on passing in the current timestamp to get an entity (only one revision as we're talking non-temporal) by its `EntityId`.

I can add a todo in the code if you'd like though, this might be worth us investigating later. One of the things I think this type _could_ be helpful for, is forcing us to make it ergonomic to accept any sort of identifier in the stdlib. Right now, some of the `get*` methods (such as `getPropertyTypeByVertexId`) take in a specific kind of identifier, which requires the user to either select the right function (if it exists), or transform an identifier they just got from a different part of the stdlib, into another identifier. We _could_ make all of the `getPropertyType*` methods take a generic "identifier" thing from this mapping object, and then you'd be forced to handle all of them, which seems like good UX. cc @CiaranMn for my intentions long term here.

libs/@blockprotocol/graph/src/internal/mutate-subgraph.ts
```diff
@@ -50,103 +54,128 @@ export const addKnowledgeGraphEdgeToSubgraphByMutation = (
...
-export const addEntitiesToSubgraphByMutation = (
-  subgraph: Subgraph,
-  entities: Entity[],
+export const addEntitiesToSubgraphByMutation = <Temporal extends boolean>(
+  subgraph: Subgraph<Temporal>,
+  entities: Entity<Temporal>[],
 ) => {
+  /*
+   * @todo This assumes that the left and right entity ID of a link entity is static for its entire lifetime, that is
+   *   not necessarily going to continue being the case
```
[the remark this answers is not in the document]
Alfred-Mountfield: I was going to raise this, it feels pretty consequential for HASH as well

libs/@blockprotocol/graph/src/internal/mutate-subgraph.ts
```diff
@@ -50,103 +54,128 @@ export const addKnowledgeGraphEdgeToSubgraphByMutation = (
...
-    const recordId = entity.metadata.recordId;
+    const entityId = entity.metadata.recordId.entityId;
+
+    const entityRevisionValidInterval: EntityValidInterval["validInterval"] =
+      subgraph.temporalAxes !== undefined
+        ? entity.metadata.temporalVersioning[
+            subgraph.temporalAxes.resolved.pinned.axis
+          ]
+        : {
+            start: { kind: "inclusive", limit: new Date(0).toISOString() },
```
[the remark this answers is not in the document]
Alfred-Mountfield: I was a bit unsure of what to put here but I'm not sure we do want the current date. This does feel like something we'd need to capture in a high-level definition of what a non-versioned subgraph actually looks like 🙃 as right now it's sort of implicitly designed in a few places across the code.

Assuming the thing is truly **non-versioned**, then it kind of makes sense to say the entities are just always valid, because then if you do end up passing in a timestamp or something in some logic somewhere else deep inside subgraph traversal, it will always return the entity.

I'm not suggesting that is what we want to do, that is perhaps hiding inconsistencies in implementation elsewhere and in a non-versioned subgraph you just should _not_ be able to pass in a timestamp (that _seems_ more ideal). But it does _seem_ like it's less likely to cause us hard to discover bugs in the meantime.

libs/@blockprotocol/graph/src/internal/mutate-subgraph.ts
```diff
@@ -50,103 +54,128 @@ export const addKnowledgeGraphEdgeToSubgraphByMutation = (
...
-      subgraph,
-      rightEntityId,
-      earliestTime,
-      {
-        kind: "HAS_RIGHT_ENTITY",
-        reversed: true,
-        rightEndpoint: { baseId: linkEntityId, timestamp: earliestTime },
-      },
-    );
+    for (const validInterval of validIntervals) {
```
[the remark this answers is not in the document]
Alfred-Mountfield: Encountered this once or twice myself in coding. But didn't have a better idea either

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
```diff
@@ -54,80 +66,82 @@ export const getEntity = (
...
+      .map(([_startTime, entityVertex]) => entityVertex)
+      .filter(isEntityVertex)
+      .map((entityVertex) => entityVertex.inner)
+      .filter((entity) =>
+        intervalContainsInterval(
+          interval,
+          entity.metadata.temporalVersioning[
+            subgraph.temporalAxes.resolved.variable.axis
+          ],
+        ),
```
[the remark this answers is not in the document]
Alfred-Mountfield: Hm, interesting (and great catch). I think we do want `intervalOverlapsInterval` instead of `Contains` here. We want to return any entity which was active over that range, the fact that it was still active after the search range doesn't matter (and would be filtered out by this).

I'll update and add a comment explaining this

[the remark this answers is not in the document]
Alfred-Mountfield: Have pushed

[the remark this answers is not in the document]
TimDiekmann: Thanks! :)

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
Alfred-Mountfield: Looking at the implementation of `compareBounds` I'm not convinced it actually changes things. 

But considering this is basically half of an `intervalOverlapsInterval` operation, I now think maybe it _should_ be `end`? It's effectively this portion (I think)

image

Alfred-Mountfield: (From your [other comment](https://github.com/blockprotocol/blockprotocol/pull/928#discussion_r1093565147))

> Maybe we should add a function `boundIsStrictlyRightOfInterval` and use that instead.

This sounds sensible. Do we want `strictlyRightOf` here? I'd assume we're also happy with the bounds being the same (e.g. inclusive at timestamp 3, and inclusive at timestamp 3)

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
Alfred-Mountfield: Sounds sensible, I'll add one

Alfred-Mountfield: Have added one (wasn't actually that many usages :( )

Alfred-Mountfield: ([commit](https://github.com/blockprotocol/blockprotocol/pull/928/commits/8c248181676a36894d88ef5ccb23badb1df9bd2e))

TimDiekmann: Thanks, I think it's better, though :)

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
TimDiekmann: For an "exclusive" this will change from "is smaller" to "is equal" but as we compare with `<=` both should be fine, yes. 

Yes, let's add a `strictlyRightOf`-like function, it makes reading it easier.

Alfred-Mountfield: Sorry I wasn't clear, what I meant by

> Do we want `strictlyRightOf` here?

Is that the operation we're doing doesn't _seem_ to be **strictly right of**.

TimDiekmann: No, we would check, if the interval is `not strictlyRightOf` I think 🙂

Alfred-Mountfield: And actually we're not checking if the bound is to the right of the interval. We're checking if the bound is to the left of or contained in the interval

TimDiekmann: Yes, this is the same as if the bound is not strictly right of the interval

Alfred-Mountfield: (We talked about this offline and landed on this) https://github.com/blockprotocol/blockprotocol/pull/928/commits/3a11838b5d74494484778721270590f661a70c30

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
Alfred-Mountfield: Handled by [Introduce strict comparison operators for intervals](https://github.com/blockprotocol/blockprotocol/pull/928/commits/3a11838b5d74494484778721270590f661a70c30)

TimDiekmann (approved):

thehabbos007 (commented):
Just leaving a couple comments I had around, still not through all of it. Feel free to merge regardless, I'll continue to look through

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -205,26 +326,76 @@ export const getRightEntityForLinkEntity = (
...
-          timestamp,
-        ),
-      };
-    },
-  ) as LinkAndRightEntities; // @todo consider fixing generics in functions called within
+        );
+
+        if (rightEntityRevisions.length !== 1) {
+          throw new Error(
+            `Querying a Subgraph without support for temporal versioning but there wasn't a unique revision for the right entity with ID: ${linkEntity.metadata.recordId.entityId}`,
```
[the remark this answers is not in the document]
Alfred-Mountfield: It would cheers

libs/@blockprotocol/graph/src/stdlib/subgraph/builder.ts
```diff
@@ -65,11 +72,20 @@ export const buildSubgraph = (
...
     roots,
     vertices: {},
     edges: {},
     depths,
+    ...(temporalAxes !== undefined
+      ? {
+          temporalAxes: {
+            initial: temporalAxes,
+            resolved: temporalAxes,
```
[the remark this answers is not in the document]
Alfred-Mountfield: Hm, it was intentional, but perhaps we want to change it to allow the caller to also pass in the `Unresolved` axes. I'm going to change it to do so.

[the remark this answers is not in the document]
Alfred-Mountfield: [Take SubgraphTemporalAxes in buildSubgraph](https://github.com/blockprotocol/blockprotocol/pull/928/commits/e00c258c2e92bde7ad360febd185db6355a05ce6)

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
```diff
@@ -54,80 +66,94 @@ export const getEntity = (
...
+          interval,
+          /*
+           these casts are safe as we check for `targetRevisionInformation === undefined` above and that's only ever
+           defined if `Temporal extends true`
+           */
+          (entity as Entity<true>).metadata.temporalVersioning[
+            (subgraph as Subgraph<true>).temporalAxes.resolved.variable.axis
+          ],
+        ),
+      );
```
CiaranMn, line 152: Looks like a lot of looping – can we not do something like this?
```suggestion
    const entities = [];
    for (const [_startTime, vertex] of typedEntries(versionObject)) {
      const entity = isEntityVertex(vertex) ? vertex.inner : null;

      if (
        entity &&
        "temporalVersioning" in entity.metadata &&
        intervalOverlapsInterval(
          interval,
          entity.metadata.temporalVersioning[
            subgraph.temporalAxes.resolved.variable.axis
          ],
        )
      ) {
        entities.push(entity);
      }
    }
    return entities;
```

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
```diff
@@ -54,80 +66,94 @@ export const getEntity = (
...
+        return intervalContainsTimestamp(
+          /*
+           these casts are safe as we check for `targetRevisionInformation === undefined` above and that's only ever
+           defined if `Temporal extends true`
+           */
+          (entity as Entity<true>).metadata.temporalVersioning[
+            (subgraph as Subgraph<true>).temporalAxes.resolved.variable.axis
+          ],
+          targetTime,
+        );
```
CiaranMn, line 103: Technically there's nothing to stop JS users passing a non-temporal subgraph and `targetRevisionInformation` but I know that worrying about that is a whole can of worms 😁

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
```diff
@@ -54,80 +66,94 @@ export const getEntity = (
 
   const revisionVersions = Object.keys(entityRevisions).sort();
 
-  let entityRevision: Entity | undefined;
-
   // Short circuit for efficiency, just take the latest
   if (targetRevisionInformation === undefined) {
-    [entityRevision] = revisionVersions
+    return revisionVersions
```
CiaranMn, line 71: Doubt it makes any performance difference but maybe marginally clearer
```ts
    const lastIndex = revisionVersions.length - 1;
    return entityRevisions[revisionVersions[lastIndex]!]!.inner;
```

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
```diff
@@ -54,80 +66,94 @@ export const getEntity = (
 
   const revisionVersions = Object.keys(entityRevisions).sort();
```
CiaranMn, line 67: Doesn't this only need to be sorted in the 'give me latest' case?

CiaranMn (commented):
Phew! Thanks @Alfred-Mountfield, fantastic work. 

My comments are mostly about naming / explaining stuff.

I started to optimise things but I think you're aware that we can reduce a lot of looping so I stopped.

I haven't looked at the TS errors you flag as I thought I'd get this in and then we can revisit the most egregious ones – let me know which you think might be worth further investigation.

libs/@blockprotocol/graph/src/internal/mutate-subgraph.ts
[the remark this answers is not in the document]
Alfred-Mountfield: I'm not sure, that's part of the issue. 

I think it's presently somewhat undefined/unexplored behaviour that's already present within the constraints of our system as we have specified them. We have encoded within our own implementations (but not the definitions of the primitives and their operations) that a link's endpoints cannot change. But the primitives do not encode that.

I think we maybe just need a task to look into this?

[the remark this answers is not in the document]
CiaranMn: The primitives don't, but if we take the message signatures in `@blockprotocol/graph` as representing what will be defined the spec, they do for now encode that the left and right entity can only be set on creation and never again.

We can create a task if we have some way of capturing why we would want to revisit that.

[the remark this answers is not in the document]
Alfred-Mountfield: For completeness, I was actually considering the message signatures and such as part of the "primitives" (I appreciate that's perhaps using the term too loosely).

I think we still have a lack of clarity within the current system because of these scenarios
1. You call `createEntity` for a link entity, set its leftEntity and rightEntity
   1. You call `updateEntity` and change its entity type to something that _is not_ a link entity type
   2. What happens? 
      1. Do we continue to return left and right entity IDs in the metadata?
      2. Do we delete the left and right entity information?   
      3. What should backends do?
    3. You now call `updateEntity` again and change its entity type to something that _is_ a link entity type (perhaps a completely different type)
        1. What do you do about the various options above?
 2.  You call `createEntity` for a _non_-link entity, you do **not** set its leftEntity and rightEntity
     1. You now call `updateEntity` and change its entity type to something that _is_ a link entity type
     2. What happens?  
          1. Do you continue to have an entity that breaks the invariants where we say all link entities must have left and right endpoints?
          2. Do you change `updateEntity` to allow setting of left and right endpoints?
          3. Do you somehow block the `updateEntity` operation by not allowing its entity type to be changed to something that's a link? There's nothing in the messages, documentation, or primitives that suggest anything EAs should do in this situation.

I think there are some viable approaches (including starting to more concretely specify validation of mutation methods, which we don't currently do to my knowledge), but these are things we need to visit rather than revisit, in my eyes.

libs/@blockprotocol/graph/src/internal/mutate-subgraph.ts
```diff
@@ -11,18 +13,20 @@ import {
 import { isEqual } from "./mutate-subgraph/is-equal.js";
 
 /**
- * Looking to build a subgraph? You probably want `import { buildSubgraph } from @blockprotocol/graph/stdlib`
+ * Looking to build a subgraph? You probably want {@link buildSubgraph} from `@blockprotocol/graph/stdlib`
```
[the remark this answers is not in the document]
Alfred-Mountfield: As you say, it wasn't written correctly (so not a direct copy paste), and I think the vast majority of people do not manually write their own imports in the world of intellisense and auto-complete. 

I do think this is an improvement for maintainability of the library as if you're clicking through documentation, you can immediately click on `buildSubgraph` and read its documentation with a single click to see if that is actually what you were trying to do. As opposed to the alternative where you have to leave whatever you were using to read this documentation in the first place, write some code to import the other function, and then read its documentation (or you have to use some IDE feature to go to the function rather than importing it).

We really are talking about some miniscule DX improvements here, but I think trying to get our docs more compliant with discoverability in aggregate will do wonders, especially if we struggle to minimise the exposure of various complex parts of the Graph service like temporal versioning.

libs/@blockprotocol/graph/src/internal/mutate-subgraph.ts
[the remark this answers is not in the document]
CiaranMn: Hmm, fair enough. I was thinking that it is impossible to set the left/right entity after creation, but you're right that this creates issues with type switching that we need to figure out what to do about.

libs/@blockprotocol/graph/src/internal/mutate-subgraph.ts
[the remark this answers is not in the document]
CiaranMn: Sure, makes sense – leave it as it is

thehabbos007 (commented):
Great work, Alfie! I've gotten through most files now and added some comments :)

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
```diff
@@ -45,6 +45,10 @@ export type Unbounded = { kind: "unbounded" };
  */
 export type TemporalBound = Unbounded | LimitedTemporalBound;
 
+/**
+ * A range of time from a given `start` {@link TemporalBound} (or lack of) to a given `end` {@link TemporalBound} (or
+ * lack of), where `start` is strictly before or equal to `end`.
+ */
 export type TimeInterval<
```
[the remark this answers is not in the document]
Alfred-Mountfield: I have tried to take this on board and applied it to a few things in the aim of consistency. Please see
- [Rename TimeInterval types](https://github.com/blockprotocol/blockprotocol/pull/928/commits/6dd35285f2ea2f0cfb68b3fb71573e83f4e4c3c9)
- [Rename Axis types](https://github.com/blockprotocol/blockprotocol/pull/928/commits/5a6b5105a4c2bef14c68c2691de3500770841273)
- [Rename Axes types](https://github.com/blockprotocol/blockprotocol/pull/928/commits/3fec1ea62d6e875f946f729c53efd0b106bdb897)

Happy to iterate on these until it makes more sense. I opted to try and centre on `Unresolved` rather than `Input`, but can also be swayed the other way. I've also put the `Unresolved` at the latter end of the type names now, which seemed more consistent if we avoid `Resolved` on the name.

libs/@blockprotocol/graph/src/types/subgraph/edges/outward-edge.ts
```diff
@@ -1,42 +1,28 @@
...
-  Reversed extends boolean = boolean,
-> = {
-  kind: EdgeKind;
-  reversed: Reversed;
-  rightEndpoint: Endpoint;
-};
+import { KnowledgeGraphOutwardEdge } from "./variants/knowledge.js";
+import { OntologyOutwardEdge } from "./variants/ontology.js";
 
 export type EntityIdAndTimestamp = {
```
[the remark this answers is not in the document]
Alfred-Mountfield: [Add doc-comment to EntityIdAndTimestamp](https://github.com/blockprotocol/blockprotocol/pull/928/commits/74d386f342876255b4ba9c7327b8d3558509a59f)

Discussions about the `ValidInterval` type continue off GitHub

libs/@blockprotocol/graph/src/types/subgraph/edges/variants/ontology.ts
```diff
@@ -0,0 +1,205 @@
...
+export type IsTypeOfEdge = Subtype<
+  GenericOutwardEdge,
+  {
+    reversed: true;
+    kind: "IS_OF_TYPE";
+    rightEndpoint: EntityValidInterval;
+  }
+>;
+
+export const isIsTypeOfEdge = (
```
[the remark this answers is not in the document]
Alfred-Mountfield: Yep, and I actually imported the wrong thing here :( which is why I was really thankful for my little tsc hacky tests below.

I've had some other ideas but they all come with other pitfalls which seem better but then end up being the same in my mind
- `isInstanceOfType`
- `hasType`

etc.

libs/@blockprotocol/graph/src/types/entity.ts
```diff
@@ -52,7 +52,7 @@ export type EntityPropertiesObject = {
   [_: BaseUri]: EntityPropertyValue;
 };
 
-type HalfClosedInterval = TimeInterval<
+type HalfClosedInterval = TimeIntervalUnresolved<
```
CiaranMn, line 55: Shouldn't this be simply `TimeInterval`? It does not have `null` in it.

libs/@blockprotocol/graph/src/types/subgraph/edges/variants/ontology.ts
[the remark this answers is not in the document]
CiaranMn: `isTypeDescribing` / `isDescribedByType` ?

`isTypeConstraining` / `isConstrainedByType` ?

libs/@blockprotocol/graph/src/stdlib/subgraph/element/map-revisions.ts
```diff
@@ -0,0 +1,43 @@
...
+      "entityId" in element.metadata.recordId
+        ? element.metadata.recordId.entityId
+        : element.metadata.recordId.baseUri;
+
+    if (revisionMap[baseId] !== undefined) {
+      revisionMap[baseId]!.push(element);
+    } else {
+      // eslint-disable-next-line no-param-reassign
+      revisionMap[baseId] = [element];
+    }
```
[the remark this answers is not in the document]
Alfred-Mountfield: ` ??= ` I had not seen that operator before, interesting.

I appreciate this is about as micro an optimisation as you can possibly get, but it seemed that it would be faster to create an array with the thing in it rather than allocate an array and then push the thing in. But I've done some more reading on JS arrays and the various implementations across the JS engines and my brain has broken, so I'm not even sure the intuition was correct. (sparse arrays, optimisations for numeric arrays, not actually allocating contiguous space despite setting initial length to 1000, etc.)

Regardless, absolutely wouldn't argue for the previous code on the basis of that tiny a potential optimisation.

[Improve non-existing array initialization](https://github.com/blockprotocol/blockprotocol/pull/928/commits/262fddead22c34c0268ce273dfdeac72bad535f8)

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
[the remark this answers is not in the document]
CiaranMn: Had a look and dropped a comment. 

Personally I would not derive `TimeInterval` from `TimeIntervalUnresolved`. I don't think avoiding specifying `start` and `end` twice is worth the conceptual weirdness of saying that `TimeInterval` is a variant of  `TimeIntervalUnresolved`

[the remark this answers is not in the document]
Alfred-Mountfield: I thought about this too.. However, I _**really**_ dislike defining sub/super types completely separately in TypeScript, it makes me very uncomfortable and has lead to too many bugs in even my recent work.

That's why a while ago we ended up discussing the `Subtype` helper you found and that I've used in some places.

I could redefine `TimeIntervalUnresolved` as an `Omit` and replacement thing I guess...

libs/@blockprotocol/graph/src/stdlib/subgraph/element/map-revisions.ts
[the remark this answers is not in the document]
CiaranMn: It's 'new' as of 2020. There are also `||=` and `&&=` but I've never had occasion to use those.

libs/@blockprotocol/graph/src/types/subgraph/edges/variants/ontology.ts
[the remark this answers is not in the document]
Alfred-Mountfield: I'd stay well clear of `constraining` because we use those for all of the ontology related edges (e.g. `constrainsPropertiesOn`), and I don't like `describing` :(

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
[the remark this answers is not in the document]
CiaranMn: > I could redefine TimeIntervalUnresolved as an Omit and replacement thing I guess...

I think this would be better – in my head that's the right way round. `TimeIntervalUnresolved` is a special variation of `TimeInterval`, not vice versa (and you don't want people worrying about it when dealing with the normal `TimeInterval` / seeing it lurking at the base of the type)

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
Alfred-Mountfield: I was fairly confused by your comment and after some googling I think I'm about as disappointed in JS as I have ever been. I cannot believe that the functional iterator methods each create a loop. That is insane.

Basically every language I've interacted with (even streams in Java) combines functional iterators that can be combined (for example multiple filters in a row, etc.) so that you are iterating a minimal amount of times and so that the compiler can actually be more confident about how to optimise the code compared to for loops with bound checks and such.

I'm going to search across the code for a variety of these now.

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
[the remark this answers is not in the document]
Alfred-Mountfield: Except it's not a variation, it's not actually a sub-type 🙃 

I'll play around with it a bit

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
CiaranMn: >  I cannot believe that every functional iterator method creates a loop. 

And an array 😁 (at least these ones)

CiaranMn: There are libraries that do what you want and you can write your own with generators, but easiest to go with the trusty `for` loop for now.

Alfred-Mountfield: [Remove deep nesting of rubbish functional iterators](https://github.com/blockprotocol/blockprotocol/pull/928/commits/d9e4be57740420e67d18fcf2e33a405d3fa8eafc)

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
Alfred-Mountfield: I've been slightly concerned about this too.. I hope they'll at least read the documentation? 😅

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
Alfred-Mountfield: Yeah fair enough, I just hate being forced into shoving `!`'s in places where iterator functions can convince the compiler it's okay

Alfred-Mountfield: [Replace slice(-1) syntax](https://github.com/blockprotocol/blockprotocol/pull/928/commits/538ac50b82e2f9d62a01b9c9383f70e2536d0cc2)

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
Alfred-Mountfield: Fixed as driveby in [Replace slice(-1) syntax](https://github.com/blockprotocol/blockprotocol/pull/928/commits/538ac50b82e2f9d62a01b9c9383f70e2536d0cc2)

libs/@blockprotocol/graph/src/types/entity.ts
Alfred-Mountfield: That makes more sense (good spot), both equivalent of course, but that's the problem with structural types 😭 

[Replace unnecessary usages of TimeIntervalUnresolved](https://github.com/blockprotocol/blockprotocol/pull/928/commits/d4abfd57510ebdf2ce7eaf7b6f276afb5c98f030)

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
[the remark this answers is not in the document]
Alfred-Mountfield: It was all vile, opted for the simple thing (which I still really don't like 😭)

[Make TimeInterval independent of TimeIntervalUnresolved](https://github.com/blockprotocol/blockprotocol/pull/928/commits/619b4219291125078e38bb9f68a1be20961c8316)

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
```diff
@@ -54,80 +66,94 @@ export const getEntity = (
...
+      .map((entityVertex) => entityVertex.inner)
+      .filter((entity) =>
+        /*
+        We want to find all entities which were active _at some point_ in the search interval. This is indicated by an
+        overlap.
+         */
+        intervalOverlapsInterval(
+          interval,
+          /*
+           these casts are safe as we check for `targetRevisionInformation === undefined` above and that's only ever
```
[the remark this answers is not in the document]
Alfred-Mountfield: It's a mistake, it should be `interval !== undefined`

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -113,90 +153,172 @@ export const getIncomingLinksForEntity = (
...
+    }
+
+    return getEntityRevisionsByEntityId(
+      subgraph as Subgraph<true>,
+      leftEntityId,
+      intersection,
+    ) as Entity<Temporal>[];
+  } else {
+    return getEntityRevisionsByEntityId(
+      subgraph as Subgraph<true>,
```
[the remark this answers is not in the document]
Alfred-Mountfield: You are very correct

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -113,90 +153,172 @@ export const getIncomingLinksForEntity = (
...
+    }
+
+    return getEntityRevisionsByEntityId(
+      subgraph as Subgraph<true>,
+      rightEntityId,
+      intersection,
+    ) as Entity<Temporal>[];
+  } else {
+    return getEntityRevisionsByEntityId(
+      subgraph as Subgraph<true>,
```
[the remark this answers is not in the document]
Alfred-Mountfield: Same as above (you are very correct), thanks for spotting! Hard to find these things and annoying that TS doesn't care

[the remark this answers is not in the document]
Alfred-Mountfield: [Fix temporal generics](https://github.com/blockprotocol/blockprotocol/pull/928/commits/5c873ea94626cb7bb956deb320d7e6863307f42a)

libs/@blockprotocol/graph/src/stdlib/subgraph/edge/link-entity.ts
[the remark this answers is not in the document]
Alfred-Mountfield: [Fix temporal generics](https://github.com/blockprotocol/blockprotocol/pull/928/commits/5c873ea94626cb7bb956deb320d7e6863307f42a)

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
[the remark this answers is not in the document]
Alfred-Mountfield: [Fix incorrect safety comment](https://github.com/blockprotocol/blockprotocol/pull/928/commits/49ee03b2c02fb76a17934ae0bbae7125b52158cb)

libs/@blockprotocol/graph/src/types/temporal-versioning.ts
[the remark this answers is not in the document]
CiaranMn: We can always change again later without affecting users – looks good for now!

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
CiaranMn: I think [getEntities](https://github.com/blockprotocol/blockprotocol/blob/d9e4be57740420e67d18fcf2e33a405d3fa8eafc/libs/%40blockprotocol/graph/src/stdlib/subgraph/element/entity.ts#L23) is also a good candidate for optimisation, but we can leave it to follow up if performance is actually an issue

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
```diff
@@ -1,51 +1,62 @@
...
  * @param latest - whether or not to only return the latest revisions of each entity
  */
-export const getEntities = (
-  subgraph: Subgraph,
+export const getEntities = <Temporal extends boolean>(
+  subgraph: Subgraph<Temporal>,
   latest: boolean = false,
-): Entity[] => {
+): Entity<Temporal>[] => {
   return Object.values(
```
CiaranMn, line 23: This outer `Object.values` is pointless – it's recreating the array that the inner `Object.values` returns

CiaranMn (approved):

libs/@blockprotocol/graph/src/stdlib/subgraph/element/entity.ts
Alfred-Mountfield: Again good spot, missed it in refactoring 
[Clarity and remove unnecessary call](https://github.com/blockprotocol/blockprotocol/pull/928/commits/64d3c6c1f2b876a3c2481779ae9cad44fba8d36c)

libs/@blockprotocol/graph/src/types/subgraph/element-mappings.ts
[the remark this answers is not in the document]
Alfred-Mountfield: This was done

PR #929, opened 2023-02-01
Serve data types and link entity type as JSON

CiaranMn:

We are introducing a new type system to the Block Protocol, whereby entity types are made up of property types, which refer to data types (e.g. text, number).

At first, entity types and property types will be manageable via UI that inserts them into the db (coming soon), but data types will be limited to a fixed number of primitive types, with no support for more.

This PR introduces the fixed, initial data types and the ability to request them as JSON at their `$id`.

It also does the same for a special entity type which is relied on to indicate which entities represent relationships/links between entities (we will migrate this to the db once support for new entity types is added, but it's useful to have hardcoded here for now).

Although this relates to the new type system, introducing it into `main` now causes no issues as it modifies no existing functionality, and it allows us to refer to these types at `blockprotocol.org` rather than on some staging deployment.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203814168312947/f) _(internal)_

## 🔍 What does this change?

- Adds the JSON for the hardcoded types
- Adds middleware which returns the type if a relevant URL is requested with "application/json" in the accept header
- Upgrades to Next 13.1, which introduces support for responding from middleware (`transpilePackages` is now also out of experimental)

- Not yet.

## ❓ How to test this?

```js
const previewDeplyoment = // preview deployment url
fetch(
  `${previewDeplyoment}/@blockprotocol/types/data-type/text/v/1`, 
  { headers: { accept: "application/json" }}
);
```


apps/site/package.json
```diff
@@ -66,7 +67,7 @@
     "mime-types": "^2.1.35",
     "mock-block-dock": "0.0.39",
     "mongodb": "^4.2.2",
-    "next": "13.0.7",
+    "next": "13.1.0",
```
kachkaev, line 70: `next@13.1.6` is out, which may contain bugfixes in `transpilePackages`. I’d suggest upgrading, unless you see reasons to stay on `13.1.0`.

apps/site/src/middleware.page.ts
```diff
@@ -4,6 +4,10 @@ import type { NextRequest } from "next/server";
 import { NextResponse } from "next/server";
 
 import { SESSION_COOKIE_NAME } from "./lib/api/middleware/constants";
+import {
+  returnTypeAsJson,
+  versionedTypeUriRegExp,
+} from "./middleware/return-types-as-json";
```
kachkaev, line 10:
```suggestion
} from "./middleware.page/return-types-as-json";
```

kachkaev (changes requested):
Playwright tests are failing because of compatibility issues with coverage collection and middleware. This can be fixable by adding `/* istanbul ignore file */` to `apps/site/src/middleware.page/**`.

Context: https://github.com/blockprotocol/blockprotocol/pull/520#discussion_r953779484

kachkaev (approved):

PR #931, opened 2023-02-02
Add support for bi-temporal versioning in MBD

Alfred-Mountfield:

#928 updated the Graph Service to have support for bi-temporal versioning, this follows through and migrates MBD to support it. This should prove as a good example, for EA developers, as to how to implement versioned subgraph traversal, and how to still support blocks that do not require/support temporal versioning in environments that do support it.

This PR currently _fully migrates_ MBD over to temporal versioning, this is not the long (or even medium) term ideal, and there are a few places in the code where non-temporal logic has been omitted due to this (highlighted with `todo`s).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543026679373/1203528958301161/f) _(internal)_

## 🚫 Blocked by

- [x] #928 

## 🔍 What does this change?

As with most of my PRs, please see commits and review _by_ commit. I have tried to lead the review process in an understandable way for the most part.

- As discussed in previous PRs, this absolutely will need changes to the docs. This is deferred to a further point, but effort has been made to have good in-code documentation.

## ⚠️ Known issues

- The previous `Subgraph` traversal logic had a really sneaky bug that meant that it could explore much further than the depths that were passed in. This could happen when there were cycles of edges in the Graph. As such, this PR reimplements the `traversal` logic with a (mostly) plain, pure recursive, implementation. This implementation should be treated almost as the pseudo-code specification of what BP `Subgraph` traversal currently is. In a follow-up we shall be trying to remove most of the recursion and **massively** speed up the traversal implementation. (These optimisations will be encapsulated in a PR and should prove helpful for other EAs looking to optimise their own implementations, and we intend to port over some of that logic back to our implementation in HASH)
- As mentioned above, temporal versioning support has not been made configurable as part of this PR, there has been a full-blown conversion of MBD to temporal versioning. This does _not_ include the bundled blocks, as it is important that temporal-version-supporting EAs can still use blocks that are not opinionated about such things. In the close future we should update MBD to make this configurable, so that people have an appropriate test environment to see how their blocks behave in a variety of EAs. This is especially important as we do not currently plan to have a temporal datastore for the BP WordPress plugin.
- Textual blocks that push a lot of `updateEntity` operations will absolutely blow-up the DB, these include our default ones. Due to capacity constraints and pragmatism, we haven't put an immense amount of effort into optimising, this does mean that MBD starts to chug if the datastore becomes too large, and we do not have an option to wipe it at the moment
  image
  - Exposing the potential side-effects of spamming with `updateEntity` operations is likely to be a positive/beneficial thing, and has identified a reason for us to update our simple example blocks with some aggregation/debounce logic to group update operations to avoid pushing such a high load onto EAs.
- The code that generates block types from type system schemas was broken with a previous PR. This will need fixing in a follow-up 
- The HTML block in MBD has not been updated to the type system, it is still accessing properties based on plain keys rather than Base URIs

## 🐾 Next steps

- Fix type generation code
- Fix the HTML block
- Publish new canary versions of these packages
- Figure out a way to expose a temporal and non-temporal interface to the graph package
- Move past pure recursive subgraph traversal in MBD
- Implement non-temporal traversal logic in MBD
- Make the temporal datastore in MBD toggleable

## 🛡 What tests cover this?

- I'm unsure, at the very least tsc and eslint. 
- I have also manually tested MBD and it seems to work as intended.

## ❓ How to test this?

Run MBD as usual, try some blocks, see how they perform

## 📹 Demo

image
image
image


libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -1,284 +1,543 @@
...
+      if (!isEntityVertex(source)) {
+        return [];
+      }
+
+      const { inner: sourceEntity } = source;
+
+      const sourceEntityId = sourceEntity.metadata.recordId.entityId;
+
+      if (reversed) {
+        // get outgoing links for entity
```
TimDiekmann, line 125: The function name is pretty descriptive, I don't think this comment is needed here as it is. I'd say we either should remove it or give a reason, why we're getting outgoing links, if `"HAS_LEFT_ENTITY"` is reversed.

libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -1,284 +1,543 @@
...
+      if (!isEntityVertex(source)) {
+        return [];
+      }
+
+      const { inner: sourceEntity } = source;
+
+      const sourceEntityId = sourceEntity.metadata.recordId.entityId;
+
+      if (reversed) {
+        // get incoming links for entity
```
TimDiekmann, line 202: Same as above

libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -1,284 +1,543 @@
...
+        return rightEntityRevisions.map((entity) => ({
+          kind: "entity",
+          inner: entity,
+        }));
+      }
+
+      return [];
+    }
+    default: {
+      throw new Error(`Currently unsupported traversal edge kind: ${edgeKind}`);
```
TimDiekmann, line 270: Is it intended, that we cannot traverse from `entity` to `entityType`? Will this be added in a follow-up?

libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -1,284 +1,543 @@
...
-                    editionId: edgeTimestamp,
-                  },
-                },
-              } = leftEntity;
+        if (isEntityVertex(neighborVertex)) {
+          // get from temporal data of the neighbor vertex
+          const entityValidInterval =
+            neighborVertex.inner.metadata.temporalVersioning[
+              traversalSubgraph.temporalAxes.resolved.pinned.axis
+            ];
```
TimDiekmann, line 328: I cannot really follow this code. What is `entityValidInterval` in this context? Why you're indexing via the `pinned` axis?

libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -1,284 +1,543 @@
...
+            element: neighborVertex,
+            elementIdentifier: neighborVertexId,
+            interval: newIntersection,
+            currentTraversalDepths: {
+              ...currentTraversalDepths,
+              [edgeKind]: {
+                ...currentTraversalDepths[edgeKind],
+                [direction]: depth - 1,
+              },
+            },
```
TimDiekmann, line 358: nit: I don't know what's common in TS, but I prefer to first write the changed values, then the fallback, so something like
```suggestion
            currentTraversalDepths: {
              [edgeKind]: {
                [direction]: depth - 1,
                ...currentTraversalDepths[edgeKind],
              },
              ...currentTraversalDepths,
            },
```

Alfred-Mountfield: That would result in the fields being overridden https://stackoverflow.com/a/63238111/14687716

libs/mock-block-dock/src/datastore/traverse.ts
Alfred-Mountfield: I agree with this, this is a left-over from a previous implementation which didn't have as descriptive function names directly below, have removed

libs/mock-block-dock/src/datastore/traverse.ts
Alfred-Mountfield: Same as above, have removed

libs/mock-block-dock/src/datastore/traverse.ts
TimDiekmann: Well, this would be suboptimal 🙃

libs/mock-block-dock/src/datastore/use-mock-datastore.ts
```diff
@@ -220,13 +225,47 @@ export const useMockDatastore = (
...
+            // Update the old entity revision's end times
+            const updatedOldRevisionEndBound = {
+              kind: "exclusive",
+              limit: currentTime,
+            } as const;
+
+            currentEntity.metadata.temporalVersioning.decisionTime.end =
+              updatedOldRevisionEndBound;
+            currentEntity.metadata.temporalVersioning.transactionTime.end =
+              updatedOldRevisionEndBound;
```
CiaranMn, line 245: We shouldn't do this because it is mutating an object in React state, which should be immutable (it might lead to weird bugs if other state updates happen around the same time)

I would do this at the top
```ts
const newSubgraph = JSON.parse(JSON.stringify(currentGraph));
```
then mutate `newSubgraph` as needed. the current assignment of `newSubgraph` below can be removed. Where we just return `currentGraph` unmodified that can stay the same, as it will skip a re-render.

The other option is to introduce a library like [immer](https://immerjs.github.io/immer/) but it seems like overkill for this one change.

libs/mock-block-dock/src/datastore/use-mock-datastore.ts
```diff
@@ -220,13 +225,47 @@ export const useMockDatastore = (
                   }
                 : undefined;
 
-            const updatedEntity: Entity = {
+            /**
+             * @todo - this is 'safe' because `getEntityRevisionFromSubgraph` returns the object from within the
+             *   `Subgraph`. If it did something like spreading and creating a new object then this wouldn't mutate
+             *   the vertex in the original graph. Are we happy with this, or should we do the messy code to refetch
+             *   it here again so we can be sure we're updating the object in the subgraph.
+             */
```
CiaranMn, line 233: How messy would it be?

Could it be something like:
```ts
newSubgraph.vertices = newSubgraph.vertices.map(vertex => {
  if (vertex.id !== idWeCareAbout) {
    return vertex;
  }
  return { ... }; // the updated object
});

File: libs/mock-block-dock/src/mock-block-dock.tsx [hunk 8]
```diff
@@ -48,8 +50,11 @@ type MockBlockDockProps = {
   blockEntityRecordId?: EntityRecordId;
   debug?: boolean;
   hideDebugToggle?: boolean;
-  initialEntities?: Entity[];
-  // initialLinkedAggregations?: LinkedAggregationDefinition[];
+  initialData?: {
+    initialTemporalAxes: QueryTemporalAxes;
```
CiaranMn, line 54: Does this need to be required? What would happen if we let it default to 'from 1970 to now'? What would happen if we didn't let people specify it at all?

libs/mock-block-dock/dev/dev-app.tsx
```diff
@@ -11,49 +15,89 @@ import { TestReactBlock } from "./test-react-block";
 
 const node = document.getElementById("app");
 
+const intervalForAllTime =
```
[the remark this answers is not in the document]
Alfred-Mountfield: I thought about it but I'm not sure about `Date(0)` yet

libs/mock-block-dock/src/datastore/hook-implementations/entity/aggregate-entities.ts
```diff
@@ -7,55 +7,98 @@ import {
...
+
+export const aggregateEntities = <Temporal extends boolean>(
+  data: AggregateEntitiesData<Temporal>,
+  graph: Subgraph<true>,
+): AggregateEntitiesResult<
+  Temporal,
+  Subgraph<Temporal, EntityRootType<Temporal>>
+> => {
+  // this cast is safe as we're only checking against undefined
+  if ((data as AggregateEntitiesData<true>).temporalAxes !== undefined) {
```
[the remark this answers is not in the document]
Alfred-Mountfield: I think this is actually a lot worse (and I originally had what you had written and removed it). 

`"temporalAxes" in data` completely breaks type checking. If we decided to rename `data.temporalAxes` to `data.timeAxes` for example, `tsc` would not warn us here at all.

I think the assertion actually slightly improves on conveying the intention here.

libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -1,284 +1,539 @@
...
-  traversalSubgraph: Subgraph,
-  elementIdentifier: GraphElementVertexId,
-  datastore: Subgraph,
-  traversalContext: TraversalContext,
-  currentTraversalDepths: PartialDepths,
-) => {
-  const unresolvedDepths = traversalContext.insert(
-    elementIdentifier,
-    currentTraversalDepths,
+const TIMESTAMP_PLACEHOLDER = `TIMESTAMP_PLACEHOLDER` as const;
```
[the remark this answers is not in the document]
Alfred-Mountfield: I had backticks because I was going to add `Date.new().toISOString()` (or some other unique identifier) within the template, and decided against the need for it. Will remove, thanks for highlighting

libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -1,284 +1,539 @@
...
+};
+
+export const isTemporalTraversalSubgraph = <
+  RootType extends SubgraphRootType<boolean> = SubgraphRootType<boolean>,
+>(
+  traversalSubgraph: TraversalSubgraph<boolean, RootType>,
+): traversalSubgraph is TraversalSubgraph<true, RootType> => {
+  // this cast is safe as we're only checking against undefined
+  return (
+    (traversalSubgraph as TraversalSubgraph<true>).temporalAxes !== undefined
```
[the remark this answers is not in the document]
Alfred-Mountfield: I don't think this will avoid a crash, because if the assertion is incorrect (and the field does not exist) then it will be equivalent to undefined `({} as any).foo === undefined`

libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -1,284 +1,539 @@
...
+  datastore: Subgraph<true>,
+  source: Vertex<true>,
+  edgeKind: EdgeKind,
+  reversed: Reversed,
+  interval: TimeInterval,
+): Vertex<true>[] => {
+  switch (edgeKind) {
+    case "HAS_LEFT_ENTITY": {
+      if (!isEntityVertex(source)) {
+        return [];
```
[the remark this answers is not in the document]
Alfred-Mountfield: Not really, this means we can avoid needing to `switch` in the calling context. This edge wasn't discovered while traversing, (as in this isn't "for edges from property type", in which case finding a "has left entity" edge would be a bug) it's just used to say "for every edge kind, get me the edges from X", in which case it's fine to say "there are no left entities from this property type".

libs/mock-block-dock/src/datastore/traverse.ts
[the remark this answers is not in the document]
Alfred-Mountfield: [Remove unnecessary template string](https://github.com/blockprotocol/blockprotocol/pull/931/commits/c02858e576fef2ce43c3d50b4b16c6995abb23a3)

libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -1,284 +1,539 @@
...
-  const unresolvedDepths = traversalContext.insert(
-    elementIdentifier,
-    currentTraversalDepths,
+const TIMESTAMP_PLACEHOLDER = `TIMESTAMP_PLACEHOLDER` as const;
+
+/**
+ * Advanced type to recursively search a type for `EntityValidInterval` and patch those occurrences by removing the
+ * definition of the "validInterval" property.
+ */
+type PatchEntityValidInterval<ToPatch extends unknown> = ToPatch extends object
```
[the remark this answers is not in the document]
Alfred-Mountfield: Good suggestion, cheers
[Rename PatchEntityValidInterval to DeepOmitValidInterval](https://github.com/blockprotocol/blockprotocol/pull/931/commits/aad4de580ad8222e9e620ab86346dee8fe52dbf1)

libs/mock-block-dock/src/datastore/use-mock-datastore.ts
Alfred-Mountfield: Ah bugger. I'll do as you suggest for now but it's really not fab to have to serialize and deserialize an entire datastore every time there's an update 🙃.

So, turns out we were already doing it.. I took a stab at fixing it without going overboard on unergonomic code
[Stop mutating react state](https://github.com/blockprotocol/blockprotocol/pull/931/commits/afd5379769edf8ef9c896a0be2a1dd0720507c5e)

libs/mock-block-dock/src/datastore/use-mock-datastore.ts
Alfred-Mountfield: See
[Stop mutating react state](https://github.com/blockprotocol/blockprotocol/pull/931/commits/afd5379769edf8ef9c896a0be2a1dd0720507c5e)

It's pretty gross and depends on you knowing to look up the entity by its starting bound of the temporal metadata associated with the variable axis of the subgraph.

libs/mock-block-dock/src/mock-block-dock.tsx
Alfred-Mountfield: I tried thinking about the consequences of this and it's a bit complicated to decide how much it matters so I opted for making it required for now.

I did write a fair chunk of code before to try and make a best guess of a temporal axis from a given data set (e.g. `entity[]`) but it's impossible to determine the correct range from a single `Subgraph` of data. Opting to make it all time is _probably_ fine, but may not be, especially if we change decisions about invariants of the subgraph and say that `rightEntity` or `leftEntity`s don't need to exist in the edge sets, etc. because at that point those invariants might be broken because of the temporal query ranges and one might need to be able to look at the ranges to determine if that's the case.

And probably a bunch more things that I'm not sure we have capacity to think about right now.

CiaranMn: We can come back to this when we look at making MBD accept non-temporal data

libs/mock-block-dock/src/datastore/use-mock-datastore.ts
```diff
@@ -220,23 +227,62 @@ export const useMockDatastore = (
                   }
                 : undefined;
 
-            const updatedEntity: Entity = {
+            // This is fairly slow, we could leverage something like `immer` to make these partial updates more
```
CiaranMn, line 230: I don't think immer would be faster and would increase bundle size more than a single deep clone function so have suggested tweaking this
```suggestion
            // @todo test performance – if an issue, test against a different deep clone strategy (e.g. [lodash's](https://www.npmjs.com/package/lodash.clonedeep))
```

Alfred-Mountfield: I've updated it, but removed the link and the `todo`. I'd prefer not to add TODOs unless we think they're actually worth doing. If we were trying to investigate slow performance I think we'd find the comment, and I don't think it being a todo makes it much more discoverable, just adds to the endless pit of lost intentions 😅 

[Improve comment](https://github.com/blockprotocol/blockprotocol/pull/931/commits/77e5ceb8ad2bd7d9871543a1d6fe2c74a77c79ac)

PR #932, opened 2023-02-02
Improve reliability of core service connection between block/embedder

nathggns:

A number of changes to ensure that messages between embedder/block are more reliably processed. See "What does this change" below. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056186/1203474494377435/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- `CoreBlockHandler` repeatedly sends `init` until it gets a response, which should happen within two microtasks in the traditional case. *We may want to consider what happens with iframes*
- `CoreBlockHandler` now only sends `init` on the first call to `initialize` (each service calls this independently, which meant  it was being sent twice when using both graph and hook services)
- `CoreEmbedderHandler` now only triggers `getInitPayload` on each service once, remembers its `initResponse`, and will resend it every time it receives `init`. 
- `CoreHandler` now queues non-init messages until after `CoreBlockHandler` or `CoreEmbedderHandler` has indicated it is ready to handle messages
- Fixed a bug in `CoreHandler` where async use of `sendMessage` for `init` would never resolve the promise, which was necessary for `CoreBlockHandler` to repeatedly send `init`. 
- Moved `initialisedRef` setting to `true` in `useServiceConstructor` higher to reduce possibility of a single service calling `initialize` multiple times 

N/A

## ⚠️ Known issues

- Each services calls `initialize` on `CoreHandler`, even if its already been `initialized` – we have to handle this ourselves
- Functions to resolve promises for requests which were sent too early and therefore will never get a response are never 'forgotten', which is possibly a memory issue, but unlikely 
- We repeat the init message every microtask until we get a response. This is fine outside a sandbox as event handlers are processed synchronously. This may be an issue in sandboxed environments. 

## 🛡 What tests cover this?

None, I think? 

## ❓ How to test this?

Revert this commit – 9caec060615adf91dbfc86e7d73dc72a19b4982c – and check console log to ensure messages are being passed and received properly.


libs/mock-block-dock/src/use-send-graph-value.ts
```diff
@@ -29,7 +29,7 @@ export const useSendGraphValue = ({
       (!sentInitially.current ||
         JSON.stringify(value) !== JSON.stringify(sentValue.current))
     ) {
-      graphService[valueName](value as any); // @todo how to maintain the union GraphValue is initially set to?
+      graphService[valueName]({ data: value as any }); // @todo how to maintain the union GraphValue is initially set to?
```
CiaranMn, line 32: This is an existing bug I had to fix to test properly – it was causing the actual content of messages to be `undefined` (e.g. the contents of `graph` in `initResponse`). Seemed like a good place to get it in.

libs/@blockprotocol/core/src/core-embedder-handler.ts
```diff
@@ -61,11 +63,18 @@ export class CoreEmbedderHandler extends CoreHandler {
...
-      data[serviceName] = serviceInstance.getInitPayload();
+    let data = this.initResponse;
+
+    if (!data) {
+      data = {};
+
+      for (const [serviceName, serviceInstance] of this.services) {
+        data[serviceName] = serviceInstance.getInitPayload();
+      }
     }
```
CiaranMn, line 74: What was the motivation behind this caching? I'm wondering if it's worth it – in theory, the payload could become out of alignment with the properties it is derived from. It might be better to just let the service provide the freshest version of its init payload when needed. 

Is it correct to say that this would only ever be called twice by a block that isn't using the new `CoreBlockHandler` implementation? Since it will only ever send `init` once

CiaranMn (approved):
Tested this and works well – I fixed a pre-existing bug to do with the messages not having the proper content.

One question but needn't block merging.

libs/@blockprotocol/core/src/core-embedder-handler.ts
nathggns: Safety/worries about spec compliance, mostly. I wasn't sure of the possible consequences of a service being initiated multiple times, so I wanted safety on both ends this wouldn't happen. Additionally because of the timing uncertainties around sandboxing, it might actually be possible the init message gets sent by `CoreBlockHandler` multiple times (as it gets sent a second time before the response to the first message gets back) so I wanted to create more certainty around how it would behave.

PR #935, opened 2023-02-02
Update FAQs

nonparibus:

- Updates our FAQs to add initial endpoint questions
- Drive-by improvements to the repo's root README


apps/site/src/_pages/docs/5_faq.mdx
```diff
@@ -134,9 +134,15 @@ Because those inserting blocks are often non-technical end users, this isn't alw
   - which blocks users are able to discover and insert within your environment
   - which blocks a user inserts are ultimately able to be run in either a sandboxed or unsandboxed fashion
 
-Today you might choose to allow/deny blocks on the basis of their authorship (published namespace), or based on their _verification status_. In the future we also plan to expose additional Þ _Hub_ metadata which may prove useful in the creating and maintaining whitelists.
+Today you might choose to allow/deny blocks on the basis of their authorship (published namespace), or based on their _verification status_.
+  
+In the future we also plan to expose:
+- additional Þ _Hub_ metadata which may prove useful in the creating and maintaining whitelists;
+- permission grants for blocks and block types, that allow for users to control the level and types of access blocks have (e.g. the ability to access information created through the Graph Service outside of themselves) or access to other services (e.g. making network requests).
```
CiaranMn, line 141: This would involve a more complicated sandboxing solution we'll need to think about further.
```suggestion
- permission grants for blocks and block types, that allow for users to control the level and types of access blocks have (e.g. the ability to access information created through the Graph Service outside of themselves) or access to other services.
```

apps/site/src/_pages/docs/5_faq.mdx
```diff
@@ -209,6 +209,30 @@ We do not yet believe it necessary to specify block-application requests in Grap
...
+- **through the Þ Hub:** in addition to blocks and types, endpoints (APIs) can also be made accessible via the Þ Hub. A limited set of endpoints will be available as of 28th February 2023, including ones providing access to _Mapbox_ and _OpenAI_.
+  
+Facilitating external API access via the Þ Hub is a new addition to the protocol, and intructions for making additional endpoints available will be added in due course. In the meantime, we're able to manually add endpoints on behalf of block developers. If there's an endpoint you'd like to use within a new or existing block, please [message us on Discord](https://blockprotocol.org/discord).
+
+</FAQ>
+  
+<FAQ question="Can blocks access external APIs while authenticating as users?">
+  
+The ability for Þ blocks to access external services while (optionally) authenticated as an individual user is a planned capability. A technical solution has been identified which will enable this securely and seamlessly across embedding environments, and we intend to introduce support in the coming months. This will enable blocks to be developed that allow for a much wider range of use-cases to be addressed.
```
CiaranMn, line 228: I would feel better about this as I don't think we've fully detailed the solution yet, and it makes the same point
```suggestion
The ability for Þ blocks to access external services while (optionally) authenticated as an individual user is a capability which we intend to introduce in the coming months. This will enable blocks to be developed that allow for a much wider range of use-cases to be addressed.
```

apps/site/src/_pages/docs/5_faq.mdx
```diff
@@ -209,6 +209,30 @@ We do not yet believe it necessary to specify block-application requests in Grap
...
+  
+**Example:** a block user has information synced to _Coda_ or _AirTable_ from multiple external locations. They want to display this data on their website, which uses WordPress. With the _Block Protocol for WordPress_ plugin, a user will be able to use a _Coda_ or _AirTable_ block which allows them to select information contained within their external account, and insert this information onto their page or post in WordPress, which can then automatically keep itself in sync with the original source. This allows publishing from a wide variety of external sources.
```
CiaranMn, line 230: Or similar – we need to edit this section anyway to remove language to 'planned capability' so we could make it clear also.
```suggestion
**Example:** a block user has information synced to _Coda_ or _AirTable_ from multiple external locations. They want to display this data on their website, which uses WordPress. With the _Block Protocol for WordPress_ plugin, a user will in the future be able to use a _Coda_ or _AirTable_ block which allows them to select information contained within their external account, and insert this information onto their page or post in WordPress, which can then automatically keep itself in sync with the original source. This allows publishing from a wide variety of external sources.
```

CiaranMn (approved):

CiaranMn (approved):

PR #936, opened 2023-02-02
Remove hub directory (switch to publishing blocks via API)

kachkaev:

This PR completes the migration from the `/hub` folder and moves all blocks to the database. Unused code is removed, and some tests are updated accordingly.

## 🔗 Related links

## 🔍 What does this change?

- Local DX is improved because blocks are mirrored from the production hub instead of built locally. This speeds up the warm-up phase for the app. Tests and Vercel builds speed up as well (especially those with cold cache).

- Tests assert more hardcoded values than before. They used to be taken from `apps/site/blocks-data.json` which was removed. Using hardcoded values in tests makes them closer to what a human would be testing, thus making the logic more resilient (we are not checking 🐞 === 🐞). If we need value flexibility in the future, we can replace string comparisons with regexps etc.

- Yes

## ⚠️ Known issues

- I'm unable to use ``blockprotocol.hashai.workers.dev`` in `blockTarballUrls` due to a chicken-and-egg situation. See Next steps.

- Firefox tests are failing, but they are failing on `main` too.

## 🐾 Next steps

1. Tell developers to delete `apps/site/blocks-data.json` and `apps/site/public/blocks` locally to make sure these files are not committed by accident.

1. Publish blocks to prod db (repeat https://github.com/hashintel/hash/actions/runs/4277684057 but via https://github.com/hashintel/hash/actions/workflows/publish-blocks-to-production.yml)

1. Replace `blockprotocol-preview.hashai.workers.dev` with `blockprotocol.hashai.workers.dev` in `blockTarballUrls` in a follow-up PR.

1. Re-publish third-party blocks using a manual workflow (TBD by @CiaranMn)

## 🛡 What tests cover this?

- CI (Playwright tests)

## ❓ How to test this?

1.  Follow updated instructions in README
1.  Observe blocks source from local DB / S3

## 📹 Demo

Screenshot 2023-02-27 at 00 38 09


apps/site/src/lib/blocks.ts
```diff
@@ -313,15 +313,6 @@ export const retrieveBlockReadme = async (
   blockMetadata: ExpandedBlockMetadata,
```
CiaranMn, line 313: @kachkaev to note that there are still some `readFileSync` /`readFile` calls above that can be removed when we remove locally-built blocks

kachkaev: Thanks! Removed in 94c0630e6446a703cb7ec53b3a2245be75d98556

apps/site/tests/shared/wrapped-playwright.ts
```diff
@@ -32,7 +32,7 @@ const tolerableSharedConsoleMessageMatches: RegExp[] = [
 // Messages that only show up in `yarn dev` (not CI)
 if (!process.env.CI) {
   tolerableSharedConsoleMessageMatches.push(
-    /Image with src "\/_next\/static\/media\/primary-helix-min\.\w+\.png" was detected as the Largest Contentful Paint \(LCP\)\. Please add the "priority" property if this image is above the fold\./, // https://nextjs.org/docs/api-reference/next/legacy/image#priority
+    /Image with src "\/_next\/static\/media\/.*" was detected as the Largest Contentful Paint \(LCP\)\. Please add the "priority" property if this image is above the fold\./, // https://nextjs.org/docs/api-reference/next/legacy/image#priority
```
kachkaev, line 35: Fixes `"[warning] Image with src \"/_next/static/media/helix-boxes.33d40be1.webp\" was detected as the Largest Contentful Paint (LCP). Please add the \"priority\" property if this image is above the fold.
+ Read more: https://nextjs.org/docs/api-reference/next/legacy/image#priority",`

Screenshot 2023-02-27 at 01 18 28

CiaranMn (approved):

PR #937, opened 2023-02-03
Add `"additionalProperties": false` to Entity Types and fix broken links definition

Alfred-Mountfield:

Entity types should comprehensively define the contents of an Entity. This should be represented within the schema, but due to an oversight we had omitted specifying `"additionalProperties": false`, this modifies the Type System to enforce it.

The `links` definition inside `EntityType` allowed `| {}` inside it. This meant that TypeScript allowed any fields within the object due to quirks with how it handles the object type. This has also been fixed (see demo below for more explanation).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203602044074391/f) _(internal)_
- [Asana task](https://app.asana.com/0/1202805690238892/1203466297702512/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Requires `"additionalProperties": false` to be present in all Entity Types.
- Updates the parsing logic in the Type System crate to correctly return an error if this is not the case
- Fixes an empty object in the `links` object causing TS to allow anything.

This crate/package at the moment is treated as the source of truth and spec, the types being updated therefore constitute a change to the docs for now.

## ⚠️ Known issues

N/A

## 🐾 Next steps

N/A

## 🛡 What tests cover this?

- Type System crate tests
- Type system node package tests

## ❓ How to test this?

1.  `yarn` should pass, rebuilding the Type System crate
1.  `yarn workspace @blockprotocol/type-system test` should pass
1.  `cargo make test` should pass inside the `type-system/crate` directory 

### Broken `links`

**Previously**
image
image

**After**
image
image

### DIFF
--- a/.changeset/loud-chicken-count.md
+++ b/.changeset/loud-chicken-count.md
@@ -1,0 +1,5 @@
+---
+"@blockprotocol/type-system": patch
+---
+
+Require `"additionalProperties": false` on entity types
--- a/.changeset/silver-dogs-march.md
+++ b/.changeset/silver-dogs-march.md
@@ -1,0 +1,5 @@
+---
+"@blockprotocol/type-system": patch
+---
+
+Fix empty object in links definition inside EntityType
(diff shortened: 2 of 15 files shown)


TimDiekmann (approved):
✅ Rust implementation LGTM

CiaranMn (approved):

PR #939, opened 2023-02-06
Fix problematic typings in `Subgraph`

Alfred-Mountfield:

The `Subgraph` object contains correct information about the keys used as identifiers for `vertices` and `edges`. Namely, that elements are identified by the following pairs 
- `(EntityId, TimestampString)`, 
- `(BaseUri, number)`

In an ideal world, this would encapsulate quite a lot more information about a correctly formed `Subgraph`. It would greatly reduce the number of type guards that would need to be called in TS, for example (`isEntityVertex`), as the code path would know if it's dealing with an ontology element or a knowledge-graph one.

Unfortunately, the `number` type confuses the heck out of TypeScript and it actually can cause code to be completely incorrect at the worst of times.

This is because when the `number` type is used as a key in an object, it is inferred as being equivalent to the string `` `${number}` `` which overlaps the string type used for entity related things. But when seeing a plain string index, TS will assume it's an entity-related thing, rather than correctly inferring it could be either as it _could_ overlap the `number` string.

As part of this PR, I investigated trying to have keys that completely do not overlap (either through additional template strings, or some form of type branding), but eventually needed to opt for removing the helpful type information. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543026679380/1203904064506712/f) _(internal)_

## 🚫 Blocked by

- [x] #931 

## 🔍 What does this change?

As usual, see commits, I will open comments directly on code that needs specific attention.

These changes are mostly internal, and as we haven't published the `Subgraph` type yet, this _change_ shouldn't need additional documentation besides the in-code one added.

## ⚠️ Known issues

Will open comments on specific parts of the code which posed known issues, otherwise N/A

## 🐾 Next steps

N/A

## 🛡 What tests cover this?

- `tsc`
- `eslint`
- Manually running MBD

## ❓ How to test this?

1.  See CI
2. Checkout the branch and ensure `yarn` works
3. Try running MBD and ensure it all behaves as expected (noting that a few of the blocks are broken on this branch)

## 📹 Demo

image
image

### DIFF
--- a/libs/@blockprotocol/graph/src/shared.ts
+++ b/libs/@blockprotocol/graph/src/shared.ts
@@ -42,3 +42,26 @@ export const typedKeys = <T extends {}>(object: T): Entry<T>[0][] => {
 export const typedValues = <T extends {}>(object: T): Entry<T>[1][] => {
   return Object.values(object);
 };
+
+/**
+ * Checks if a given string is exactly a non-negative integer.
+ *
+ * For example, it will accept strings such as:
+ * - "0"
+ * - "1"
+ * - "94818981"
+ *
+ * And it will not accept strings such as:
+ * - "0.0"
+ * - "1.0"
+ * - "-1"
+ * - "foo"
+ * - "0foo"
+ * - "1.1"
+ *
+ * @param {string} input
+ */
+export const stringIsNonNegativeInteger = (input: string) => {
+  const asInteger = Number.parseInt(input, 10);
+  return asInteger.toString() === input && asInteger >= 0;
+};
--- a/libs/@blockprotocol/graph/src/shared.ts
+++ b/libs/@blockprotocol/graph/src/shared.ts
@@ -27,6 +27,18 @@ type Entry<T extends {}> = T extends readonly [unknown, ...unknown[]]
  *
  * Source: https://dev.to/harry0000/a-bit-convenient-typescript-type-definitions-for-objectentries-d6g
  */
-export function typedEntries<T extends {}>(object: T): ReadonlyArray<Entry<T>> {
+export const typedEntries = <T extends {}>(
+  object: T,
+): ReadonlyArray<Entry<T>> => {
   return Object.entries(object) as unknown as ReadonlyArray<Entry<T>>;
-}
+};
+
+/** `Object.values` analogue which returns a well-typed array */
+export const typedKeys = <T extends {}>(object: T): Entry<T>[0][] => {
+  return Object.keys(object) as Entry<T>[0][];
+};
+
+/** `Object.values` analogue which returns a well-typed array */
+export const typedValues = <T extends {}>(object: T): Entry<T>[1][] => {
+  return Object.values(object);
+};
--- a/libs/@blockprotocol/graph/src/shared.ts
+++ b/libs/@blockprotocol/graph/src/shared.ts
@@ -61,7 +61,9 @@ export const typedValues = <T extends {}>(object: T): Entry<T>[1][] => {
  *
  * @param {string} input
  */
-export const stringIsNonNegativeInteger = (input: string) => {
+export const stringIsNonNegativeInteger = (
+  input: string,
+): input is `${number}` => {
   const asInteger = Number.parseInt(input, 10);
   return asInteger.toString() === input && asInteger >= 0;
 };
--- a/libs/@blockprotocol/graph/src/types/ontology.ts
+++ b/libs/@blockprotocol/graph/src/types/ontology.ts
@@ -31,3 +31,12 @@ export const isOntologyTypeRecordId = (
     typeof recordId.version === "number"
   );
 };
+
+/**
+ * The second component of the [{@link BaseUri}, RevisionId] tuple needed to identify a specific ontology type vertex
+ * within a {@link Subgraph}. This should be the version number as a string.
+ *
+ * Although it would be possible to create a template literal type, this confuses TypeScript when traversing the
+ * {@link Subgraph} in generic contexts, whereby it then thinks any string must relate to a {@link EntityVertex}.
+ */
+export type OntologyTypeRevisionId = string; // we explicitly opt not to use `${number}`
--- a/libs/@blockprotocol/graph/src/types/ontology.ts
+++ b/libs/@blockprotocol/graph/src/types/ontology.ts
@@ -25,7 +25,6 @@ export const isOntologyTypeRecordId = (
     typeof recordId === "object" &&
     "baseUri" in recordId &&
     typeof recordId.baseUri === "string" &&
-    /** @todo - This means we need to have initialized the type system */
     validateBaseUri(recordId.baseUri).type === "Ok" &&
     "version" in recordId &&
     typeof recordId.version === "number"
--- a/libs/@blockprotocol/graph/src/types/subgraph/edges.ts
+++ b/libs/@blockprotocol/graph/src/types/subgraph/edges.ts
@@ -9,18 +9,14 @@ export * from "./edges/kind.js";
 export * from "./edges/outward-edge.js";
 export * from "./edges/variants.js";
 
-/** @todo - Re-express these and `Vertices` as `Record`s? */
+export type OntologyRootedEdges = Record<
+  BaseUri,
+  Record<number, OntologyOutwardEdge[]>
+>;
 
-export type OntologyRootedEdges = {
-  [typeBaseUri: BaseUri]: {
-    [typeVersion: number]: OntologyOutwardEdge[];
-  };
-};
-
-export type KnowledgeGraphRootedEdges = {
-  [entityId: EntityId]: {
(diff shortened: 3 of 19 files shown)


libs/@blockprotocol/graph/src/types/subgraph/edges.ts
```diff
@@ -9,18 +10,17 @@ export * from "./edges/kind.js";
...
-  [entityId: EntityId]: {
-    [edgeFirstCreatedAt: Timestamp]: KnowledgeGraphOutwardEdge[];
-  };
-};
-
-export type Edges = OntologyRootedEdges & KnowledgeGraphRootedEdges;
+// We technically want to intersect (`&`) the types here, but as their property keys overlap it confuses things and we
+// end up with unsatisfiable values like `EntityVertex & DataTypeVertex`. While the union (`|`) is semantically
+// incorrect, it structurally matches the types we want.
+export type Edges = OntologyRootedEdges | KnowledgeGraphRootedEdges;
```
Alfred-Mountfield, line 26: This is really frustrating. If we use `&` as we'd like to, then TS thinks the keys overlap and produces an unsatisfiable type
image

As they’re index signature property thingymajigsm it treats them as if they were a same key and then distributes the `&` all the way down through the `string` properties. Generally this distributive property works out (if you actually know the keys) for example in this where it combines the two `y` properties but leaves `x` and `z` as they were



But because of the index signature types it just assumes that the keys are the same and combines the constraints.

TypeScript's structural typing does mean that `x: Foo` is assignable to `Foo | Bar` and `x: Bar` is assignable to `Foo | Bar`, _and importantly_ `x: Foo & Bar` is assignable to `Foo | Bar`, so `OntologyRootedEdges | KnowledgeGraphRootedEdges` works fine for us as a type.

libs/@blockprotocol/graph/src/types/subgraph/vertices.ts
```diff
@@ -99,7 +99,7 @@ export const isOntologyTypeVertexId = (
     /** @todo - This means we need to have initialized the type system */
     validateBaseUri(vertexId.baseId).type === "Ok" &&
     "revisionId" in vertexId &&
-    typeof vertexId.revisionId === "number"
+    !Number.isNaN(Number(vertexId.revisionId))
```
Alfred-Mountfield, line 101: Is this equivalent?

libs/@blockprotocol/graph/src/types/ontology.ts
```diff
@@ -25,9 +25,17 @@ export const isOntologyTypeRecordId = (
     typeof recordId === "object" &&
     "baseUri" in recordId &&
     typeof recordId.baseUri === "string" &&
-    /** @todo - This means we need to have initialized the type system */
     validateBaseUri(recordId.baseUri).type === "Ok" &&
```
Alfred-Mountfield, line 28: `validateBaseUri` was migrated a while ago to have a pure TS implementation.

libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -50,12 +56,13 @@ export type TraversalSubgraph<
   RootType extends SubgraphRootType<Temporal> = SubgraphRootType<Temporal>,
 > = Omit<Subgraph<Temporal, RootType>, "edges"> & {
   edges: PatchEntityValidInterval<
-    OntologyRootedEdges & {
-      [entityId: EntityId]: Record<
-        typeof TIMESTAMP_PLACEHOLDER,
-        KnowledgeGraphOutwardEdge[]
-      >;
-    }
+    | OntologyRootedEdges
```
Alfred-Mountfield, line 59: Same as with `Subgraph`, now requires changing from `&` to `|`

libs/mock-block-dock/src/datastore/traverse.ts
```diff
@@ -429,31 +437,27 @@ export const finalizeSubgraph = <
     const variableAxis = traversalSubgraph.temporalAxes.resolved.variable.axis;
 
     const validRanges = Object.fromEntries(
-      Object.entries(traversalSubgraph.vertices).map(
-        ([baseId, revisionMap]) => {
-          const sortedRevisions = Object.keys(revisionMap).sort();
-
-          /** @todo - This cast is needed because TS is confused again and thinks these must always be entity vertices */
```
Alfred-Mountfield, line 436: Could remove a lot of these when moving to `typedEntries` and `typedKeys` 🎉

libs/@blockprotocol/graph/src/types/subgraph/edges.ts
thehabbos007: Does this have any implicatiosn for the consumers of `Edges`?

libs/@blockprotocol/graph/src/types/subgraph/vertices.ts
thehabbos007: Is it problematic to use `Number` here?
```js
Number(); // 0
Number(null); // 0
Number(""); // 0 
```

Whereas 
```js
parseInt() // NaN
parseInt(null) // NaN
parseInt("") // NaN
```

Although we would then have this..
```js
parseInt("0not a number") // 0
Number("0not a number"); // Nan
```

thehabbos007 (commented):
Haven't tested yet, but looks nice so far

libs/@blockprotocol/graph/src/types/subgraph/vertices.ts
Alfred-Mountfield: If we want to make sure it's exactly an integer, (without accepting `.0` making it a float) then I think our two options are a regex, or this. 
image

Alfred-Mountfield: Seems that this check is better than the regex ([crappy benchmark here](https://jsben.ch/6wCQN))

Alfred-Mountfield: [Introduce and use stringIsNonNegativeInteger](https://github.com/blockprotocol/blockprotocol/pull/939/commits/9dd1b30162bebaa5dc183a01b0fdc4d284cf918d)

libs/@blockprotocol/graph/src/types/subgraph/edges.ts
Alfred-Mountfield: Yes, but the only way I feel like I can answer that is by saying look at the rest of the diff of the PR :( 

It's all sort of weird obscure stuff to do with how well TS can infer/constrain types while traversing it.

libs/@blockprotocol/graph/src/types/subgraph/vertices.ts
```diff
@@ -114,21 +115,23 @@ export const isEntityVertexId = (
...
-};
-
-export type KnowledgeGraphVertices<Temporal extends boolean> = {
-  [entityId: EntityId]: {
-    [entityVersion: EntityRevisionId]: KnowledgeGraphVertex<Temporal>;
-  };
-};
+export type OntologyVertices = Record<
+  BaseUri,
+  Record<OntologyTypeRevisionId, OntologyVertex>
```
CiaranMn, line 124: It will probably be useful to have separate types for each level in these maps at some point. It would make it a bit easier to write typeguards like this. Not suggesting any action now, we can add them as and when we need them.
```ts
const isEntityVersionObject = <Temporal extends boolean>(
  versionObject:
    | Record<EntityRevisionId, KnowledgeGraphVertex<Temporal>>
    | Record<OntologyTypeRevisionId, OntologyVertex>,
): versionObject is Record<
  EntityRevisionId,
  KnowledgeGraphVertex<Temporal>
> => {
  const sampleVersion = Object.values(versionObject)[0];
  if (!sampleVersion) {
    throw new Error("version object is empty");
  }
  return isEntityVertex(sampleVersion);
};
```

Alfred-Mountfield: We could definitely consider this at some point :)

libs/@blockprotocol/graph/src/shared.ts
```diff
@@ -27,6 +27,41 @@ type Entry<T extends {}> = T extends readonly [unknown, ...unknown[]]
...
+ * - "0.0"
+ * - "1.0"
+ * - "-1"
+ * - "foo"
+ * - "0foo"
+ * - "1.1"
+ *
+ * @param {string} input
+ */
+export const stringIsNonNegativeInteger = (input: string) => {
```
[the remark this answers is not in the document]
Alfred-Mountfield: Good call, I don't see a harm in adding this as it'll usually be downcast straight away back to `string` anyway
[Improve typing of stringIsNonNegativeInteger](https://github.com/blockprotocol/blockprotocol/pull/939/commits/0506eba8b8e78192c03a707d8cc597d2c4a93d00)

PR #940, opened 2023-02-06
Accept ontology types in `buildSubgraph`

Alfred-Mountfield:

A while ago we added a `buildSubgraph` method which constructed a `Subgraph` object from a given `Entity[]`. This extends that method to accept collections of ontology types which are then added to a `Subgraph` with the appropriate edge-kinds.

At the moment `mock-block-dock` does not support returning ontology types within `Subgraph`s. This is fine for the majority of blocks, but is problematic if blocks wish to concern themselves with validation of entities, want to be able to modify types (such as an incoming component for creating/editing ontology types). As such, for now this is meant as a utility for some EA implementations, but in a follow-up in the medium-term this will ideally show the new expected state whereby every `Subgraph` will be able to contain ontology elements should the depths facilitate it.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203543026679380/1203845052955907/f) _(internal)_

## 🚫 Blocked by

- [ ] #939 

## 🔍 What does this change?

See commits as I moved some things around, but generally
- adds new iteration methods to the type system package to get IDs referenced within types
- restructures `mutate-subgraph`
- adds new mutation (add*) methods for all of the other elements of the Subgraph 
- expands `buildSubgraph` to take in arrays of the other elements of the graph
- hotfixes mock-block-dock

- This specific change does not as this is an 'internal' API for the moment.

## ⚠️ Known issues

- mock-block-dock does not support creation/updating/querying of ontology types

## 🐾 Next steps

- Update `mock-block-dock` and `Subgraph` traversal code to support ontology types

## 🛡 What tests cover this?

- Besides `tsc` and `eslint`, none at present

## ❓ How to test this?

1. This is an internal API at the moment, and we shall be using it, so we can probably err on the side of a code-review for now, and check the functionality in usage, updating as necessary, until we're ready to expose and document this appropriately.

## 📹 Demo

N/A


libs/@blockprotocol/type-system/src/native/entity-type.ts
```diff
@@ -0,0 +1,48 @@
+import { EntityType, VersionedUri } from "../../wasm/type-system";
+
+/**
+ * Returns all the IDs of all types referenced in a given property type.
+ *
+ * @param {EntityType} entityType
+ */
+export const getReferencedIdsFromEntityType = (
```
Alfred-Mountfield, line 8: I would have liked to just expose the `Rust` implementations of these functions, but we're trying to avoid requiring all importers of the `graph` package from having to initialise the type system at the moment..

Perhaps it's fine to require initialisation of the type system for our internal API methods?

cc. @CiaranMn for thoughts, unsure if this will complicate it for environments like WP?

CiaranMn: WP is going to have to use `buildSubgraph` in the browser. It won't need ontology types, but I think the addition of anything WASM-dependent here would necessitate shipping the WASM to the browser even if it isn't used? So you might not have a crash, but you're adding a lot to the bundle size. If that's true I'd avoid it.

Alfred-Mountfield: Correct. I still think we can get the bundle size down quite a lot, and can benefit from caching of the WASM module, but without that investment this seems like the correct compromise for now then

TimDiekmann (approved):
Code logic LGTM ✅

PR #944, opened 2023-02-07
Restore uploadFile to mock-block-dock

CiaranMn:

Makes mock-block-dock support `uploadFile` again

I've guessed at the URIs for the types on blockprotocol.org for `File` - to be confirmed / updated later.

This unblocks @luisbettencourt's further work on an address/map block.

## Next steps

Release a canary version so Luís can test this in his block


libs/mock-block-dock/src/datastore/use-mock-datastore.ts
```diff
@@ -567,89 +559,79 @@ export const useMockDatastore = (
...
-        //       url: result.toString(),
-        //     };
-        //     return Promise.resolve({ data: returnData });
-        //   }
-        //
-        //   throw new Error("Couldn't read your file");
-        // }
-        // throw new Error("Unreachable.");
+        const { description } = data;
+        const file = "file" in data ? data.file : null;
```
Alfred-Mountfield, line 563: I do wonder if we should design the type to avoid us needing an `in` check, really don't like leaving the support of TS

libs/mock-block-dock/src/datastore/use-mock-datastore.ts
```diff
@@ -567,89 +559,79 @@ export const useMockDatastore = (
...
+        const url = "url" in data ? data.url : null;
+        if (!file && !url?.trim()) {
+          throw new Error("Please provide either a valid URL or file below");
```
Alfred-Mountfield, line 566: This error message seems a little strange if a block called `uploadFile`

libs/@blockprotocol/graph/src/types/file.ts
```diff
@@ -1,15 +1,18 @@
...
-export type UploadFileReturn = {
-  entityId: EntityId;
-  url: string;
-  mediaType: FileMediaType;
+// @todo -
+export type FileEntityProperties = {
+  "https://blockprotocol.org/@blockprotocol/types/property-type/description/v/1"?: string;
+  "https://blockprotocol.org/@blockprotocol/types/property-type/filename/v/1": string;
+  "https://blockprotocol.org/@blockprotocol/types/property-type/mime-type/v/1": string;
+  "https://blockprotocol.org/@blockprotocol/types/property-type/url/v/1": string;
```
Alfred-Mountfield, line 13: Shouldn't these be BaseUri's? (This is why I'd really like `satisfies` on types 🙃 , you want this to be a subtype of `EntityPropertiesObject` I assume, although to be fair `BaseUri` is just a string at the moment so I guess it wouldn't have helped you)

libs/mock-block-dock/src/datastore/use-mock-datastore.ts
```diff
@@ -567,89 +559,79 @@ export const useMockDatastore = (
...
+            resolvedUrl = readFileResult.toString();
+          } catch (err) {
+            throw new Error("Could not upload file");
+          }
+        }
+
+        const mimeType = mime.lookup(filename) || "application/octet-stream";
+
+        const newEntityProperties: FileEntityProperties = {
+          "https://blockprotocol.org/@blockprotocol/types/property-type/description/v/1":
```
Alfred-Mountfield, line 604: Same as above, pretty sure you want these to be `BaseUri`s

libs/mock-block-dock/src/datastore/use-mock-datastore.ts
CiaranMn: Added typeguards in https://github.com/blockprotocol/blockprotocol/pull/944/commits/8d53e9ca352a899dc85521206bebc44072dd72a1

libs/mock-block-dock/src/datastore/use-mock-datastore.ts
CiaranMn: Fixed in https://github.com/blockprotocol/blockprotocol/pull/944/commits/b6c9ab49f83299d20c7220b0dcc68e338fe732b8

Alfred-Mountfield (approved):

PR #946, opened 2023-02-08
Add property type methods

CiaranMn:

Adds the type for property type methods to `@blockprotocol/graph`, which we'll require for the type editor.

I've added the ontology resolve depths as a consequence, optional for now to avoid people having to pass them when e.g. making a `getEntity` request. We should decide what to do about this properly.


libs/@blockprotocol/graph/src/graph-block-handler.ts
```diff
@@ -41,7 +41,13 @@ export class GraphBlockHandler<Temporal extends boolean>
   implements
     Omit<
       BlockGraphMessages<Temporal>,
-      "createEntityType" | "updateEntityType" | "deleteEntityType"
+      | "createEntityType"
+      | "updateEntityType"
+      | "deleteEntityType"
```
Alfred-Mountfield, line 46: Should we just remove this method?

libs/@blockprotocol/graph/src/types/ontology/property-type.ts
```diff
@@ -12,3 +19,29 @@ export type PropertyTypeWithMetadata = {
...
 };
+
+export type AggregatePropertyTypesData = {
+  graphResolveDepths?: Partial<
+    Pick<
+      Subgraph<true>["depths"],
+      "constrainsValuesOn" | "constrainsPropertiesOn"
+    >
+  >;
+};
```
Alfred-Mountfield, line 30: Just to flag, this now makes this method inconsistent with `AggregateEntityTypes`. Also I've noticed you've made it `Partial`. 

Previously we had opted to make the whole depths object required, because we encountered too many scenarios where a dev absent-mindedly didn't think about what they actually needed. I appreciate that introduces a very large amount of boilerplate, so I could be swayed to us having them as optional. 

I think I actually was swayed in some previous discussion which is why `graphResolveDepths` is now optional in `getEntity` and `aggregateEntities`. However, it is not `Partial` (which _seems_ better to me if we're allowing it to be optional and presumably going to **specify** that it is 0 if you leave it undefined). So, all of that was a roundabout way to say:

**If we keep this can you go and change the other methods to be consistent with this approach?**

libs/@blockprotocol/graph/src/types/ontology/property-type.ts
```diff
@@ -12,3 +19,29 @@ export type PropertyTypeWithMetadata = {
...
+export type AggregatePropertyTypesResult = {
+  results: Subgraph<true, PropertyTypeRootType>;
+};
+
+export type GetPropertyTypeData = {
+  propertyTypeId: VersionedUri;
+};
+
+export type CreatePropertyTypeData = {
+  propertyType: PropertyType;
```
Alfred-Mountfield, line 41: In HASH this is represented as
```ts
export type CreatePropertyTypeRequest = {
  propertyType: Omit<PropertyType, "$id">;
};
```

I think we should look at doing the same here. Callers (blocks) _**cannot**_ freely determine type IDs, they have no mechanism of knowing what URLs an EA is capable of serving.

libs/@blockprotocol/graph/src/types/ontology/property-type.ts
```diff
@@ -12,3 +19,29 @@ export type PropertyTypeWithMetadata = {
...
+};
+
+export type UpdatePropertyTypeData = {
+  propertyTypeId: VersionedUri;
+  propertyType: PropertyType;
+};
```
Alfred-Mountfield, line 47: Similar comment as above. A caller cannot determine the type ID, and generally the EA should be responsible for ensuring the version was incremented appropriately.

libs/@blockprotocol/graph/src/types/subgraph/graph-resolve-depths.ts
```diff
@@ -1,11 +1,22 @@
...
   outgoing: number;
 }
 
 /** @todo - Add documentation */
-/** @todo - expand this with ontology - related edges */
 export type GraphResolveDepths = {
   hasLeftEntity: EdgeResolveDepths;
   hasRightEntity: EdgeResolveDepths;
+
+  // @todo decide if these should be required
```
Alfred-Mountfield, line 15: I think these should be required, but we can use `Partial` in function parameters.

For now, when we _create_ one of these complete objects for `Subgraph`s in `mock-block-dock` or otherwise we can just fill in a depth of `0` for the ones we don't support at the moment?

Alfred-Mountfield (changes requested):

libs/@blockprotocol/graph/src/graph-block-handler.ts
CiaranMn: Yeah – done in https://github.com/blockprotocol/blockprotocol/pull/946/commits/49e085829cbe652b1b8cae115a451722b6aeda63

libs/@blockprotocol/graph/src/types/ontology/property-type.ts
CiaranMn: Yes - https://github.com/blockprotocol/blockprotocol/pull/946/commits/086f506ad7e3054eed4c6896363cce4d53a06145

libs/@blockprotocol/graph/src/types/ontology/property-type.ts
CiaranMn: Done in https://github.com/blockprotocol/blockprotocol/pull/946/commits/086f506ad7e3054eed4c6896363cce4d53a06145

libs/@blockprotocol/graph/src/types/ontology/property-type.ts
CiaranMn: Done in https://github.com/blockprotocol/blockprotocol/pull/946/commits/086f506ad7e3054eed4c6896363cce4d53a06145

Although I think a legitimate alternative for updates is to allow it (to save callers omitting it, since some might send it anyway).

libs/@blockprotocol/graph/src/types/subgraph/graph-resolve-depths.ts
CiaranMn: Yep – done in https://github.com/blockprotocol/blockprotocol/pull/946/commits/12ecc76c552c4df3e6767f625425bf474b2dfbaa

Alfred-Mountfield (approved):

PR #948, opened 2023-02-08
Fix bugs in the Graph service stdlib

Alfred-Mountfield:

A few sneaky bugs slipped by in previous PRs that introduced temporal versioning code for the stdlib. As identified in those PRs, testing has been deferred to unblock other work in the immediate term, and we were going to discover bugs through development usage, that has happened, and this PR fixes those bugs.

A follow-up will add a block that will make testing these things more easy, although where to put that block is currently unclear, that in-dev block was used to check that the fixes here were correct.

Furthermore,  previously, when traversing a subgraph, the subgraph lib expected there to be left and right entities for all link entities. This was largely based on the reasoning that it's impossible to create a link entity that does not have a left or right endpoint. However, depending on your depths (e.g. `hasLeftEntity: { incoming: 0, outgoing: 0 }`) you will not get back that vertex within the `Subgraph`. This broke an invariant in the stdlib code. This PR removes that incorrect invariant, and hoists it to the EA (mock-block-dock) where that invariant still _is_ an invariant.

There are still some unresolved questions about if we want to treat missing things as `undefined` or `[]` return values, or if we should throw errors like "NOT_FOUND".

This also drive-by renames a typo.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643281/1203928892175779/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

See commits

Not as of yet.

## ⚠️ Known issues

N/A

## 🐾 Next steps

Make the block that uses all of the BP graph methods available somewhere

## 🛡 What tests cover this?

`tsc`, `eslint`, etc.

## ❓ How to test this?

Try loading a block in MBD that creates and gets links.

## 📹 Demo

Example from the block that's currently in-dev. Previously querying this link entity would result in an error and 0 edges.
image

### DIFF
--- a/libs/@blockprotocol/graph/src/types/temporal-versioning.ts
+++ b/libs/@blockprotocol/graph/src/types/temporal-versioning.ts
@@ -12,7 +12,7 @@ export type Timestamp = string;
 /**
  * @todo - doc
  */
-export type TemporalAxes = "transactionTime" | "decisionTime";
+export type TemporalAxis = "transactionTime" | "decisionTime";
 
 /**
  * The bound of a time-interval that is either exclusively or inclusively limited by a `Timestamp`
@@ -91,7 +91,7 @@ export type BoundedTimeInterval = TimeInterval<
  * {@link PinnedTemporalAxisUnresolved}.
  */
 export type VariableTemporalAxisUnresolved<
-  Axis extends TemporalAxes,
+  Axis extends TemporalAxis,
   StartBound extends TemporalBound | null = TemporalBound | null,
   EndBound extends LimitedTemporalBound | null = LimitedTemporalBound | null,
 > = {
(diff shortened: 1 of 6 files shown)


libs/@blockprotocol/graph/src/types/temporal-versioning.ts
```diff
@@ -12,7 +12,7 @@ export type Timestamp = string;
 /**
  * @todo - doc
  */
-export type TemporalAxes = "transactionTime" | "decisionTime";
+export type TemporalAxis = "transactionTime" | "decisionTime";
```
TimDiekmann, line 15: Thanks for this drive-by fix 🙂

TimDiekmann (approved):

CiaranMn (approved):

PR #880, opened 2023-01-11
Introduce 0.3 specifications

CiaranMn:
## 🌟 What is the purpose of this PR?

Introduces major changes to the Block Protocol specification and the packages that implement it.

The documentation and specification in `apps/site` in this branch describe how things work with these changes introduced.

Versus the current specification and implementation, the key changes:

1. Implements the new **type system** in relevant packages and describes it in the specification (initial [RFC](https://github.com/blockprotocol/blockprotocol/blob/main/rfcs/text/0352-graph-type-system.md))
2. Introduces a new `Subgraph` object which replaces how blocks previously received entity graphs
3. Renames 'services' – i.e. the functional groups of the specification which describe the content and purpose of messages exchanged between applications and blocks – to 'modules', because this also...
4. Introduces a new 'Service Module', which handles blocks requesting that embedding applications make requests to external services on their behalf, e.g. MapBox, OpenAI
  - `apps/site` (i.e. blockprotocol.org once this is merged) also gains a subscription system whereby users can sign up to a single account to handle payment for such external API calls – this will be used by embedding applications (e.g. HASH, WordPress) to implement making the requests to external services, assuming the user of the application has an eligible account

## 🔗 Related links

- The best way to understand the new system is to visit the `/docs` path of the preview deployment.

## 🚫 Blocked by

- [ ] Fix integration tests https://github.com/blockprotocol/blockprotocol/pull/1036

## 🔍 What does this change?

- See summary above, and individual commit list for PRs.

## 📜 Does this require a change to the docs?

- It includes changes to docs.

## 🐾 Next steps

- `canary` release of packages
- Testing
- `latest` release of packages

## 🛡 What tests cover this?

- Existing integration tests.

## ❓ How to test this?

1. Intensive use of the new packages and modules.

## 📹 Demo

- See the docs in the preview deployment. Various individual PRs have demos of changes listed above.

PR #949, opened 2023-02-08
Temporarily remove pagination from `aggregate*` methods

Alfred-Mountfield:

With the introduction of the `Subgraph` type, the previous implementation of paginated results for "aggregate" queries no longer makes sense. 

Paginating a graph is a difficult task, as there isn't a discrete index one can use in a stable-ordered _list_ of items, and a cursor needs to contain a potentially unbounded size of information that is yet to be resolved for future queries.

We intend to revisit this, potentially just allowing cursor-based pagination over the `roots` of a `Subgraph`, although ideally with a different solution that will be better suited to blocks that need control over the message sizes.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203919127849757/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Removes all `page` related fields
- Removes the pagination implementation in `mock-block-dock`

- The spec will need to be updated to identify this change. This will happen prior to the merge of #880.

## ⚠️ Known issues

N/A

## 🐾 Next steps

- N/A at this time

## 🛡 What tests cover this?

- Unknown aside from `tsc` and `eslint`, blocks using aggregations should continue to work

### DIFF
--- a/libs/mock-block-dock/src/util.ts
+++ b/libs/mock-block-dock/src/util.ts
@@ -241,20 +241,13 @@ export function filterAndSortEntitiesOrTypes<
 ): FilterResult<Temporal> {
   const { operation } = payload;
 
-  const pageNumber = operation?.pageNumber || 1;
-  const itemsPerPage = operation?.itemsPerPage || 10;
   const multiSort = operation?.multiSort ?? [{ field: "updatedAt" }];
   const multiFilter = operation?.multiFilter;
 
   const appliedOperation = {
-    pageNumber,
-    itemsPerPage,
     multiFilter,
     multiSort,
   };
-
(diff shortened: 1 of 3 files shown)


CiaranMn (approved):

libs/mock-block-dock/src/util.ts
```diff
@@ -264,18 +257,14 @@ export function filterAndSortEntitiesOrTypes<
...
     multiSort,
-  }).slice(startIndex, endIndex) as Elements;
+  });
   return {
     results,
     operation: {
       ...appliedOperation,
-      totalCount,
-      pageCount,
     },
```
CiaranMn, line 268: This can just be
```ts
    operation: appliedOperation

Alfred-Mountfield: Oops, thanks

PR #954, opened 2023-02-08
`@blockprotocol/service` package with mapbox method definitions

benwerner01:
## 🌟 What is the purpose of this PR?

This PR adds a canary version of the `@blockptocol/service` package. The structure of the package is strongly influenced by the design of the prior `@blockprotocol/graph` and `@blockprotocol/hook` packages.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643281/1203880493107349/f) _(internal)_

## 🔍 What does this change?

See commits.

## 📜 Does this require a change to the docs?

- this package should be accompanied with documentation in the BP site ([asana task](https://app.asana.com/0/0/1204004000163859/f))

## ⚠️ Known issues

- [Find way to stop copy pasting mapbox types without causing an error in the `typescript-json-schema` package](https://app.asana.com/0/0/1204004000163864/f)

## 🐾 Next steps

- [Add `ServiceElementBase` class to `@blockprotocol/service` module to support custom elements](https://app.asana.com/0/0/1204003986334469/f)
- [add definition for service errors](https://app.asana.com/0/0/1204004000163863/f)

## 🛡 What tests cover this?

None.

## ❓ How to test this?

The functionality of this PR is best tested as part of https://github.com/blockprotocol/blockprotocol/pull/966

PR #966, opened 2023-02-09
Add `@blockprotocol/service` to `mock-block-dock`

benwerner01:

This PR adds the service module to `mock-dock-block`.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643281/1203940085640294/f) _(internal)_
- [Allow overriding callbacks for mock block dock](https://app.asana.com/0/1203358502199087/1203946711018582/f)

## 🔍 What does this change?

See commits.

Not that I'm aware of.

## 🛡 What tests cover this?

None.

## ❓ How to test this?

1. Run the BP site in your local developer environment with seeded users
2. Create/obtain your BP API key and add it as an `BP_API_KEY` environment variable to your `.env.local`
3. Run the HASH internal API in your local dev environment
4. Run the  mock-dock-block local developer environment
5. Use the test block to make a request to the mapbox API

## 📹 Demo

https://user-images.githubusercontent.com/42802102/220391775-1bba1f78-fda1-4acb-9c06-c26938cbe826.mov


nonparibus:
Total aside: I assume the `Mapbox Suggest API request` results returned for London are all in California as geo-tailored to your present location, @benwerner01? I would hope for more sensible results back in London (England).

benwerner01:
> Total aside: I assume the `Mapbox Suggest API request` results returned for London are all in California as geo-tailored to your present location, @benwerner01? I would hope for more sensible results back in London (England).

Yes, I assume mapbox by default users the source IP address to infer the location of the user.

libs/mock-block-dock/dev/dev-app.tsx
```diff
@@ -40,12 +40,20 @@ const blockEntityMap = {
 
 const DevApp: FunctionComponent = () => {
   const [testBlockType, setTestBlockType] = useState<TestBlockType>("react");
+  const [blockProtocolApiKey, setBlockProtocolApiKey] = useState<string>();
```
CiaranMn, line 43: Why does this need to be in state?

libs/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -254,6 +275,23 @@ export const MockBlockDock: FunctionComponent<MockBlockDockProps> = ({
...
+        /**
+         * Registration can error when the user switches between preview and debug mode.
+         * Registration is attempted with the old service, which has been destroyed.
+         * It then succeeds with the new one.
+         * @todo can we avoid this error?
+         */
+      }
+    }
+  }, [serviceModule, serviceModuleCallbacks]);
+
```
CiaranMn, line 294: @nathggns I think we can get rid of this and the re-registration below, right? The hooks should handle it for us.

@benwerner01 we recently removed the need for this, I believe – I appreciate you are following the existing pattern, which we should've cleaned up

libs/mock-block-dock/src/service-module-callbacks.ts
```diff
@@ -0,0 +1,113 @@
+import { EmbedderServiceMessageCallbacks } from "@blockprotocol/service";
+import axios from "axios";
+
+const bpApiBaseURL = process.env.BP_API_BASE_URL ?? "http://localhost:3000/api";
```
CiaranMn, line 4: I'm not sure about `localhost` as the default – most people developing blocks will not also have the `blockprotocol` repo and site running. I think we should default to supporting people who can just use the production API instance (that'll be fine for us most of the time too, unless we want to debug the API itself). I've also suggested a tweak to the environment variable as we use `BLOCK_PROTOCOL_SITE_HOST` elsewhere in the repo (in the publish script)

```suggestion
const bpApiBaseURL = `${process.env.BLOCK_PROTOCOL_SITE_HOST ?? "http://localhost:3000}/api`;
```

libs/mock-block-dock/src/service-module-callbacks.ts
```diff
@@ -0,0 +1,113 @@
...
+  blockProtocolApiKey?: string;
+  providerName: string;
+  methodName: string;
+  payload: any;
+}): Promise<{ data: any }> => {
+  const { providerName, methodName, payload, blockProtocolApiKey } = params;
+
+  if (!blockProtocolApiKey) {
+    throw new Error(
+      `A block protocol API key is required to make calls to the "${methodName}" method of the BP service module`,
```
CiaranMn, line 20:
```suggestion
      `Visit ${bpSiteHost}/dashboard to generate a Block Protocol API key, which is required to make calls to the "${methodName}" method of the BP service module. Set it as BLOCK_PROTOCOL_API_KEY in your environment.`,
```

CiaranMn (changes requested):
Looking good, @benwerner01 – a few suggestions around environment variables

libs/mock-block-dock/dev/dev-app.tsx
CiaranMn: i.e. you could just pass `process.env.BLOCK_PROTOCOL_API_KEY` directly as a prop for the same effect

libs/mock-block-dock/dev/dev-app.tsx
```diff
@@ -40,12 +40,20 @@ const blockEntityMap = {
...
 
   const blockEntity = blockEntityMap[testBlockType];
 
   let blockDefinition;
   let blockType: "html" | "custom-element" | "react" | undefined;
 
+  useEffect(() => {
+    // Retrieve the blockprotocol API key environment variable
+    if (process.env.BP_API_KEY) {
+      setBlockProtocolApiKey(process.env.BP_API_KEY);
```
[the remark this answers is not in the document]
benwerner01: Addressed in https://github.com/blockprotocol/blockprotocol/pull/966/commits/f1559ef85d4029c7c6469afb8c674632d7d16d13

libs/mock-block-dock/dev/dev-app.tsx
benwerner01: Good point, have addressed in https://github.com/blockprotocol/blockprotocol/pull/966/commits/d1f2423c0b696dbcaa34ad525dc298df3e415475

libs/mock-block-dock/src/service-module-callbacks.ts
benwerner01: I think "Visit ${bpSiteHost}/dashboard to generate a Block Protocol API key" is a good addition, however I'm not sure whether it makes sense to include "Set it as BLOCK_PROTOCOL_API_KEY in your environment" as this would only be true for the developer environment of MBD. It's my understanding that the `MockBlockDock` can be imported in other applications, that may retrieve a value for the `blockProtocolApiKey` prop differently. Perhaps something like:
```suggestion
      `Visit ${bpSiteHost}/dashboard to generate a Block Protocol API key, which is required to make calls to the "${methodName}" method of the BP service module, and ensure it is passed to the "MockBlockDock" component.`,
```

libs/mock-block-dock/src/service-module-callbacks.ts
benwerner01: That's a good point, I've modified the default value to be `blockprotocol.org` and made use of `BLOCK_PROTOCOL_SITE_HOST` in https://github.com/blockprotocol/blockprotocol/pull/966/commits/83aea17a2ea8d7c07a1dae77d0c54e7d97235be5

libs/mock-block-dock/src/mock-block-dock.tsx
benwerner01: That's good to know, thank you. I've removed this `useEffect` in https://github.com/blockprotocol/blockprotocol/pull/966/commits/8d35f8e19029db8341683617c3639839f3696484. @nathggns let me know if I can remove the one for the `graphService`.

nonparibus (approved):
Legal ✅

libs/mock-block-dock/src/service-module-callbacks.ts
CiaranMn: That makes sense – let's add the prop name as well
```suggestion
      `Visit ${bpSiteHost}/dashboard to generate a Block Protocol API key, which is required to make calls to the "${methodName}" method of the BP service module, and ensure it is passed to the "MockBlockDock" component as `blockProtocolApiKey`.`,
```

CiaranMn (approved):
I've rebased this after rebasing #954 and it should be good to go with the error message change in

libs/mock-block-dock/src/mock-block-dock.tsx
nathggns: All modules used via react now keep their callbacks up to date, so you can remove this

PR #971, opened 2023-02-13
Add statistics reporting to R2 block downloading

thehabbos007:

This PR adds statistics reporting to our R2 worker. We still provide aggregate statistics in MongoDB as these are used to show download in the website. The new statistics reporting will be used for internal analytics.

The reporting is disabled until we set the environment variables in the workers control panel on CF.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/0/1203657549452631/f) _(internal)_

## 🔍 What does this change?

- Adds disabled-by-default reporting for every block download enriched with specific block metadata.

## 🛡 What tests cover this?

- None, I had to test by disabling everything to run locally. I tried using [miniflare](https://miniflare.dev/) to mock the storage server, but it isn't ideal still. We should look into setting up a jest test suite that uses miniflare so we can properly test this worker. 
- We can probably also use Cloudflare KV instead of MongoDB for the aggregate statistics, but since we're using MongoDB everywhere else, this isn't really useful at all right now.

## ❓ How to test this?

I commented out all bucket/mongodb related things and made HTTP requests by hand. Not ideal. I used a HTTP request inspector/bin site to make sure the request was made and that the worker wouldn't crash if the request is erroneous

### DIFF
--- a/infra/cloudflare-r2-wrangler/blockprotocol-preview/wrangler.toml
+++ b/infra/cloudflare-r2-wrangler/blockprotocol-preview/wrangler.toml
@@ -5,6 +5,10 @@ compatibility_date = "2022-06-30"
 account_id = "c6499786332a3d2fb35419a7803ab7aa"
 workers_dev = true
 
+[vars]
+DATA_URL = ""
+DATA_WRITE_KEY = "something-probably-secret"
+
 [[r2_buckets]]
 binding = "BLOCKPROTOCOL_BUCKET"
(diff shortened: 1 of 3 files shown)


CiaranMn (approved):

thehabbos007:
Merging this now - shouldn't affect anything as the env var defaults for the `data_url` value is empty.
We don't need to wrap payload `batch` as we can supply the absolute URL in the env var

PR #973, opened 2023-02-13
Miscellaneous updates to the Graph service

Alfred-Mountfield:

- In using the canary versions it became apparent that the library could benefit from hoisting and exposing its interval comparison function.
- the addition of an `ownedById` on ontology type records was an oversight, and doesn't make sense in a lot of contexts in EAs (a type may not be owned by any user of the system.
- `getOutgoingLinkAndTargetEntities` took a `Timestamp` instead of a `TimeInterval`, this was inconsistent with the other methods


TimDiekmann (approved):
LGTM

PR #981, opened 2023-02-15
Replace entity type editor + API endpoints for new type system

CiaranMn:

Introduce the new entity type editor to blockprotocol.org, using the new package created as part of https://github.com/hashintel/hash/pull/2035

This allows the creation and updating of entity types and property types using the new type system.

## 🔗 Related links

- [Overarching task](https://app.asana.com/0/1200211978612931/1201589261424227/f) _(internal)_

## 🔍 What does this change?

- Replaces the existing `Entity` model class with individual functions for creating and updating entity and property types - https://github.com/blockprotocol/blockprotocol/pull/981/commits/bfac02e43f618e25f742313d0815d50b1af5b850
  - there was a concern a while ago that having classes negatively affects API lambda cold start time, because they mean that every endpoint gets all the code involved in the class. I don't think we have validated this but we don't really need a class anyway
- Uses the new `@hashintel/type-editor` package to replace the entity type editor - https://github.com/blockprotocol/blockprotocol/pull/981/commits/0ca299304d5bb0fd6c5ab3972ac38f14d5e0e2cd
- Update the 'send type as JSON' middleware to handle the new types, and to default to sending JSON unless text/html is requested ([which browsers reliably will](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values)) - https://github.com/blockprotocol/blockprotocol/pull/981/commits/bfac02e43f618e25f742313d0815d50b1af5b850
- Driveby: set cookie domain so that logging into any one preview deployment means you're logged in to them all (avoids issues with using domain aliases in Vercel) - https://github.com/blockprotocol/blockprotocol/pull/981/commits/c163b8a30d7ec36ee78f6bc4ed019fde15155e21

- We will need to explain the usage of the type editor and types.

## ⚠️ Known issues

- The API currently returns all versions of all types when getting options to load into the type editor. The type editor itself only requires (a) the latest types to show in dropdowns, and (b) those referred to by the entity type loaded into it. At some point we should reduce fetching by updating the data requested to only (a) and (b), and then further to make (a) a dynamic request that happens when the user types in the dropdown.
- Needs better loading state
- Other known issues in next steps...

## 🐾 Next steps

With internal tasks

**Fixes required in `@hashintel/type-editor`**
- [Implement readonly mode](https://app.asana.com/0/1203179076056209/1203623179643234/f)
- [Only show latest version in entity type dropdowns](https://app.asana.com/0/1203358502199087/1203957373155149/f) (e.g. for setting constraint on link destination)

**Fixes required in blockprotocol.org**
- [Update type listing/editing page](https://app.asana.com/0/1203358502199087/1203801600565920/f)
- [A few visual improvements](https://app.asana.com/0/1203358502199087/1203979552136012/f)
  - Make the edit bar sticky
  - Check dropdown positioning
  - Fix any other visual oddities
- We should probably review the various uses of 'Schema' and the urls to do with entity types in the site, and make sure it makes sense for the new type system.
- There is also no 'Types' tab on the user's dashboard, and they must either go to their profile or click a 'Manage types' card to find them. This is a bit weird since they do have a 'Blocks' tab on their dashboard

## 🛡 What tests cover this?

- Two in `test/integration` (schema creating and schema editing) which are probably broken and can be fixed in follow up

## ❓ How to test this?

1.  View the preview deployment
2. Try creating and viewing types
3. Try fetching / curling the versioned URI to get JSON

## 📹 Demo

https://user-images.githubusercontent.com/37743469/219171123-edc3edc6-3505-4edf-9a63-15fc05ff0095.mp4


apps/site/src/pages/api/types/property-type/get.api.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    if (!dbRecord) {
+      return res.status(404).json(
+        formatErrors({
+          msg: "Cannot find property type",
+        }),
+      );
+    }
+
+    res.status(200).send({ propertyType: dbRecord.propertyTypeWithMetadata });
```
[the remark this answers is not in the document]
CiaranMn: This is not HTML and it is not directly rendered.

apps/site/src/pages/api/types/property-type/aggregate.api.ts
```diff
@@ -0,0 +1,31 @@
...
+    const { db } = req;
+    const { latestOnly } = req.query;
+
+    const dbRecords = await getPropertyTypes(db, {
+      latestOnly: latestOnly === "true",
+    });
+
+    res.status(200).send({
+      propertyTypes: dbRecords.map((record) => record.propertyTypeWithMetadata),
+    });
```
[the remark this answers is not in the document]
CiaranMn: This is not HTML and it is not directly rendered.

apps/site/src/pages/api/types/entity-type/aggregate.api.ts
```diff
@@ -0,0 +1,31 @@
...
+    const { db } = req;
+    const { latestOnly } = req.query;
+
+    const dbRecords = await getEntityTypes(db, {
+      latestOnly: latestOnly === "true",
+    });
+
+    res.status(200).send({
+      entityTypes: dbRecords.map((record) => record.entityTypeWithMetadata),
+    });
```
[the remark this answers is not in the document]
CiaranMn: This is not HTML and it is not directly rendered.

apps/site/src/pages/api/types/entity-type/get.api.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    if (!dbRecord) {
+      return res.status(404).json(
+        formatErrors({
+          msg: "Cannot find entity type",
+        }),
+      );
+    }
+
+    res.status(200).send({ entityType: dbRecord.entityTypeWithMetadata });
```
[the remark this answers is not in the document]
CiaranMn: This is not HTML and it is not directly rendered.

apps/site/src/components/pages/blocks/block-form-styles.tsx
```diff
@@ -1,17 +1,17 @@
-import { experimental_sx as sx, Paper, styled } from "@mui/material";
+import { Paper, styled } from "@mui/material";
 
-export const BlockFormContainer = styled(Paper)(
-  sx({ flex: 1, width: "100%", p: 6 }),
+export const BlockFormContainer = styled(Paper)(({ theme }) =>
+  theme.unstable_sx({ flex: 1, width: "100%", p: 6 }),
 );
```
CiaranMn, line 5: This is a consequence of updating to MUI 5.11.x, which we also did in `@hashintel/type-editor`

apps/site/src/components/service-message-list.tsx
```diff
@@ -1,25 +1,25 @@
 import {
+  JsonObject,
   ServiceDefinition,
   ServiceMessageDefinition,
 } from "@blockprotocol/core";
 import { Box, Typography } from "@mui/material";
 import { Fragment, FunctionComponent } from "react";
 
-import { JsonSchema } from "../lib/json-schema";
```
CiaranMn, line 8: I removed the `json-schema` file – we should settle on a canonical JsonSchema type and use it where appropriate. This doesn't really matter for now.

apps/site/src/pages/api/types/entity-type/aggregate.api.ts
```diff
@@ -0,0 +1,31 @@
+import { EntityTypeWithMetadata } from "@blockprotocol/graph";
```
CiaranMn, line 1: The property type and entity type versions of these endpoints and db implementations is very similar. It could probably be deduplicated.

thehabbos007 (commented):
Thank Ciaran, this is great! the cookie change seems good to me, I went through the rest to see if there was something. Would be good to get a frontend review too. I'll be testing this locally and add any problems I run into :)

apps/site/src/components/modal/create-schema-modal.tsx
```diff
@@ -69,8 +69,7 @@ export const CreateSchemaModal: FunctionComponent<CreateSchemaModalProps> = ({
       if (error) {
         setApiErrorMessage(error.message);
       } else if (data) {
```
thehabbos007, line 71: (Comment for line 39)
We allow spaces in the schema titles (we do in HASH). Should the types here behave different in that regard?

apps/site/src/pages/api/types/entity-type/shared/db.ts
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  const latestVersion =
+    existingEntityType.entityTypeWithMetadata.metadata.recordId.version;
+  const versionUpdateRequestAt = extractVersion(versionedUri);
+
+  if (versionUpdateRequestAt !== latestVersion) {
+    const updateRequestAgainstVersion = extractVersion(versionedUri);
+    throw new Error(
+      `Can only request an update using the latest type version, which is ${latestVersion} – the provided URI used ${updateRequestAgainstVersion}`,
```
CiaranMn, line 185:
```suggestion
    throw new Error(
      `Can only request an update using the latest type version, which is ${latestVersion} – the provided URI used ${versionUpdateRequestAt}`,
```

apps/site/src/pages/api/me.api.ts
```diff
@@ -1,6 +1,7 @@
 import { createBaseHandler } from "../../lib/api/handler/base-handler";
 import { SESSION_COOKIE_NAME } from "../../lib/api/middleware/constants";
 import { SerializedUser } from "../../lib/api/model/user.model";
+import { isProduction } from "../../lib/config";
```
CiaranMn, line 4:
```suggestion
```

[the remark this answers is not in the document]
CiaranMn: Good spot - removed in https://github.com/blockprotocol/blockprotocol/pull/981/commits/846805358b49aacfe247284405834ffde1c027f1

apps/site/src/pages/api/types/entity-type/shared/schema.ts
```diff
@@ -0,0 +1,49 @@
...
+import { SystemDefinedProperties } from "../../shared/constants";
+
+export const generateEntityTypeWithMetadata = (data: {
+  author: `@${string}`;
+  schema: Omit<EntityType, SystemDefinedProperties>;
+  version: number;
+}): EntityTypeWithMetadata => {
+  const { author, schema: incompleteSchema, version } = data;
+
+  if (!incompleteSchema.title) {
```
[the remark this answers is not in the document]
CiaranMn: Yes, good idea - https://github.com/blockprotocol/blockprotocol/pull/981/commits/7deceee6736d5a42e184001bfe1d8cb6c5bd0e5a

apps/site/src/pages/api/types/property-type/shared/schema.ts
```diff
@@ -0,0 +1,44 @@
...
+import { SystemDefinedProperties } from "../../shared/constants";
+
+export const generatePropertyTypeWithMetadata = (data: {
+  author: `@${string}`;
+  schema: Omit<PropertyType, SystemDefinedProperties>;
+  version: number;
+}): PropertyTypeWithMetadata => {
+  const { author, schema: incompleteSchema, version } = data;
+
+  if (!incompleteSchema.title) {
```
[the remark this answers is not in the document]
CiaranMn: Thank you - https://github.com/blockprotocol/blockprotocol/pull/981/commits/e3969f273d8a29299176922a3167da87959844a4

apps/site/src/middleware.page/return-types-as-json.ts
```diff
@@ -32,6 +35,19 @@ export const versionedTypeUriRegExp =
 const validateVersionedUri = (uri: string): uri is VersionedUri =>
   !!uri.match(versionedTypeUriRegExp);
 
+const getTypeByVersionedUri = (
+  versionedUri: VersionedUri,
+  kind: "entity-type" | "property-type",
+) =>
+  fetch(
+    `${FRONTEND_URL}/api/types/${kind}/get?versionedUri=${versionedUri}`,
```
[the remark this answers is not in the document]
CiaranMn: This is the middleware that's processing requests made to that URI – it needs to call the API path (this path) to get the object from the database (cannot call the db directly as the driver uses APIs that the Vercel edge function doesn't support).

The most we could do here is replace `FRONTEND_URL` with the origin of `versionedUri`

thehabbos007 (commented):
Tried out the editor locally, pretty cool! 
I have one direct comment re. the title, and then also something I assume you're aware of, but worth pointing out

The domain is `hash.ai` in the tooltips on the editor. I suppose it's something to do with the editor component and not any of the added code here
image

apps/site/src/components/modal/create-schema-modal.tsx
CiaranMn: Good flag – updated in https://github.com/blockprotocol/blockprotocol/pull/981/commits/5f8ac4a852469707588c10bf083b699bcd5de3b3 to just trim and remove most special characters (we can add them back in by request)

CiaranMn:
> The domain is `hash.ai` in the tooltips on the editor. I suppose it's something to do with the editor component and not any of the added code here image

Thanks @thehabbos007! I hadn't in fact spotted that, so many thanks for doing so. I've created a [task](https://app.asana.com/0/1203623179643287/1203975175898856/f) (internal) to fix this in the type editor package

apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-edit-bar.tsx
```diff
@@ -0,0 +1,79 @@
+import {
```
Comment (luisbettencourt) [inline apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-edit-bar.tsx:1]: I think we should probably hide the `Coming Soon` banner in this screen. And maybe use a similar styling for it?

image

Comment (luisbettencourt) [reply apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-edit-bar.tsx:1]: Actually will the `Coming Soon` banner stick around after the 28th?

apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-form.tsx
```diff
@@ -0,0 +1,249 @@
+import {
```
Comment (luisbettencourt) [inline apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-form.tsx:1]: I know you are aware there are visual oddities that require fixing and I don't suggest we tackle them in this PR but just flagging for posterity. Some font sizes in this menu are quite different from the original designs.

image

image

apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-form.tsx
```diff
@@ -0,0 +1,249 @@
...
+  EntityType,
+  EntityTypeWithMetadata,
```
Comment (luisbettencourt) [inline apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-form.tsx:3]: Also, probably a problem with the [extract type editor PR](https://github.com/hashintel/hash/pull/2035) and not with this one, but when editing custom array property type's min and max values this error is being thrown:
```Uncaught ReferenceError: setImmediate is not defined```

luisbettencourt (approved):
BP is coming together! Looks good to me.

Made a couple of suggestions but approving anyway as I think they can be addressed in a follow-up.

apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-edit-bar.tsx
Comment (CiaranMn) [reply apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-edit-bar.tsx:1]: The `Coming Soon` banner will be removed when we merge the `0.3` branch so this will solve itself

Comment (CiaranMn) [reply apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-edit-bar.tsx:1]: Well, we still need to introduce the code that does that – I'll remove in follow up

apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-form.tsx
Comment (CiaranMn) [reply apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-form.tsx:3]: Thanks Luís – I'll have a look at this in the type editor, it will be a missing polyfill.

Comment (CiaranMn) [reply apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page/entity-type-form.tsx:3]: Done in https://github.com/hashintel/hash/pull/2035/commits/87059b991f8d29efdd4b730a150a0d84e24709cf

PR #985, opened 2023-02-16
Rename services to modules

CiaranMn:

We're introducing the ability for blocks to call external APIs, which we want to refer to as 'services'. 

'Services' is what we currently call the different functional groups that make up the specification/protocol (e.g. 'graph service' for dealing with entities and links between them).

We're renaming those functional groups to 'modules' to free up the name 'services' for external APIs/integrations.

Since this will require updating a bunch of imports for consumers, I took the opportunity to ensure that module types are consistently in the form `[Module][Source][Type]`, e.g. `GraphEmbedderMessageCallbacks`. It was previously inconsistent, and I often found myself failing to find `EmbedderGraphMessageCallbacks`. The most salient piece of information in a given file is more likely to be the name of the module.


kachkaev (commented):
Can we add a Changesets file to ensure we’ve communicated breaking changes?

kachkaev:
CI does not run for this PR. Shall we change the list of target branches in that yaml to avoid accidental merges with failed linting?

CiaranMn:
> Can we add a Changesets file to ensure we’ve communicated breaking changes?

@kachkaev done in https://github.com/blockprotocol/blockprotocol/pull/985/commits/a3219015de024b2c6c9bf24a3c133b565b092452

We'll want to look at the changesets again when merging `0.3` and make sure they capture the scale of changes 

> CI does not run for this PR. Shall we change the list of target branches in that yaml to avoid accidental merges with failed linting?

I think we should do this early next week, but at the moment linting fails on `0.3` so no PRs will pass – I can fix it up and then add the linting for the final changes

kachkaev (approved):
Approving based on diff despite E2E test failures ([in attempt 1](https://github.com/blockprotocol/blockprotocol/actions/runs/4196359212/attempts/1)). Happy to re-approve if E2E tests fail because of a real problem and we need to fix it before merging.

nonparibus:
Can we set up redirects so that any old URLs that have `service` in the name point to the new ones? I'm specifically thinking of the spec pages.

CiaranMn:
> Can we set up redirects so that any old URLs that have `service` in the name point to the new ones? I'm specifically thinking of the spec pages.

Redirect added in https://github.com/blockprotocol/blockprotocol/pull/985/commits/238a75fe56c318b4dc31175904f602043394c87c – I've done this by specific paths rather than a pattern matching anything with 'service' as it avoids unintentionally catching future docs path that talk about the new 'service' (and there are only two to redirect)

PR #987, opened 2023-02-16
Make fields in `VersionedUri` public

TimDiekmann:

In the HASH Graph Query Layer, we use `VersionedUri` quite often and it's impossible to destructure it into a `BaseUri` and a version.
While not necessarily every combination is valid, any `base_uri` with any `version` is a well-formed `VersionedUri`, it does not need to hold any invariants.
While other languages usually prefer "constructors", "getters", and "setters" (which are also missing), Rust has the concept of borrowing, so it's not possible to partially move out a variable of a struct by a function without consuming the struct. This implies if the `VersionedUri` has to be split into `BaseUri` and `u32`, we need something like
```rust
fn into_base_uri_and_version(self) -> (BaseUri, u32);
```
or 
```rust
fn into_parts(self) -> (BaseUri, u32);
```
which is (a) verbose to write and (b) does not retain the parameter names. It's easier to access `.base_uri` and `.version directly`. In addition, it's possible to modify the variables in place (e.g. `.version += 1`), which we do in our update procedure.

## 🔍 What does this change?

- Make the member variables `pub`
- Remove `new` and "getters"

### DIFF
--- a/libs/@blockprotocol/type-system/crate/src/ontology/entity_type/repr.rs
+++ b/libs/@blockprotocol/type-system/crate/src/ontology/entity_type/repr.rs
@@ -101,7 +101,7 @@ impl TryFrom<EntityType> for super::EntityType {
             .try_into()
             .map_err(ParseEntityTypeError::InvalidLinks)?;
 
-        if entity_type_repr.additional_properties == true {
+        if entity_type_repr.additional_properties {
             return Err(ParseEntityTypeError::InvalidAdditionalPropertiesValue);
         }
 
--- a/libs/@blockprotocol/type-system/crate/src/ontology/uri/mod.rs
+++ b/libs/@blockprotocol/type-system/crate/src/ontology/uri/mod.rs
@@ -85,30 +85,11 @@ impl<'de> Deserialize<'de> for BaseUri {
 //  if we can then we should delete wasm::VersionedUriPatch
 #[derive(Debug, Clone, PartialEq, Eq, Ord, PartialOrd, Hash)]
 pub struct VersionedUri {
-    base_uri: BaseUri,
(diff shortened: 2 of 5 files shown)


libs/@blockprotocol/type-system/crate/src/ontology/entity_type/repr.rs
```diff
@@ -101,7 +101,7 @@ impl TryFrom<EntityType> for super::EntityType {
             .try_into()
             .map_err(ParseEntityTypeError::InvalidLinks)?;
 
-        if entity_type_repr.additional_properties == true {
+        if entity_type_repr.additional_properties {
```
TimDiekmann, line 104: Drive-by: clippy complained here

libs/@blockprotocol/type-system/crate/src/ontology/uri/mod.rs
```diff
@@ -85,30 +85,11 @@ impl<'de> Deserialize<'de> for BaseUri {
...
-    version: u32,
+    pub base_uri: BaseUri,
+    pub version: u32,
 }
 
 impl VersionedUri {
-    /// Creates a new [`VersionedUri`] from the given `base_uri` and `version`.
-    ///
-    /// # Errors
-    /// - `ParseBaseUriError` if the given URI string is invalid
```
TimDiekmann, line 96: This was not correct at all

Alfred-Mountfield (approved):

Alfred-Mountfield:
This was something we'd been meaning to do for a while, no sense keeping ownership of the inner values away from users of the struct. Cheers!

TimDiekmann:
Is this targeting the correct branch?

PR #988, opened 2023-02-18
Upload original tarballs alongside individual block files

kachkaev:

This PR updates block publishing process. It makes original tarballs available on R3, alongside individual files. This allows us to mirror tarballs between hub instances, thus unblocking the completion of https://github.com/blockprotocol/blockprotocol/pull/936. Uploaded tarballs are especially helpful for local dev and allow us to quickly warm up a local instance with blocks that are already published on blockprotocol.org.

## 🔗 Related links

## 🐾 Next steps

- Supplement `block-uploads` key on S3 with `block-contents` (also versioned). We can make the existing `blocks` key redundant after that.

### DIFF
--- a/apps/site/src/lib/api/blocks/s3.ts
+++ b/apps/site/src/lib/api/blocks/s3.ts
@@ -277,3 +277,25 @@ export const validateExpandAndUploadBlockFiles = async ({
 
   return { expandedMetadata };
 };
+
+export const uploadOriginalTarballToS3 = async ({
+  pathWithNamespace,
+  tarballFilePath,
+  version,
+}: {
+  pathWithNamespace: string;
+  tarballFilePath: string;
+  version: string;
+}) => {
+  await getS3Client().send(
+    new PutObjectCommand({
+      Bucket: getS3Bucket(),
+      Body: await fs.readFile(tarballFilePath),
+      ContentType: "application/x-gtar",
+      Key: resolveS3ResourceKey(
+        "block-uploads",
+        `${stripLeadingAt(pathWithNamespace)}/${version}.tar.gz`,
+      ),
+    }),
+  );
+};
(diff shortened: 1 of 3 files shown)


apps/site/src/lib/api/blocks/s3.ts
```diff
@@ -277,3 +277,25 @@ export const validateExpandAndUploadBlockFiles = async ({
...
+  tarballFilePath: string;
+  version: string;
+}) => {
+  await getS3Client().send(
+    new PutObjectCommand({
+      Bucket: getS3Bucket(),
+      Body: await fs.readFile(tarballFilePath),
+      ContentType: "application/x-gtar",
+      Key: resolveS3ResourceKey(
+        "block-uploads",
```
CiaranMn, line 296: I'm not sure about this folder name – it's not obvious why we have a `blocks-upload` for tarballs and `blocks` for the rest of the block files (which are also uploads) – maybe we can put these tarballs in the regular `blocks` folder, but using the same path you have here and stop wiping the folder when updating the block? We will be introducing version subfolders imminently anyway which will involve no more deleting of files, so we can start keeping them around.

CiaranMn: It seems unnecessary to split them up like this and have two trees which each represent author/name/version, one with the original tarball and one with the extracted files.

What about something like this?

```
blocks/[author]/[name]/[version]/contents/[...file]
blocks/[author]/[name]/[version]/tarball/[file].tar.gz
```

or
```
blocks/[author]/[name]/[version]/contents/[...file]
blocks/[author]/[name]/[version]/tarball.tar.gz
```

kachkaev: Splitting uploads and contents into tiers somewhat makes sense. Contents can be extracted from uploads at any time. Only uploads need to be mirrored between hubs (e.g. remote / local)

CiaranMn (approved):

apps/site/src/lib/api/blocks/s3.ts
CiaranMn: We spoke about this and agreed to split them up so that the uploads tree can be backed up independently of the extracted

PR #992, opened 2023-02-20
Type System improvements

Alfred-Mountfield:

Through usage we've discovered a number of small improvements we could make.
- The `VersionedUri` template string encourages people to put an additional slash when adding a `BaseUri`. This is incorrect as a `BaseUri` _must_ already have a trailing slash, as such, one of the slashes in the template string have been removed and the type improved slightly.
- The `default` field in `EntityType` isn't really useful in its current form, ideally users could define defaults on a per-property basis, this will require more thought and development.
- The `requiredLinks` field is now redundant in `EntityType` as there is no longer a choice for `type: "array"` of links. As such, a `required` link can be represented by `minItems: 1`.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203522138014423/f) _(internal)_

## 🚫 Blocked by

- [ ] #990 

## 🔍 What does this change?

See description

At the present, no. Although an addendum to the RFC should be added soon.

## ⚠️ Known issues

N/A

## 🐾 Next steps

- Update docs for 0.3

## 🛡 What tests cover this?

- (In `libs/@blockprotocol/type-system/crate`)
  - `cargo make lint`
  - `cargo make test` 
- (In the root)
  - `yarn workspace @blockprotocol/type-system build`
  - `yarn workspace @blockprotocol/type-system test`
 
## ❓ How to test this?

Follow instructions above

### DIFF
--- a/libs/@blockprotocol/type-system/crate/src/ontology/data_type/mod.rs
+++ b/libs/@blockprotocol/type-system/crate/src/ontology/data_type/mod.rs
@@ -200,7 +200,7 @@ mod tests {
         let data_type_ref = DataTypeReference::new(uri.clone());
 
         data_type_ref
-            .validate_uri(uri.base_uri())
+            .validate_uri(&uri.base_uri)
             .expect("failed to validate against base URI");
     }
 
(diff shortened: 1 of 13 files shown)


TimDiekmann (approved):
LGTM 🎉

PR #565, opened 2022-09-04
Remove hard dependency on R2 storage

kachkaev:
This PR replaces R2-specific logic with S3, which is an underlying API protocol. Blocks and avatars are placed into the same S3 bucket. Local dev is simplified by running [MinIO](https://min.io/) alongside MongoDB as an external service. MinIO is also used in integration tests.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202886867026855

PR #993, opened 2023-02-20
Fix access to block assets in a local S3 instance

kachkaev:

When publishing block to a local bucket, I noticed that its contents were not available at http://localhost:3000/hub.

Screenshot 2023-02-20 at 18 12 41

Opening direct asset links was showing this error:

Screenshot 2023-02-20 at 18 25 18

It turned out that port 9001 was for MinIO's [web console](https://min.io/docs/minio/linux/administration/minio-console.html#static-vs-dynamic-port-assignment), so I changed port to 9000 in our CI config and docs. However, accessing objects at port `9000` did not help either:

Screenshot 2023-02-20 at 18 24 49

After the investigation, I discovered that bucket permissions for local dev were not configure correctly. This PR fixes that.

## 🔗 Related links

- https://github.com/blockprotocol/blockprotocol/pull/565

- See diff

## ⚠️ Known issues

- Existing developers need to open  `apps/site/.env.local` and manually make this change:
  ```diff
  -S3_BASE_URL=http://localhost:9001/dev-bucket
  +S3_BASE_URL=http://localhost:9000/dev-bucket
  ```
  It will be also necessary to run `yarn dev:seed-db`. Alternatively, it should be possible to opn MinIO at http://localhost:9001, login with `dev-access-key` / `dev-secret-key` and change bucket access from _Private_ to _Public_ in the UI.
  

## 🐾 Next steps

- Publish HASH blocks
- Migrate published blocks to local dev server

## 🛡 What tests cover this?

- Manual

## ❓ How to test this?

1.  Update `S3_BASE_URL` `apps/site/.env.local`
1.  Call `yarn dev:seed-db`
1.  Create API token for `@alice`
1.  `cd` to any of the blocks and build it. Don't forget to prepend block name with `@alice/*` before doing that
1. Publish that block locally:
   ```
   BLOCK_PROTOCOL_SITE_HOST=http://localhost:3000 BLOCK_PROTOCOL_API_KEY=b10ck5.??.?? npx blockprotocol@latest  publish --yes
   ```
1. Confirm that the block works and its icon and preview are visible

### DIFF
--- a/apps/site/scripts/reset-s3-bucket.ts
+++ b/apps/site/scripts/reset-s3-bucket.ts
@@ -57,6 +57,7 @@ const script = async () => {
   await s3Client.send(
     new CreateBucketCommand({
       Bucket: getS3Bucket(),
+      ACL: "public-read",
     }),
   );
 
--- a/apps/site/scripts/reset-s3-bucket.ts
+++ b/apps/site/scripts/reset-s3-bucket.ts
@@ -3,6 +3,7 @@ import {
   DeleteBucketCommand,
   DeleteObjectsCommand,
   ListObjectsV2Command,
+  PutBucketPolicyCommand,
 } from "@aws-sdk/client-s3";
 import chalk from "chalk";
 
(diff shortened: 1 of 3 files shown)


apps/site/scripts/reset-s3-bucket.ts
```diff
@@ -54,9 +55,34 @@ const script = async () => {
     }
   }
 
+  const Bucket = getS3Bucket();
```
kachkaev, line 58: This file runs during local dev only, prod logic is not affected

CiaranMn (approved):

PR #995, opened 2023-02-20
Remove 0.3 coming soon banner

CiaranMn:

Remove the '0.3 coming soon' banner, since merging this branch will introduce the 0.3 version of the spec.


luisbettencourt (approved):

nonparibus:
@CiaranMn We should persist this component in a Storybook. We could probably set one up for the Þ pretty rapidly (but should wait until after the 28th to do so).

PR #1004, opened 2023-02-21
Fix mock-block-dock datastore visualization

Alfred-Mountfield:

At some point in the previous changes the datastore visualization was broken in `mock-block-dock`. This PR fixes it, and improves upon the previous implementation by overriding the edge labels to make them more readable.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203994531454209/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Stop using `RecordId` as the ID's of the nodes, instead opting to use `VertexId`
- Improve a lot of the typings to rely on the actual typings of the library
- Improve the display label of edges by using a prettified name of the edge-kind, rather than the exceptionally long ID

- I don't believe so

## ⚠️ Known issues

- This implementation was always a quick addition to add some functionality, there's a fair bit of UX work we can do to improve the component, including configuration options, filtering, etc.
- The usability of the graph is fairly strongly compromised due to relying on auto-generated entity IDs and a quickly-implemented `getLabel` function we've written. We have plans to think about label properties on entities, or other approaches, and we could benefit from that here.

## 🐾 Next steps

N/A

## 🛡 What tests cover this?

None to my knowledge

## ❓ How to test this?

- Checkout [Expand react block to allow creating links](https://github.com/blockprotocol/blockprotocol/commit/f2ad2c9a979dae0eddd062ac054486476cec4889)
- Create some entities and links
- Click on an entity in the visualization to see the direct neighbors are highlighted
- Delete entities to see that edges and such are correctly deleted

## 📹 Demo

https://user-images.githubusercontent.com/25749103/220348735-67abe9f7-ac17-41f6-97e3-e5d8e26fbd54.mov

### DIFF
--- a/libs/mock-block-dock/dev/test-react-block.tsx
+++ b/libs/mock-block-dock/dev/test-react-block.tsx
@@ -1,12 +1,11 @@
-import { Entity } from "@blockprotocol/graph";
+import { extractBaseUri } from "@blockprotocol/graph";
 import {
   type BlockComponent,
   useGraphBlockModule,
 } from "@blockprotocol/graph/react";
 import { getRoots } from "@blockprotocol/graph/stdlib";
 import { useHook, useHookBlockModule } from "@blockprotocol/hook/react";
-import { extractBaseUri } from "@blockprotocol/type-system/slim";
-import { useMemo, useRef, useState } from "react";
+import { useMemo, useRef } from "react";
 
 import { propertyTypes } from "../src/data/property-types";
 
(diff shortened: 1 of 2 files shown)


thehabbos007 (approved):
Looks good to me and works locally. I'm wondering if we want to make the labels show on click for any of the nodes? Not sure if that complicates things
I.e. only when i click the top entity do I get the edge labels


https://user-images.githubusercontent.com/6988668/220367961-84bdda90-2141-4bb3-9dbb-da6b8f005eb9.mp4

Alfred-Mountfield:
> I'm wondering if we want to make the labels show on click for any of the nodes? Not sure if that complicates things

We can do this as a future extension as we revisit the UI :)

PR #1006, opened 2023-02-21
Register callbacks before initializing core handler

nathggns:

Currently, all operations in modules requiring the core handler be created (which doesn't happen until the element is available) are queued and processed after the core handler is created and initialised. This includes registering callbacks, which means any callbacks meant to catch messages sent when the core handler is initialised will not be registered in time.

This fixes that by processing `registerCallback` and `removeCallback` calls prior to initialising the core handler. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204003835300584/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

Creates a new high-priority queue which `registerCallback` and `removeCallback` handlers are placed into if called before the call handler is available. This queue is processed immediately after creating the core handler, but before calling initialise. This should ensure all callbacks are set up as expected before calling initialise.

N/A

## ⚠️ Known issues

N/A

## 🐾 Next steps

N/A

## 🛡 What tests cover this?

N/A

## ❓ How to test this?

Revert 684b4cbe9ce162830a04d127aae9057d20944d00, built the core package, and ensure you see the `blockEntitySubgraph` logged in mock block dock


libs/@blockprotocol/core/src/module-handler.ts
```diff
@@ -25,6 +25,14 @@ export abstract class ModuleHandler {
    */
   private coreQueue: CoreHandlerCallback[] = [];
 
+  /**
+   * If we register callbacks prior to creating the core handler, we want to
+   * register those on the core handler once it is available, but before we
+   * call initialise on it, to ensure callbacks which would catch messages sent
+   * during initialise are registered. To enable that, we have a separate queue.
+   */
+  private preCoreInitialiseQueue: CoreHandlerCallback[] = [];
```
Alfred-Mountfield, line 34:
```suggestion
  /**
   * If we register callbacks prior to creating the core handler, we want to
   * register those on the core handler once it is available, but before we
   * call initialize on it, to ensure callbacks which would catch messages sent
   * during initialize are registered. To enable that, we have a separate queue.
   */
  private preCoreInitializeQueue: CoreHandlerCallback[] = [];
```

libs/@blockprotocol/core/src/module-handler.ts
```diff
@@ -99,6 +107,8 @@ export abstract class ModuleHandler {
       throw new Error("Could not initialize – missing core handler");
     }
 
+    this.processCoreCallbackQueue(this.preCoreInitialiseQueue);
```
Alfred-Mountfield, line 110:
```suggestion
    this.processCoreCallbackQueue(this.preCoreInitializeQueue);
```

libs/@blockprotocol/core/src/module-handler.ts
```diff
@@ -190,6 +200,16 @@ export abstract class ModuleHandler {
...
 
+  /**
+   * When adding/removing callbacks before calling the core handler is
+   * available, we want to queue these in a queue which will be processed before
+   * calling initialising once we create the core handler, to ensure callbacks
+   * are properly set up before calling initialise
+   */
+  private getRelevantQueueForCallbacks() {
+    return this.coreHandler ? this.coreQueue : this.preCoreInitialiseQueue;
+  }
```
Alfred-Mountfield, line 211:
```suggestion
  /**
   * When adding/removing callbacks before calling the core handler is
   * available, we want to queue these in a queue which will be processed before
   * calling initializing once we create the core handler, to ensure callbacks
   * are properly set up before calling initialize
   */
  private getRelevantQueueForCallbacks() {
    return this.coreHandler ? this.coreQueue : this.preCoreInitializeQueue;
  }
```

libs/@blockprotocol/core/src/module-handler.ts
CiaranMn: 🇺🇸

CiaranMn (approved):
Testing this on the custom element template and it's fixed the issue – agree with Alfie's suggestions for consistency but looks good otherwise

PR #1008, opened 2023-02-21
Fix entity default schema

CiaranMn:

`null` is not a valid value for `EntityType["links"]`, but that's what it was being set as in the db if no value for it was provided.

This introduces a fallback to `{}`, which is valid, and prevents nullish fields being set on an `EntityType` to avoid further such traps (unless there is some future reason to allow `null` for any value).


Alfred-Mountfield (approved):
I think it might be worth considering changing the type system to not have optional properties if we do this.

Otherwise, as a user I'd assume best practice would be to leave `links` undefined as opposed to setting it to `{}`.

PR #1013, opened 2023-02-22
Fix missing generic in export

Alfred-Mountfield:

A generic was missed when re-exporting a type

### DIFF
--- a/libs/@blockprotocol/graph/src/temporal/main.ts
+++ b/libs/@blockprotocol/graph/src/temporal/main.ts
@@ -3,7 +3,7 @@
  * This defines the main types and type-guards used when working with the Graph module.
  */
 
-import {BaseUri} from "@blockprotocol/type-system/slim";
+import { BaseUri } from "@blockprotocol/type-system/slim";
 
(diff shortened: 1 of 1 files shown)


libs/@blockprotocol/graph/src/temporal/main.ts
```diff
@@ -3,7 +3,7 @@
  * This defines the main types and type-guards used when working with the Graph module.
  */
 
-import { BaseUri } from "@blockprotocol/type-system/slim";
+import {BaseUri} from "@blockprotocol/type-system/slim";
```
CiaranMn, line 6: This is probably a formatting error

Alfred-Mountfield: Hm, I wonder why it wasn't picked up by my IDE

Alfred-Mountfield: Seems like prettier had become badly configured. Fixed it, cheers

CiaranMn (approved):

PR #1020, opened 2023-02-23
RFC - Link Entities

Alfred-Mountfield:

This RFC introduces the concept of "link entities", removing the outdated concept of the link primitive from the Block Protocol graph model, instead having a special-case of entities that can link between other entities.

This is (in my mind) a beautiful solution that achieves a few key goals while actually reducing the surface area and implementation complexity:
- Links can now have properties associated with them, such as a date for a marriage between two people
- Links can be referenced in the system, allowing the creation of higher-order-abstractions such as `Path`s and `Group`s

**Note:** these changes _have actually made their way_ into the current newest version of the spec. This was done due to some constraints on resources at the time, and we would like to ensure the decision is captured in a persistent manner, and are retroactively making sure that there is a place to discuss these changes if issues / new ideas should arise.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1203250866216846/f) _(internal)_


rfcs/text/1020-link-entities.md
```diff
@@ -0,0 +1,385 @@
...
+    }
+  },
+  "required": ["leftEntityId", "rightEntityId"],
+  "additionalProperties": false
+}
+```
+
+#### "Left" and "Right"
+
+At the present, the system specifies uni-directional links, that is, links that go in a single direction from one entity to another entity.
```
nonparibus, line 326: I don't think uni/bi/multidirectionality need hyphenating. I know we decided this for bi/multitemporality. Cc @CiaranMn

Alfred-Mountfield: Google seems to suggest it does not, but I wouldn't worry _toooo_ much about grammar, this is just an RFC 😅 

```suggestion
At the present, the system specifies unidirectional links, that is, links that go in a single direction from one entity to another entity.
```

rfcs/text/1020-link-entities.md
```diff
@@ -0,0 +1,385 @@
...
+```
+
+The fields within `"linkData"` are as follows:
+
+- `"leftEntityId"`: The `entityId` of the entity on the "left" hand side of the link entity. In most scenarios this will be the "source" of the link entity.
+- `"rightEntityId"`: The `entityId` of the entity on the "right" hand side of the link entity. In most scenarios this will be the "target" of the link entity.
+- `"leftToRightOrder"`: A value used to order this link entity against other link entities of the same type with the same `leftEntityId`.
+- `"rightToLeftOrder"`: A value used to order this link entity against other link entities of the same type with the same `rightEntityId`.
+
+The choice of "left" and "right" as opposed to "source" and "target" is an optimistic attempt to leave room for an extension of the system in the future to allow for "bi-directional" links (where a "source" and "target" would not make sense).
```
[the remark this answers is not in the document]
Alfred-Mountfield: Changing to a hypergraph (where an edge can connect to multiple vertices on both ends) would be a pretty substantial change, and not one I've heard a great argument for at the moment. This seems better modelled (to me) with high-order-abstractions such as a `Group`, etc. as that is more extensible.

However, I don't see how allowing multiple endpoints would actually support bidirectional links, if you have

```
Sources   Link         Destinations
[Bob, ] --Friend Of--> [Charlie, ] 
```

I don't see how that gives us the bidirectionality, were you imagining us then doing something like this?
```
Sources          Link         Destinations
[Bob, Charlie] --Friend Of--> [Charlie, Bob] 
```

Because that would seem to imply `Bob` is friends with `Bob`, and if we established explicit mappings within the lists (adding a _**lot**_ of complexity), then we're basically back to just having an ordered list of links (`[Bob --Friend of--> Charlie, Charlie --Friend of--> Bob]`)

rfcs/text/1020-link-entities.md
```diff
@@ -0,0 +1,385 @@
...
```
Alfred-Mountfield, line 216:
```suggestion
The choice of "left" and "right" as opposed to "source" and "target" is an optimistic attempt to leave room for an extension of the system in the future to allow for "bidirectional" links (where a "source" and "target" would not make sense).
```

rfcs/text/1020-link-entities.md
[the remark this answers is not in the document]
nonparibus: I was imagining that, and that you would simply exclude/ignore self-referential links. Is there an instance an entity would ever actually want to link to itself?

rfcs/text/1020-link-entities.md
```diff
@@ -0,0 +1,385 @@
...
+      ┌──► │B│ ───┐
+┌─┐   │    └─┘    │   ┌─┐
+│A│ ──┤           ├─► │D│
+└─┘   │    ┌─┐    │   └─┘
+      └──► │C│ ───┘
+           └─┘
+```
+
+Now, we could apply the same strategy as above, and create a "Path" entity type, and an associated "Has Element" link type.
+They could then use the _order_ of the link to maintain an ordered collection of the entities that make up the path.
```
CiaranMn, line 96: The order of what link? I'm not following this example (including the diagrams below), maybe easier to talk sync. This should be really clear as it's leaned on in the rationale

CiaranMn (commented):
A few comments – mainly I think the 'path' explanation could be clearer

rfcs/text/1020-link-entities.md
```diff
@@ -0,0 +1,385 @@
...
+At the present, the following types of edges exist:
+
+- Entity Type -> Link Type (constrains the types of links that can come from an entity)
+- Entity Type -> Entity Type (constrains the types of entities that can be linked to from an entity)
+- Entity Type -> Property Type (constrains the properties that can appear on an entity)
+- Property Type -> Property Type (constrains the types of properties that can appear within an object in another property)
+- Property Type -> Data Type (constrains the type of data that can be stored within a property)
+- Entity -> Entity Type (describes the associated entity type of an entity)
+- Entity -> Link (the outgoing or incoming link of an entity)
+- Link -> Entity (the source or target entity of a link)
```
Alfred-Mountfield, line 45:
```suggestion
At the present, the following kinds of edges exist (as well as their respective reversed edges, which are omitted for brevity):

- Entity Type -> Link Type (constrains the types of links that can come from an entity)
- Entity Type -> Entity Type (constrains the types of entities that can be linked to from an entity)
- Entity Type -> Property Type (constrains the properties that can appear on an entity)
- Property Type -> Property Type (constrains the types of properties that can appear within an object in another property)
- Property Type -> Data Type (constrains the type of data that can be stored within a property)
- Entity -> Entity Type (describes the associated entity type of an entity)
- Entity -> Link (identifies a link as one that is outgoing from this entity)
- Link -> Entity (identifies an entity as the destination of this link)
```

rfcs/text/1020-link-entities.md
```diff
@@ -0,0 +1,385 @@
...
+       └──┬──┘
+          │
+          │
+ ┌───Has element───┐
+ │     │     │     │
+ │     │     │     │
+ ▼     ▼     ▼     ▼
+┌─┐   ┌─┐   ┌─┐   ┌─┐
+│A│   │B│   │C│   │D│
+└─┘   └─┘   └─┘   └─┘
```
Alfred-Mountfield, line 69:
```suggestion
                    ┌─────┐
     ┌──────────────┤Group├──────────────┐
     │              └┬───┬┘              │
     │               │   │               │
     │               │   │               │
     │           ┌───┘   └───┐           │
     │           │           │           │
     │           │           │           │
     │           │           │           │
Has element Has element Has element Has element
     │           │           │           │
     │           │           │           │
     ▼           ▼           ▼           ▼
    ┌─┐         ┌─┐         ┌─┐         ┌─┐
    │A│         │B│         │C│         │D│
    └─┘         └─┘         └─┘         └─┘
```

rfcs/text/1020-link-entities.md
```diff
@@ -0,0 +1,385 @@
...
+
+At the present, the following types of edges exist:
+
+- Entity Type -> Link Type (constrains the types of links that can come from an entity)
+- Entity Type -> Entity Type (constrains the types of entities that can be linked to from an entity)
+- Entity Type -> Property Type (constrains the properties that can appear on an entity)
+- Property Type -> Property Type (constrains the types of properties that can appear within an object in another property)
+- Property Type -> Data Type (constrains the type of data that can be stored within a property)
+- Entity -> Entity Type (describes the associated entity type of an entity)
+- Entity -> Link (the outgoing or incoming link of an entity)
```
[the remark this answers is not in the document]
CiaranMn: Looks good, thank you

rfcs/text/1020-link-entities.md
```diff
@@ -0,0 +1,385 @@
...
+
+## Creating Higher-order abstractions through composition of the primitives of the type system
+
+### Groups
+
+Due to the compositional design of the primitives of the Type System, it's possible to create high-order abstractions within the system.
+For example, one could want to define "groups" of entities within their system, in such a way that they would be programmatically resolvable through blocks, _with constraints guaranteed by the specification_.
+
+One could currently do this by creating a "Group" entity type, and an associated "Has Element" link type.
+They could then create a specific "Group" entity, and link it to the entities they wish to be part of the group:
```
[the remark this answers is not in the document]
Alfred-Mountfield: Have added a separate comment to adjust it, I think it will end up being too big to be readable but I'll commit it and see first

rfcs/text/1020-link-entities.md
```diff
@@ -0,0 +1,389 @@
...
+     │           │           │           │
+     │           │           │           │
+Has element Has element Has element Has element
+     │           │           │           │
+     │           │           │           │
+     ▼           ▼           ▼           ▼
+    ┌─┐         ┌─┐         ┌─┐         ┌─┐
+    │A│         │B│         │C│         │D│
+    └─┘         └─┘         └─┘         └─┘
+```
```
Alfred-Mountfield, line 74:
```suggestion
```text
                      ┌─────┐
     ┌────────────────┤Group├────────────────┐
     │                └┬───┬┘                │
     │                 │   │                 │
     │                 │   │                 │
     │            ┌────┘   └────┐            │
     │            │             │            │
     │            │             │            │
     │            │             │            │
Has element  Has element   Has element  Has element
     │            │             │            │
     │            │             │            │
     ▼            ▼             ▼            ▼
    ┌─┐          ┌─┐           ┌─┐          ┌─┐
    │A│          │B│           │C│          │D│
    └─┘          └─┘           └─┘          └─┘
```
```

File: rfcs/text/1020-link-entities.md (hunk 8 above)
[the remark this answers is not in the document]
Comment (Alfred-Mountfield) [reply rfcs/text/1020-link-entities.md]: image

Looks a little like a weird crab animal thing but it worked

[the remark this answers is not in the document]
Comment (CiaranMn) [reply rfcs/text/1020-link-entities.md]: 🦀

File: rfcs/text/1020-link-entities.md [hunk 10]
```diff
@@ -0,0 +1,385 @@
...
+┌─┐   ┌─┐   ┌─┐   ┌─┐
+│A│   │B│   │C│   │D│
+└─┘   └─┘   └─┘   └─┘
+```
+
+Being able to create such abstractions is a powerful feature of the Type System, as users are able to implicitly benefit from some of the following scenarios:
+
+- A user updates entity `A`, the link is still associated with the new version
+- A user deletes entity `A`, the Embedding Application (EA) knows that a link has been invalidated and is able to orchestrate/trigger a resolution flow
+- A user redefines the destination constraints of `Has element` links on the `Group` entity type to disallow entities of type `D`.
```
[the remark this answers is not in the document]
Alfred-Mountfield: Explaining them feels a bit superfluous but happy to do so if you think it's confusing to not have, or otherwise helpful to do so.

- The first one is implicit in the Block Protocol rather than the type system, updating an entity does not change its `EntityId`, identity is preserved, and links are based on identity.
- Similarly, with the second one, the application is _explicitly required to be aware of those primitives_, I do not believe "I think this applies to most if not all ways of modelling relational data" is quite true in this case. The point is that if users start creating their own relationship definitions, e.g. a block decides to just store an `EntityId` in a property, the application is not _explicitly_ aware of that link, and doesn't know it's invalidated. Sure, it could _theoretically_ be introspecting every single property ever and trying to intelligently identify relationships, but that's not an expectation of EAs.
- Your explanation of what it's due to is correct. As that's not actually a _part_ of this RFC, I do think it perhaps doesn't need to be fully explained as it's captured as a core behaviour of the system and explaining it serves to extend the RFC with calling out to other features.. 

I'll open a new comment with a suggestion because unfortunately the range in this one is missing the last line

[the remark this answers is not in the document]
Alfred-Mountfield: Does this help:
[Add some supporting information to example situations](https://github.com/blockprotocol/blockprotocol/pull/1020/commits/0fe9b290b2ae5c55bf75fd691865e12e4de2db0c)

rfcs/text/1020-link-entities.md
Alfred-Mountfield: The order of the `Has element` links that are displayed directly below

rfcs/text/1020-link-entities.md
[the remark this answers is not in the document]
CiaranMn: Yes, thank you

rfcs/text/1020-link-entities.md
Alfred-Mountfield: https://github.com/blockprotocol/blockprotocol/pull/1020/commits/1d0294fa7f8e3ee53c19fd9524a5e50f841a51f4

https://github.com/blockprotocol/blockprotocol/pull/1020/commits/bb2b1fae04ba7a252be85444b5535741f2cbaccc

Alfred-Mountfield: [Make styling of diagrams consistent](https://github.com/blockprotocol/blockprotocol/pull/1020/commits/1d0294fa7f8e3ee53c19fd9524a5e50f841a51f4)


[Add clarification](https://github.com/blockprotocol/blockprotocol/pull/1020/commits/bb2b1fae04ba7a252be85444b5535741f2cbaccc)

CiaranMn: The clarification makes it much better, thank you

CiaranMn (approved):

TimDiekmann (dismissed):

TimDiekmann (approved):

thehabbos007 (approved):

PR #1021, opened 2023-02-23
Update HTML block template

Alfred-Mountfield:

The HTML block template had fallen out of sync with the recent changes. This PR _mostly_ fixes them, although some of the code around the block entity type, and property types, will need updating once we merge `0.3` into main and have stable type URLs on the website.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204029869185044/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

See diff, it's small

To my knowledge, no.

## ⚠️ Known issues

- There's a chicken and egg problem with the `esm.sh` part of the template at the moment, it requires a release of the code to be able to point to the package.
- There currently isn't a stable prod URL for types to be hosted at, so I have opted for adjusting the URLs to be "example.com" for now, and modifying the `block-schema` to be an entity type definition. Really the `schema` property of the `block-metadata` should just be the Versioned URL of the schema (which we can have once the changes hit main)
- ~~Trying to run the block template gives a strange error about an object being passed as children to the react component (seems that the props aren't being correctly inserted into `MockBlockDock`)~~ this was caused by our inline comments..

## 🐾 Next steps

Fixup the schema and types definitions, and esm.sh link, when changes hit `main`

## 🛡 What tests cover this?

Unsure if any

## ❓ How to test this?

- Currently it's broken, see known issues above


libs/block-template-html/block-metadata.json
```diff
@@ -11,7 +11,7 @@
   },
   "protocol": "0.2",
   "displayName": "Template",
-  "examples": [{ "name": "World" }],
+  "examples": [{ "http://example.com/types/property-type/name/": "World" }],
```
CiaranMn, line 14: Could use the same as in our other templates, which is a real URL

libs/block-template-html/block-schema.json
```diff
@@ -1,11 +1,14 @@
 {
+  "$id": "http://example.com/types/entity-type/example/v/1",
```
CiaranMn, line 2: Ditto other comment – could use `Thing` from the other templates

libs/block-template-html/block-schema.json
```diff
@@ -1,11 +1,14 @@
...
   "properties": {
-    "name": {
-      "type": "string"
+    "http://example.com/types/property-type/name/": {
+      "$ref": "http://example.com/types/property-type/name/v/1"
     }
   },
-  "required": ["name"],
+  "required": ["http://example.com/types/property-type/name/"],
+  "additionalProperties": false,
```
CiaranMn, line 12: 🧐

libs/block-template-html/block-schema.json
CiaranMn: In fact we should probably just delete this file, it is a relic from when blocks bundled their schemas

libs/block-template-html/block-metadata.json
Alfred-Mountfield: Happy to do that now

libs/block-template-html/block-metadata.json
```diff
@@ -11,7 +11,7 @@
   },
   "protocol": "0.2",
```
CiaranMn, line 12:
```suggestion
  "protocol": "0.3",
```

Alfred-Mountfield: Should we even have this file or should it be in `package.json` now?

CiaranMn: This has a different bundling script which doesn't do the 'grab things from package.json'. so we potentially could but would need to update the bundling to generate this file

Alfred-Mountfield: Gotcha, keep the file for now then

Alfred-Mountfield: [Remove redundant files, use real types](https://github.com/blockprotocol/blockprotocol/pull/1021/commits/20ab0d7034706eb82f4e916d10e57f3563a5387e)

libs/block-template-html/block-metadata.json
Alfred-Mountfield: [Remove redundant files, use real types](https://github.com/blockprotocol/blockprotocol/pull/1021/commits/20ab0d7034706eb82f4e916d10e57f3563a5387e)

CiaranMn (approved):

CiaranMn (approved):

PR #1024, opened 2023-02-23
Fix TypeScript and linting errors in 0.3

CiaranMn:

Prepare the `0.3` branch for merging into `main` by fixing TypeScript and linting errors.

It also starts running the `CI` workflow on PRs into `0.3` so we can monitor it (it's not a required check still).

Tests are expected to fail. I will investigate in follow up, potentially just skipping the ones that are problematic in order to do an initial merge.


libs/block-scripts/scripts/codegen.js
```diff
@@ -2,12 +2,17 @@ import { writeFileSync } from "node:fs";
...
 import chalk from "chalk";
 import fs from "fs-extra";
 
 import { blockRootDirPath } from "../shared/paths.js";
 
+/**
+ * @param {string} url
+ * @returns {boolean}
+ */
+const isVersionedUrlValid = (url) => /(.+\/)v\/(\d+)(.*)/.test(url);
```
CiaranMn, line 14: The `type-system` package has DOM typings in it. This package only has `Node` types. Technically it is correct for this to be an error because `type-system` might use APIs that are unavailable in this command line tool, and there is no building or polyfilling going on here. So it would be inaccurate to just add the DOM types to the TS config for this package.

I've opted to just do a minimal URL validation instead of sinking more time into worrying about it.

libs/block-scripts/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "@local/tsconfig/node.json",
-  "include": ["bin.js", "scripts", "shared"]
+  "include": ["bin.js", "scripts", "shared"],
+  "compilerOptions": {
+    "skipLibCheck": true
```
CiaranMn, line 5: Even with `validateVersionedUrl` changed, `graph` is still throwing errors because of its dependency on `type-system`. I'm not sure if there's a better way of approaching this. We have `skipLibCheck: true` in the frontend already.

libs/mock-block-dock/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   "extends": "@local/tsconfig/dom+jsx.json",
-  "include": ["assets.d.ts", "src", "dev"]
+  "include": ["assets.d.ts", "src", "dev"],
+  "compilerOptions": {
+    "skipLibCheck": true
+  }
```
CiaranMn, line 6: `echarts` has a bunch of type errors which this allows us to ignore. I don't know why they weren't happening before or if there's a better way of fixing this rather than effectively ignoring the problem (and typings in every other library).

libs/mock-block-dock/src/service-module-callbacks.ts
```diff
@@ -34,7 +34,13 @@ const callExternalApiMethod = async (params: {
...
 export const constructServiceModuleCallbacks = (params: {
   blockProtocolApiKey?: string;
-}): EmbedderServiceMessageCallbacks => {
+}): Omit<
+  EmbedderServiceMessageCallbacks,
+  "openaiCreateImage" | "openaiCompleteText"
+> & {
+  openaiCreateImage?: never;
+  openaiCompleteText?: never;
+} => {
```
CiaranMn, line 43: These callbacks will be added in a future PR, and this type will then error when they are and can be reverted

apps/site/tests/integration/schemapage.test.ts
```diff
@@ -38,20 +36,20 @@ test("schema page should contain key elements", async ({ page }) => {
 
   await expect(
     page.locator(
-      "text=You can use these entity types as the expected value for a property in another schema.",
+      "text=You can use these entity types as the expected value for a property in another ",
```
Alfred-Mountfield, line 39: I appreciate you're just fixing the test but where does this occur. How can an entity type be an expected value for a property?

Alfred-Mountfield (commented):
Few comments but nothing that necessarily needs to block. Looking forward to a green CI again!

libs/block-scripts/tsconfig.json
CiaranMn: Yes, it's the WASM stuff

apps/site/tests/integration/schemapage.test.ts
CiaranMn: Oh, I think this was a mass replace in the file that went awry. It was probably something to do with links

CiaranMn: It's looking for copy that doesn't exist and we can clean up when fixing the tests (this PR is just making TS compile)

Alfred-Mountfield (commented):
`tsc` and `eslint` are failing due to the site-map by the way

CiaranMn:
> tsc and eslint are failing due to the site-map by the way

@Alfred-Mountfield  Was just looking at that – I don't know why it isn't being generated, the warm up repo step is skipping some actions. Will try fix in a future PR

PR #1025, opened 2023-02-23
Fix turbo pipeline

CiaranMn:

Fix the following GitHub actions:

1. CI / Linting
2. CI / NPM Packages

- This restores a missing `codegen` running in `yarn prepare` – this is required to trigger `@apps/site`'s `codegen`, because `@apps/*` are filtered out of the turbo build command in `yarn prepare`. Without it, we don't get `site-data.json`
- Vendors in GeoJSON types which were being a pain in `@blockprotocol/service`
- Don't import from `src/lib/config.ts` because doing so creates a Vercel error for some reaosn


libs/@blockprotocol/service/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@blockprotocol/service",
-  "version": "0.0.1-canary.9",
+  "version": "0.0.2",
```
Alfred-Mountfield, line 3: Do we want to be doing these manually or should `changesets` be handling this for us?

libs/@blockprotocol/service/src/mapbox-types.ts
```diff
@@ -1,4 +1,173 @@
-import type { GeoJSON } from "geojson";
+/**
```
Alfred-Mountfield, line 1: Do we want to say where we vendored these in from?

package.json
```diff
@@ -36,7 +36,7 @@
     "dev": "yarn workspace @apps/site dev",
     "dev:seed-db": "yarn workspace @apps/site dev:seed-db",
     "dev:db": "yarn workspace @apps/site dev:db",
-    "prepare": "patch-package --error-on-warn && turbo run build \"--filter=!./apps/*\""
+    "prepare": "patch-package --error-on-warn && turbo run codegen && turbo run build \"--filter=!./apps/*\""
```
Alfred-Mountfield, line 39: I'm trying to think of a better way around this and I can't think of one right now but this does feel like it's not aligning with how we should be using Turbo.

If the prepare step is dependent on codegen, isn't that something that should be represented in the Turbo dependency graph?

Also why do we need "site-data.json" if we're not building the site?

turbo.json
```diff
@@ -29,14 +29,6 @@
...
       "outputs": []
     },
-    "block-template-custom-element#codegen": {
-      "dependsOn": ["^build"],
-      "outputs": ["src/types.gen.ts"]
-    },
-    "block-template-react#codegen": {
-      "dependsOn": ["^build"],
-      "outputs": ["src/types.gen.ts"]
-    },
```
Alfred-Mountfield, line 39: Why did we remove these?

package.json
CiaranMn: > why do we need "site-data.json" if we're not building the site?

We don't `build` it but we do `lint:tsc` it, and it depends on the file to compile.

> If the prepare step is dependent on codegen, isn't that something that should be represented in the Turbo dependency graph?

Yes – I tried add this but it didn't work for some reason:
```json
    "lint:tsc": {
      "dependsOn": ["^build"],
      "outputs": []
    },
```
and because this is only restoring part of the `prepare` command that is in `main` anyway (which I removed in an earlier PR), I thought we could come back to it later, since it's an existing issue (merging this will mean no diff on this line from `0.3` to `main`)

Alfred-Mountfield: I'm fine with us revisiting later :)

Out of curiosity did you try:
```
    "lint:tsc": {
      "dependsOn": ["^codegen"],
      "outputs": []
    },
```

or 

```
    "lint:tsc": {
      "dependsOn": ["codegen"],
      "outputs": []
    },
```

I appreciate it's finicky to test

CiaranMn: This may be related to the fact that we exclude the frontend from the `turbo build` command. In any case it's extremely tedious and time consuming to test.

Alfred-Mountfield (approved):

libs/@blockprotocol/service/package.json
CiaranMn: It should, be to date this package has been published manually while being developed (which makes sense to get it off the ground). We can let `changesets` take over soon.

package.json
CiaranMn: I didn't try either of those things, no. We can do so as part of a general 'clean up dependency graphs in turbo' thing.

We can run `turbo something --graph` to get a visual representatoin to help debug

PR #1027, opened 2023-02-23
Don't wipe R2 folders when uploading new block files

CiaranMn:

We previously wiped the block's source file folder in R2 before uploading the latest version. This has the effect of immediately removing the previous version's source file, which is bad if someone is relying on it (e.g. has stored it in a page somewhere).

This PR stops doing that. While it means there will be an accumulation of old source files, they won't do any harm and anything from the new files that clashes with the old will overwrite them. This doesn't apply to source files as they have a hash in the filename.

This will all become irrelevant when we start persisting previous versions, which will be in the next few weeks.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203979552136008/f) _(internal)_

### DIFF
--- a/apps/site/src/lib/api/blocks/s3.ts
+++ b/apps/site/src/lib/api/blocks/s3.ts
@@ -1,10 +1,6 @@
 import path from "node:path";
 
-import {
-  DeleteObjectsCommand,
-  ListObjectsV2Command,
-  PutObjectCommand,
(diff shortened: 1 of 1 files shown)


kachkaev (approved):
It's somewhat unfortunate that we need to use this hack, but I understand why it’s there given the deadlines.

AFAIU, CI failures are unrelated to this PR.

PR #1030, opened 2023-02-24
Add "Tax ID" field to billing page

benwerner01:

This PR adds a "Tax ID" section to the billing page for users with a stripe subscription.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1203880493107350/f) _(internal)_

## 🔍 What does this change?

See commits.

Not that I'm aware of.

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

1. Checkout the branch
2. Add the `NEXT_PUBLIC_BILLING_FEATURE_FLAG=1` env var to your `app/site/.env.local` to enable the billing feature flag
3. Run the local developer environment for the BP site
4. Checkout the HASH internal API repo on the `bw/set-tax-id` branch (if it hasn't merged yet)
5. Follow the instructions in the internal API README to run the local developer environment, except for the mongo db external service (as the BP site is already running mongo db)
6. Login as a user and navigate to `/settings/billing`
7. Ensure the logged in user is subscribed to a paid subscription tier (by upgrading to "hobby" or "pro" for example)
8. Observe the "Tax ID" input and ensure you're able to input a valid Tax ID for a region

## 📹 Demo

https://user-images.githubusercontent.com/42802102/221115934-db7a26bb-817b-4012-93ba-1e64e2353730.mov


apps/site/src/pages/settings/billing-settings-panel/tax-id-section.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const [existingTaxId, setExistingTaxId] = useState<StripeTaxId | null>();
+
+  const taxIdInputRef = useRef<HTMLInputElement>(null);
+  const [taxIdValue, setTaxIdValue] = useState<string>("");
+  const [errorMessage, setErrorMessage] = useState<ReactNode>();
+  const [loading, setLoading] = useState<boolean>();
+  const [stripeTaxIdType, setStripeTaxIdType] =
+    useState<StripeTaxIdType | null>(null);
+
+  const fetchTaxId = useCallback(async () => {
```
CiaranMn, line 39: This is only used inside an effect, it could just be inside that function

CiaranMn (approved):
One non-blocking comment

nonparibus:
**Non-blocking observation:** under the `United Kingdom` dropdown listings it's not clear that `EU VAT` refers to Northern Irish VAT numbers, while `GB VAT` refers to the rest of the UK (whose tax IDs are in the example shown as being entered in the EU-style format, which is to say prefixed with `GB`).

The useful distinguishing information _has_ been included in the `StripeTaxIdType` `description` data, though, so we should expose this to users. I've designed a dropdown (based on one found in the HASH entity type editor) that incorporates space for the descriptive sentence to be included as well. See: [Figma](https://www.figma.com/file/Fpmde3E9nJMb4VLkPTYgrM/blockprotocol.org?node-id=5601%3A41955&t=QLsdcYnjqb7op9bd-1) & [implementation ticket](https://app.asana.com/0/0/1204022529039335/f) (_internal links_)

apps/site/src/pages/settings/billing-settings-panel/tax-id-section.tsx
benwerner01: Yes, but because it's async it'd require something like `void (async () => {})()`, is that still preferable?

nonparibus (approved):

apps/site/src/pages/settings/billing-settings-panel/tax-id-section.tsx
CiaranMn: Ah, true. It's fine as it is

CiaranMn (approved):

PR #1032, opened 2023-02-24
Save exact versions when adding dependencies. Pin react

CiaranMn:

The block sandbox (and therefore tests) is failing because of a React version mismatch. The underlying issue is that esm.sh has a recent bug whereby version ranges pull the most recent version regardless of tag, which is installing `react-dom@18.3.0@next` alongside `react@18.2.0`. 

This PR pins React to `18.2.0` which should fix it.

It also adds an `.npmrc` with `save-exact=true` so that we save exact versions always, for additional safety against potential version shifting (we use unstable APIs from libraries, for example, which might break outside of semver).


luisbettencourt (approved):

luisbettencourt (dismissed):

Alfred-Mountfield (approved):

apps/site/src/pages/api/rewrites/sandboxed-block-demo.api.ts
```diff
@@ -40,8 +40,12 @@ const handler: NextApiHandler = async (req, res) => {
 
   const mockBlockDockVersion = packageJson.dependencies["mock-block-dock"];
 
-  const reactVersion =
-    blockMetadata.externals?.react ?? packageJson.dependencies.react;
+  // // @todo restore this when esm.sh treats version ranges correctly
+  // //   it current pulls non-latest versions of packages if a range is supplied
+  // const reactVersion =
+  //   blockMetadata.externals?.react ?? packageJson.dependencies.react;
+
```
Alfred-Mountfield, line 47:
```suggestion
  // @todo restore this when esm.sh treats version ranges correctly
  //    it currently pulls non-latest versions of packages if a range is supplied
  // const reactVersion =
  //   blockMetadata.externals?.react ?? packageJson.dependencies.react;

```

CiaranMn: This is intentional, so that you can select the whole block and uncomment it, and still be left with a comment and code. Maybe it's a bit pointless but it's not doing any harm and isn't a formatting error

Alfred-Mountfield: I'm a bit confused. When you restore it you want the comment still there?

Alfred-Mountfield: You want it to look like
```
  // @todo restore this when esm.sh treats version ranges correctly
  //   it current pulls non-latest versions of packages if a range is supplied
  const reactVersion =
    blockMetadata.externals?.react ?? packageJson.dependencies.react;
```

once it's uncommented?

Alfred-Mountfield: If so, what is the "restore this" referring to

CiaranMn: I have applied your suggestion, thank you

Alfred-Mountfield (approved):

PR #1034, opened 2023-02-24
Update Graph module schema specs

TimDiekmann:

Brings the `graph-module.json` definitions in line with methods defined on BP


libs/@blockprotocol/graph/src/graph-module.json
```diff
@@ -181,512 +170,10 @@
...
-      "data": {
-        "type": "object",
-        "properties": {
-          "link": {
-            "description": "The created link",
-            "$ref": "https://blockprotocol.org/types/modules/graph/link"
-          }
-        },
-        "required": ["link"]
+        "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/subgraph"
```
Alfred-Mountfield, line 173: Does this match the type in the package?

It looks like it returns
```ts
{
  data:  {
    results: Subgraph<EntityRootType>;
    operation: AggregateOperationInput
  }
}
```

whereas here you're just returning 
```ts
{
  data: Subgraph<EntityRootType>
}
```

libs/@blockprotocol/graph/src/graph-module.json
```diff
@@ -163,14 +158,8 @@
...
-        "type": "object",
-        "properties": {
-          "operation": {
-            "description": "The aggregation operation to apply",
-            "$ref": "https://blockprotocol.org/types/modules/graph/aggregation-operation"
-          }
-        },
-        "required": ["operation"]
+        "description": "The aggregation operation to apply",
+        "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/aggregation-operation"
```
Alfred-Mountfield, line 162: Shouldn't this be something like this?

```ts
"data": {
  "description": "The aggregation operation to apply",
  "type": "object",
  "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/aggregation-operation",
  "properties": {
    "graphResolveDepths":  {
       "$ref": ...
    }
  }
}
```

(I think it would be clearer if `aggregation-operation` didn't include the `"operation"` key in it)

apps/site/public/types/modules/graph/0.3/schema/aggregation-operation.json
```diff
@@ -1,87 +1,125 @@
...
-                  },
-                  "operator": {
-                    "$ref": "#/$defs/FilterOperatorRequiringValue"
+        "multiFilter": {
+          "type": "object",
+          "properties": {
+            "filters": {
+              "type": "array",
+              "items": {
+                "anyOf": [
```
Alfred-Mountfield, line 16: I know this was here previously but I _think_ we actually want `oneOf`, an individual item can't be both (or at least shouldn't be)

apps/site/public/types/modules/graph/0.3/schema/aggregation-operation.json
```diff
@@ -1,87 +1,125 @@
...
+  "$defs": {
+    "Field": {
+      "type": "array",
       "items": {
-        "properties": {
-          "desc": {
-            "type": "boolean"
-          },
-          "field": {
+        "anyOf": [
```
Alfred-Mountfield, line 93: Similarly `oneOf` would be more correct here, `string & number` is unsatisfiable

apps/site/public/types/modules/graph/0.3/schema/aggregation-operation.json
```diff
@@ -1,87 +1,125 @@
...
+      "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/partial-graph-resolve-depths"
+    }
+  },
+  "required": [
+    "operation"
+  ],
+  "additionalProperties": false,
+  "$defs": {
+    "Field": {
+      "type": "array",
```
Alfred-Mountfield, line 91:
```suggestion
      "description": "the path to a JSON field, expressed as a list of path components where strings are object fields and numbers are array indices",
      "type": "array",
```

apps/site/public/types/modules/graph/0.3/schema/aggregation-operation.json
```diff
@@ -1,87 +1,125 @@
...
+                    "type": "object",
+                    "properties": {
+                      "field": {
+                        "$ref": "#/$defs/Field"
+                      },
+                      "operator": {
+                        "$ref": "#/$defs/FilterOperatorRequiringValue"
+                      },
+                      "value": {
+                      }
```
Alfred-Mountfield:
```suggestion
                      "value": {
                        "$comment": "any JSON value allowed"
                      }
```

apps/site/public/types/modules/graph/0.3/schema/aggregation-operation.json
TimDiekmann: I was not sure, if `oneOf` implies, that all elements in the array have to be the same. It appears it has no effect

apps/site/public/types/modules/graph/0.3/schema/aggregation-operation.json
```diff
@@ -1,87 +1,125 @@
...
     },
-    "multiSort": {
+    "graphResolveDepths": {
+      "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/partial-graph-resolve-depths"
+    }
+  },
+  "required": [
+    "operation"
+  ],
+  "additionalProperties": false,
```
Alfred-Mountfield, line 88: I think we can remove all of these. To my knowledge we aren't treating it as an error if people send more fields in the messages, we just ignore them.

libs/@blockprotocol/graph/src/graph-module.json
TimDiekmann: Good spot, I fixed it in [Fix aggregation operation](https://github.com/blockprotocol/blockprotocol/pull/1034/commits/61e6b7a5d4ea03cba90759ae2f0128d9598e4396)

libs/@blockprotocol/graph/src/graph-module.json
TimDiekmann: Should be fixed in [Fix aggregation operation](https://github.com/blockprotocol/blockprotocol/pull/1034/commits/61e6b7a5d4ea03cba90759ae2f0128d9598e4396)

apps/site/public/types/modules/graph/0.3/schema/aggregation-operation.json
Alfred-Mountfield: Neither was I, I had to google it before writing this comment 😁

libs/@blockprotocol/graph/src/graph-module.json
Alfred-Mountfield: Oh I see you had it in the actual aggregation-operation schema

apps/site/public/types/modules/graph/0.3/schema/aggregation-operation.json
TimDiekmann: Fixed in [Apply suggestions from Code Review](https://github.com/blockprotocol/blockprotocol/pull/1034/commits/2e5bd9dc964ed1713eb1ec846335a89862f8e4a5)

apps/site/public/types/modules/graph/0.3/schema/aggregation-operation.json
TimDiekmann: Fixed in [Apply suggestions from Code Review](https://github.com/blockprotocol/blockprotocol/pull/1034/commits/2e5bd9dc964ed1713eb1ec846335a89862f8e4a5)

libs/@blockprotocol/graph/src/graph-module.json
```diff
@@ -14,21 +14,22 @@
...
+            "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/versioned-url"
           },
           "properties": {
             "description": "The properties of the entity to create",
-            "type": "object"
+            "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/property-type-object"
           },
-          "links": {
+          "linkData": {
             "description": "Links to create along with the entity",
```
Alfred-Mountfield, line 24:
```suggestion
            "description": "Link data if the entity is a link entity",
```

apps/site/public/types/modules/graph/0.3/schema/aggregation-operation.json
TimDiekmann: Commit, which removes the majority of these: [Remove additionalProperties where not needed](https://github.com/blockprotocol/blockprotocol/pull/1034/commits/470035c00d8995d6754600e88f9477a12240034a)

libs/@blockprotocol/graph/src/graph-module.json
```diff
@@ -58,17 +52,33 @@
...
-            "type": "string"
+        "allOf": [
+          {
+            "properties": {
+              "entityId": {
+                "description": "The entityId of the entity to update",
+                "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/entity-id"
+              },
+              "entityTypeId": {
+                "description": "The entityTypeId of the type of the entity to update",
```
Alfred-Mountfield, line 62:
```suggestion
                "description": "The entityTypeId of the updated entity",
```

libs/@blockprotocol/graph/src/graph-module.json
```diff
@@ -58,17 +52,33 @@
...
+                "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/versioned-url"
+              },
+              "properties": {
+                "description": "The properties of the entity to update",
```
Alfred-Mountfield, line 66: Let's go back to this

```suggestion
                "description": "The new properties object to assign to the entity",
```

Alfred-Mountfield: It seems clearer to me that it's overwriting the properties object

libs/@blockprotocol/graph/src/graph-module.json
```diff
@@ -131,10 +129,14 @@
         "properties": {
```
Alfred-Mountfield, line 127: Too far from the diff but can we update the description of this message to be
"Request to retrieve a subgraph rooted at a specific entity"

apps/site/public/types/modules/graph/0.3/schema/subgraph.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://blockprotocol.org/types/modules/graph/0.3/schema/subgraph",
+  "title": "Subgraph",
```
Alfred-Mountfield, line 4: We should add a `$comment` saying this is a non-exhaustive type and users should consult the documentation and/or the `blockprotocol/graph` package

libs/@blockprotocol/graph/src/graph-module.json
```diff
@@ -163,14 +158,20 @@
       "respondedToBy": "aggregateEntitiesResponse",
       "sentOnInitialization": false,
       "data": {
+        "description": "The aggregation operation to apply",
```
Alfred-Mountfield, line 158:
```suggestion
        "description": "The results of an entity aggregation, and the operation that was used to produce it.",
```

libs/@blockprotocol/graph/src/graph-module.json
```diff
@@ -696,20 +204,20 @@
       "data": {
```
Alfred-Mountfield, line 204: We're missing a variant here:
```ts
     "data": {
        "oneOf": [
          {
            "type": "object",
            "properties": {
              "description": {
                "type": "string"
              },
              "name": {
                "type": "string"
              },
              "url": {
                "type": "string",
                "format": "uri",
                "$comment": "The URL to take the file from, if not being uploaded directly. 
              }
            },
            "required": [
              "url"
            ]
          },
          {
            "type": "object",
            "properties": {
              "description": {
                "type": "string"
              },
              "name": {
                "type": "string"
              },
              "file": {
                "$comment": "the file blob, if being uploaded directly."
              }
            },
            "required": [
              "file"
            ]
          }
        ]
      },
```

libs/@blockprotocol/graph/src/graph-module.json
```diff
@@ -39,14 +40,7 @@
...
-        "properties": {
-          "entity": {
-            "description": "The newly created entity",
-            "$ref": "https://blockprotocol.org/types/modules/graph/entity"
-          }
-        },
-        "required": ["entity"]
+        "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/entity"
       },
       "errorCodes": ["FORBIDDEN", "INVALID_INPUT"]
```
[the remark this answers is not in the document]
TimDiekmann: [Add "NOT_IMPLEMENTED" to error codes](https://github.com/blockprotocol/blockprotocol/pull/1034/commits/e4ca32064e40fc5c833caa45589f043238e83f6a)

libs/@blockprotocol/graph/src/graph-module.json
TimDiekmann: [Update description for getEntity](https://github.com/blockprotocol/blockprotocol/pull/1034/commits/d132aa2ecfa566c6c7d1bd854b6330aad06942d9)

apps/site/public/types/modules/graph/0.3/schema/subgraph.json
TimDiekmann: [Add comment on non-exhaustiveness for the subgraph schema](https://github.com/blockprotocol/blockprotocol/pull/1034/commits/a9db89eec4890e960c64291327a4de5a2662e471)

libs/@blockprotocol/graph/src/graph-module.json
TimDiekmann: [Add missing variant of uploadFile](https://github.com/blockprotocol/blockprotocol/pull/1034/commits/e601b5cad362ba2559d51f8b5ff144b9a186ae17)

Alfred-Mountfield (commented):
Thanks Tim, looks good to me, we'll wait for the target branch to merge first

Alfred-Mountfield (approved):

PR #1036, opened 2023-02-25
Fix integration tests

CiaranMn:

Update copy and flows to fix the integration tests.

I'm skipping two tests for now that are being a pain and need further time invested in them. [Task to restore them](https://app.asana.com/0/1203633819073575/1204059034603087)


apps/site/src/components/module-message-list.tsx
```diff
@@ -55,40 +55,43 @@ const ModuleMessageData: FunctionComponent<{
             </tr>
           </thead>
           <tbody>
-            {Object.entries(
-              data.properties ?? ({} as Record<string, JsonObject>),
```
CiaranMn, line 59: This assertion is unnecessary

Alfred-Mountfield (approved):

PR #1037, opened 2023-02-25
Docs/spec: update hook service, cover block verification, update metadata type

CiaranMn:

Bundling a few smaller fixes needed for 0.3

See the commits.

---
  - https://app.asana.com/0/0/1203701650452070
  - https://app.asana.com/0/0/1203946355023398
  - https://app.asana.com/0/0/1203978096566296

### DIFF
--- a/libs/@blockprotocol/core/src/types.ts
+++ b/libs/@blockprotocol/core/src/types.ts
@@ -40,6 +40,8 @@ export type BlockMetadataRepository =
     }
   | string;
 
+type VersionedUrl = `${string}v/${number}`;
+
 export type BlockMetadata = {
   /**
    * The name of the author of the block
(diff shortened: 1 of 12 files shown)


libs/@blockprotocol/core/src/types.ts
```diff
@@ -40,6 +40,8 @@ export type BlockMetadataRepository =
     }
   | string;
 
+type VersionedUrl = `${string}v/${number}`;
```
CiaranMn, line 43: `core` has very few dependencies and I would like to keep it that way (especially when we might have other services that don't need to bring in potential build complications from the type system package)

CiaranMn:
I am merging this because it introduces bug fixes which have caused CI regressions and I'm trying to fix them. 

Happy to take feedback in follow up – please comment on the diff and ping me if you have done so

PR #1038, opened 2023-02-25
Fix type caching, metaschema fetching, error code. Change dashboard link

CiaranMn:

Fixes and improvements:
- consistently reload user profile data when visiting user profile page
- in `/dashboard`, have 'create entity type' button go to user's profile types section (was hard to find otherwise)
- in `/@user/overview`, don't limit types and blocks to 4 a piece (can be confusing and appear that some are missing)
- fix rewrites to metaschema files
- fix generic `UNKNOWN_ERROR` code to `INTERNAL_ERROR` for consistency with spec
- consistently check URL length

### DIFF
--- a/.changeset/blockprotocol-version-bump.md
+++ b/.changeset/blockprotocol-version-bump.md
@@ -1,0 +1,5 @@
+---
+"blockprotocol": patch
+---
+
+bump `node-fetch` version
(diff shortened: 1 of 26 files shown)


Alfred-Mountfield (approved):
Cheers! Have tested locally and the path rewrite works, I also can't reproduce the unpleasant caching behaviour I was seeing before so I think the fix works.

PR #1039, opened 2023-02-25
Rename 'aggregate' and 'aggregation' to 'query'

Alfred-Mountfield:

The graph module has 2 main flavours of methods to access data, `get*` and `aggregate*` (e.g. `getEntity`, and `aggregateEntities`). I believe "aggregate" was chosen to try and imply an operation over many elements, however through usage it has proven less intuitive than hoped. It is possible to misinterpret it as an operation that behaves like a `reduce` function whereby the 'aggregate result' is a singular value like that produced by `sum`, `avg`, or `max`.

In this PR we rename "aggregate" to a simple "query", although this is less specific, and somewhat vaguely competes with "get" (`get`'s are a kind of query after all), we hope this will simplify some of the terminology and aid intuition.

In the future, when "linked aggregations" are reimplemented, we'll likely call them something like "linkedQueries" or "savedQueries" or some other variant.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1204038928975501/f) _(internal)_

## 🚫 Blocked by

N/A

I believe I have updated the relevant parts of the docs

## ⚠️ Known issues

- I've noticed the sandbox demo is falling out of date, we should probably look to update it or remove it

## 🐾 Next steps

- Publish new versions so blocks can be updated in time for the 0.3 launch

## 🛡 What tests cover this?

All standard tests across the repo, this was a wide-spanning change

## ❓ How to test this?

- Check CI
- Check the deployment still works
- Check mock-block-dock still works
  - Checkout this commit for a provided example: [Add mock-block-dock demonstration](https://github.com/blockprotocol/blockprotocol/pull/1039/commits/ec198618d9c63be412259bdaf88788e8d066e37d)


CiaranMn (approved):

nathggns:
Think I need to update shuffle block in response to this – it uses aggregate https://github.com/hashintel/hash/pull/2120 – will do so

PR #1040, opened 2023-02-25
Improve developing blocks guide

Alfred-Mountfield:

Adds some more guidance on how to call `getEntity` and `queryEntities` as well as a bit more information on the `graphResolveDepths` object.

### DIFF
--- a/apps/site/src/_pages/docs/1_developing-blocks.mdx
+++ b/apps/site/src/_pages/docs/1_developing-blocks.mdx
@@ -259,11 +259,95 @@ whether they are connected to the block entity at all, and far away they are (wh
 
 There are messages for exploring the data available in the embedding application:
 
-- `queryEntities` allows you submit a query which filters the entities in the data store.
+#### Retrieving the subgraph rooted at a specific entity
 
-You can browse the available entities in order to display them, or create links between them.
+If you know of a specific entity that you want to retrieve more detailed information on, you can call `getEntity` with its `entityId`.
+When doing this, you can provide how much of the graph you want to explore around the entity, by providing `graphResolveDepths`.
 
-If you know of a specific entity that you want to retrieve more detailed information on, you can call `getEntity`.
+For example, to retrieve the subgraph rooted at the entity, with its links and their destinations:
+
+```js
+const getEntityResponse = await graphModule.getEntity({
+  data: {
+    entityId: someEntityId,
+    graphResolveDepths: {
+      hasLeftEntity: {
+        incoming: 1,
+        outgoing: 0,
+      },
+      hasRightEntity: {
+        incoming: 0,
+        outgoing: 1,
+      }
+    }
+  }
+});
+
+if (getEntityResponse.errors) {
+  // handle errors
+}
+
+const { data: entitySubgraph } = getEntityResponse;
+```
+
+Here `graphResolveDepths` specifies two edge kinds,
+- `hasLeftEntity` - the edge that appears between a link entity and its left entity
+- `hasRightEntity` - the edge that appears between a link entity and its right entity
+
+For each edge kind, we specify the `incoming` and `outgoing` depths to resolve, where:
+- for `hasLeftEntity`
+  - `outgoing` means there is an outgoing edge _from_ the left entity _to_ the link entity, and defines the number of times to explore the left entity of link entities in a given branch of traversal.
+  - `incoming` means there is an incoming edge _from_ the link entity _to_ the left entity, which can be thought of as an "outgoing link entity" to the entity being traversed. The depth defines the number of times to explore outgoing link entities of entities in a given branch of traversal.
+- for `hasRightEntity`
+  - `outgoing` means there is an outgoing edge _from_ the right entity _to_ the link entity, and defines the number of times to explore the right entity of link entities in a given branch of traversal.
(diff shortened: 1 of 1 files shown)


apps/site/src/_pages/docs/1_developing-blocks.mdx
```diff
@@ -259,11 +259,140 @@ whether they are connected to the block entity at all, and far away they are (wh
...
+  - `outgoing` means there is an outgoing edge _from_ the right entity _to_ the link entity, and defines the number of times to explore the right entity of link entities in a given branch of traversal.
+  - `incoming` means there is an incoming edge _from_ the link entity _to_ the right entity, which can be thought of as an "incoming link entity" to the entity being traversed. The depth defines the number of times to explore incoming link entities of entities in a given branch of traversal.
+
+Therefore, the parameters in the example mean:
+
+- `(hasLeftEntity, incoming, 1)` - explore outgoing link entities to a depth of 1
+- `(hasLeftEntity, outgoing, 0)` - do not explore the left entity of link entities
+- `(hasRightEntity, incoming, 0)` - do not explore incoming link entities
+- `(hasRightEntity, outgoing, 1)` - explore the right entity of link entities to a depth of 1
+
```
CiaranMn, line 313:
````suggestion

In other words, these depth parameters mean the returned subgraph will contain the entity queried for, plus any outgoing link entities, plus the entities those link entities point to (formally, the entities which are "on the right" of the link entity).

To retrieve _incoming_ links as well as outgoing links, the values would be as follows:
```js
      ...
      hasLeftEntity: {
        incoming: 1,
        outgoing: 1,
      },
      hasRightEntity: {
        incoming: 1,
        outgoing: 1,
      },
      ...
```

````

File: apps/site/src/_pages/docs/1_developing-blocks.mdx [hunk 2]
```diff
@@ -259,11 +259,140 @@ whether they are connected to the block entity at all, and far away they are (wh
...
+if (getEntityResponse.errors) {
+  // handle errors
+}
+
+const { data: entitySubgraph } = getEntityResponse;
+```
+
+Here `graphResolveDepths` specifies two edge kinds,
+
+- `hasLeftEntity` - the edge that appears between a link entity and its left entity
```
CiaranMn, line 295:
```suggestion
- `hasLeftEntity` - the edge that appears between a link entity and its left entity (which can be thought of as the 'source' of the link)
```

apps/site/src/_pages/docs/1_developing-blocks.mdx
```diff
@@ -259,11 +259,140 @@ whether they are connected to the block entity at all, and far away they are (wh
...
+- `hasRightEntity` - the edge that appears between a link entity and its right entity
```
CiaranMn, line 296:
```suggestion
- `hasRightEntity` - the edge that appears between a link entity and its right entity (which can be thought of as the 'destination' of the link)
```

apps/site/src/_pages/docs/1_developing-blocks.mdx
```diff
@@ -259,11 +259,140 @@ whether they are connected to the block entity at all, and far away they are (wh
...
+
+For each edge kind, we specify the `incoming` and `outgoing` depths to resolve, where:
```
CiaranMn, line 298: I got worried that this gets pretty heavy for an explanation on 'how do I get stuff linked from my block entity', and is the first time that the word 'traversal' appears in this document, so I attempted to explain it.

But I may have made it worse😭 
```suggestion
The process of exploring a graph is known as 'traversal'. This involves visiting a vertex (an element in the graph – in our case, an entity), and then following edges from it (connections to other vertices) based on the parameters of the query. The parameters specify which kinds of edges to follow, and how many times they should be followed for any given journey through the graph. Each vertex visited might have multiple edges that can be followed from it , which means that each query actually involves many different journeys through the graph, which we refer to as 'branches' of the traversal.

For each edge kind, we specify the `incoming` and `outgoing` depths to resolve, where:
```

Alfred-Mountfield: It's a tough one, I was trying to avoid blowing it up too much too

apps/site/src/_pages/docs/1_developing-blocks.mdx
Alfred-Mountfield: > In other words, these depth parameters mean the returned subgraph will contain the entity queried for, plus any outgoing link entities, plus the entities those link entities point to (formally, the entities which are "on the right" of the link entity).

Happy to add this but I was trying to not be too verbose (and it's already very long). The start of the explanation says this:

> For example, to retrieve the subgraph rooted at the entity, with its links and their destinations:

And
> To retrieve _incoming_ links as well as outgoing links, the values would be as follows: 
> [...]

I opted not to write this because it was already too big and scary. I don't think we need to over-explain too much when we have this literally a line or two above

>- `(hasLeftEntity, outgoing, 0)` - do not explore the left entity of link entities
> - `(hasRightEntity, incoming, 0)` - do not explore incoming link entities

CiaranMn: Fair enough, leave it out

CiaranMn (approved):

PR #1045, opened 2023-02-26
Add "Payment History" section to the billing page

benwerner01:

This PR adds a "Payment History" section to the billing page.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1203781148500078/f) _(internal)_

## 🔍 What does this change?

- adds a `PaymentHistory` component

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

1. Checkout the branch and setup the local dev environment
2. Run the internal api locally on the `bw/payment-history` branch (if it hasn't merged yet)
3. Login and subscribe to a subscription tier
4. Observe the billing history section on the billing page

## 📹 Demo

image

---
  - https://app.asana.com/0/0/1204060610003539


apps/site/src/pages/settings/billing-settings-panel/payment-history-section.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+import { Link } from "../../../components/link";
+import { LinkButton } from "../../../components/link-button";
+import {
+  getDateFromStripeDate,
+  internalApi,
+} from "../../../lib/internal-api-client";
+import { priceToHumanReadable } from "../../shared/subscription-utils";
+
+const formatInvoiceDate = (date: Date) => {
```
luisbettencourt, line 29: Is there a reason not to use `dateToHumanReadable` with the `-` delimitator here?

luisbettencourt (changes requested):
Noticed that we are hiding the section when the payment history is empty, instead of showing this:
image
Was this left out intentionally?

Otherwise, tested it and is functionally working as expected. Some discrepancies in styling from the designs but that will come in a follow-up PR, right?

apps/site/src/pages/settings/billing-settings-panel/payment-history-section.tsx
benwerner01: Yes, because the order of the date items is reversed. But have unified the methods to make this re-usable in https://github.com/blockprotocol/blockprotocol/pull/1045/commits/5fe7a074ec1fce723bdc6cbb8a26faa089f32c89

benwerner01:
> Noticed that we are hiding the section when the payment history is empty, instead of showing this

Thanks for catching, have addressed in https://github.com/blockprotocol/blockprotocol/pull/1045/commits/5949837ddbb5ca52176df42cb0d99737bc01034d. You should now see:

image

> Some discrepancies in styling from the designs but that will come in a follow-up PR, right?

There are some discrepancies from the designs (no CSV format, a missing button, etc) but those should've been confirmed with David in figma commetns.

luisbettencourt (dismissed):
Thank you for addressing the changes @benwerner01

luisbettencourt (approved):

PR #1048, opened 2023-02-26
Bug fixes: Dock environment variables and default entity. Schema code generator. Use production Thing type

CiaranMn:

Two fixes in `mock-block-dock`:

1. Make the BP site host a configuration prop that can be passed to mock-block-dock (it previously relied on `process` being defined, which causing a crash) – I've made it consistent with how `blockProtocolApiKey` is passed through

2. Handle no initial block entity being provided. This was originally crashing because the `entityRecordIdOfEntityForBlock`  state was initialised with a missing entity. It now correctly defaults to the first entity in the initial data, if no `blockEntityRecordId` is provided. I've also updated this to always provide the initial mock entities, as there was previously no way of specifying a block entity and having the additional mock data, and blocks should be resilient to additional things being in the datastore anyway. We could add a prop to turn this off, or do something fancier to do with enabling initial data generation, but is a slight improvement.

Also:

3. Fix entity type meta schema in the TS-from-schema generator

4. Replace the staging `Thing` with production `Thing` in block templates / docs


nathggns:
This should fix a crash I'm currently experiencing. Will see if I can publish a canary version

nathggns:
Okay – wasn't able to publish a working canary version. Will have to come back to it when this is merged

Alfred-Mountfield (approved):

PR #1051, opened 2023-02-27
Update `blockTarballUrls` to production R2

kachkaev:

Follows #936 (see next steps)


apps/site/src/lib/blocks.ts
```diff
@@ -45,16 +45,6 @@ export type ExpandedBlockMetadata = BlockMetadata & {
   verified?: boolean;
 };
 
-// The contents of the JSON file users provide when adding a block via PR, stored in the Hub/ folder
-export interface StoredBlockInfo {
```
kachkaev, line 49: Bonus: removed unused type

CiaranMn (approved):

PR #1054, opened 2023-02-27
External service error handling

CiaranMn:

This is mainly to improve error handling in the site and packages related to the new Service Module:
1. Make the return of the site's `/api/external-service-method` route consistent with other routes by using the shared `formatErrors` function
2. Add a `FORBIDDEN` code to all service module messages – previously, there was no error code present, which meant no errors were permitted to be returned from callbacks
3. Catch errors in Mock Block Dock's implementation of service callbacks and send to the block as BP-structured errors

Drive-bys:
1. Delete 'blocks' tab (internal [Slack thread](https://hashintel.slack.com/archives/C02LG39FJAU/p1677022137504389?thread_ts=1677004305.523959&cid=C02LG39FJAU))
4. Handle URLs ending in a trailing slash when reporting block URL after publishing


apps/site/src/pages/api/external-service-method.api.ts
```diff
@@ -45,8 +45,22 @@ export default createApiKeyRequiredHandler<
     } catch (error) {
       const axiosError = error as AxiosError;
```
Alfred-Mountfield, line 46: Do we want to introduce and use this here

```ts
const isErrorAxiosError = (error: Error): error is AxiosError =>
  (error as AxiosError).isAxiosError;
```

apps/site/src/pages/api/external-service-method.api.ts
```diff
@@ -45,8 +45,22 @@ export default createApiKeyRequiredHandler<
...
+        formatErrors({
+          msg:
+            data &&
+            typeof data === "object" &&
+            "errors" in data &&
+            (data.errors as any[])?.[0].message
+              ? (data.errors as any[])?.[0].message
+              : "An unknown error occurred.",
+          code: status.toString(),
+        }),
```
Alfred-Mountfield, line 63: If this is guaranteed to be backend code, do we want to log the error using `util.inspect` here? Slightly concerned about us having absolutely no information if it doesn't match this weird type you're checking against here.

libs/@blockprotocol/service/src/types.ts
```diff
@@ -241,21 +241,23 @@ export type MapboxRetrieveStaticMapData = {
  */
 export type MapboxRetrieveStaticMapResponseData = string;
 
+export type ServiceMessageError = "FORBIDDEN";
```
Alfred-Mountfield, line 244: Do we not want an `"UNIMPLEMENTED"` message? Or `"INTERNAL_ERROR"` like we have in the graph module?

libs/mock-block-dock/src/service-module-callbacks.ts
```diff
@@ -23,20 +29,39 @@ const callExternalApiMethod = async (params: {
 
   if (!blockProtocolApiKey) {
     throw new Error(
-      `Visit ${bpSiteHost}/dashboard to generate a Block Protocol API key, which is required to make calls to the "${methodName}" method of the BP service module, and ensure it is passed to the \`MockBlockDock\` component as \`${blockProtocolApiKey}\`.`,
+      `Visit ${bpSiteHost}/dashboard to generate a Block Protocol API key, which is required to make calls to the "${methodName}" method of the BP service module, and ensure it is passed to the \`MockBlockDock\` component as blockProtocolApiKey.`,
```
Alfred-Mountfield, line 32: Not within the scope of this PR but it'd be really good to have this be visually shown to the user through the DOM rather than buried in the console somewhere like it was last time I tried this out

Alfred-Mountfield: Will this end up called on the Hub if a user doesn't have an API key? Or do we handle that differently because `mock-block-dock` passes control to the app?

Alfred-Mountfield (approved):
None of my comments were non-blocking, so I'm approving so you can merge if you view them as something that can be handled in follow ups

Alfred-Mountfield:
Apparently I shouldn't have approved, I didn't see auto-merge was enabled... well, we can address in follow-ups, no big deal

libs/@blockprotocol/service/src/types.ts
Alfred-Mountfield: I believe the `service-module.json` should be updated to have these too?

libs/mock-block-dock/src/service-module-callbacks.ts
CiaranMn: We will handle that differently because we will implement our own callbacks – this file will be unused

apps/site/src/pages/api/external-service-method.api.ts
CiaranMn: We could `Sentry.captureException` here (this weird type is Axios which I'd love to get rid of)

apps/site/src/pages/api/external-service-method.api.ts
CiaranMn: Yes

apps/site/src/pages/api/external-service-method.api.ts
Alfred-Mountfield: As long as it won't fail on circular references (which is why I suggested `util.inspect`), we've ran into a lot of issues with JSON.stringifying an axios error

libs/@blockprotocol/service/src/types.ts
CiaranMn: As a side note, if services do not add an error code, they don't even get the default ones – I briefly looked at changing that but it was going to mess around with `core` types so I decided not to.

Whoever picks up implementing these follow ups could look at that too, or discuss it with someone

PR #1061, opened 2023-02-27
Prepare for block versioning

kachkaev:

This PR introduces version-based block metadata and contents storage. It follows #988.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056186/1202539910143057/f) _(internal)_

## 🔍 What does this change?

- Files for newly uploaded blocks are stored on S3 under `block-contents/author/name/version` rather than `blocks/author/name`

- Each block's metadata can exist in the database multiple times

No

## ⚠️ Known issues

- DB queries are not optimized for high load (which is fine for now)

- There are no version tags, if a block with a smaller semver is published on top of a block with a larger semver, the new block won't be the `latest`. A similar problem would occur if someone publishes a new canary version on top of an existing one. E.g. `2.0.0-alpha.1` would be shown as ‘latest’ instead of `1.4.2`.

## 🐾 Next steps

- Immutability (can be implemented as a feature flag by restricting new uploads with the same `author / name / version` triple)

- UI to display block versions (currently showing semver-highest block versions everywhere for BC)

- Query optimization (we may benefit from it after we get 100s of unique block versions in the hub)

## 🛡 What tests cover this?

- CI

## ❓ How to test this?

1.  Try publishing multiple versions locally
2. Ensure no duplicates in the hub


CiaranMn:
Going to close this and port over the work into a new PR when we're ready to finish the feature – thanks @kachkaev for the foundation here!

PR #1069, opened 2023-02-28
Introduce hosted _Remote File_ entity type

Alfred-Mountfield:

This expresses the return type of the `uploadFile` graph module message as an entity type, which is now accessible at
https://blockprotocol.org/@blockprotocol/types/entity-type/remote-file/v/2

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1204013463058318/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Replaces the placeholder URLs with the new hosted one
- Renames some of the type and variable names to be consistent

- An update to the graph module json (and therefore the messages that are rendered in the spec) has been included

## ⚠️ Known issues

- We'd like to generate the properties and base URL definitions to keep them in sync with the remote. Unfortunately our code generator imports from `@blockprotocol/graph` which makes it hard to use _within_ the graph, especially as we want to generate a temporally agnostic `Entity`
  -  See [WIP - try and generate file type](https://github.com/blockprotocol/blockprotocol/pull/1069/commits/70e66d2dedc34e6ec7578c35f17696d92d1b1579) for the start of an attempt

## 🐾 Next steps

N/A

## 🛡 What tests cover this?

- Static code analysis

## ❓ How to test this?

Unknown if we currently have a block available that uses `uploadFile`


libs/@blockprotocol/graph/src/shared/types/file.ts
```diff
@@ -26,13 +26,17 @@ export const isFileData = (fileData: UploadFileData): fileData is FileData =>
   "file" in fileData;
 
 // @todo - auto-generate this from File type in blockprotocol.org when available
```
CiaranMn, line 28:
```suggestion
// @todo - auto-generate this from File type in blockprotocol.org
```

libs/@blockprotocol/graph/src/shared/types/file.ts
```diff
@@ -26,13 +26,17 @@ export const isFileData = (fileData: UploadFileData): fileData is FileData =>
...
   "https://blockprotocol.org/@blockprotocol/types/property-type/mime-type/": string;
-  "https://blockprotocol.org/@blockprotocol/types/property-type/url/": string;
+  "https://blockprotocol.org/@blockprotocol/types/property-type/file-url/": string;
 };
 
-export type FileEntity = Entity<true, FileEntityProperties>;
+export type RemoteFileEntity<Temporal extends boolean> = Entity<
+  Temporal,
+  RemoteFileEntityProperties
+>;
```
CiaranMn, line 39: Shouldn't this overwrite `entityTypeId` to have a `const` of its Versioned URL?

CiaranMn (approved):

PR #1072, opened 2023-02-28
Update WordPress plugin page for launch

nonparibus:

- Updates 'coming soon' copy to be... more now.

## 🔗 Related links

- [Block Protocol for WordPress plugin](https://wordpress.org/plugins/blockprotocol/)

## 🐾 Next steps

- If a user has already entered an email address before, we currently display the `https://wordpress.org/plugins/blockprotocol/` link in plaintext to them. This should be changed to an actual hyperlink.
- Tests around the email signup flow would be good.

---
  - https://app.asana.com/0/0/1204059903170098


nonparibus:
We'll update the video soon, separately. And expand copy to incorporate more.

TimDiekmann (approved):
Compared staging to production and looks good to me

PR #1078, opened 2023-02-28
Add Upgrade banner in account dropdown

luisbettencourt:

This PR adds the upgrade banner to the account dropdown (when a user is logged in).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204059903170056/f) _(internal)_

## 📹 Demo

image


nonparibus:
@luisbettencourt Please can we hide this whole upgrade thing behind the billing feature flag if we're not already? (Even if we flip the toggle on that now, this is functionality that should be grouped with the other stuff so we can test/disable it in future, if required)

nonparibus:
@luisbettencourt As a `Free` user I'm not actually seeing this. Using this dev deployment: https://blockprotocol-qhibjqua2.stage.hash.ai/dashboard

luisbettencourt:
> In addition to the above, @luisbettencourt, as a `Free` user I'm not actually seeing this in the navbar dropdown. Tested on this dev deployment: https://blockprotocol-qhibjqua2.stage.hash.ai/dashboard

I'm not being able to reproduce this, can you check on the newest deployment please?

nonparibus:
> I'm not being able to reproduce this, can you check on the newest deployment please?

Will do when the preview env deploys!

nonparibus:
@luisbettencourt Still not seeing this, even on latest build. Any idea what I'm doing wrong?

Capture 2023-03-01 at 05 22 32@2x

luisbettencourt:
Are you sure the user is free? 
image

nonparibus:
@luisbettencourt Yes. It says I am a free user on the billing page and I have never gone through the upgrade flow.

nonparibus:
@luisbettencourt Confirmed this works fine from another email address (non-HASH). Maybe it is that... will approve.

nonparibus (approved):

nonparibus:
@luisbettencourt So that we can get a minimal version of this in tonight, in follow-up please let's make sure...

- proper FE code review of this takes place (checking above assumption about root cause of why I couldn't see it)
- we make the whole purple banner bit clickable, while retaining the hover effect over the 'upgrade' button (currently the only clickable part)

luisbettencourt:
Figured it out, your account was flagged with `stripeSubscriptionTier: pro` but `stripeSubscriptionStatus: incomplete`. Must check against both, will open a PR follow-up shortly. @nonparibus

PR #1080, opened 2023-03-01
Catch errors in revalidating pages. Update `block-metadata.json` types

CiaranMn:

Small fixes:
1. We generate a JSON Schema for `block-metadata.json` which lives in [JSON Schema Store](https://github.com/SchemaStore/schemastore/blob/b4eccc2fb5ad76fd9c0a70fa67228e5d65e2b562/src/api/json/catalog.json#L658) and use to validate the file in editors. But it seems that the JSON Schema-from-TS generator turns a string template such as `VersionedUri` until an `array` (or something is), leading to an incorrect error. This makes it a `string` again so we don't have misleading editor hints

Screenshot 2023-03-01 at 03 29 11

2. We revalidate statically generated pages when publishing blocks to the hub (so that they are updated with the new/updated block), but have recently made `hub/` not statically generated. This resulted in an internal error being returned to users using the CLI publishing tool. This adds error handling, putting it in Sentry where we'll notice it if the same thing happens again, without showing an error to users.

3. Proper 404 if fetching a missing type as JSON


apps/site/src/middleware.page/return-types-as-json.ts
```diff
@@ -67,7 +67,7 @@ export const returnTypeAsJson = async (request: NextRequest) => {
 
   const kind = url.match(versionedTypeUrlRegExp)?.[1];
 
-  let type: DataType | PropertyType | EntityType =
+  let type: DataType | PropertyType | EntityType | null =
```
Alfred-Mountfield, line 70: Wouldn't this be `undefined` rather than `null`?

libs/@blockprotocol/core/src/types.ts
```diff
@@ -101,9 +99,9 @@ export type BlockMetadata = {
    */
   repository?: BlockMetadataRepository | null;
   /**
-   * The versioned URL to the block's schema
+   * The versioned URL to the block's schema, e.g. https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2
    */
-  schema: VersionedUrl;
+  schema: string;
```
Alfred-Mountfield, line 104: :(

Alfred-Mountfield (approved):

apps/site/src/middleware.page/return-types-as-json.ts
CiaranMn: It might possibly be `undefined` here yes, but it will be turned into `null` if it doesn't become a type below. Could be improved but not today

libs/@blockprotocol/core/src/types.ts
CiaranMn: I know :(

CiaranMn: We can patch the generated JSON schema to bring it back

PR #1085, opened 2023-03-01
Sort items in user profile and add icons for link entity types

thehabbos007:
This PR adds some small visual fixes
- Sorts types & blocks alphabetically
- Adds link icon for link entity types in overview tab

The types tab does not currently have any extra info about the type kind, so I didn't add anything there as I was unsure if we had designs for that

## 📹 Demo

image


apps/site/src/components/pages/user/overview-card.tsx
```diff
@@ -19,6 +19,29 @@ type OverviewCardProps = {
...
+    return (
+      <>
+        Entity Type
+        <LinkIcon
+          sx={{
+            marginLeft: "0.25rem",
+            width: "1rem",
+            height: "1rem",
+          }}
+        />
```
TimDiekmann, line 35: I wonder if we want to put the 🔗 to the left instead. (no strong opinion)
We could also (and/or?) use it as an image.

thehabbos007: (haven't pushed code yet)
image

Looks fine, thoughts?

TimDiekmann: This is better, thanks @thehabbos007! 
I agree with Ciaran that we should keep the image, so that change sufficient.

thehabbos007: 9852976a96f546f94bea4309f1bd38736a22ad39

CiaranMn (approved):

PR #1087, opened 2023-03-01
Add copyable URL to `Entity Type` editor header

thehabbos007:

This PR adds the entity type URL to the entity type editor and allows clicking to copy. This design seemed the simplest to make the type URL prominent for when developing blocks, but it could be missing the point. Regardless we can always move the copying logic else where on an icon button for example

Blocked by https://github.com/hashintel/hash/pull/2142 so we can bump the verions of the table

## 📹 Demo

https://user-images.githubusercontent.com/6988668/222495515-7d192dee-9f81-4821-b3cc-575861f3ec43.mp4

---
  - https://app.asana.com/0/0/1204096735845490


nonparibus (approved):
Nicely done. Design approved ✅

nonparibus:
@thehabbos007 For this on-screen URL, could we drop the font size slightly and use `JetBrains Mono` to be consistent with how we render full URLs in the docs and elsewhere? Thank you! We'll get further styling tweaks done in follow-up.

thehabbos007:
How does this look @nonparibus 
image

apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx
```diff
@@ -190,53 +207,85 @@ const EntityTypePage: NextPage = () => {
...
+                      <Typography variant="bpHeading4" color="inherit">
+                        (v{latestVersionNumber} available)
+                      </Typography>
+                    </Link>
+                  )}
+                </Stack>
+                <Box>
+                  {entityType?.schema.$id && (
+                    <Tooltip
+                      arrow
```
Comment (luisbettencourt) [inline apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx:266]:
```suggestion
```
@nonparibus Do we want the arrow here? Asking because we don't normally use it in other tooltips

image

image

apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx
```diff
@@ -190,53 +207,85 @@ const EntityTypePage: NextPage = () => {
...
+                        href="#"
+                        onClick={copyEntityTypeId}
+                        variant="bpCode"
+                        fontSize={10}
+                        sx={{
+                          color: ({ palette }) => palette.purple[500],
+                          "&:hover": {
+                            color: ({ palette }) => palette.purple[600],
+                          },
+                        }}
```
Comment (luisbettencourt) [inline apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx:282]: Nitpick:
```suggestion
                        sx={({ palette }) => ({
                          color:  palette.purple[500],
                          "&:hover": {
                            color: palette.purple[600],
                          },
                        })}
```

luisbettencourt (changes requested):
Thank you for this QOL improvement @thehabbos007!
Requested changes just to confirm the arrow is intentional, has my approval if so

luisbettencourt (approved):
Thank you @thehabbos007

PR #1092, opened 2023-03-01
Explicitly set 'verified' status on blocks when created

CiaranMn:

Make sure a `verified` status is set on block creation. Lists some starting trusted authors whose blocks are automatically verified. See https://blockprotocol.org/docs/faq#how-does-block-verification-work


luisbettencourt (approved):
Ran `yarn dev:db` and `yarn dev:seed-db` and the blocks were populated in the DB with the verified field set to true.

PR #1097, opened 2023-03-02
Prevent layout shift with docs sidebar + other UX improvements

nathggns:

We hide and show the docs sidebar on mobile to swap it for a mobile navigation. Currently, we do that using a `useMediaQuery` for "above medium". Because media queries are not available while server side rendering, this means the sidebar is hidden until hydration (and worse than that, as hydration must match server render, there's a necessary second pass when using useMediaQuery). 

I thought I could easily fix that by swapping the media query for a CSS one instead of a JavaScript one (i.e, rendering the full sidebar even on mobile, just hiding it with CSS). I have done that, but that had the consequence of rendering the sidebar contents on server when we didn't do that before. That was causing a hydration error as we need access to the URL hash to work out which link in the sidebar is "active", and that hash is not available on the server. I've fixed that by hiding the sidebar contents while server side rendering, swapping them for a skeleton instead.

I've also removed a few other uses of useMediaQuery and other non-ideal code which to remove layout shift on docs, and fixed some other UX issues while here:

- Removed the accidental spacing above the docs sidebar
- Corrected the spacing on the righthand side of the sidebar
- Removed the layout shift when opening/hiding the mobile menu (which had been caused by hiding breadcrumbs)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1203455535857863/f) _(internal)_
- [Asana task 2](https://app.asana.com/0/1203623179643287/1204086167143649/f)

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- [ ] Write this
- New `NavbarContainer` component with some functionality hoisted to that from `Navbar`. This is in order to have a component on which to apply some CSS variables used by both the Navbar and the Sidebar
- Made the code calculating breadcrumb height more reliable using a resize observer (it was just reading from the DOM node directly in render, which meant it could be incorrect). This height is used to position the page content, and this fixes a bug where page content would appear too high as the valuable was incorrect
- Now always render breadcrumbs, just hide them if not on mobile using a CSS media query, to fix a layout shift. We don't hide them when opening mobile nav either to prevent layout shift – instead, the mobile nav appears on top of them
- Height of the navbar now set using a CSS media query instead of JS, to fix a layout shift
- Desktop/mobile nav now swapped using CSS and not JS, to fix a layout shift
- Render a skeleton instead of the sidebar contents when server side rendering (for docs, not for settings, as settings is server-side safe)
- Opened pages is synced with the current route outside of an effect as it doesn't need to be an effect
- Recognised that sidebar links may be display hidden when calculating its parent for sidebar scroll position – fixing a potential crash
- Spacing around sidebar is now set by parent to fix the docs bug 

No

## ⚠️ Known issues

The breadcrumbs/auto selecting active link are broken – this is a pre-existing bug tracked by https://app.asana.com/0/1203179076056209/1204085034317822/f 

## 🐾 Next steps

Fix breadcrumbs

## 🛡 What tests cover this?

N/A

## ❓ How to test this?

Load the page on a slow network and ensure no layout shift

Ensure appears visually correct

Some not insignificant structure changes in this PR, so worth doing a visual check across the app.

## 📹 Demo

https://user-images.githubusercontent.com/719814/222470739-cb1e1cae-8bea-49df-9383-c5953107bc31.mov


luisbettencourt (approved):
After talking with @nonparibus about this we agreed to change the number of skeletons shown to resemble the menu when reaching docs from the navbar link. 
Also fixed a problem with the menu shrinking too much for smaller window sizes.

PR #1099, opened 2023-03-02
Create GH cache after merging a Pull Request

TimDiekmann:

GH Actions uses the cache from the same branch and falls back to the base branch. The base branch after creating a PR is the `main` branch, so if the CI is not running on a `push` to main it cannot be created

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643290/1204099711044327/f) _(internal)_

## 🔍 What does this change?

- Run CI on `push` to main

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,6 @@ env:
 jobs:
   linting:
     name: Linting
-    if: ${{ github.event_name == 'pull_request' }}
     runs-on: ubuntu-20.04
 
     steps:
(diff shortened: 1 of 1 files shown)


Alfred-Mountfield (approved):
Seems like a generally positive thing to do. I do wonder if this would increase our cache usage and potentially cause us to go over (more frequently?)

TimDiekmann:
It should use less cache usage as a new cache won't be created if a cache hit happens, so in theory, a Pull Request keeps using the cache from `main`

PR #1100, opened 2023-03-02
Run CI on any Pull Request (or merge queue event)

TimDiekmann:

The CI currently runs only on pull request targeting `0.3` and `main` while the former does not exist anymore.
When stacking PR, it's currently not possible to check linting as the CI does not run at all.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643290/1204099711044327/f) _(internal)_

## 🔍 What does this change?

- Enable CI runs for all pull request
- Drive-by: Add `merge_group` as target, **this should enable us to use the merge queue**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main, 0.3]
+  merge_group:
 
 env:
   TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}


.github/workflows/ci.yml
```diff
@@ -3,7 +3,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main, 0.3]
+  merge_group:
```
Alfred-Mountfield, line 6: Should this not be nested?

Also does this mean CI is _**required**_? Because I think a few of the jobs don't work if not targeting main

TimDiekmann: > Should this not be nested?

No, `merge_group` is completely separated from `pull_request`

> Also does this mean CI is _**required**_? Because I think a few of the jobs don't work if not targeting main

No. the `required` status will only be added when targeting a protected branch. If you, however, targeting a protected branch without running the CI you're not able to merge that PR at all.

Alfred-Mountfield (approved):

PR #1101, opened 2023-03-02
Download `wasm-pack` instead of compiling it

TimDiekmann:

Compiling `wasm-pack` is really slow. Instead of compiling it, this will change the behavior of `cargo-make` to download it instead as officially recommended

### DIFF
--- a/libs/@blockprotocol/type-system/crate/Makefile.toml
+++ b/libs/@blockprotocol/type-system/crate/Makefile.toml
@@ -174,8 +174,11 @@ private = true
 install_crate = { crate_name = "cargo-nextest", version = "0.9.28", binary = "cargo", test_arg = ["nextest", "--version"] }
 
 [tasks.install-wasm-pack]
-private = true
-install_crate = { crate_name = "wasm-pack", version = "0.10.3", binary = "wasm-pack", test_arg = ["--version"] }
+script = """
+if [ ! -f ~/.cargo/bin/wasm-pack ]; then
+  curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh -s -- -f
+fi
+"""
 
 [tasks.install-wasm-opt]
 private = true
--- a/libs/@blockprotocol/type-system/crate/Makefile.toml
+++ b/libs/@blockprotocol/type-system/crate/Makefile.toml
@@ -70,7 +70,7 @@ args = ["build", "--profile", "${CARGO_MAKE_CARGO_PROFILE}", "@@split(CARGO_BUIL
 extend = "node"
 cwd = "${CARGO_MAKE_WORKSPACE_WORKING_DIRECTORY}"
 args = ["--loader", "ts-node/esm", "./../scripts/build-wasm.ts"]
-dependencies = ["install-wasm-pack", "install-wasm-opt"]
+dependencies = ["install-wasm-opt"]
 
 ################################################################################
 ## Lints                                                                      ##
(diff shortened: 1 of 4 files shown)


libs/@blockprotocol/type-system/crate/Makefile.toml
```diff
@@ -174,8 +174,11 @@ private = true
 install_crate = { crate_name = "cargo-nextest", version = "0.9.28", binary = "cargo", test_arg = ["nextest", "--version"] }
 
 [tasks.install-wasm-pack]
-private = true
-install_crate = { crate_name = "wasm-pack", version = "0.10.3", binary = "wasm-pack", test_arg = ["--version"] }
+script = """
```
Alfred-Mountfield, line 177: Is there a reason you made this not `private`?

Also I appreciate this install approach is recommended but unfortunately a very large amount of the WASM-Pack docs are actually severely out of date. There _were_ large problems with installing wasm-pack like this on MacOS (maybe only M1), I _think_ this is the associated issue https://github.com/rustwasm/wasm-pack/issues/952 but there actually seems to be some progress made on it.

I'm not sure I have capacity to test it right now, so I think we have two options
- Make sure that installing through this method works fine on M1, and that the README recommends you install it this way (ideally just through a command using `cargo make`
- Make sure that this is limited to CI, jobs you can run from the README don't accidentally call this, and make sure this job is very difficult to accidentally call

TimDiekmann: I tested this locally in a few variations and everything work as expected.

Removing the `private = true` was an oversight, but I also don't see a reason to not have it public

Alfred-Mountfield: Do we know how users should update wasm-pack after installing it like this?

TimDiekmann: They can update it by just running this command

Alfred-Mountfield: Does this command _check_ for updates, or does it just reinstall every time?

Wondering if `cargo make build` is going to install every time for example (as this is a dependency to the task)

TimDiekmann: My comment was badly phrased, sorry. This will only install if `wasm-pack` was not installed, this is what the check does). The user would need to remove the `wasm-pack` before. Writing this reveals that this is not optimal, I can change this to check for a specific version and install it if the version does not match (That's mainly the behavior of `cargo-make`).

Alfred-Mountfield: If that's not too hard to do, I'd appreciate that

TimDiekmann: Should be doable fairly quick 🚀

TimDiekmann: Solved by using `wasm-pack` as dev-dependency

libs/@blockprotocol/type-system/package.json
```diff
@@ -69,6 +69,7 @@
     "rollup": "3.5.1",
     "ts-jest": "29.0.5",
     "tslib": "2.4.1",
-    "typescript": "4.9.4"
+    "typescript": "4.9.4",
+    "wasm-pack": "0.10.3"
```
Alfred-Mountfield, line 73: Have you removed your local setup and tested this quite a bit?

I remember this was one of the first approaches we tried and it had similar problems on M1 and such. (I also just really don't like the idea of loading a rust binary through `yarn` but _if_ it works I'm fine with it)

Also who maintains this distribution? Who are we dependent on updates from

TimDiekmann: Yes, everything is tested locally. `which wasm-pack` points to `.node_modules`.

The package is maintained by https://github.com/rustwasm

Alfred-Mountfield: Hm, let's try it out and see if there are issues then :)

Alfred-Mountfield (dismissed):

PR #1104, opened 2023-03-02
Use `wasm-opt` from `wasm-pack`

TimDiekmann:

There is no point in compiling a huge binary to reduce the size of the WASM file by 300 kb, if it's only used locally. This significantly slows down CI times. This PR adjusts the `build` command to not optimize the WASM file. The optimization is still run by the `prepublishOnly` job.

### DIFF
--- a/libs/@blockprotocol/type-system/package.json
+++ b/libs/@blockprotocol/type-system/package.json
@@ -45,7 +45,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "yarn clean && yarn build:wasm && yarn optimize && yarn build:bundle",
+    "build": "yarn clean && yarn build:wasm && yarn build:bundle",
     "build:bundle": "rollup -c --bundleConfigAsCjs",
     "build:wasm": "cd crate && cargo make build",
     "clean": "rimraf ./dist/",
--- a/libs/@blockprotocol/type-system/package.json
+++ b/libs/@blockprotocol/type-system/package.json
@@ -53,8 +53,7 @@
     "fix:eslint": "eslint --fix .",
     "lint:eslint": "eslint --report-unused-disable-directives .",
     "lint:tsc": "tsc --noEmit",
-    "optimize": "(cd crate && cargo make install-wasm-opt) && wasm-opt -Os wasm/type-system_bg.wasm -o tmp.wasm && mv tmp.wasm wasm/type-system_bg.wasm",
-    "prepublishOnly": "yarn build:wasm && yarn test && yarn optimize && yarn build:bundle",
+    "prepublishOnly": "yarn build:wasm && yarn test && yarn build:bundle",
     "test": "jest"
   },
   "devDependencies": {
--- a/libs/@blockprotocol/type-system/package.json
+++ b/libs/@blockprotocol/type-system/package.json
@@ -53,7 +53,7 @@
     "fix:eslint": "eslint --fix .",
     "lint:eslint": "eslint --report-unused-disable-directives .",
     "lint:tsc": "tsc --noEmit",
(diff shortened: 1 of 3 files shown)


libs/@blockprotocol/type-system/package.json
```diff
@@ -45,15 +45,15 @@
...
+    "build": "yarn clean && yarn build:wasm && yarn build:bundle",
     "build:bundle": "rollup -c --bundleConfigAsCjs",
     "build:wasm": "cd crate && cargo make build",
     "clean": "rimraf ./dist/",
     "compressed-size": "yarn build && find dist -iname '*.js' -exec npx terser@latest --compress --mangle --output {} -- {} \\;",
     "fix:eslint": "eslint --fix .",
     "lint:eslint": "eslint --report-unused-disable-directives .",
     "lint:tsc": "tsc --noEmit",
-    "optimize": "wasm-opt -Os wasm/type-system_bg.wasm -o tmp.wasm && mv tmp.wasm wasm/type-system_bg.wasm",
+    "optimize": "(cd crate && cargo make install-wasm-opt) && wasm-opt -Os wasm/type-system_bg.wasm -o tmp.wasm && mv tmp.wasm wasm/type-system_bg.wasm",
```
Alfred-Mountfield, line 56: How is this called by prepublish?

TimDiekmann: I haven't found a location, where this job is called at all 🙃 

However, I fixed it by calling it from `wasm-pack`, so we don't need to install it at all.

thehabbos007 (approved):

libs/@blockprotocol/type-system/package.json
Alfred-Mountfield: Ah, sorry to disappoint but we actually install our own version of `wasm-opt` because the one distributed with `wasm-pack` is missing some pretty substantial optimisation routes. I'll see if I can dig up the relevant issues

TimDiekmann: It results in the exact same binary size, so I doubt we used a different version of `wasm-opt`

Alfred-Mountfield: I'm looking into this, it might have been `binaryen` or something else. But this was definitely behaviour for a long time, and actually we were heavily considering moving away from `wasm-pack` to just `wasm-bindgen` and `wasm-opt` ourselves.

The crate was pretty unmaintained for a long time (and very out of date docs). I've just checked the repo and it _seems_ to have had more activity recently so it's possible we no longer need to do this. 

I appreciate the number might seem the most important thing here, but the fact that they are same right now does not mean they will continue to be, so I'd at least like to find out what I am remembering first, and then we can merge this if we're happy it's addressed or mitigated :)

Alfred-Mountfield: Found a few comments confirming this, but I thought I stumbled across a resource previously that went more in-depth than this:

> Though wasm-pack seems to use a slightly older version by default, using a [newer version](https://github.com/WebAssembly/binaryen) gives slightly smaller results for me. If you have it in the PATH, wasm-pack will use that version instead.

- https://github.com/firasdib/Regex101/issues/1208#issuecomment-1320750056

> You're not using the latest wasm-opt version (wasm-pack seems rather slow to update this stuff). Though it only saves an extra 4K for me. But I also already get to 423 KB without it, not sure what causes the difference, we both seem to be on the latest nightly compiler and wasm-pack 🤔

- https://github.com/firasdib/Regex101/issues/1208#issuecomment-1321018200

Alfred-Mountfield: Considering the recent activity, I think this is probably worthwhile for the CI savings, thanks!

Alfred-Mountfield (approved):

PR #1105, opened 2023-03-02
Use shared cache for Rust jobs

TimDiekmann:

The Rust jobs is the same for all kind of jobs, we can share the cache

### DIFF
--- a/.github/actions/warm-up-repo/action.yml
+++ b/.github/actions/warm-up-repo/action.yml
@@ -26,6 +26,7 @@ runs:
       uses: Swatinem/rust-cache@v2
       with:
         workspaces: libs/@blockprotocol/type-system/crate
+        shared-key: "shared"
 
     - run: yarn install
       shell: bash


.github/actions/warm-up-repo/action.yml
```diff
@@ -26,6 +26,7 @@ runs:
       uses: Swatinem/rust-cache@v2
       with:
         workspaces: libs/@blockprotocol/type-system/crate
+        shared-key: "shared"
```
Alfred-Mountfield, line 29: Do you know the implication of this if we are making changes to the Rust setup (dependencies or something else) in multiple PRs?

What would happen if we added a new crate?

TimDiekmann: This only sets up for the workspace above, so this would not affect any other crates at all. We need to revisit caching when adding a new crate in any case.

Alfred-Mountfield: And
> Do you know the implication of this if we are making changes to the Rust setup (dependencies or something else) in multiple PRs

TimDiekmann: The cache cannot be shared across multiple PRs, it can only re-use the cache from the base branch.

Alfred-Mountfield: Ah I understand, this is shared between _jobs_ on the same PR (sorry I know you wrote that in the PR description I just didn't understand).

Alfred-Mountfield (approved):

PR #1106, opened 2023-03-02
Fix block page revalidation

CiaranMn:

Certain pages in the site are statically built by NextJS, and are supposed to be rebuilt:

1. On a timer (the value for `revalidate` returned from `getStaticPaths`)
2. On a trigger, e.g. when a block is published or updated

The revalidation trigger for the individual block page had the wrong path, which meant the block page wasn't being rebuilt. Because the timer was set to 30 minutes, this meant that any publishing changes were only reflected when the next 30 minute timer began again.

This PR fixes the path for the block page and lowers the 'at most' timer to 3 minutes in case we manually edit things in the db or something else that wouldn't trigger the revalidation.

I had recently added exception capturing in Sentry precisely to warn us about paths being incorrect – I'm not sure why this hasn't shown up.

---
  - https://app.asana.com/0/0/1204086165607412


luisbettencourt (approved):
Is there a good way to test this?

PR #1111, opened 2023-03-03
Abort on `panic!`

TimDiekmann:

We don't `panic!` in our code. We only have a few location where we `.expect()` that a (previously validated) URL is a valid URL.

This reduces the size of the WASM binary by 74 KB

### DIFF
--- a/libs/@blockprotocol/type-system/crate/Cargo.toml
+++ b/libs/@blockprotocol/type-system/crate/Cargo.toml
@@ -32,6 +32,7 @@ wasm-bindgen-test = "0.3.13"
 [profile.release]
 lto = true
 opt-level = 's'
+panic = 'abort'
 
 [package.metadata.wasm-pack.profile.release]
 wasm-opt = false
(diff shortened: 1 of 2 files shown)


Alfred-Mountfield (commented):
Going to leave this PR in review until we make greater size gains for now I think. In the context of the current binary this is relatively small. 

If we do manage to shrink the size further (currently limited by URLs and Regex support) then this will actually be quite significant and, as discussed offline, this is probably worth the trade-off as we ideally won't have any panics within this code (contingent on external dependency behaviour)

Alfred-Mountfield:
> In light of the huge relative size reduction in https://github.com/blockprotocol/blockprotocol/pull/1113 does this PR https://github.com/blockprotocol/blockprotocol/pull/1111 now merit consideration as being sufficiently material?

@nonparibus Tim and I had already discussed this offline, we were waiting for the upstream branches to merge

Alfred-Mountfield (approved):
We've had a scan through the code and there shouldn't be risks of panicking

PR #1113, opened 2023-03-03
Remove `regex` crate and use `rsplit("v/")` instead

TimDiekmann:

We currently use a regex to parse a versioned URL from a string, but just searching for `"v/"` is sufficient. This reduces the binary size of WASM by ~638 KB.


libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
```diff
@@ -115,38 +114,18 @@ impl FromStr for VersionedUrl {
     type Err = ParseVersionedUrlError;
 
     fn from_str(url: &str) -> Result<Self, ParseVersionedUrlError> {
-        static RE: LazyLock<Regex> =
-            LazyLock::new(|| Regex::new(r#"(.+/)v/(\d+)(.*)"#).expect("regex failed to compile"));
-
-        if url.len() > 2048 {
```
Alfred-Mountfield, line 121: This should not have been removed, perhaps should happen before the split

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
```diff
@@ -115,38 +114,18 @@ impl FromStr for VersionedUrl {
...
-        }
-
-        Ok(Self {
-            base_url: BaseUrl::new(base_url.to_owned())
-                .map_err(ParseVersionedUrlError::InvalidBaseUrl)?,
-            version: u32::from_str(version)
-                .map_err(|error| ParseVersionedUrlError::InvalidVersion(error.to_string()))?,
-        })
+        url.rsplit_once("v/").map_or(
+            Err(ParseVersionedUrlError::MissingVersion),
```
Alfred-Mountfield, line 118: I think this should be an `IncorrectFormatting` error

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
```diff
@@ -115,38 +114,18 @@ impl FromStr for VersionedUrl {
...
+            |(base_url, version)| {
+                Ok(Self {
+                    base_url: BaseUrl::new(base_url.to_owned())
+                        .map_err(ParseVersionedUrlError::InvalidBaseUrl)?,
+                    version: version.parse::<u32>().map_err(|error| {
```
Alfred-Mountfield, line 123: If this is an empty string I think we should return `MissingVersion`

libs/@blockprotocol/type-system/crate/src/ontology/data_type/mod.rs
```diff
@@ -206,7 +206,9 @@ mod tests {
                   "type": "string"
                 }
             ),
-            ParseDataTypeError::InvalidVersionedUrl(ParseVersionedUrlError::AdditionalEndContent),
+            ParseDataTypeError::InvalidVersionedUrl(ParseVersionedUrlError::InvalidVersion(
+                "invalid digit found in string".to_owned(),
```
Alfred-Mountfield, line 210: We should change this to something _like_ "version suffix wasn't a valid non-negative integer", ideally we'd also return what the invalid component was

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
Alfred-Mountfield: (see below)

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
Alfred-Mountfield: Also ideally if there is a number, and then more stuff (for example query params) we would warn about `AdditionalEndContent` still

Alfred-Mountfield (changes requested):
Tests have not been updated

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
Alfred-Mountfield: Oh we do this in `BaseUrl` as well.. perhaps this is fine

libs/@blockprotocol/type-system/crate/src/ontology/data_type/mod.rs
TimDiekmann: We cannot return the exact component but at least the string, which was failing to parse

Alfred-Mountfield: That's what I meant :) The component after `v/`

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
TimDiekmann: Supporting `AdditionalEndContent` is possible but hard as we would search for any non-digit character on our own. Having additional content would result in an error like
> invalid version `12-a-b-c`: invalid digit found in string

Alfred-Mountfield: > is possible but hard as we would search for any non-digit character on our own. Having additional content would result in an error like

It shouldn't be too hard right? **Once the version parse fails** Can't we just iterate over the characters after the `v/` separator and return the slice after the first non-digit?

TimDiekmann: If you pass in `/v/123,456` (accidentally used `,` as thousand separators), you would get `,456` as additional content. Unless you have an overflow error in parsing it would always result in `AdditionalEndContent`. I don't see a reason why we need to catch this, this is basically the same as the `InvalidDigit` error kind returned from parsing.

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
TimDiekmann: I don't have a strong opinion on this, we can still have it in place as the length is different from the one in `BaseUrl`

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
TimDiekmann: Should be fixed

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
Alfred-Mountfield: I personally think this is actually quite helpful, it identifies `,456` as the problematic bit, and especially if there's a query param (which I think is far more likely than a comma) you will see something like ``"Additional end content `?q=abstract%20api`"``

TimDiekmann: > If this is an empty string I think we should return MissingVersion

Should be fixed

libs/@blockprotocol/type-system/crate/src/ontology/data_type/mod.rs
TimDiekmann: Should be fixed

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
Alfred-Mountfield: Let's add it because it's technically what the spec suggests

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
```diff
@@ -115,38 +114,31 @@ impl FromStr for VersionedUrl {
...
+        url.rsplit_once("v/").map_or(
+            Err(ParseVersionedUrlError::IncorrectFormatting),
+            |(base_url, version)| {
+                Ok(Self {
+                    base_url: BaseUrl::new(base_url.to_owned())
+                        .map_err(ParseVersionedUrlError::InvalidBaseUrl)?,
+                    version: version.parse::<u32>().map_err(|error| match error.kind() {
+                        IntErrorKind::Empty => ParseVersionedUrlError::MissingVersion,
+                        IntErrorKind::InvalidDigit => {
+                            let i = version.find(|c: char| !c.is_numeric()).unwrap_or(0);
```
Alfred-Mountfield, line 126: If the index is `0` I think we might want to go with InvalidVersion` again 😅

Alfred-Mountfield: Because that would imply there weren't any numerics right? So it's not just additional end content after a valid one

Alfred-Mountfield (changes requested):
The typescript tests still haven't been updated (I recognise I wasn't clear last time, apologies)

I'd be pretty surprised if they still pass without needing updating

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
TimDiekmann: Should be fixed now 🚀

Alfred-Mountfield: Nice, thank you!

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
Alfred-Mountfield: Can we re-add it?

libs/@blockprotocol/type-system/src/native/url.ts
```diff
@@ -56,12 +56,6 @@ const versionedUrlRegExp = /(.+\/)v\/(\d+)(.*)/;
 export const validateVersionedUrl = (
   url: string,
 ): Result<VersionedUrl, ParseVersionedUrlError> => {
-  if (url.length > 2048) {
```
Alfred-Mountfield, line 59: Can we re-add this?

libs/@blockprotocol/type-system/test/url.test.ts
```diff
@@ -84,9 +84,9 @@ const invalidVersionedUrlCases: [string, ParseVersionedUrlError][] = [
   ],
   ["http://example.com", { reason: "IncorrectFormatting" }],
   ["http://example.com/v/", { reason: "IncorrectFormatting" }],
-  ["http://example.com/v/0.2", { reason: "AdditionalEndContent" }],
+  ["http://example.com/v/0.2", { reason: "AdditionalEndContent", inner: ".2" }],
   ["http://example.com/v//20", { reason: "IncorrectFormatting" }],
-  ["http://example.com/v/30/1", { reason: "AdditionalEndContent" }],
+  ["http://example.com/v/30/1", { reason: "AdditionalEndContent", inner: "/1" }],
   ["http://example.com/v/foo", { reason: "IncorrectFormatting" }],
```
Alfred-Mountfield, line 90: Can we add a test which checks `InvalidVersion`?

Also wondering if we want to add a test which checks the `TooLong` 😅

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
TimDiekmann: Done

libs/@blockprotocol/type-system/crate/src/ontology/url/mod.rs
```diff
@@ -114,6 +114,10 @@ impl FromStr for VersionedUrl {
     type Err = ParseVersionedUrlError;
 
     fn from_str(url: &str) -> Result<Self, ParseVersionedUrlError> {
+        if url.len() > 2048 {
```
Alfred-Mountfield, line 117: Can we add this to `BaseUrl`? As it was before

TimDiekmann: It's still in `BaseUrl::validate_str`

Alfred-Mountfield: Ah apologies!

PR #1115, opened 2023-03-03
Cancel previous jobs when newer commits were added

TimDiekmann:

We don't need to run CI checks for commits which are outdated.

## 🔗 Related links

- [Slack thread](https://hashintel.slack.com/archives/C02TWBTT3ED/p1677854234827139) _(internal)_


Alfred-Mountfield (approved):
🚀

PR #1116, opened 2023-03-03
Update dependencies in `block-scripts`

Alfred-Mountfield:

Some new dependencies were released a while ago, these include support for the new(ish) `satisfies` operator in TypeScript.


.changeset/quiet-suns-eat.md
```diff
@@ -0,0 +1,5 @@
+---
```
Alfred-Mountfield, line 1: @CiaranMn, I've added a changeset, do I need to do a follow-up PR after release to change local dependency versions on `block-scripts` to the updated version? Or does the changeset release thing handle that

CiaranMn: changesets does that

CiaranMn (dismissed):

CiaranMn (approved):

PR #1123, opened 2023-03-05
Add Block Protocol WordPress plugin

CiaranMn:
## 🌟 What is the purpose of this PR?

Adds the Block Protocol WordPress plugin. 

## 🔗 Related links

- [Plugin homepage](https://blockprotocol.org/wordpress)
- [WordPress plugin directory listing](https://wordpress.org/plugins/blockprotocol/)

- [Asana task](https://app.asana.com/0/1203358502199087/1203994362939599/f) _(internal)_

## 🔍 What does this change?

- Adds the files for the plugin. I've mirrored the structure of its [Subversion repository](https://plugins.trac.wordpress.org/browser/blockprotocol) to make it easier to sync changes on new release, but we can alter this if we want.
- The plugin is NOT added as a Yarn workspace. There are some type incompatibilities that need resolving.

## ⚠️ Known issues

- Because this is not a Yarn workspace, linting and TS checks are not yet automated (but are passing). Formatting is covered by the root commands.

## 🐾 Next steps

- Make this a Yarn workspace.
- GitHub action to handle releasing new versions of the plugin.

## ❓ How to test this?

1.  Follow the instructions in the plugin's readme to get set up – they should be self-explanatory (and need amending if they are not)

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204086165607416

PR #1126, opened 2023-03-06
Add unsupported database version notice to WordPress plugin

thehabbos007:

This PR adds a notice if the database used for WordPress is unsupported in the admin settings - we also totally disable the migrations and admin-part of the post viewing to disable any issues. If the admin can't even make blocks, the user-facing post won't run into errors.
We use a generic message to prevent any abuse in exposing the exact DB version through the notice.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643290/1204120050952648/f) _(internal)_

## 🚫 Blocked by

- #1123 

## 🔍 What does this change?

- Adds a new check for the DB version using our documented versions
- Adds early bails for admin init and migrations
- Adds `admin_notice` for when the DB version is unsupported

## ❓ How to test this?

Change docker compose "db" service to 
```yaml
  db:
    platform: linux/amd64
    image: mariadb:10.2.6
    ports:
      - "3306:3306"
    volumes:
      - db_data:/var/lib/mysql
    restart: always
    environment:
      MYSQL_ROOT_PASSWORD: somewordpress
      MYSQL_DATABASE: wordpress
      MYSQL_USER: wordpress
      MYSQL_PASSWORD: wordpress
```
(Yes that MariaDB version uses MYSQL for the env vars)
Start the plugin dev env as usual using the readme instructions

## 📹 Demo

Using MariaDB 10.2.6, should be the same for MySQL < 8
image

### DIFF
--- a/libs/wordpress-plugin/plugin/trunk/server/block-db-table.php
+++ b/libs/wordpress-plugin/plugin/trunk/server/block-db-table.php
@@ -63,6 +63,11 @@ function block_protocol_migrate()
 {
   $saved_version = (int) get_site_option('block_protocol_db_migration_version');
 
+  // Don't apply migrations if the DB version is unsupported.
+  if (!block_protocol_is_database_supported()) {
+    return;
+  }
+
   if ($saved_version < 2) {
     block_protocol_migration_1();
     update_site_option('block_protocol_db_migration_version', 2);
--- a/libs/wordpress-plugin/plugin/trunk/server/block-db-table.php
+++ b/libs/wordpress-plugin/plugin/trunk/server/block-db-table.php
@@ -1,7 +1,7 @@
 <?php
 
-const MINIMUM_MYSQL_VERSION = "8.0.0";
-const MINIMUM_MARIADB_VERSION = "10.2.7";
+const BLOCK_PROTOCOL_MINIMUM_MYSQL_VERSION = "8.0.0";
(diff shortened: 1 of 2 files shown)


libs/wordpress-plugin/plugin/trunk/server/block-db-table.php
```diff
@@ -1,5 +1,24 @@
 <?php
 
+const MINIMUM_MYSQL_VERSION = "8.0.0";
+const MINIMUM_MARIADB_VERSION = "10.2.7";
```
CiaranMn, line 4: Global scope 😁 
```suggestion
const BLOCK_PROTOCOL_MINIMUM_MYSQL_VERSION = "8.0.0";
const BLOCK_PROTOCOL_MINIMUM_MARIADB_VERSION = "10.2.7";
```

CiaranMn (changes requested):
I was unable to test this because it seems this container cannot be emulated as an AMD64 on my machine (M2 chip)

> wordpress-db-1         | runtime: failed to create new OS thread (have 2 already; errno=22)

I don't think there's an easy fix for this so I will trust that it works – one change requested on variable scoping

libs/wordpress-plugin/plugin/trunk/server/block-db-table.php
thehabbos007: Addressed in 32165b7a2670623af0c61f7f03c6a6d3778a111e Thank you!

CiaranMn (approved):

PR #1127, opened 2023-03-06
Introduce `Feature Showcase` block

Alfred-Mountfield:

This PR introduces the first functional self-contained block within the BlockProtocol repository (caveats are to exclude the templates, and the blocks shipped with MBD).

This block is meant to serve as a foundation of providing examples for _all_ functionalities _in one_. While it doesn't currently meet that purpose, it is the hopeful end-goal. This should prove as a good demonstration of how to interact with the various Block Protocol modules, and how to handle generated types within block code. It is a good playground for establishing block best-practices, and will also be a good candidate for automated, and manual, testing flows.

This block currently utilises two types, `Person`, and `Company`, which have been created in a new `@examples` namespace. Although they were made alongside this block, their long term intention is to provide a _**very**_ simple set of example types that can be used for tutorials and guides in the future.

This introduces a novel pattern for trying to keep local state up to date when the datastore is modified. Unfortunately it is not a good replacement for the introduction of actual data subscriptions within the Block Protocol spec, if the datastore is mutated _outside_ of the block then the input fields will not refresh. But for now it does improve the experience a fair bit.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203767287750341/1203488186089072/f) _(internal)_

## 🚫 Blocked by

- [x] #1223 

## 🔍 What does this change?

Please see commits and ask questions if anything is unclear, I have attempted to keep the history clean so that it can be referred to in the future for certain processes (such as adapting a new block from the templates to work within the monorepo).

- When/if we look to publishing this block (and not just using internally in the repo for testing) we should expand the READMEs and accompanying documentation to discuss how to orchestrate its publishing.

## ⚠️ Known issues

- I was unable to make this block `"type": "module"` as it breaks the `block-scripts` logic around the `example-graph.ts` file as when compiled it creates a `.js` file which is then wrongly imported as an ESM script, rather than a CommonJS one.

## 🐾 Next steps

- Utilise the new types from `@examples` within MockBlockDock to replace the `example.com` ones that are currently present (this will allow us to generate types as well)
- Address the ESM issue outlined above
- Look towards block publishing should we wish
- Utilise this block within some automated tests (Playwright for example?)
- Add this to the local DB seeding logic when running the site and hub
- Expand this block as we expand the spec (for example with type mutation methods such as `createEntityType`)

## 🛡 What tests cover this?

- `eslint`
- `tsc`
- Manual tests

## ❓ How to test this?

- Run the block `yarn workspace @blocks/aio-demo dev` and test it out and see how it performs
- Run the block and test it in MockBlockDock.


CiaranMn (approved):

nonparibus (approved):

thehabbos007 (approved):
Thank you, I tested locally and it is working as expected!

PR #1137, opened 2023-03-07
Return full HASH error when error is thrown in internal API

benwerner01:

This PR updates the custom error handling of the `/external-service-method` endpoint to return the full HASH error when an error is returned by the internal API. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204117110538083/f) _(internal)_

## ⚠️ Known issues

- Because the wordpress plugin currently depends on a `msg` field being returned in in an `errors` array, this field has been left untouched by this PR for backwards compatibility purposes. We may therefore want to update how the wordpress plugin reads error messages (see [this follow-up task](https://app.asana.com/0/1202805690238892/1204117110538079/f))

## 🛡 What tests cover this?

Manual testing

## ❓ How to test this?

1. Run the developer environment of the BP site and the internal API locally
2. Call an external service method using `/external-service-method` and cause an error to be thrown by the external service. For example, by providing an empty string as the `searchText` for a `suggestAddress` method call

## 📹 Demo

image


CiaranMn (approved):

apps/site/src/pages/api/blocks/index.api.ts
```diff
@@ -23,12 +24,24 @@ export type ApiBlockSearchResponse = {
...
+    if (!errors.isEmpty()) {
+      return res.status(400).json(
+        formatErrors(
+          ...errors.array().map((validationError) => ({
+            message: validationError.msg,
+            ...validationError,
+          })),
+        ),
+      );
+    }
```
benwerner01, line 43: @CiaranMn have added the change we discussed in standup, but noticed that we aren't throwing any of the validation errors in the `/api/blocks` method when they occur. Have added the necessary handler here, and marked the query params as optional (as to my understanding they all should be). Here's an example of what a validation error would look like if the `json` parameter isn't JSON:
image

CiaranMn (approved):

CiaranMn (approved):

PR #1094, opened 2023-03-01
Handle service module blocks in the hub (part 1)

CiaranMn:
## 🌟 What is the purpose of this PR?

Some blocks now call services that incur usage costs (in the form of BP credits or real money if the user is on a paid account and has exhausted their credits). This normally requires an API key, which isn't present in the Block Hub. Those blocks are therefore non-functional for the time being.

This PR shows a modal if a block tries to call an external service informing the user of the fact. 

A follow-up PR tomorrow will offer logged-in users the chance to use their credits (or money) on testing the block.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643287/1204086167143590) _(internal)_

## 🐾 Next steps

- Part 2 will offer logged-in users the chance to allow this block to make calls, so they can test it properly.

## ❓ How to test this?

1.  In the preview deployment, go to a service-calling block's page (e.g. AI Text, AI Image, Address)
2. Use it
3. See modal

## 📹 Demo

Screenshot 2023-03-01 at 19 54 54

PR #1138, opened 2023-03-07
Handle service module blocks in the hub (part 2)

CiaranMn:

#1094 introduced a modal where users would be told that a block was trying to make a call to an external service.

This PR introduces the option for logged-in users to allow the request to happen. The user's choice is saved on a per-block basis for the session (i.e. closing the tab will forget it).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203845052955973/f) _(internal)_

## 🔍 What does this change?

- introduce an `allowCookieFallback` option to the API key middleware, which will let the request proceed if no API key is provided but there is a cookie-authenticated user.
- Update `sandboxed-block-demo-api.ts` (the code for the iFrame) to sent an id with each service message it sends to the parent, and wait for a response – the block is given a promise which is resolved when that response is received.
- Update `sandboxed-block-demo.tsx` and `service-module-modal.tsx` (the parent window code) to offer logged-in users the choice to actually process the request from the block, with the appropriate response sent back to the iFrame
- Reorganises a few files so that they respect the fractal tree folder structure we are moving towards, whereby files which are private to another file are stored in a subfolder named after that file

## ⚠️ Known issues

- In testing this I noticed an issue whereby if you edit a block's properties (e.g. `Code`) inside the block, editing it from the JSON editor will not take effect. This appears to be an issue with `mock-block-dock` which I will investigate separately.

## Next steps

It would be nice to inform the user how many credits they had left in the modal, to inform their decision as to whether to use them on testing the block or not.

## ❓ How to test this?

1.  Try the`AI Text` block in the preview deployment

## 📹 Demo

Screenshot 2023-03-07 at 20 00 25


apps/site/src/components/pages/hub/block-data-container/sandboxed-block-demo.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+const generateBlockServicePermissionKey = (blockPath: string) =>
+  `allow-${blockPath}-service-usage`;
+
+const setServiceUsagePermission = (blockPath: string, allowed: boolean) => {
+  sessionStorage.setItem(
+    generateBlockServicePermissionKey(blockPath),
+    allowed.toString(),
+  );
+};
```
benwerner01, line 29: This doesn't block merging, but I wonder if we just want to save this in the page's state instead. If I'm browsing through the blocks the hub, make the decision to not allow requests from a block, continue browsing the hub in the same tab and then later change my mind, it's not obvious how I go about changing my preference.

At that point it's also not obvious that requests are failing, and that that's why the block isn't functioning as intended.

Perhaps we could either store the preference in the pages state so that it isn't persisted across redirects in the same tab, or we can have some kind of message fade-in under the block reminding the user that the external service message requests are being blocked because of a decision they made previously, ideally with something like an "Undo" button?

benwerner01 (approved):
Thanks Ciaran, this looks good to me and is working as expected. Had one thought about the current UX, but shouldn't block merging 👍

apps/site/src/components/pages/hub/block-data-container/sandboxed-block-demo.tsx
CiaranMn: It's a fair point – I've created a [follow-up task](https://app.asana.com/0/1203623179643290/1204138539312008/f) to do this tomorrow

PR #1147, opened 2023-03-09
Add example of Hook Module usage in custom element block

CiaranMn:

Adds an example of using the Hook Module in a custom element block.

---
  - https://app.asana.com/0/0/1204164203672395

### DIFF
--- a/libs/mock-block-dock/dev/test-custom-element-block.ts
+++ b/libs/mock-block-dock/dev/test-custom-element-block.ts
@@ -25,7 +25,7 @@ export class TestCustomElementBlock extends BlockElementBase {
 
     const paragraph = this.renderRoot.querySelector(`#${paragraphId}`);
     if (!paragraph || !(paragraph instanceof HTMLParagraphElement)) {
-      throw new Error("No paragraph for hook service found in element DOM");
+      throw new Error("No paragraph for hook module found in element DOM");
     }
 
     void this.hookModule.hook({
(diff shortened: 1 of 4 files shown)


libs/mock-block-dock/dev/test-custom-element-block.ts
```diff
@@ -1,12 +1,44 @@
 import { extractBaseUrl } from "@blockprotocol/graph";
 import { BlockElementBase } from "@blockprotocol/graph/custom-element";
 import { getRoots } from "@blockprotocol/graph/stdlib";
+import { HookBlockHandler } from "@blockprotocol/hook";
 // eslint-disable-next-line import/no-extraneous-dependencies
```
Alfred-Mountfield, line 5: Do you know the reason for this import by the way? Can we add a reason to it
```suggestion
// eslint-disable-next-line import/no-extraneous-dependencies -- ?
```

Alfred-Mountfield (approved):
Question that could be addressed as a driveby, but unrelated to diff. ✅

libs/mock-block-dock/dev/test-custom-element-block.ts
CiaranMn: The reason for the ignore is that eslint was complaining about `lit` being imported here when it is listed in `devDependencies`, because eslint did not know that this folder is not part of the distributed library. I have updated the linting config in https://github.com/blockprotocol/blockprotocol/pull/1147/commits/499304b6e311d7d8005313ab0dfdd162f766be26 to exclude development files from the rule.

Alfred-Mountfield: Thank you ✨

Alfred-Mountfield (approved):

PR #1150, opened 2023-03-09
Hide entity editor for certain blocks in the WordPress plugin

CiaranMn:

We show a form in the WordPress plugin settings sidebar which is generated from the block's schema.

For some blocks, this form is potentially confusing and may lead to the block breaking if properties are edited in unexpected ways when it is in certain states.

The best solution to this is for blocks to be able to specify which properties should be editable externally, so that the plugin needs no knowledge of any particular block, but in the short term this hides the editor for blocks where it may be a problem.


luisbettencourt (approved):
Working locally for me 
image

PR #1153, opened 2023-03-10
Add Graph stdlib helper functions for Ontology Type references

thehabbos007:

This PR adds various helper methods to get ontology types referenced by an ontology type. I've added:
- `getEntityTypesReferencedByEntityType` (edges `InheritsFrom`, `ConstrainsLinksOn` and `ConstrainsLinkDestinationsOn` )
- `getPropertyTypesReferencedByPropertyType` (`edge ConstrainsPropertiesOn`)
- `getDataTypesReferencedByPropertyType` (edge `ConstrainsValuesOn`)

I've separated the edge union to make the 'helper helper' function have more accurate typing: `StrictOntologyOutwardEdge`

These new functions have been exported and use the same "shared" helper to simplify their implementation

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643290/1204128625385540/f) _(internal)_

## 🔍 What does this change?

- See start of description and commits.

## 🐾 Next steps

- Implementing traversals for these types [Asana task](https://app.asana.com/0/0/1204128625385541/f)

### DIFF
--- a/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/entity-type.ts
+++ b/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/entity-type.ts
@@ -1,52 +1,12 @@
-import {
-  BaseUrl,
-  extractBaseUrl,
-  extractVersion,
-  VersionedUrl,
-} from "@blockprotocol/type-system/slim";
+import { VersionedUrl } from "@blockprotocol/type-system/slim";
 
 import {
   isConstrainsPropertiesOnEdge,
   isInheritsFromEdge,
-  OntologyOutwardEdge,
-  OntologyTypeRevisionId,
   OntologyTypeVertexId,
-  OutwardEdge,
-  StrictOntologyOutwardEdge,
   Subgraph,
 } from "../../../types/subgraph.js";
-
-// Helper higher-order function to simplify edge extraction
-const getXTypesReferencedByYType = (
-  subgraph: Subgraph<boolean>,
-  entityTypeId: OntologyTypeVertexId | VersionedUrl,
-  outwardEdgePredicate: (
-    outwardEdge: OutwardEdge<boolean>,
-  ) => outwardEdge is StrictOntologyOutwardEdge,
-): OntologyTypeVertexId[] => {
-  let baseUrl: BaseUrl;
-  let revisionId: OntologyTypeRevisionId;
-
-  if (typeof entityTypeId === "string") {
-    baseUrl = extractBaseUrl(entityTypeId);
-    revisionId = extractVersion(entityTypeId).toString();
-  } else {
-    baseUrl = entityTypeId.baseId;
-    revisionId = entityTypeId.revisionId;
-  }
-
-  const outwardEdges = subgraph.edges[baseUrl]?.[
-    revisionId
-  ] as OntologyOutwardEdge<boolean>[];
-
-  if (outwardEdges === undefined) {
-    return [];
-  }
-
-  return outwardEdges
-    .filter(outwardEdgePredicate)
-    .map((outwardEdge) => outwardEdge.rightEndpoint);
-};
+import { getXTypesReferencedByYType } from "./shared.js";
 
 /**
  * Gets identifiers for all `PropertyType`s referenced within a given `EntityType` schema by searching for
--- a/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/entity-type.ts
+++ b/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/entity-type.ts
@@ -31,13 +31,13 @@ export const getPropertyTypesReferencedByEntityType = (
 
 /**
  * Gets identifiers for all `EntityType`s referenced within a given `EntityType` schema by searching for
- * "InheritsFrom" `Edge`s from the respective `Vertex` within a `Subgraph`.
+ * "InheritsFrom", "ConstrainsLinksOn" and "ConstrainsLinkDestinationsOn" `Edge`s from the respective `Vertex` within a `Subgraph`.
  *
  * @param subgraph {Subgraph} - The `Subgraph` containing the type tree of the `EntityType`
  * @param entityTypeId {OntologyTypeVertexId | VersionedUrl} - The identifier of the `EntityType` to search for
- * @returns {OntologyTypeVertexId[]} - The identifiers of the `EntityType`s inherited from the `EntityType`
+ * @returns {OntologyTypeVertexId[]} - The identifiers of the `EntityType`s referenced from the `EntityType`
  */
-export const getEntityTypesInheritedFromEntityType = (
+export const getEntityTypesReferencedByEntityType = (
   subgraph: Subgraph<boolean>,
   entityTypeId: OntologyTypeVertexId | VersionedUrl,
 ): OntologyTypeVertexId[] =>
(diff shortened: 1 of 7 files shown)


libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/entity-type.ts
```diff
@@ -24,27 +22,30 @@ import {
...
+  subgraph: Subgraph<boolean>,
+  entityTypeId: OntologyTypeVertexId | VersionedUrl,
+): OntologyTypeVertexId[] =>
+  getOntologyTypesReferencedBy(
+    subgraph,
+    entityTypeId,
+    (outwardEdge): outwardEdge is StrictOntologyOutwardEdge =>
+      isInheritsFromEdge(outwardEdge) ||
+      isConstrainsLinksOnEdge(outwardEdge) ||
+      isConstrainsLinkDestinationsOnEdge(outwardEdge),
```
thehabbos007, line 50: Do we want these to be separated?

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/shared.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * This is a helper function to extract ontology edges given an `outwardEdgePredicate` function.
+ *
+ * @internal
+ * @param subgraph {Subgraph} - The `Subgraph` containing the underlying ontology types
+ * @param ontologyTypeId {OntologyTypeVertexId | VersionedUrl} - The identifier of the `EntityType` to search for
+ * @param outwardEdgePredicate {(outwardEdge: OutwardEdge<boolean>) => outwardEdge is StrictOntologyOutwardEdge} - The predicate to filter edges by
+ * @returns {OntologyTypeVertexId[]} - The resulting endpoints of the filtered edges
+ */
+export const getOntologyTypesReferencedBy = (
```
Alfred-Mountfield, line 26: The name of this is probably fine, I just keep wondering if we can name it to be something like `getRightEndpointsForEdgeKind` or something.. but while still maintaining that this is about '`StrictOntologyOutwardEdge`'s

libs/@blockprotocol/graph/src/shared/types/subgraph/edges/variants/ontology.ts
```diff
@@ -180,7 +180,7 @@ export const isIsTypeOfEdge = <Temporal extends boolean>(
   return outwardEdge.kind === "IS_OF_TYPE" && outwardEdge.reversed;
 };
 
-export type OntologyOutwardEdge<Temporal extends boolean> =
+export type StrictOntologyOutwardEdge =
```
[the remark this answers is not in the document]
Alfred-Mountfield: I'd maybe call this something along the lines of `ToOntologyOutwardEdge`

I was going to suggest `OntologyToOntologyOutwardEdge` (and `KnowledgeGraphToOntologyOutwardEdge` or something for the other one, but really these outward edges don't constrain the left element. 

But then again that is the point of this file... 🙃 It is implying that these are intended to be used _from_ ontology elements, even if the actual `OutwardEdge` types don't constrain that.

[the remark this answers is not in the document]
Alfred-Mountfield: Maybe `OntologyToOntologyOutwardEdge` is fine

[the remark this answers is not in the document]
thehabbos007: f7dcf36483c5e640c058237dbf8afc39fdaa0406

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/shared.ts
```diff
@@ -0,0 +1,55 @@
...
+  subgraph: Subgraph<boolean>,
+  ontologyTypeId: OntologyTypeVertexId | VersionedUrl,
+  outwardEdgePredicate: (
+    outwardEdge: OutwardEdge<boolean>,
+  ) => outwardEdge is StrictOntologyOutwardEdge,
```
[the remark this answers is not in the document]
thehabbos007: image

It gives a type error because the endpoint is incompatible
(`StrictOntologyOutwardEdge` has been renamed to `OntologyToOntologyOutwardEdge`)

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/shared.ts
thehabbos007: It starts to get quite long if we write it out in full `getRightEndpointsForOntologyToOntologyOutwardEdgeKind` :sweat_smile:

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/entity-type.ts
Alfred-Mountfield: Agreed, but talking about the return type of this function generally (not the predicate), if the caller only wants `ConstrainsLinksOn` edges, they'd need to refilter the output of this if we don't return separate collections.

(Splitting the _implementation_ up into 3 separate functions would end up with a traversal for each predicate as you say though so we probably shouldn't do that).

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/shared.ts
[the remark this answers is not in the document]
Alfred-Mountfield: Nice :)

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/shared.ts
Alfred-Mountfield: Why is naming things so hard 😭

thehabbos007: What do you think about `getRightOntologyEndpointsForOntologyOutwardEdge`?

Alfred-Mountfield: Or `getOntologyEndpointsForOntologyOutwardEdge` (as it's an outward edge)

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/entity-type.ts
Alfred-Mountfield: Should we split the return value?

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/shared.ts
thehabbos007: 06074e224df3f88dbebcadde3d84b17649b61b07

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/entity-type.ts
thehabbos007: Split them up in 09ccca724d78f531c83ffbf51bc6a647657a484e. The names/descriptions may need adjusting though

Alfred-Mountfield: But now to get all of the references we will require 3 traversals:
> We're not doing individual traversals for each predicate I guess

That's what I meant by this:
> (Splitting the implementation up into 3 separate functions would end up with a traversal for each predicate as you say though so we probably shouldn't do that).

I was suggesting we should split the _output_ like we have for the non-subgraph parallel implementation.

Alternatively we could have a method per edge kind, **and** a method to get all of the ontology ones (which is _not_ implemented as just calling all the methods per edge kind as that'd be triple the traversals)

thehabbos007: I completely misunderstood this, my bad. I see now that you explicitly said to split up the return values

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/entity-type.ts
```diff
@@ -24,7 +28,31 @@ import {
...
+    subgraph,
+    entityTypeId,
+    isConstrainsPropertiesOnEdge,
+  );
+
+type EntityTypeReferences = {
+  inheritsFrom: OntologyTypeVertexId[];
+  constrainsLinksOn: OntologyTypeVertexId[];
+  constrainsLinkDestinationsOn: OntologyTypeVertexId[];
+};
```
thehabbos007, line 42: Should we export this? I don't see any strong reasons to

Alfred-Mountfield (approved):

PR #1156, opened 2023-03-10
Add helper function for fully recomputing edges in Subgraph

thehabbos007:

This PR changes how we update entities to take into account that edges may need to recompute. It does so by removing all edges and recompute them through the `infer` functions.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643290/1204145497706392/f) _(internal)_

## 🔍 What does this change?

- Added new `inferSubgraphEdges` helper to recompute all edges of the graph given the vertices
- Delete edges on update
- Recompute the edges on update through `inferSubgraphEdges`

## ❓ How to test this?

I suggest the following test setup:
- checkout `asa/temp-updateentity-mbd` at commit `5c8668d98716c0431939df090ddb323c7a3dedbd`
- Build `@blockprotocol/graph` and `mock-block-dock`
- `yarn dev` from the block at `blocks/aio-demo` 

Operations on the subgraph through the dev block should result in the right edges being present.

## 📹 Demo

This demo may be hard to follow without a description.
What happens here is I've seeded entity types into the demo and added 1 person, 1 org and a link between them.
I show the state of the edges where the `Person` has `IS_OF_TYPE` edges pointing to the `Person` entity type in the beginning. I am going to change the `Person` to have the `Thing` entity type instead. When this update happens, the edges are fully recomputed. The, previously, `Person`  entity is now a `Thing` with its edges adjusted.

https://user-images.githubusercontent.com/6988668/224391588-09598f52-2e9f-4ac7-9d6a-0899f76fe562.mp4

### DIFF
--- a/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
+++ b/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
@@ -26,7 +26,7 @@ import {
   Subgraph,
   SubgraphTemporalAxes,
 } from "../../types.js";
-import { typedEntries, typedKeys } from "../../util.js";
+import { typedEntries } from "../../util.js";
 import { getVertexIdForRecordId } from "./vertex-id-for-element";
 
 /**
--- a/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
+++ b/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
@@ -184,35 +184,12 @@ export const inferSubgraphEdges = <Temporal extends boolean>(
       },
       [baseId, revisionObject],
     ) => {
-      const revisionIds = typedKeys(revisionObject);
-      let firstEntry = null;
-      if (revisionIds[0] && revisionIds[0] in revisionObject) {
-        firstEntry = revisionObject[revisionIds[0]];
-      }
-
-      const newVertexIds = revisionIds.map((revisionId) => ({
-        baseId,
-        revisionId,
-      }));
-
-      switch (firstEntry?.kind) {
-        case "dataType":
-          acc.dataTypeVertexIds.push(...newVertexIds);
-          break;
-        case "propertyType":
-          acc.propertyTypeVertexIds.push(...newVertexIds);
-          break;
-        case "entityType":
-          acc.entityTypeVertexIds.push(...newVertexIds);
-          break;
-        case "entity":
-          acc.entityVertexIds.push(...newVertexIds);
-          break;
-        default:
-          throw new Error(
-            `Internal implementation error, unexpected vertex: ${firstEntry}`,
-          );
-      }
+      typedEntries(revisionObject).forEach(([revisionId, vertex]) => {
+        acc[`${vertex.kind}VertexIds`].push({
+          baseId,
+          revisionId,
+        });
+      });
 
       return acc;
     },
--- a/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
+++ b/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
@@ -182,12 +182,12 @@ export const inferSubgraphEdges = <Temporal extends boolean>(
         entityTypeVertexIds: OntologyTypeVertexId[];
         entityVertexIds: EntityVertexId[];
       },
-      [baseId, entry],
+      [baseId, revisionObject],
     ) => {
-      const revisionIds = typedKeys(entry);
+      const revisionIds = typedKeys(revisionObject);
       let firstEntry = null;
-      if (revisionIds[0] && revisionIds[0] in entry) {
-        firstEntry = entry[revisionIds[0]];
+      if (revisionIds[0] && revisionIds[0] in revisionObject) {
+        firstEntry = revisionObject[revisionIds[0]];
       }
 
       const newVertexIds = revisionIds.map((revisionId) => ({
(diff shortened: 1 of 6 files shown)


libs/mock-block-dock/src/datastore/use-mock-datastore/temporal.ts
```diff
@@ -351,11 +354,12 @@ export const useMockDatastore = (
             /**
              * @todo this update logic may be incorrect, we have to do some
              * testing to ensure it behaves as expected */
```
Alfred-Mountfield, line 356: Is this TODO still required?

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
```diff
@@ -151,3 +154,84 @@ export const buildSubgraph = <Temporal extends boolean>(
...
+  // Construct object with vertex ids for each vertex kind
+  const vertexIds = typedEntries(subgraph.vertices).reduce(
+    (
+      acc: {
+        dataTypeVertexIds: OntologyTypeVertexId[];
+        propertyTypeVertexIds: OntologyTypeVertexId[];
+        entityTypeVertexIds: OntologyTypeVertexId[];
+        entityVertexIds: EntityVertexId[];
+      },
+      [baseId, entry],
```
Alfred-Mountfield, line 185: I find `entry` pretty confusing. Maybe we could call it `mappedRevisions`, `revisionObject`, `revisionVertices`, or something else

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
```diff
@@ -151,3 +154,84 @@ export const buildSubgraph = <Temporal extends boolean>(
...
+    ) => {
+      const revisionIds = typedKeys(entry);
+      let firstEntry = null;
+      if (revisionIds[0] && revisionIds[0] in entry) {
+        firstEntry = entry[revisionIds[0]];
+      }
```
Alfred-Mountfield, line 191: I genuinely found this quite hard to read and it took me a few goes to understand it :(

Maybe something like this?
```suggestion
      const sampleVertex = typedValues(entry).pop();

      const revisionIds = typedKeys(entry);
```

or if you're concerned about `Object.values` allocating a new array
```suggestion
      const revisionIds = typedKeys(entry);
      
      const sampleVertexRevisionId = revisionIds[0];
      const sampleVertex = sampleVertexRevisionId ? entry[sampleVertexRevisionId]! : null;
      }
```

Unless that second suggestion doesn't work? Bit confused about why you're having to do the `in` check, you should be able to assert it's non-null (considering you just checked it) but I didn't think you'd have to assert it's actually a valid index argument

Alfred-Mountfield: Alternatively we just rewrite this to be over `typedEntries` in the first place, and map the `newVertexIds` from that

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
```diff
@@ -151,3 +154,84 @@ export const buildSubgraph = <Temporal extends boolean>(
...
+
+      const newVertexIds = revisionIds.map((revisionId) => ({
+        baseId,
+        revisionId,
+      }));
+
+      switch (firstEntry?.kind) {
+        case "dataType":
```
Alfred-Mountfield, line 199: Can't you use the type-guards for the vertices? `isDataTypeVertex`, etc.

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
```diff
@@ -151,3 +154,84 @@ export const buildSubgraph = <Temporal extends boolean>(
...
+          break;
+        case "entityType":
+          acc.entityTypeVertexIds.push(...newVertexIds);
+          break;
+        case "entity":
+          acc.entityVertexIds.push(...newVertexIds);
+          break;
+        default:
+          throw new Error(
+            `Internal implementation error, unexpected vertex: ${firstEntry}`,
```
Alfred-Mountfield, line 213: Probably worth `JSON.stringify`ing it if you're putting it in the error message

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
thehabbos007: I agree, `revisionObject` seems fitting here, thanks! fc5a9aa632769336dce86700eac46429da4affe6

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
thehabbos007: I commited this because it was so succinct 72616b0d6735ff341c47a4a14c4c5f8b3c19ff42
If you dislike it too much we can change it, but the template string actually fully type checks on the `kinds` we define and errors if anything is wrong for the types we've set. What do you think? :laughing:

Alfred-Mountfield: I thought I'd dislike it a lot more than I did, but upon looking at the plain file without the diff, I think it's fairly reasonable.

If it's properly type checked, I'm good with it :)

Alfred-Mountfield (approved):

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
Alfred-Mountfield: (Now redundant)

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/builder.ts
Alfred-Mountfield: (Now redundant)

Alfred-Mountfield (approved):

PR #1157, opened 2023-03-10
Fix block message init retry and custom element rendering

CiaranMn:

Fixes a few issues I noticed while building a custom element block:

1. The `CoreBlockHandler` includes a retry on its `init` message if it doesn't receive an immediate response, but this can cause an infinite loop with rendering custom elements (not while running them in dev mode, but once built). This PR fixes that by moving the retry from a `queueMicrotask` to a `setTimeout`. I chose not to bring a dependency to make it `setImmediate`, which would be a bit quicker middleground, on the basis that for most use cases it isn't vital that init messages are exchanged in the fastest possible time, and avoiding `setImmediate` saves adding another 1KB to every block.

2. The logic for renaming custom element tags was broken – I haven't actually encountered any duplication of custom element tag names but spotted this while looking for loops.

3. The property name key in the example object in the custom element template was not a base URI

## ❓ How to test this?

1.  You can test fix (1) by building the custom element template and then publishing it to a locally-running Hub, and checking its page works. Compared to doing the same thing in `main`, where the Hub will freeze.

2. You can test (2) by running `yarn dev` in `mock-block-dock`, and then (before selecting the custom element block) registering something to occupy its name:
```js
// browser console
customElements.define("test-custom-element-block", HTMLElement);
// you can repeat this with `test-custom-element-block1" etc 
// but will need to choose a different HTMLXElement for each
```
now switch to the custom element block, and inspect the DOM to see that it's had a number appended.

3. You can probably just look at 3 to check its correctness

---
  - https://app.asana.com/0/0/1204164203672391

### DIFF
--- a/libs/block-template-react/package.json
+++ b/libs/block-template-react/package.json
@@ -59,7 +59,7 @@
     "displayName": "Template",
     "examples": [
       {
-        "name": "World"
+        "https://blockprotocol.org/@blockprotocol/types/property-type/name/": "World"
       }
     ],
     "icon": "public/omega.svg",
--- a/.changeset/plenty-camels-crash.md
+++ b/.changeset/plenty-camels-crash.md
@@ -1,0 +1,5 @@
+---
+"block-template-custom-element": patch
+---
+
+correct property key in example object
--- a/.changeset/sharp-stingrays-scream.md
+++ b/.changeset/sharp-stingrays-scream.md
@@ -1,0 +1,5 @@
+---
+"@blockprotocol/core": patch
+---
+
+update block init message retry from microtask to macrotask
--- a/.changeset/sharp-tools-bow.md
+++ b/.changeset/sharp-tools-bow.md
@@ -1,0 +1,5 @@
+---
+"mock-block-dock": patch
+---
+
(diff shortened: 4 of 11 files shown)


libs/block-template-react/package.json
```diff
@@ -59,7 +59,7 @@
     "displayName": "Template",
     "examples": [
       {
-        "name": "World"
+        "https://blockprotocol.org/@blockprotocol/types/property-type/name/": "World"
```
benwerner01, line 62: I'm assuming it's not an issue that `https://blockprotocol.org/@blockprotocol/types/property-type/name/` is not a real type, and that we plan on publishing common types like this when we have a type nexus

benwerner01 (approved):
Thanks Ciaran, have tested the first two changes locally and they seem to be working as expected, and the code looks good to me 👍

libs/block-template-react/package.json
CiaranMn: https://blockprotocol.org/@blockprotocol/types/property-type/name/v/1 should actually exist - it's a property on https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2

PR #1158, opened 2023-03-11
Introduce improved Type System codegen utility for blocks

Alfred-Mountfield:

This introduces a completely rewritten codegen script for use in blocks. The new script is faster, more extensible, supports many more edge-cases, and is more customisable. 

Some of the notable improvements:

- Hoisting shared identifiers to a `shared.ts` file, this avoids many duplicated definitions of common types (especially data types such as `Text`)
- Removing duplicating documentation - the previous codegen utility could result in orphaned duplicated documentation strings lying around after their types (which were duplicates of others) were deleted
- The output is now idempotent, which makes the generation suitable for checking as part of CI (ensuring that the generated types stay up to date if the codegen utility is improved, or a user updates the type they're generating)
- Much better name conflict resolution. Previously we would identify duplicate entity type titles and append an incrementing number to them (or add a version to one of them), this instead intelligently differentiates conflicting type names by their class of type (data type, property type, entity type), their base URL, and their versions.
- Support for generating multiple types, to multiple files. Previously the codegen utility in `block-scripts` only supported generating a single schema for the source entity type.
- Link entities are now affirmed as having `linkData` (whereas previously this was left as optional)

This also fixes the dependency graph within the repo (at least related to `build` and `codegen`), which now allows us to use `codegen` rather than `schema` as the `yarn script` as the bug to do with site building has been fixed. This was due to us poorly saying that every single build required codegen, as opposed to individual ones.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056186/1203984686668764/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203193968491715/1204155628030421/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203179076056186/1204155628030424/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203179076056186/1204155628030426/f) _(internal)_
- [Asana task](https://app.asana.com/0/1203179076056186/1203843256013361/f) _(internal)_

## 🚫 Blocked by

- [ ] Check spec for any updates that need to be made

## 🔍 What does this change?

Please see commits in order. It is a pretty huge PR but I was attempting to reach near-parity before submitting one. I have gone to considerable effort to make sure the commits are clean and help guide understanding in the design.

- I'm unsure if we already mentioned the codegen in the docs, if we did we should update it
- As mentioned above, I think we do need to update the spec to mention the changes to the `package.json` as I believe we'd mentioned the `schema` keyword

## ⚠️ Known issues

- N/A

## 🐾 Next steps

- Generate a helper look-up map of URLs (versioned and perhaps base) to prettified constants so users can stop dealing with the long URLs and instead just use `name` or `columnId`, etc.
- Make the logging level configurable to blocks
- Add the type's IDs to the doc comments so that you know where the generated type came from
- Move the documentation from `<EntityName>Properties` to `<EntityName>` where it's more useful

## 🛡 What tests cover this?

- `tsc`
- `eslint`

These should be checked in the templates as part of CI now, whereas they weren't previously

## ❓ How to test this?

1.  Try looking at the definition of the `CodegenParameters` type, and add some new elements to one of the block templates `package.json`
1. Look at the generated types
1. Also see [[DROP] - Example local usage](https://github.com/blockprotocol/blockprotocol/commit/72d7a2db00002f55f833d98e16fbaaff9ea56ed2) for an example (although this was done locally in `@blockprotocol/graph` as opposed to in one of the templates. It does show how we should update the `turbo.json` though)

## 📹 Demo

- [Use new codegen in React block template](https://github.com/blockprotocol/blockprotocol/commit/eda7aa5e8bb6df8d1c6d2cc993be51f8c6ed2fca)
- [Use new codegen in Custom element block template](https://github.com/blockprotocol/blockprotocol/commit/42d5d596c26ac9c7a6bfedd88100bff1c1958b9e)
- [[DROP] - Example local usage](https://github.com/blockprotocol/blockprotocol/commit/72d7a2db00002f55f833d98e16fbaaff9ea56ed2)


libs/@blockprotocol/graph/src/codegen/initialize/traverse.ts
```diff
@@ -0,0 +1,165 @@
...
+  const addFetchPromise = (fetchPromise: Promise<void>) => {
+    fetchQueue.push(fetchPromise);
+
+    void fetchPromise.then(() => {
+      fetchQueue.splice(fetchQueue.indexOf(fetchPromise), 1);
+    });
+  };
+
+  // Somewhat concurrent fetch queue, keep exploring as long as there is an in-flight request or there are more to
+  // explore
```
Alfred-Mountfield, line 109: 🏎️ 💨

libs/@blockprotocol/graph/src/codegen/parameters.ts
```diff
@@ -0,0 +1,241 @@
...
+        overrides?: {
+          [versionedUrl: string]: string;
+        };
+      };
+  /* @todo - Add support for generating Base URL aliases? */
+  /* @todo - Make the resolve depth configurable */
+  temporal?: boolean;
+};
+
+export const validateCodegenParameters = (
```
Alfred-Mountfield, line 41: This is pretty horrible but the error reports are nice.. perhaps we can just leverage some codegen utility to generate the shape and appropriate type-guard

libs/@blockprotocol/graph/src/codegen/postprocess/generate-block-link-target-aliases.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      if (
+        typedValues(context.filesToDefinedIdentifiers).find(
+          (definedIdentifier) =>
+            definedIdentifier.has(entityLinkAndTargetsIdentifier),
+        )
+      ) {
+        context.logDebug(
+          `Generating block entity links and targets alias ${typeName} for file ${fileName}`,
+        );
```
Alfred-Mountfield, line 32: I threw this together somewhat quickly to get this PR up because it had been lagging. I don't like this, it could potentially be made less gross by at least defining this template somewhere shared but I'd really like to not rely on checking if this thing has been created already or manually recreating the identifier for it.

Perhaps it needs another map on the context 🙃

libs/@blockprotocol/graph/src/codegen/parameters.ts
```diff
@@ -0,0 +1,241 @@
...
+
+export type ProcessedCodegenParameters = Omit<
+  Required<CodegenParameters>,
+  "targets"
+> & {
+  targets: {
+    [fileName: string]: {
+      versionedUrls: VersionedUrl[];
+      blockEntity?: VersionedUrl;
+    };
```
Alfred-Mountfield, line 190: I opted for a different representation here because it's easier to deal with later on, but I wanted to avoid making users copy paste the same Versioned URL twice in the package.json, as opposed to just being able to write an inline flag like `blockEntity: true`

thehabbos007 (commented):
Leaving a partial review, most of the way through, looking at postprocessing. Looks great so far!

libs/@blockprotocol/graph/src/codegen/initialize/traverse.ts
```diff
@@ -0,0 +1,165 @@
...
+
+  const fetchQueue: Promise<void>[] = [];
+
+  const addFetchPromise = (fetchPromise: Promise<void>) => {
+    fetchQueue.push(fetchPromise);
+
+    void fetchPromise.then(() => {
+      fetchQueue.splice(fetchQueue.indexOf(fetchPromise), 1);
+    });
+  };
```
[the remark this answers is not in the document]
Alfred-Mountfield: Yep, couldn't rely on `Promise.all` or something as that'd block on a subset of the queue and won't let it continue to be dynamically updated as requests come in. And to my knowledge, you can't poll promises to check their results without awaiting them. So it seemed clean to me to get the promise to remove itself from the queue on completion

libs/@blockprotocol/graph/src/codegen/preprocess/transform-type-names.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+    return typeName;
+  } else {
+    if (isTypescriptKeyword(titleCase)) {
+      throw new Error(
+        `Generated type name "${titleCase}" is a TypeScript keyword`,
+      );
+    }
+    return titleCase;
+  }
```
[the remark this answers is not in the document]
Alfred-Mountfield: Good catch, there previously was some other logic in here which benefited from separated branches.

I'll re-add the explanatory comment though and commit it, cheers

libs/@blockprotocol/graph/src/codegen/preprocess/transform-type-names.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  typedEntries(context.allTypes).forEach(([typeId, type]) => {
+    const override = typeNameOverrides[typeId];
+    const typeNameFromTitle = generateValidTypeScriptIdentifierFromTitle(
+      override ?? type.title,
+    );
+
+    if (override && typeNameFromTitle !== override) {
+      context.logWarn(
+        `Type name override of "${override}" for "${typeId}" isn't in Title Case, using "${typeNameFromTitle}" instead.`,
```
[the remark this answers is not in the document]
Alfred-Mountfield: Oof good catch. I think I had previously actually been converting it to Title Case and then allowing the JSON schema compiler to turn it into a type name as it wanted (which it did by turning it into PascalCase I think).

However I've decided to take greater control over this, as if we define the `title` as a valid typescript identifier the compiler doesn't seem to mutate it at all, which means we can use the rewritten `title` at other points in the pipeline _as_ the identifier, rather than needing to maintain some mapping.

I'll try and find a way to explain this slightly better in the code, and then update to replace Title Case with PascalCase in messages and variable names. Thanks!

libs/@blockprotocol/graph/src/codegen/initialize/traverse.ts
```diff
@@ -0,0 +1,165 @@
...
+    const typeUrl = this.exploreQueue.values().next().value;
+    if (typeUrl) {
+      this.exploreQueue.delete(typeUrl);
+      this.explored.add(typeUrl);
+    }
+    return typeUrl;
+  }
+}
+
+export const traverseAndCollateSchemas = async (
```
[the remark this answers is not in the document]
Alfred-Mountfield: [Make processing step filenames consistent](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/a04180b363f204798bf453d8aad1c4b9d7c61d8b)

libs/@blockprotocol/graph/src/codegen/preprocess/transform-type-names.ts
```diff
@@ -0,0 +1,154 @@
...
+    return titleCase;
+  }
+};
+
+/**
+ * **Rewrites** the `title`s of the types to be locally (per file) unique valid javascript identifiers.
+ *
+ * Clashes are dealt with per class of type, and then per Base Url, and then per revision.
+ */
+export const rewriteTypeTitles = (context: PreprocessContext) => {
```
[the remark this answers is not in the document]
Alfred-Mountfield: [Make processing step filenames consistent](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/a04180b363f204798bf453d8aad1c4b9d7c61d8b)

libs/@blockprotocol/graph/src/codegen/initialize/traverse.ts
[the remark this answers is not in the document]
Alfred-Mountfield: [Add clarifying comment](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/2dd4906d94569a1ee0132156214d66e97f7b3f87)

libs/@blockprotocol/graph/src/codegen/preprocess/transform-type-names.ts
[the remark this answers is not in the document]
Alfred-Mountfield: [Clean-up type title transformation](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/9420d6ada452776cf2fc202e0a565cb1910af9cb)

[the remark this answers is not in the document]
Alfred-Mountfield: I re-read the documentation of the top level function in this file and it does say that we rewrite the titles to valid JS identifiers so I think it's probably fine

libs/@blockprotocol/graph/src/codegen/preprocess/transform-type-names.ts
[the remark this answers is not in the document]
Alfred-Mountfield: [Clean-up type title transformation](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/9420d6ada452776cf2fc202e0a565cb1910af9cb)

I re-read the documentation of the top level function in this file and it does say that we rewrite the titles to valid JS identifiers so I think it's probably fine

libs/@blockprotocol/graph/src/codegen/postprocess/generate-block-link-target-aliases.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        const identifier = "BlockEntityOutgoingLinksAndTargets";
+
+        const blockEntityLinkAndTargetTypesFragment = `\nexport type ${identifier} = ${entityLinkAndTargetsIdentifier}\n`;
```
thehabbos007, line 36: We won't run into clasing identifiers because we only allow one block entity type per file, right?

libs/@blockprotocol/graph/src/codegen/postprocess/prepend-banner-comments.ts
```diff
@@ -0,0 +1,20 @@
+import { mustBeDefined } from "../../shared/util/must-be-defined";
+import { typedKeys } from "../../shared/util/typed-object-iter.js";
+import { PostprocessContext } from "../context/postprocess.js";
+
+const bannerComment = () => `/**
+ * This file was automatically generated – do not edit it.
+ * See the respective URLs for the Block Protocol Types this was generated from.
+ */
```
thehabbos007, line 8: Would it make sense to add the types that the file contains in this comment or any of the others? I appreciate that we may have these URLs in other structures within the generated code, but maybe linking to the types in the docs may be helpful?

libs/@blockprotocol/graph/src/codegen/context/postprocess.ts
```diff
@@ -0,0 +1,205 @@
...
+      console.log(`DEBUG: ${message}`);
+    }
+  }
+
+  logTrace(message: string) {
+    if (this.logLevel === "trace") {
+      // eslint-disable-next-line no-console
+      console.log(`TRACE: ${message}`);
+    }
+  }
```
thehabbos007, line 137: This is shared between the different contexts. Do we want to centralize this along with the todo at the top?

thehabbos007 (commented):
Very nice. Thanks for addressing the comments above as well. 
I'm happy with the changes, but I'll let Ciaran have a look as well!

libs/@blockprotocol/graph/src/codegen/postprocess/prepend-banner-comments.ts
Alfred-Mountfield: An alternative to that is raised in a `todo` in the code and in the "Next Steps" of the PR :)

Basically to put the versioned URLs of the types into their doc-comments "Generated from http:..."

libs/@blockprotocol/graph/src/codegen/postprocess/generate-block-link-target-aliases.ts
Alfred-Mountfield: Correct, and that's checked as part of validating the `CodegenParameters`, and we only hoist "locally importable" identifiers to `shared.ts` so these will be unique per generated file if they're defined.

libs/@blockprotocol/graph/src/codegen/context/postprocess.ts
Alfred-Mountfield: The actual implementations are a line or two so I don't see turning them into functions as helpful. The other way I could see us centralising it is by making some context that all the others `extend` but that seems kind of superfluous..

Really the solution is to use `winston` or something for logging, in which case you generally need a `logger` object to pass around which is why I put the methods in the classes for now as opposed to some standalone thing.

libs/block-template-react/package.json
```diff
@@ -19,13 +19,13 @@
   },
   "scripts": {
     "build": "block-scripts build",
+    "codegen": "rimraf ./src/types && block-scripts codegen && yarn format && yarn fix:eslint",
```
Alfred-Mountfield, line 22: Calling format and fix is causing problems in our CI 🙃 
https://github.com/blockprotocol/blockprotocol/actions/runs/4406199711/jobs/7718107593#step:6:198

@CiaranMn should I just remove the eslint call, or should I try investing time into our bespoke package chore thing to account for this?

CiaranMn: Just remove it. I'd also scope the `format` call to only the autogenerated file(s), otherwise running `codegen` is going to cause a diff in other files too, which seems unexpected.

CiaranMn: Also seems like it might be bad to remove `src/types` if the user has added other files there? can we filter all these to extension (if we still have `.gen.ts`)?

Alfred-Mountfield: > I'd also scope the format call to only the autogenerated file(s),
> Also seems like it might be bad to remove src/types if the user has added other files there? can we filter all these to extension (if we still have .gen.ts)?

I thought about scoping these, but the users are responsible for picking the file names and root folder now (in the config below) so it would be difficult to automate..

CiaranMn: You could output the files affected from the `codegen` script so it can be piped into the `format`? or just put the `format` inside the codegen script

CiaranMn: or `format` could read from the codegen config (but returning the generated files might generally be useful)

CiaranMn (commented):
I haven't looked at the internal logic or tested the outputs but have some comments on the user interface

libs/block-template-custom-element/package.json
```diff
@@ -57,13 +58,23 @@
...
+        "https://blockprotocol.org/@blockprotocol/types/property-type/name/": "World"
       }
     ],
     "icon": "public/omega.svg",
     "image": "public/block-preview.png",
     "name": "block-template-custom-element",
     "protocol": "0.3",
-    "schema": "https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2"
+    "codegen": {
+      "targetRoot": "src/types",
```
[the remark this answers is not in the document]
Alfred-Mountfield: I'll swap it to `outputFolder`, cheers

libs/block-template-custom-element/src/types/thing.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * This file was automatically generated – do not edit it.
+ * See the respective URLs for the Block Protocol Types this was generated from.
```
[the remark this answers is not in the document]
Alfred-Mountfield: See [my response here](https://github.com/blockprotocol/blockprotocol/pull/1158#discussion_r1134045067), and the next steps in the PR description

libs/block-template-custom-element/package.json
```diff
@@ -57,13 +58,23 @@
...
+      "targets": {
+        "thing.ts": [
+          {
+            "versionedUrl": "https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2",
```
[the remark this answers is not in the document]
Alfred-Mountfield: Maybe Type ID is better than versioned URL.

`typeId`?
`typeIdToGenerate`?
`sourceTypeId`?

libs/block-template-custom-element/package.json
```diff
@@ -57,13 +58,23 @@
...
       {
-        "name": "World"
+        "https://blockprotocol.org/@blockprotocol/types/property-type/name/": "World"
       }
     ],
     "icon": "public/omega.svg",
     "image": "public/block-preview.png",
     "name": "block-template-custom-element",
     "protocol": "0.3",
-    "schema": "https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2"
```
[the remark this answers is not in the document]
Alfred-Mountfield: > have to update the block's schema URL in two places, which we should avoid.

This was my main motivation for the current format ([see this comment](https://github.com/blockprotocol/blockprotocol/pull/1158#discussion_r1132987188)), so I feel fairly strongly about this too.

> I am instinctively attracted to (1) because it is simpler for the block use case where you don't need anything but the block + immediate subgraph generated (essentially operating as it does now, perhaps with a file name change), and specifying the more complicated codegen object is for use cases where you need multiple types generated and want to specify filenames and so on.

I actually disagree with this, I think two competing formats and needing to understand how to migrate from one to the other (or use both at once?) is more confusing and _less_ simple. We expect the vast majority of blocks to be created from the template (or following some future tutorials), which will explicitly give users the `codegen` boilerplate, so I think the simpler block cases boil down to "change the only URL you see in this object to your URL" and it's functionally the same difficulty as having `schema`.

As such I'd prefer if we go with this:

 > We drop this key and take the value for schema in block-metadata.json from whatever entry in codegen is marked as blockEntity: true.

To be able to do that I think we'd need to stop what I've done at the moment where I'm allowing a `blockEntity: true` _per_ file. My intentions there was that technically this codegen doesn't need to only be ran for a single block, one could define one codegen for a whole suite of blocks for example (using the file mapping would make that fairly easy). However, I don't see a huge benefit for that, and I guess we can continue to allow that in codegen, but just disallow it when parsing a config from a block's `package.json`. If you're happy with that I'll go and implement this

libs/block-template-custom-element/package.json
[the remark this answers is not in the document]
Alfred-Mountfield: [Rename targetRoot to outputFolder](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/e06176a51c2ce85c00fd36b69fb8acaf00e56856)

libs/@blockprotocol/graph/src/codegen/compile/remove-placeholder-types.ts
```diff
@@ -0,0 +1,19 @@
+import { typedEntries } from "../../shared/util/typed-object-iter.js";
+import { CompileContext } from "../context/compile.js";
+import { CompiledTsType } from "../shared.js";
+
+const removePlaceholderDefinitionInCompiledTsType = (
+  compiledTsType: CompiledTsType,
+): CompiledTsType => compiledTsType.replace(/^.* = "PLACEHOLDER"$/gm, "");
```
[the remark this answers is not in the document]
Alfred-Mountfield: [Reuse placeholder const](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/516e773ae8c86afa9a3276400548d2b2275edced)

libs/block-template-custom-element/package.json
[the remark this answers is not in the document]
Alfred-Mountfield: [Opt for Type ID instead of Versioned URL in most places](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/3d7d819963407738be9696e6ee2f390feac841b2)

libs/block-template-custom-element/package.json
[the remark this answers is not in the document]
CiaranMn: Yes, sounds good. Can you update the `build` script at the same time to populate `schema` from the new place please?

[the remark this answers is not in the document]
Alfred-Mountfield: I've applied a constraint of only accepting one `blockEntity` definition in the `package.json.`
[Only allow one blockEntity within block's codegen parameters](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/10d6f99a865411c21ccbffd97b58157271ef3397)

You can checkout [Add demonstration of multiple codegen blockEntity defs](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/7bd695aa341c291c91f3124e38119fba64b16f58) to see it working: image


> This is the block's declared schema which is used as the value for schema in block-metadata.json, which is generated by block-scripts/build.
> [...]
> Yes, sounds good. Can you update the build script at the same time to populate schema from the new place please?

Was a bit unsure of how to do this as `generate-dist-block-metadata.js` seems to just set `schema: "block-schema.json"`, but I'm guessing that's implicitly overridden by `...blockprotocol`.

[Use blockEntity type URL from codegen object in block-metadata](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/38b5246b58a055a8703679c9394e9476df63cca8)

Seems to work:
image

libs/block-template-custom-element/src/types/thing.ts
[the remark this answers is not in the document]
Alfred-Mountfield: I'll opt to remove it from the banner now then, as I think we should get the PR moving and the prior implementation was missing the URLs for the majority of the types (it only had it for the one root entity type)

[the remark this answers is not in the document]
Alfred-Mountfield: [Temporarily remove referral to URLs from banner comment](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/b15d70c85c0a9b7e2aa74bbff5051ce30b43009e)

libs/block-template-react/package.json
Alfred-Mountfield: [Output affected files on stdout and use as input for formatting](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/98cf8fa0230049e81e049b5b9f78caa386cf8f3e)

Alfred-Mountfield: [Clean outputDir from within codegen](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/2c764b7c21dab05f2c3749dd85df0a4d3cb10195)

libs/block-scripts/shared/generate-dist-block-metadata.js
```diff
@@ -41,6 +41,16 @@ export const generateDistBlockMetadata = async (extra) => {
     ...extra,
   };
 
+  /* @todo - deprecate support for the old `schema` definitions and migrate to this being the primary implementation */
```
CiaranMn, line 44: Let's just delete `schema: "block-schema.json"` now – we don't support it

libs/block-template-react/package.json
CiaranMn: Formatting looks good. I'm still wary of the combination of (a) nuking the `outputFolder` and (b) having it default in the template to something as generic as `src/types`. What about making the default folder something more specific, e.g. `src/types/generated` or something?  Then people can choose to simplify it if they want. The current set up prevents people from using `src/types` without changing the output.

CiaranMn: Also wonder if we should call the default file `block.ts` or `block-entity.ts`? otherwise initial set up for a block (or swapping the block entity) means immediately replacing both the filename and the URL

libs/@blockprotocol/graph/src/codegen/postprocess/generate-link-and-target-definitions.ts
```diff
@@ -0,0 +1,237 @@
...
+ * Generates the union of all outgoing link and target definitions for a given entity
+ *
+ * @param entityName
+ * @param linkAndTargetIdentifiers
+ */
+const linkAndTargetsUnionDefinition = (
+  entityName: string,
+  linkAndTargetIdentifiers: string[],
+): IdentifierToCompiledDefinition => {
+  const identifier = `${entityName}OutgoingLinksAndTargets`;
```
CiaranMn, line 79: As this is a union of single items, I would name it in the singular. This type describes a single outgoing link and target (which can be satisfied in multiple different ways).

Then you don't get oddities likes `FooOutgoingLinksAndTargets[]` which looks like an array of arrays but is an array of single link/target objects.

```suggestion
  const identifier = `${entityName}OutgoingLinkAndTarget`;
```

libs/@blockprotocol/graph/src/codegen/postprocess/generate-link-and-target-definitions.ts
```diff
@@ -0,0 +1,237 @@
...
+ */
+const individualOutgoingLinkAndTargetDefinition = (
+  sourceName: string,
+  linkIdentifier: string,
+  targetIdentifiers: [string, ...string[]],
+): IdentifierToCompiledDefinition => {
+  const identifier = `${sourceName}${linkIdentifier}Links`;
+
+  const targetUnion = targetIdentifiers.join(" | ");
+  const compiledContents = `export type ${identifier} = { linkEntity: ${linkIdentifier}; target: ${targetUnion} }`;
```
CiaranMn, line 32: Here we've encoded the `rightEntity` as the `target`, and this type is therefore inconsistent with the typings of a subgraph. 

I think this should be `rightEntity`, unless we want to change `rightEntity` to `target` everywhere else

libs/block-scripts/shared/generate-dist-block-metadata.js
Alfred-Mountfield: [Remove outdated block-schema.json reference](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/aff5a04fa9cee68ba4203d0f312d0779a958a6fd)

Note that I left the TODO. Can we completely drop `blockprotocol.schema` or will that break existing blocks or the hub or anything?

CiaranMn: It won't break anything, it's only read by `block-scripts`

libs/block-template-react/package.json
Alfred-Mountfield: I've changed it to `src/types/generated` as you suggest. I've left it as `thing.ts` because I think it's good to get someone to replace the filename and get into the thought of separating types into well-named files as they need to. `block-entity.ts` feels so generic and isn't a pattern I'd want to encourage personally. I don't feel super strongly so I can rename it if you do
[Change default codegen outputFolder to src/types/generated](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/498eac132cf838737726429d49b0bb53f84c0961)

libs/@blockprotocol/graph/src/codegen/postprocess/generate-link-and-target-definitions.ts
Alfred-Mountfield: I had named it that way before, and then found it confusing in usage and swapped it around for some reason.. I wish I knew why.

Regardless, definitely agree with the general convention on union namings:
[Rename OutgoingLinksAndTargets union to OutgoingLinkAndTarget](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/9f0576eb32a34230090296de4daf0fa0d82c6625)

libs/@blockprotocol/graph/src/codegen/postprocess/generate-link-and-target-definitions.ts
Alfred-Mountfield: So it is, I changed these about 5 times trying to figure out which one was consistent, clearly I missed something in the refactoring. (Yet again I really wish we could use `satisfies` on types to describe a subtype)

Applied as driveby in
[Rename OutgoingLinksAndTargets union to OutgoingLinkAndTarget](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/9f0576eb32a34230090296de4daf0fa0d82c6625)

libs/block-scripts/shared/generate-dist-block-metadata.js
Alfred-Mountfield: Well it will break block building if someone updates their `block-scripts` version and doesn't update their `package.json`. But I guess trying to support the legacy `package.json` format for that doesn't matter anyway as `codegen` will break.

Will remove the TODO

CiaranMn: Yes – this is why I suggest updating the error messagr from `block-scripts` to explain (or point to docs that do)

Alfred-Mountfield: See [Replace schema with blockEntityType and inject that into codegen …](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/d8129045ce1ba4622c0b3c3d3cb191ec276698be) for updated code around this

I'm working on docs at the moment

libs/block-template-react/package.json
Alfred-Mountfield: I ended up changing my mind about this while writing the docs, as I realised not setting the `schema` explicitly was quite strange and we were depending on it being in a somewhat unrelated thing of `codegen`.

I've renamed `schema` to `blockEntityType` within packages which use the `package.json` approach, and updated the `codegen` to allow that to be injected:
[Replace schema with blockEntityType and inject that into codegen …](https://github.com/blockprotocol/blockprotocol/pull/1158/commits/d8129045ce1ba4622c0b3c3d3cb191ec276698be)

apps/site/src/_pages/docs/2_working-with-types.mdx
```diff
@@ -65,3 +65,59 @@ fetch("https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2")
 ```shell
 curl https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2
 ```
+
+### TypeScript types for entities
+
+Types within the Type System are representable within TypeScript, and as such it is possible to create TypeScript representations of entity types.
+There is a provided `codegen` utility in the graph module package which provides an implementation of such generation.
```
CiaranMn, line 72: I was originally going to suggest Graph Module but I think this makes more sense
```suggestion
There is a provided `codegen` utility in the `@blockprotocol/graph` package which provides an implementation of such generation.
```

apps/site/src/_pages/docs/2_working-with-types.mdx
```diff
@@ -65,3 +65,59 @@ fetch("https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2")
...
+For relatively simple cases, a wrapper around this is exposed through `block-scripts` which is able to parse the metadata about your block and generate types for you.
+An example usage of this is included within the `react` and `custom-element` block templates.
+
+The `block-scripts codegen` utility is configured through the `codegen` object in a block's `package.json`.
+
```
CiaranMn, line 77: This line looks like an earlier version of the following lines
```suggestion
```

apps/site/src/_pages/docs/2_working-with-types.mdx
```diff
@@ -65,3 +65,59 @@ fetch("https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2")
...
+- a field `outputFolder` which describes the folder which will contain the files with the generated types
+- **all contents of this folder are deleted when the codegen is run**
+- a field `targets` which is an object which is a mapping of files to the types that should be generated within them
+- keys of `targets` are the names of the files to be generated (e.g. `person.ts`)
+- values of `targets` are an array where the items are objects which satisfy the following:
+- contain the key `sourceTypeId` with a value of an ID of a type in the Type System (e.g. `https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2`), _or_
+- contain the key `blockEntityType` with value `true`
+- this says to generate types for the block entity type (given by the `blockEntityType` property in the `blockprotocol` object in `package.json`), alongside type aliases to identify it as such (e.g. `type BlockEntity = ...`)
+- there can be **one at most** of these entries in the whole of the `codegen` object
```
CiaranMn, line 91: Suggestion for nesting and some combination / shortening of the `blockEntityType` bullets which I thought could be a bit clearer

I'm also not clear that we need to place any limit on how many times `blockEntityType` is provided in `codegen` – removing the constraint would simplify the docs and implementation a bit, unless we think it worth maintaining
```suggestion
- a field `outputFolder` which describes the folder which will contain the files with the generated types
  - **all contents of this folder are deleted when the codegen is run**
- a field `targets` which is an object which is a mapping of files to the types that should be generated within them
  - keys of `targets` are the names of the files to be generated (e.g. `person.ts`)
  - values of `targets` are an array where the items are objects which satisfy the following:
    - contain the key `sourceTypeId` with a value of an ID of a type in the Type System (e.g. `https://blockprotocol.org/@blockprotocol/types/entity-type/thing/v/2`), _or_
    - contain the key `blockEntityType` with value `true`, which if present will generate types for the type given by the `blockEntityType` property in the `blockprotocol` object in `package.json`, alongside type aliases for it such as `BlockEntity`
      - there can be **one at most** of these entries in the whole of the `codegen` object
```

libs/block-template-custom-element/package.json
```diff
@@ -19,7 +19,7 @@
   },
   "scripts": {
     "build": "block-scripts build",
-    "codegen": "rimraf ./src/types && block-scripts codegen && yarn format && yarn fix:eslint",
+    "codegen": "set -o pipefail; block-scripts codegen | xargs -I {} yarn prettier --write {}",
```
CiaranMn, line 22: A bit unfortunate that this command will no longer work on Windows terminal but think it's fine to require use of WSL for now

CiaranMn (commented):
Just a few comments on the added docs, looks good to go otherwise

apps/site/src/_pages/docs/2_working-with-types.mdx
Alfred-Mountfield: > I'm also not clear that we need to place any limit on how many times blockEntityType is provided in codegen – removing the constraint would simplify the docs and implementation a bit, unless we think it worth maintaining

I don't think we necessarily _need_ to either, but I think it'd be a bit confusing to start to have competing (duplicate) `BlockEntity` definitions. Also the codegen implementation will complain if you try and define `BlockEntity` twice in the same file (as that would error), so we'd still need to explain it's once per file, and at that point I reckon just limit it to once

libs/block-template-custom-element/package.json
Alfred-Mountfield: I guess you mean `cmd` or powershell?

Regardless this was broken on plain `sh` as you pointed out to me offline 😭, so I've compromised back to `&&` and just formatting everything. It's unpleasant and causes side-effects but it's very easy to locate, understand, and disable for anyone who is bothered by that behaviour.

CiaranMn: Seems like a fine compromise for now, can revisit if needed

PR #1161, opened 2023-03-13
Add latency simulation to `mock-block-dock`

Alfred-Mountfield:

Delays in responses to methods from an EA can **very easily** result in strange errors in components. Such errors were very difficult to identify during block development because MockBlockDock responded very quickly to requests.

This PR introduces a new prop which allows the user to configure a latency range whereby responses from the Graph Module will delay for a random time in ms from within that range.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056186/1204144427699229/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- Introduces a `simulateLatency` prop in MBD
- Adjusts the Graph module callbacks to wait for a random time within the given latency range
- Updates the templates to provide the parameter

- I'm unsure if there is documentation about MBD props, I have updated the templates and given an accompanying message

## ⚠️ Known issues

- A lot of our demo blocks called an `update` on every key-press. This was already known as being problematic (as it would absolutely spam the EA with updates, and should be batched), but the latency issues with re-renders causing the input to reset is unfortunate and luckily highlights the need for this PR.

## 🐾 Next steps

- Update existing blocks (including templates) to account for the issues uncovered by this

## 🛡 What tests cover this?

- `tsc`
- `eslint`
- Manually running the templates

## ❓ How to test this?

1. Try running MBD or the templates
2. Adjust the latency parameter (and remove it) and see the differences

## 📹 Demo

https://user-images.githubusercontent.com/25749103/224759478-f60bfe37-63f3-44bc-9a8d-adf9f91438e6.mov

### DIFF
--- a/libs/block-template-custom-element/src/dev.tsx
+++ b/libs/block-template-custom-element/src/dev.tsx
@@ -55,6 +55,11 @@ const DevApp = () => {
       initialData={{
         initialEntities: [testEntity],
       }}
+      simulateLatency={{
+        // configure this to adjust the range of artificial latency (in ms)
+        min: 50,
+        max: 200,
+      }}
       // hideDebugToggle <- uncomment this to disable the debug UI entirely
       // initialEntities={[]} <- customise the entities in the datastore (blockEntity is always added, if you provide it)
       // initialEntityTypes={[]} <- customise the entity types in the datastore
--- a/libs/block-template-custom-element/src/dev.tsx
+++ b/libs/block-template-custom-element/src/dev.tsx
@@ -55,7 +55,7 @@ const DevApp = () => {
       initialData={{
         initialEntities: [testEntity],
       }}
-      simulateLatency={{
+      simulateDatastoreLatency={{
         // configure this to adjust the range of artificial latency (in ms)
         min: 50,
         max: 200,
--- a/libs/block-template-custom-element/src/dev.tsx
+++ b/libs/block-template-custom-element/src/dev.tsx
@@ -56,7 +56,7 @@ const DevApp = () => {
         initialEntities: [testEntity],
       }}
       simulateDatastoreLatency={{
-        // configure this to adjust the range of artificial latency (in ms)
+        // configure this to adjust the range of artificial latency in responses to datastore-related requests (in ms)
         min: 50,
         max: 200,
       }}
--- a/libs/block-template-html/dev/index.html
+++ b/libs/block-template-html/dev/index.html
@@ -55,8 +55,13 @@
             initialData=${initialData}
             blockEntityRecordId=${blockEntity.metadata.recordId}
             debug
+            simulateLatency=${{
+              min: 100,
+              max: 200,
+            }}
           />
         `;
+        // modify `simulateLatency` above to adjust the range of artificial latency (in ms)
         // remove 'debug' above to start with the debug UI minimised. You can also toggle it in the UI
         // hideDebugToggle <- add 'hideDebugToggle' to disable the debug UI entirely
         // readonly <- add this above to start your block in readonly mode. You can also toggle it in the UI
--- a/libs/block-template-html/dev/index.html
+++ b/libs/block-template-html/dev/index.html
@@ -61,7 +61,7 @@
             }}
           />
         `;
-        // modify `simulateDatastoreLatency` above to adjust the range of artificial latency (in ms)
+        // modify `simulateDatastoreLatency` above to adjust the range of artificial latency in responses to datastore-related requests  (in ms)
         // remove 'debug' above to start with the debug UI minimised. You can also toggle it in the UI
         // hideDebugToggle <- add 'hideDebugToggle' to disable the debug UI entirely
         // readonly <- add this above to start your block in readonly mode. You can also toggle it in the UI
--- a/libs/block-template-html/dev/index.html
+++ b/libs/block-template-html/dev/index.html
@@ -55,13 +55,13 @@
             initialData=${initialData}
             blockEntityRecordId=${blockEntity.metadata.recordId}
             debug
-            simulateLatency=${{
+            simulateDatastoreLatency=${{
               min: 100,
               max: 200,
             }}
           />
         `;
-        // modify `simulateLatency` above to adjust the range of artificial latency (in ms)
+        // modify `simulateDatastoreLatency` above to adjust the range of artificial latency (in ms)
         // remove 'debug' above to start with the debug UI minimised. You can also toggle it in the UI
         // hideDebugToggle <- add 'hideDebugToggle' to disable the debug UI entirely
         // readonly <- add this above to start your block in readonly mode. You can also toggle it in the UI
--- a/libs/block-template-react/src/dev.tsx
+++ b/libs/block-template-react/src/dev.tsx
@@ -44,7 +44,7 @@ const DevApp = () => {
         initialEntities: [testEntity],
       }}
       simulateDatastoreLatency={{
-        // configure this to adjust the range of artificial latency (in ms)
+        // configure this to adjust the range of artificial latency in responses to datastore-related requests (in ms)
         min: 50,
         max: 200,
       }}
--- a/libs/block-template-react/src/dev.tsx
+++ b/libs/block-template-react/src/dev.tsx
@@ -43,6 +43,11 @@ const DevApp = () => {
       initialData={{
         initialEntities: [testEntity],
       }}
+      simulateLatency={{
+        // configure this to adjust the range of artificial latency (in ms)
+        min: 50,
+        max: 200,
+      }}
       // hideDebugToggle <- uncomment this to disable the debug UI entirely
       // initialEntities={[]} <- customise the entities in the datastore (blockEntity is always added, if you provide it)
       // initialEntityTypes={[]} <- customise the entity types in the datastore
--- a/libs/block-template-react/src/dev.tsx
+++ b/libs/block-template-react/src/dev.tsx
@@ -43,7 +43,7 @@ const DevApp = () => {
       initialData={{
         initialEntities: [testEntity],
       }}
-      simulateLatency={{
+      simulateDatastoreLatency={{
         // configure this to adjust the range of artificial latency (in ms)
         min: 50,
         max: 200,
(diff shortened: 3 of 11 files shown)


libs/mock-block-dock/src/use-mock-block-props.ts
```diff
@@ -104,7 +106,11 @@ export const useMockBlockPropsNonTemporal = (
 
   const [readonly, setReadonly] = useDefaultState<boolean>(args.readonly);
 
-  const mockDatastore = useMockDatastoreNonTemporal(mockData, readonly);
+  const mockDatastore = useMockDatastoreNonTemporal(
+    mockData,
+    readonly,
+    args.simulateLatency,
+  );
```
Alfred-Mountfield, line 113: I've currently only simulated latency on Graph module callbacks. I'm not sure the best way to go about adding it to the Hook module or service module (or if its needed for those)

cc: @CiaranMn

nonparibus:
Shall we add an in-UI toggle for this? Fine to dump in top-right for now as "Simulate latency", and we can clean up later.

libs/mock-block-dock/dev/dev-app.tsx
```diff
@@ -158,6 +158,7 @@ const DevApp: FunctionComponent = () => {
           initialEntities: Object.values(blockEntityMap),
         }}
         temporal={false}
+        simulateLatency={{ min: 50, max: 200 }}
```
Alfred-Mountfield, line 161: @nonparibus, threading this to allow for inline comments.

> Shall we add an in-UI toggle for this? 

There are quite a few props that I think would be nice to configure through the UI, but it's a bit unclear to me when we do which at the moment. I haven't looked into it but it might be that at the moment they are either in the UI _or_ they are configurable through the caller of MBD (e.g. the Hub) rather than both. @CiaranMn is that the case? If it's in the UI I think it disappears in preview mode right, and therefore is managed a different way and doesn't go through the `MockBlockDock` component props.

> Fine to dump in top-right for now as "Simulate latency", and we can clean up later.

I think we'd definitely want the ranges to be configurable so it'd involve slightly more work than just a toggle.

If we want this I'd suggest a follow-up

libs/block-template-html/dev/index.html
```diff
@@ -55,8 +55,13 @@
             initialData=${initialData}
             blockEntityRecordId=${blockEntity.metadata.recordId}
             debug
+            simulateLatency=${{
+              min: 100,
+              max: 200,
+            }}
           />
         `;
+        // modify `simulateLatency` above to adjust the range of artificial latency (in ms)
```
CiaranMn, line 64:
```suggestion
        // modify `simulateLatency` above to adjust the range of artificial latency in responses to Graph module requests (in ms)
```

libs/mock-block-dock/src/use-mock-block-props.ts
Alfred-Mountfield: That was my reasoning too. Should I rename it to `simulateGraphLatency`? Or `simulateDatastoreLatency` or something?

I wasn't completely confident about the hook-module bit, I appreciate most of it should be the EA frontend managing things (where latency should be close to 0?) but I wasn't sure if there are still side-effects that could benefit from having an artificial slowdown somewhere between a block request and the EA

libs/mock-block-dock/dev/dev-app.tsx
CiaranMn: `readonly` is an example of a property which can be set via both prop and via in-UI toggle. "Preview mode" just controls whether the debug panel at the bottom appears or not, it doesn't affect the value of any properties set via it.

I think it's fine to wait, since we have a sensible default that there shouldn't be any reason to mess with often.

libs/mock-block-dock/src/datastore/use-mock-datastore/shared.ts
```diff
@@ -0,0 +1,28 @@
...
+import { randomNumberFromRange } from "../../util";
+
+type AnyAsyncFunction = (...args: any[]) => Promise<any>;
+
+export const useCallbackWithLatency = <T extends AnyAsyncFunction>(
+  callback: T,
+  deps: DependencyList,
+  simulateLatency?: { min: number; max: number },
+): T => {
+  // eslint-disable-next-line react-hooks/exhaustive-deps -- it's not smart enough to deal with the cast
```
CiaranMn, line 12: I'm not sure it's the cast – it's got no way of knowing what `callback` is doing and what the contents of `deps` is and whether the two are related at all.

I think the bigger issue though is that `useCallbackWithLatency` removes any checking of dependencies for any function passed into it, i.e. if you go and empty out a dependency array from `updateEntity`, there is no error.

I can think of two options to deal with this:

1. We keep the existing functions defined exactly as they are now (which keeps the checking of the dependency array), and then do something like `const updateEntityWithLatency = useCallbackWithLatency(...)` which is the function that gets returned from `use-mock-datastore`

2. Rather than latency being implemented by wrapping datastore functions, it's just a function that we call inside them, e.g.
```ts
  const sleepForLatency = useCallback(async () => {
    const waitFor = randomNumberFromRange(
      simulateLatency?.min ?? 0,
      simulateLatency?.max ?? 0,
    );
    return new Promise((resolve) => {
      setTimeout(resolve, waitFor);
    });
  }, [simulateLatency]);

  const queryEntities: GraphEmbedderMessageCallbacks["queryEntities"] =
    useCallback(
      async ({ data }) => {
        await sleepForLatency();
        // etc
```

libs/mock-block-dock/src/use-mock-block-props.ts
CiaranMn: The EA does update the entity properties in the datastore as it's editing it via the hook service, and it uses `updateEntity` [to do so](https://github.com/blockprotocol/blockprotocol/blob/db4cf8d422b881e52113aa52467d53115270e2b3/libs/mock-block-dock/src/hook-portals.tsx#L46) (just reminded myself of this) – so it should get whatever latency in that automatically.

Maybe `simulateDatastoreLatency`?

Alfred-Mountfield: Cool, and I've verified `updateEntity` comes from the actual `graphModuleCallbacks` as opposed to the underlying `updateEntity` implementation, so as you say it should get the latency automatically.

[Rename simulateLatency to simulateDatastoreLatency](https://github.com/blockprotocol/blockprotocol/pull/1161/commits/fc5af94920fec918061a06b0f3db7d0022f52041)

libs/mock-block-dock/src/mock-block-dock.tsx
```diff
@@ -590,6 +595,7 @@ const MockBlockDockTemporal: FunctionComponent<
  * @param [props.initialData.initialLinkedQueries] - The linkedQuery DEFINITIONS to include in the data store (results will be resolved automatically)
  * @param [props.readonly=false] whether the block should display in readonly mode or not
  * @param [props.serviceModuleCallbacks] overrides the default service module callbacks
+ * @param {{min: number, max: number}} [props.simulateLatency] simulate latency in responses where each delay is randomly chosen from the supplied range
```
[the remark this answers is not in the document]
Alfred-Mountfield: [Update latency docs](https://github.com/blockprotocol/blockprotocol/pull/1161/commits/f0f30cb1c0b048874a1fb8ec1496dcc68d701698)

libs/block-template-html/dev/index.html
Alfred-Mountfield: [Update latency docs](https://github.com/blockprotocol/blockprotocol/pull/1161/commits/f0f30cb1c0b048874a1fb8ec1496dcc68d701698)

libs/block-template-react/src/dev.tsx
```diff
@@ -43,6 +43,11 @@ const DevApp = () => {
       initialData={{
         initialEntities: [testEntity],
       }}
+      simulateLatency={{
+        // configure this to adjust the range of artificial latency (in ms)
```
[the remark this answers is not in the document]
Alfred-Mountfield: [Update latency docs](https://github.com/blockprotocol/blockprotocol/pull/1161/commits/f0f30cb1c0b048874a1fb8ec1496dcc68d701698)

libs/block-template-custom-element/src/dev.tsx
```diff
@@ -55,6 +55,11 @@ const DevApp = () => {
       initialData={{
         initialEntities: [testEntity],
       }}
+      simulateLatency={{
+        // configure this to adjust the range of artificial latency (in ms)
```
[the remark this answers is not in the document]
Alfred-Mountfield: [Update latency docs](https://github.com/blockprotocol/blockprotocol/pull/1161/commits/f0f30cb1c0b048874a1fb8ec1496dcc68d701698)

libs/mock-block-dock/src/datastore/use-mock-datastore/shared.ts
Alfred-Mountfield: > I'm not sure it's the cast – it's got no way of knowing what callback is doing and what the contents of deps is and whether the two are related at all.

If you remove the cast it actually does stop complaining about it and only complains about the fact that we're spreading `deps`.

With cast:
image

Without cast:
image
image


I agree with the general point that it doesn't know what's inside callback, but my thought was that's fine honestly because the dependencies in _this_ function are obvious and we can ensure them ourselves without relying on eslint. Unfortunately you've now made me realise that this hook is pretty dangerous because we're not getting the `deps` checked in the callers, because the eslint rule looks for `useCallback`, so it doesn't check the parameters of `useCallbackWithLatency`.
image

As such I'm going to go for the plain approach (which is what I started with) that you suggest as (2). My only concern with it is how easy it would be to miss calling `sleepForLatency`, because we've got to manually call it in every callback which feels uncomfortable.

[Call the wait function manually in every callback](https://github.com/blockprotocol/blockprotocol/pull/1161/commits/4ff8d5f2ac42a92efc59df9297fb0e727aa9a5ef)

libs/mock-block-dock/src/datastore/use-mock-datastore/shared.ts
```diff
@@ -0,0 +1,28 @@
...
+      if (simulateLatency) {
+        const waitFor = randomNumberFromRange(
+          simulateLatency.min,
+          simulateLatency.max,
+        );
+        await new Promise((resolve) => {
+          setTimeout(resolve, waitFor);
+        });
+      }
+      return await callback(...args);
```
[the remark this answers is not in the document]
Alfred-Mountfield: (now redundant)

CiaranMn (approved):

PR #1162, opened 2023-03-13
Remove Mock Block Dock CommonJS build output

CiaranMn:

We were building `mock-block-dock` for both CommonJS and ESM output, but the CommonJS output doesn't work because it has an ESM dependency ,`@lit-labs/react`.

We could fix this by using some other bundler, but since MBD is intended as a local development tool rather than being bundled up in other apps, it doesn't seem worth it, and just outputting ESM simplifies things.

This PR is a byproduct of trying to figure out why the Block Hub doesn't work in incognito, which seems to be something to do with esm.sh (since just importing MBD directly into the Hub works – it's only when it's loaded via esm.sh that it fails).

### DIFF
--- a/.changeset/thirty-coats-smell.md
+++ b/.changeset/thirty-coats-smell.md
@@ -1,0 +1,5 @@
+---
+"mock-block-dock": patch
+---
+
+remove (broken) commonjs build output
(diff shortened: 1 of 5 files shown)


Alfred-Mountfield (commented):
Lint is broken 🙃

Alfred-Mountfield (approved):

PR #1169, opened 2023-03-14
Move 'can block call services' decision from session storage to component state

CiaranMn:

Implement [this PR suggestion](https://github.com/blockprotocol/blockprotocol/pull/1138#discussion_r1129265623) to not persist a user's decision as to whether a block in the hub can call external services in session storage, as it may confuse users if they revisit the block and it doesn't work – instead, store it in component state that resets every time they revisit the block's page.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203623179643290/1204138539312008/f) _(internal)_

## ❓ How to test this?

1.  Visit e.g. the AI text block in the preview deployment, use it, see modal, make choice
2. Visit another block, revisit AI text block, see modal again

### DIFF
--- a/apps/site/src/components/pages/hub/block-data-container/sandboxed-block-demo.tsx
+++ b/apps/site/src/components/pages/hub/block-data-container/sandboxed-block-demo.tsx
@@ -33,9 +33,8 @@ export const SandboxedBlockDemo: FunctionComponent<SandboxedBlockProps> = ({
   const loadedRef = useRef(false);
   const iframeRef = useRef<HTMLIFrameElement>(null);
 
-  const [serviceUsageIsPermitted, setServiceUsageIsPermitted] = useState<
-    boolean | undefined
-  >(undefined);
+  const [serviceUsageIsPermitted, setServiceUsageIsPermitted] =
+    useState<boolean>();
 
   const [serviceModuleMessage, setServiceModuleMessage] =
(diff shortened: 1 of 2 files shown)


apps/site/src/components/pages/hub/block-data-container/sandboxed-block-demo.tsx
```diff
@@ -53,6 +33,10 @@ export const SandboxedBlockDemo: FunctionComponent<SandboxedBlockProps> = ({
   const loadedRef = useRef(false);
   const iframeRef = useRef<HTMLIFrameElement>(null);
 
+  const [serviceUsageIsPermitted, setServiceUsageIsPermitted] = useState<
+    boolean | undefined
+  >(undefined);
```
benwerner01, line 38: Nit: this is equivalent
```suggestion
  const [serviceUsageIsPermitted, setServiceUsageIsPermitted] =
    useState<boolean>();
```

benwerner01 (dismissed):
Thanks Ciaran, working as expected 👍

apps/site/src/components/pages/hub/block-data-container/sandboxed-block-demo.tsx
CiaranMn: Fair enough, updated

PR #1182, opened 2023-03-16
Fix nested folder creation for codegen output

CiaranMn:

Fixes creation of an `outputFolder` in codegen where it is a nested folder and one or more intermediary folders also does not exist.

### DIFF
--- a/.changeset/sixty-eels-thank.md
+++ b/.changeset/sixty-eels-thank.md
@@ -1,0 +1,5 @@
+---
+"@blockprotocol/type-system": patch
+---
+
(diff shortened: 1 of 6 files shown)


Alfred-Mountfield (dismissed):
Dangit. 🚀

Alfred-Mountfield (approved):

Alfred-Mountfield (changes requested):
Wait did we mean to change the react template or is that a polluted commit

Alfred-Mountfield (approved):

PR #1184, opened 2023-03-16
Test that codegen runs successfully in block templates in CI

CiaranMn:

Adds a test to the `CI` workflow that checks that `yarn codegen` in the published block template packages runs without error.

This only checks that `yarn codegen` runs successfully _for the templates_ using__ the _code in the repo_. If there is a discrepancy between what gets published to npm and what is in the repo (e.g. some dependency does not get published) then errors may occur in the wild that are not checked by this test. We need another test in the Version Packages PR which uses npm-published packages where they are not being bumped, and local code where they are being bumnped

## ❓ How to test this?

1.  ~The CI step should fail until #1182 is merged. This verifies it is catching something we weren't previously. It should then pass, or highlight further issues.~ It does not fail because it's using local code for all packages, and one of the issues with #1182 is fixing was that one package wasn't published to npm


libs/mock-block-dock/src/version.ts
```diff
@@ -1 +1 @@
-export const MOCK_BLOCK_DOCK_VERSION = "0.1.1";
+export const MOCK_BLOCK_DOCK_VERSION = "0.1.2";
```
Alfred-Mountfield, line 1: Does this only happen after a PR or something? I seem to have applied this as a drive-by once or twice as well and I'm unsure if we're just missing it in linting or

Alfred-Mountfield (dismissed):

libs/mock-block-dock/src/version.ts
CiaranMn: This is a `codegen` in `mock-block-dock`. I'm not sure why it doesn't get run when the MBD version is _actually_ updated. It seems to just happen in branches afterwards.

Alfred-Mountfield: Is it to do with `changesets` managing our version bumps but probably not committing the results of codegen?

CiaranMn: Oh, yes, that explains it

Alfred-Mountfield (approved):

PR #1181, opened 2023-03-16
Improve WordPress plugin error handling and messages, set version to 0.0.4

CiaranMn:
## 🌟 What is the purpose of this PR?

Improved WordPress plugin error handling and bug fixes for the `0.0.4` release:
1. Add `Toast` component for better user-facing errors than default WordPress admin notices / snackbar: deduplicated and visible wherever user is scrolled to in page
2. Show toast when entities cannot be created or retrieved (rather than error being in browser console only)
3. Fix a bug in 'find entity location' logic by retrieving all pages to look in (rather than the default 5 pages)
4. Don't show block in selector if DB version is unsupported
5. Add label to custom block controls in settings sidebar
6. Set a shorter db connection timeout in Block Protocol API (we are seeing intermittent errors in connecting to the db, and the default timeout is 30s)
7. Reduce timeout in plugin for connecting to the BP API from 10s to 5s (if 5s is reached the API is definitely not going to respond successfully)

Also sets the necessary metadata for the `0.0.4` release, and documents the publishing process.

## 🔗 Related links

- [Asana task for better error toasts](https://app.asana.com/0/1203358502199087/1204145931939225/f) _(internal)_
- [Asana task for custom block setting labelling](https://app.asana.com/0/1203358502199087/1204182825551978/f) _(internal)_

## ⚠️ Known issues

- Intermittent timeouts in connection to MongoDB – no luck in finding cause so far (sometimes none a day, sometimes a few).

## ❓ How to test this?

1.  Failed entity retrieval can be tested by adding a block to a page and then deleting the entity it is associated with in the db and refreshing
2. External service calls can be tested by hitting a usage limit (easiest to do by retrieving static maps repeatedly) and then trying another request
3. Block settings panel can be checked by selecting a BP block and looking at righthand settings sidebar

## 📹 Demo

https://user-images.githubusercontent.com/37743469/225738021-c5603866-063e-43f7-ba41-31203b5199df.mp4

PR #1188, opened 2023-03-16
Add client-side Sentry React instrumentation to Wordpress Plugin

thehabbos007:

**Context:** we're trying to debug a number of user errors that are proving elusive.

This PR adds client-side React instrumentation for React to the admin view of post editing. The plugin aims to be a good citizen on the installation by not overwriting existing configurations if there are any.

The Sentry configuration includes console error/warn/assert logging so we can track bugs. The reporting only happens for JS files matching the `plugins/blockprotocol` regex.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778935/1204197991701675/f) _(internal)_

## 🔍 What does this change?

- Changes how Sentry is configured to extract common settings for client/server-side
- Conditionally add Sentry reporting on the admin editing view
- Add new webpack entry for `sentry`
- Add new DSN for client side sentry along with its config

## ⚠️ Known issues

- We are unable to capture errors if other plugins/scripts take over the sentry instance - differently from the WP plugin where we manually instruments some errors.

## 🐾 Next steps

Release after #1181 is merged

## ❓ How to test this?

1.  Checkout the branch, start WP instance with newly built JS. 
1.  Configure BP plugin with an API key
1.  Go to a post to edit it, issue a `console.error("Boom!")` in the console of the post editor page
1.  Confirm that event was tracked in sentry.

You can also trigger an error from react by adding 
```tsx
      <button onClick={() => methodDoesNotExist()}>Break the world</button>;
```
to the `edit.tsx` return and click the button after creating a block. Sentry should also pick this up.

## 📹 Demo

<details><summary>Running the above test steps</summary>

image

image

</details>

<details><summary>Screenshots of how Sentry behaves with enabling/disabling telemetry</summary>

image

image

</details>

I've also tested the plugin with another JS-side WP Sentry extension. Our config is simply overwritten. We do try not to create a sentry instance if we can see there's already one.


libs/wordpress-plugin/plugin/trunk/block-protocol.php
```diff
@@ -212,6 +212,22 @@ function block_protocol_init()
     return;
   }
 
+	// register and enqueue the sentry plugin if plugin analytics are enabled
```
CiaranMn, line 215: We should probably move this into the `block_protocol_editor_assets` function below because that's where the other client-side JS gets loaded – we don't need to load Sentry on non-editor pages (e.g. settings page) because we don't load any `blockprotocol` JS there. It can also go below the `get_block_protocol_blocks` check because they won't even be able to select blocks if it's failed

thehabbos007: Trying to test this change, but seems Sentry is being a little slow right now

CiaranMn (approved):
Great, thank you!

PR #1189, opened 2023-03-16
Fix Hub menu responsiveness

luisbettencourt:

This PR fixes the mobile view in the hub page.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056186/1204177293864932/f) _(internal)_

## 🔍 What does this change?

- Changed the tabs to use MUI Tabs, so we can easily change the vertical orientation to horizontal on mobile;
- Made it so the list and menu are full width in mobile.

## 📹 Demo

Before:
image

After:
image


nonparibus (dismissed):
This looks much better. Please can we make sure when stacked vertically that the spacing between the "Types" and "Blocks" tabs matches the Figma? ([internal link](https://www.figma.com/file/Fpmde3E9nJMb4VLkPTYgrM/blockprotocol.org?node-id=3703%3A23502&t=ZTNm2XfBGfBaEKE5-1)) There's quite a bit more whitespace expected than is present. _Reminder: **Services** will be added as a third tab shortly, but does not need adding as part of this PR._

CiaranMn (approved):

PR #1191, opened 2023-03-17
Silence .env warnings, remove blockEntityType from metadata

CiaranMn:

A couple of tweaks to the output of `block-scripts`:
1. Silence warning if no `.env` file is present – otherwise users see `Failed to load ./.env` every time they run `yarn dev` if they aren't using a `.env` file, which might seem like a problem 
2. Don't include the `blockEntityType` property from `package.json[blockprotocol]` in `block-metadata.json` – its value should be taken for `schema` in that file, rather than appearing twice under different keys

### DIFF
--- a/libs/block-scripts/shared/generate-dist-block-metadata.js
+++ b/libs/block-scripts/shared/generate-dist-block-metadata.js
@@ -27,6 +27,9 @@ export const generateDistBlockMetadata = async (extra) => {
     ? await fs.readJson(variantsJsonPath)
     : undefined;
 
+  const schema = blockprotocol.blockEntityType;
+  delete blockprotocol.blockEntityType;
+
   const blockMetadata = {
     name,
     version,
@@ -36,9 +39,9 @@ export const generateDistBlockMetadata = async (extra) => {
     license,
     externals: peerDependencies,
     variants,
+    schema,
     ...blockprotocol,
     ...extra,
-    schema: blockprotocol.blockEntityType,
   };
 
   return writeFormattedJson(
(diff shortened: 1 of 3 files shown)


libs/block-scripts/shared/generate-dist-block-metadata.js
```diff
@@ -27,6 +27,9 @@ export const generateDistBlockMetadata = async (extra) => {
     ? await fs.readJson(variantsJsonPath)
     : undefined;
 
+  const schema = blockprotocol.blockEntityType;
```
Alfred-Mountfield, line 30: Interesting, I'd actually thought about this and forgot to ask.

Do we maybe just want to stop spreading `...blockprotocol` and be explicit about what we do want to appear in `block-metadata.json`?

Unless we validate it doesn't have extra properties, users can put whatever they want in the object

Alfred-Mountfield (approved):
One idea, but happy to merge as is

libs/block-scripts/shared/generate-dist-block-metadata.js
CiaranMn: We could do, but they can put what they want in `block-metadata.json` after it's generated anyway. We only take the properties we want when populating the metadata that we save when a block is published.

Alfred-Mountfield: I guess it just feels a little strange to single out this specific property to clean-up. But I appreciate that's because it's *guaranteed* to be messy every time

CiaranMn: Yes – this is something we have introduced. I'm happy to update to remove the spread when we're next updating this but would rather not hold it up for that as we'd need to be careful we got it right

PR #1192, opened 2023-03-17
Allow "free" users to update payment method

benwerner01:

This PR enables anyone with an active stripe subscription (including those with a "free" subscription tier) to edit their payment method.

It also fixes an issue where a runtime error is thrown for users with an inactive stripe subscription in the `PaymentHistorySection` component.

## 🔗 Related links

- [Update BP site to let user change payment method when they have a "free" account](https://app.asana.com/0/1204000740778935/1204188703131678/f) _(internal)_
- [Catch errors when a user without a subscription loads billing page](https://app.asana.com/0/1204000740778938/1204173853214856/f) _(internal)_

## ⚠️ Known issues

- when a user has an active stripe subscription on the "free" tier, they are not able to use the existing upgrade subscription flow ([asana task](https://app.asana.com/0/0/1204188703131699/f))

## 🛡 What tests cover this?

Manual testing.

## ❓ How to test this?

1. Checkout the branch and go to the billing page as new user with an inactive stripe subscription
2. Ensure no runtime error is thrown on the billing page
3. Upgrade the user to the "hobby" or "pro" tier using test stripe card details
4. Go to the stripe dashboard, and update the user's subscription manually to a "free" tier by removing the "hobby" tier subscription item and adding the "free" tier subscription item to the stripe subscription (you can watch me do this in the demo video of [this PR](https://github.com/hashintel/internal-api/pull/86))
5. Go back to the billing page, and observe that the user is on the "free" tier and is able to update their payment information

## 📹 Demo

image

### DIFF
--- a/apps/site/src/pages/settings/billing-settings-panel/billing-overview.tsx
+++ b/apps/site/src/pages/settings/billing-settings-panel/billing-overview.tsx
@@ -74,6 +74,10 @@ export const BillingOverviewPanelPage: FunctionComponent = () => {
     [user],
   );
 
+  const subscriptionStatusIsActiveOrPastDue =
+    stripeSubscriptionStatus === "active" ||
+    stripeSubscriptionStatus === "past_due";
+
   const currentSubscriptionTierIsPaid = isPaidSubscriptionTier(
     currentSubscriptionTier,
   );
@@ -122,12 +126,12 @@ export const BillingOverviewPanelPage: FunctionComponent = () => {
         <Grid item md={6} sm={12} sx={{ position: "relative" }}>
           <Link
             href={
-              currentSubscriptionTierIsPaid
+              subscriptionStatusIsActiveOrPastDue
                 ? paymentMethodsPanelPageAsPath
                 : "#"
             }
             onClick={(event) => {
-              if (!currentSubscriptionTierIsPaid) {
+              if (!subscriptionStatusIsActiveOrPastDue) {
                 event.preventDefault();
                 /**
                  * @todo: figure out why calling `scrollIntoView` directly results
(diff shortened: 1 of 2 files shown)


benwerner01:
@nonparibus 

> 1. Does this mean the tile that previously showed where the "Change payment method" tile now sits for free users will never be shown?

For `"free"` users with an active stripe subscription, yes. This is to allow `"free"` users who are getting charged for external service overage usage have a way of managing their payment methods. As far as I'm aware, there isn't currently another way of accessing the "Payment Methods" panel.

> 2 - What does the dashboard look like for a user with an active free account, but no card attached?

If you're referring to what the billing dashboard looks like for a `"free"` user that has an inactive stripe subscription, and therefore has not cards attached to their account, then the "Upgrade to get more" button is displayed as it was previously.

If you're referring to a `"free"` user that has an active stripe subscription, that requires them to have payment methods associated with their account. So they should never be in a state where they don't have a card attached to their stripe subscription. 

> 3 - ...and how do those users add a card in the first place? Solely via the upgrade flow?

Currently there is no flow for users to set a payment method before they have upgraded to the `"hobby"` or `"pro"` subscription tier, as the stripe subscription is created when they first subscribe to a paid subscription tier. Therefore currently the only way to add a payment method to your account would be to subscribe to a paid subscription tier, or to use the "Payment Method" panel to add other payment methods once the user has an active stripe subscription.

I'm aware that we will likely want to extend the behaviour so that users are able to start paying for their overage use _before_ subscribing to a paid subscription tier. As discussed in [this prior thread](https://github.com/hashintel/internal-api/pull/86#issuecomment-1473528899), this could be done by creating an active stripe subscription for all users. It could also be done as part of some other flow, where the user adds a payment method for the first time. We could repurpose the existing "Payment Method" panel for that, or create some other dedicated flow.

If you'd like I can create a follow-up ticket for this, but I'm not sure how highly this behaviour ranks in the other list of priorities currently assigned to me.

luisbettencourt (approved):

PR #1198, opened 2023-03-18
Add description editing to the entity type editor

luisbettencourt:

This PR adds the functionality to allow editing the description of an entity type.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1204080201525391/f) _(internal)_

## 🚫 Blocked by

- [x] https://github.com/hashintel/hash/pull/2231

## 🔍 What does this change?

- Added `description` to the type-editor form context in https://github.com/hashintel/hash/pull/2231;
- Added Input controlled by `react-hook-form` to allow editing the description of the entity type;
- Added a collapse animation to the edit bar.

## ⚠️ Known issues

The Input field is not perfect and I think we want a similar approach to the `EditableField` component that is being used in the `How-To` and `Address` blocks. That component is currently in the hashintel design-system and is going to be moved to the blockprotocol design-system. When it's added we should also use it here.

## 📹 Demo

demo


nathggns:
There is UX issue with the edit bar in Block Protocol which we fixed in HASH in https://github.com/hashintel/hash/pull/1439 which is that the whole page shifts when you make a change as the edit bar appears. Not strictly related to this PR, but we should fix it at some point. Ticket tracking it at https://app.asana.com/0/0/1204270465466589/f

nathggns (dismissed):

nonparibus (approved):

nathggns (approved):

PR #1199, opened 2023-03-18
Use `just` over `cargo-make`

TimDiekmann:

We want to switch away from `cargo-make` due to several reasons. The most important reason is that the `Makefile.toml` is really hard to read and the output is pretty noisy.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778938/1204207858039652/f) _(internal)_
- [just.systems](https://just.systems)

## 🔍 What does this change?

- port over and slightly adjust the jobs from `Makefile.toml` to `.justfile`

The readme was adjusted

## 🐾 Next steps

- Generate SARIF files and upload them to GitHub

## 🛡 What tests cover this?

- The CI will use the `just` to build the `wasm` package

## ❓ How to test this?

Run `yarn install` and execute the steps defined in the readme

## 📹 Demo

When just invoking `just`, this is the current output
```
Usage: just <recipe>
Available recipes:
    build                         # Builds all configured targets of the workspace
    format                        # Runs auto-formatter on the type system crate
    lint                          # Runs linters on the type system package
    test                          # Runs tests on the workspace
    doc                           # Creates the documentation for the workspace
For further information, run 'just --help'
```


libs/@blockprotocol/type-system/crate/README.md
```diff
@@ -13,10 +13,10 @@
```
Alfred-Mountfield, line 13: Can't quite comment on it, but do we require `cargo-make` at all any more?

Alfred-Mountfield (commented):
Not against us trying this out, especially because `cargo make` was causing us problems.

I'm just not sure I'm convinced that `.justfile`s _are_ much more readable.. Especially when they're mostly this simple, I'm wondering if we should just use `package.json` `scripts` like we do for most other things in the repo. It'd be more consistent.

Can approve after the comments are addressed, as I'm fine with us testing it out, but perhaps we should consider the above.

libs/@blockprotocol/type-system/crate/README.md
TimDiekmann: I removed it from the action, thanks for flagging

PR #1207, opened 2023-03-21
Edit bar improvements in entity type editor 

nathggns:

The Edit Bar, which displays when you have pending changes to an entity type, can appear out of view. This PR makes it sticky, so it scrolls with the viewport. It also ensures the header doesn't hide when on the entity type editor, to make it easier to know where to position the edit bar, and associated visual changes.

**Importantly, it also changes the URL management for entity types from a push to a replace**, and removes the scroll up. This is to prevent the scroll up when publishing new versions, which wasn't evident before as you had to be scrolled up to access the edit bar anyway. This also prevents history being polluted with previous versions, and matches the behaviour in HASH. There is a minor chance replace may be used instead of push when that isn't desirable – i.e, if you've changed to a whole new entity type instead of just a new version, but that isn't possible in the current UI. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1203979552136012/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- The header now always displays on the type editor
- URL is now updated using replace instead of push, with scroll disabled. The reasons for this are explained above.
- Removed a redundant call to router.replace when publishing the initial version of entity types. 
- Edit bar is now sticky below the header, and its contents are centered to match the header

No

## ⚠️ Known issues

N/A

## 🐾 Next steps

We may want to make the edit bar animate in, as in HASH. 

## 🛡 What tests cover this?

None

## ❓ How to test this?

Visual approval

---
  - https://app.asana.com/0/0/1204203279487398

### DIFF
--- a/apps/site/src/components/navbar.tsx
+++ b/apps/site/src/components/navbar.tsx
@@ -250,10 +250,11 @@ const Navbar: FunctionComponent<
 
   const isHomePage = generatePathWithoutParams(hydrationFriendlyAsPath) === "/";
   const isDocs = hydrationFriendlyAsPath.startsWith("/docs");
+  const isTypeEditor = hydrationFriendlyAsPath.includes("/types/entity-type/");
   const [mobileNavVisible, setMobileNavVisible] = useMobileNavVisible();
 
   const { scrolledPast, isNavbarHidden } = useScrollingNavbar(
-    isDocs,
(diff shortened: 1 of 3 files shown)


yusufkinatas (approved):
LGTM 👍

Issue #1166, opened 2023-03-14
[BP-3] Fix entity type table display on mobile


CiaranMn:
Users can create types, and the detail of those types is shown in two tables (one for properties on the type, and one for links from the type) – for example at [https://blockprotocol.org/@hash/types/entity-type/address-block/v/2](https://blockprotocol.org/@hash/types/entity-type/address-block/v/2)

This view is currently broken on mobile, because the tables are too wide for the screen. The goal is to implement a working mobile view for this page.

The form itself is generated from a library, and its internal elements should not be styled as part of this solution. One potential and acceptable solution is to wrap the tables with something with a max width, allow a horizontal scrollbar to appear, meaning the user can swipe / scroll to see more of the tables.

Closed by [https://github.com/blockprotocol/blockprotocol/pull/1210](https://github.com/blockprotocol/blockprotocol/pull/1210)

<sub>From [SyncLinear.com](https://synclinear.com) | [BP-3](https://linear.app/hash/issue/BP-3/fix-entity-type-table-display-on-mobile)</sub>

PR #1210, opened 2023-03-22
[BP-3] Fix entity type table display on mobile

gitstart-blockprotocol:
🌟 What is the purpose of this PR?
---
This PR fixes the entity type table display on mobile

🔗  Related links
---
- [Github](https://github.com/blockprotocol/blockprotocol/issues/1166)

❓How to test this?
---
https://www.loom.com/share/171b3fd67e0744b98bfd3dce00c8fe04

---
This code was written and reviewed by GitStart Community. Growing future engineers, one PR at a time.

---
  - https://app.asana.com/0/0/1204147538988935

### DIFF
--- a/apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx
+++ b/apps/site/src/pages/[shortname]/types/entity-type/[...slug-maybe-version].page.tsx
@@ -382,7 +382,7 @@ const EntityTypePage: NextPage = () => {
                 </Box>
               ) : null}
 
-              <Box component="section">
+              <Box component="section" sx={{ overflowX: "auto" }}>
                 <EntityTypeForm
                   author={shortname as `@${string}`}
                   entityType={entityType}
--- a/apps/site/src/pages/_app.page.tsx
+++ b/apps/site/src/pages/_app.page.tsx
@@ -1,5 +1,6 @@
 import "../styles/nprogress.css";
 import "../styles/prism.css";
+import "../styles/table-component.css";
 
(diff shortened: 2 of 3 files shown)


nathggns:
Looking at this now

nathggns (approved):

nonparibus:
@gitstart-blockprotocol Just to flag that this solution ignored the instructions [in the original ticket](https://github.com/blockprotocol/blockprotocol/issues/1166), and went about the fix by way of modifying styling within a class in the type editor that we had instructed not to (which is used all over the place in the app). The consequence of this PR merging was broken styles across several parts of our site that we're now fixing in follow-up.

This solution should not originally have been approved, which was our oversight (as well as highlights the deficiencies of our lack of testing on the public site, as well as poor separation of concerns around styling). cc @nathggns (fix is in https://github.com/blockprotocol/blockprotocol/pull/1228)

gitstart-blockprotocol:
> @gitstart-blockprotocol Just to flag that this solution ignored the instructions [in the original ticket](https://github.com/blockprotocol/blockprotocol/issues/1166), and went about the fix by way of modifying styling within a class in the type editor that we had instructed not to (which is used all over the place in the app). The consequence of this PR merging was broken styles across several parts of our site that we're now fixing in follow-up.
> 
> This solution should not originally have been approved, which was our oversight (as well as highlights the deficiencies of our lack of testing on the public site, as well as poor separation of concerns around styling). cc @nathggns (fix is in #1228)

Sorry about this @nonparibus, this was an oversight from our side and we'll ensure this wouldn't happen again!

PR #1215, opened 2023-03-24
Fix schema path in `entity.json`

TimDiekmann:

We accidentally used `versioned-url` instead of `base-url`. This fixes it.

### DIFF
--- a/apps/site/public/types/modules/graph/0.3/schema/entity.json
+++ b/apps/site/public/types/modules/graph/0.3/schema/entity.json
@@ -8,7 +8,7 @@
       "type": "object",
       "$comment": "This should be constrained by the associated entity type",
       "propertyNames": {
-        "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/versioned-url"
+        "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/base-url"
       }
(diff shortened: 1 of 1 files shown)


Alfred-Mountfield (approved):
It's a shame the Jetbrains JSON Schema integration doesn't seem to work for the newer drafts of JSON schema. We might have picked this up more easily if we'd looked at an entity with this metaschema.

I'm working on a little JSON schema dev-env right now anyway, maybe I can find a way to make it robust and we can use it as part of testing in CI or something

TimDiekmann:
> It's a shame the Jetbrains JSON Schema integration doesn't seem to work for the newer drafts of JSON schema. We might have picked this up more easily if we'd looked at an entity with this metaschema.

I don't see how any validator would be able to pick this up, both types have almost the same schema, only the `$id` field, and the description is different. Both are required to be a `uri`.

Alfred-Mountfield:
> I don't see how any validator would be able to pick this up, both types have almost the same schema, only the $id field, and the description is different. Both are required to be a uri.

That's good point I was thinking about the template string in TypeScript which we have for `VersionedUrl` which enforces the format, I guess we don't have that pattern on the versionedUrl metaschema at the moment

PR #1217, opened 2023-03-24
Update GitHub labeler

nonparibus:
## Changes

- Removes deprecated label (`area: hub`)
- Renames `area: docs` to `area: copy` and updates matching rules around this
- Implements various new rules including automatic tagging of testing-related changes as `area: testing`

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,8 +5,8 @@
   - "apps/**"
   - "!apps/site/src/_pages/**"
 
-"area: copy":
-  - "apps/site/src/_pages/**"
+"area: blocks":
+  - "blocks/**"
 
 "area: dependencies":
   - "**/patches/**"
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -2,18 +2,15 @@
   - "apps/site/**"
 
 "area: apps":
-  - "apps/**"
+  - ["apps/**", "!apps/site/src/_pages/**"]
```
TimDiekmann, line 5: I'm unsure about this [syntax](https://github.com/actions/labeler#usage), but needs this to be
```suggestion
  - all: ["apps/**", "!apps/site/src/_pages/**"]
```
or
```suggestion
  - "apps/**"
  - "!apps/site/src/_pages/**"
```
?
I think the latter is the one we want: Trigger in `apps/**` but does not trigger by `apps/site/src/_pages/**`

nonparibus: Let's try it

TimDiekmann (dismissed):

TimDiekmann (approved):
Pushed a prettier-fix

PR #1222, opened 2023-03-27
Add better compatibility for older MySQL versions

thehabbos007:

This PR adds a "native" PHP-based graph traversal function that can be dynamically used at runtime if the database is at version (MySQL 8+|MariaDB 10.2.2+)

I haven't changes the minimum version of MariaDB in this PR as `10.2.7` introduced JSON, at `10.2.2` we have migration errors which hopefully can be addressed/adjusted in a follow-up.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204000740778938/1204217937089582/f) _(internal)_

## 🔍 What does this change?

- Add new manual graph traversal in PHP
- Dynamically decide what method to use at runtime
- Change version checks in the code to allow checking versions ad-hoc per feature

- The docs can now specify supporting a lower version of MySQL for general usage of the app (Have adjusted)

## ⚠️ Known issues

- MariaDB version requirements are a little unclear at the moment, ideally we'd be able to use `longtext` instead of `json` there to allow older version support, but we need to see what we can do about that.

## 🐾 Next steps

- [Implement queryEntities for MariaDB - Asana task](https://app.asana.com/0/1203179076056186/1204069269252814/f)

## 🛡 What tests cover this?

- Manual test, see video demo below

## ❓ How to test this?

I recommend doing a full end-to-end test with this change. This includes:
- Setting the MySQL version to 5.7.9 (notice not 5.7.8 because 5.7.8 cannot be upgraded to version 8.x directly)
- Setting up the WP instance as usual, configuring an API key and creating 3 code blocks
- Manually adding a relationship in the DB that is something like `code block 1 <- code block 2 -> code block 3`
- Validate the response of `GET /blockprotocol/entities/*CODE BLOCK 1 ENTITY ID*` with 1 depths returns this little subgraph
- Upgrade the MySQL version in the docker-compose to `8` and restart the instance
- Validate that the page still loads and that the above response is identical.

You can also test with changing the `has_right_outgoing` depth to be `0` and only return the link entity

## 📹 Demo

https://user-images.githubusercontent.com/6988668/227957971-2c430c50-2623-4a5b-a64b-5401f9e55075.mp4

### DIFF
--- a/libs/wordpress-plugin/README.md
+++ b/libs/wordpress-plugin/README.md
@@ -108,7 +108,7 @@ Within `trunk` there are:
 
 - **WordPress:** you must be running at least WordPress 5.6 (due to Composer autoload)
 - **PHP 7.4 or later:** your server must be running at least PHP 7.4. PHP 8.0 and 8.1 are also supported.
-- **MySQL 5.7.8+ _or_ MariaDB 10.2.7+:** your database must be using at least MySQL 5.7.8. Be aware, the previous version of MySQL (5.6 and 5.7) respectively reach(ed) end of life in February 2021 and October 2023. You should upgrade to MySQL 8 now to continue to receive security updates, as well as to use the Block Protocol within WordPress. As for MariaDB, although the plugin would work with an old version, we strongly recommend using the newest version available to you.
+- **MySQL 5.7.8+ _or_ MariaDB 10.2.7+:** your database must be using at least MySQL 5.7.8. Be aware, the previous version of MySQL (5.6 and 5.7) respectively reach(ed) end of life in February 2021 and October 2023. You should upgrade to MySQL 8 now to continue to receive security updates, as well as to get the best Block Protocol experience within WordPress. As for MariaDB, although the plugin would work with an older version, we strongly recommend using the newest version available to you.
 - **HTTPS:** your webhost must support HTTPS in order for the Block Protocol to properly function.
 
 To check what your WordPress instance supports, please navigate to `Admin -> Tools -> Site Health -> Info` and then click into either `Server` (for PHP) or `Database` (for MySQL/MariaDB).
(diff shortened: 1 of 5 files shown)


libs/wordpress-plugin/plugin/trunk/server/block-db-table.php
```diff
@@ -65,7 +83,7 @@ function block_protocol_migrate()
 
   if ($saved_version < 2) {
     block_protocol_migration_1();
-    update_site_option('block_protocol_db_migration_version', 2);
+    block_protocol_set_migration_version_to(2);
```
thehabbos007, line 86: We previously set the migration version regardless of success. We now check if there's a DB failure before setting the number.

libs/wordpress-plugin/README.md
```diff
@@ -108,7 +108,7 @@ Within `trunk` there are:
...
 - **WordPress:** you must be running at least WordPress 5.6 (due to Composer autoload)
 - **PHP 7.4 or later:** your server must be running at least PHP 7.4. PHP 8.0 and 8.1 are also supported.
-- **MySQL 8 _or_ MariaDB 10.2.7+:** your database must be using at least MySQL 8 (due to use of recursive CTE queries). Be aware, the previous version of MySQL (5.6 and 5.7) respectively reach(ed) end of life in February 2021 and October 2023. You should upgrade to MySQL 8 now to continue to receive security updates, as well as to use the Block Protocol within WordPress. As for MariaDB, although the plugin would work with an old version, we strongly recommend using the newest version available to you.
+- **MySQL 5.7.8+ _or_ MariaDB 10.2.7+:** your database must be using at least MySQL 5.7.8. Be aware, the previous version of MySQL (5.6 and 5.7) respectively reach(ed) end of life in February 2021 and October 2023. You should upgrade to MySQL 8 now to continue to receive security updates, as well as to get the best Block Protocol experience within WordPress. As for MariaDB, although the plugin would work with an older version, we strongly recommend using the newest version available to you.
```
nonparibus, line 111:
```suggestion
- **MySQL 5.7.8+ _or_ MariaDB 10.2.7+:** your database must be using at least MySQL 5.7.8. Be aware, MySQL 5.6 and 5.7 respectively reach(ed) end of life in February 2021 and October 2023. You should upgrade to MySQL 8 now to continue to receive security updates, as well as to get the best Block Protocol experience within WordPress. As for MariaDB, although the plugin would work with an older version, we strongly recommend using the newest version available to you.
```

CiaranMn (approved):

PR #1223, opened 2023-03-27
Improve usability of `example-graph.json`

Alfred-Mountfield:

We previously had conflicting approaches for how blocks could supply example data, these were
- the `examples` field in the `block-metadata.json` (or the associated field in `blockprotocol` in the `package.json`)
- the `variants.examples` field  in the`block-metadata.json` (or the associated field in `blockprotocol` in the `package.json`)
- the `variants.properties` field in the `block-metadata.json` (or the associated field in `blockprotocol` in the `package.json`)
- the `example-graph.json` file which could be shipped alongside the block

This PR simplifies some of the options by mostly standardizing on the `example-graph` in most cases. In doing so, it improves the ergonomics of defining an `example-graph` by accepting a script that generates the JSON, rather than just a plain-text file. In doing so, this allows users to benefit from the codegen within blocks, and type checking. 

Trying to centralize on `example-graph` means
- It is possible to create sample block entities which are link entities (previously it'd be impossible to define the `linkData` or the accompanying links)
- It is more extensible, we could allow blocks to provide many entities (e.g. variants) and instead find a way of making the `blockEntityRecordId` configurable or provided in a map
- We can get rid of the `examples` keyword, which simplifies the API surface and reduces the number of decisions users have to make and understand
- If a user wants to start adding more entities to the example graph they don’t have to move from one representation to another, they just extend their existing one
- It allows us to have one place to update if we did want to start supporting multiple example sets (as opposed to variants)

This PR additionally modifies the site and hub to appropriately load the example data shipped with blocks, fixing numerous bugs that had slipped in recently.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1204080201525383/f) _(internal)_

## 🚫 Blocked by

See unresolved comments, I have opened a few to do with open questions around error handling and validation

## 🔍 What does this change?

- Adjusts the webpack config to produce a CommonJS example-graph script (otherwise `block-scripts` struggles)
- Adjusts the `example-graph` `block-scripts` script to allow for an example-graph script, cleaning up after it
- Removes the `examples` field from block metadata
- Adjusts the block templates to use `example-graph.ts` files instead of `examples`
- Updates the block sandboxing and surrounding code on the Hub to correctly parse the data provided by blocks

- Some of the accompanying docs and spec have been updated as part of this

## ⚠️ Known issues

- I am unsure of how we currently validate the files pushed by users, I believe if the metadata is malformed the hub throws a top-level error
- The distinction between `variants` and the other ways of defining `examples` is unclear, we should look to unify them
- It is still impossible to provide link entities under `variants`
- In general the `properties` viewer/editor on the hub isn't very intuitive and we're planning a rework soon. It doesn't update to reflect changes the block makes for example.
- ~~The Snackbar crashes the app~~ (fixed)

## 🐾 Next steps

- TBD

## 🛡 What tests cover this?

- Unsure besides static analysis

## ❓ How to test this?

1. `yarn dev:db`
2. `yarn dev:seed-db`
3. `yarn dev`
4. Check the existing blocks on the hub
5. `cd libs/block-template-react`
6. Run the publishing flow locally
7. Check the block template renders correctly and it takes it from the `example-graph`

## 📹 Demo

image
image

https://user-images.githubusercontent.com/25749103/228214599-ef65577a-5fb3-4933-ae67-4b93f2a7f8e7.mov


apps/site/src/components/pages/hub/block-data-container.tsx
```diff
@@ -53,143 +70,218 @@ export const BlockDataContainer: FunctionComponent<BlockDataContainerProps> = ({
...
+        setEntity(exampleEntity);
+      } else {
+        throw new Error(
+          `Unable to find an entity for block ${metadata.name} in the example graph`,
+        );
       }
+    } else if (!metadata.variants && !metadata.examples) {
+      // The block hasn't supplied an example graph or a schema, so as long as there aren't any variants or examples
+      // we can pass it anything as the type should be unconstrained. We can't pass in a variant or an example as we
+      // don't know what the type should be.
```
Alfred-Mountfield, line 150: This captures one of the strange edge-cases we're being forced to handle due to `variants` and `examples` just having `properties`. (It also means we can't create link entities)

cc: @CiaranMn

apps/site/src/components/pages/hub/block-data-container.tsx
```diff
@@ -53,143 +70,218 @@ export const BlockDataContainer: FunctionComponent<BlockDataContainerProps> = ({
...
+            decisionTime: interval,
+          },
+        },
+      };
+
+      setEntity(exampleThingEntity);
+    } else {
+      throw new Error(
+        `Unable to create a new entity for block ${metadata.name} from its \`examples\` or \`variants\` as it doesn't have an associated schema`,
+      );
```
Alfred-Mountfield, line 176: I'm unsure of how to do error handling in this file in general, I saw some code about an "alert snackbar"..

apps/site/src/components/pages/hub/block-data-container.tsx
```diff
@@ -53,143 +70,218 @@ export const BlockDataContainer: FunctionComponent<BlockDataContainerProps> = ({
...
+                ...prevEntity,
+                properties: parsedProperties,
+              };
+            }
+          }
+
+          return null;
         });
+      } catch (err) {
+        /* @todo - what's the correct way to handle errors here */
```
Alfred-Mountfield, line 248: See https://github.com/blockprotocol/blockprotocol/pull/1223/files#r1149552502

apps/site/src/components/pages/hub/block-data-container.tsx
```diff
@@ -53,143 +70,218 @@ export const BlockDataContainer: FunctionComponent<BlockDataContainerProps> = ({
...
-            limit: new Date(0).toISOString(),
-          },
-          end: {
-            kind: "unbounded",
-          },
-        } as const;
-      };
+    if (!entity) {
+      /* @todo - we need should handle this case and display a loading skeleton or error message */
+      return [undefined, []];
```
Alfred-Mountfield, line 263: This just ends up with an error in the console where something tries to access `blockEntity` but it doesn't exist. I'm not sure fixing this is within the scope of this PR unless I get some assistance

apps/site/src/components/pages/hub/hub-utils.tsx
```diff
@@ -9,7 +9,8 @@ export type BlockSchema = Record<string, any>;
 
 /** @todo possibly extend this type */
 export type BlockExampleGraph = {
```
Alfred-Mountfield, line 11: This feels like it should be defined somewhere else, alongside an actual validation function, as otherwise we're just asserting that things exist all over the place and pages crash when they don't.

libs/block-scripts/shared/block-assets-plugin.js
```diff
@@ -19,7 +19,7 @@ export class BlockAssetsPlugin {
         "assets-manifest.json",
       );
 
-      const defaultManifest = { "main.js": "main.js" };
+      const defaultManifest = { "main.cjs": "main.cjs" };
```
Alfred-Mountfield, line 22: I had to change the file-endings to `.cjs` because the `package.json` was polluting the scripts and saying they were ES Module scripts. 

This wasn't a problem when publishing the block, but it did end up being a problem when trying to import `example-graph.js`

libs/block-scripts/shared/generate-dist-example-graph.js
```diff
@@ -4,17 +4,65 @@ import fs from "fs-extra";
...
+    delete assetsManifest["example-graph.cjs"];
+
+    await fs.writeJson(assetsManifestPath, assetsManifest, { spaces: 2 });
+    return;
+  } else if (await fs.pathExists(sourcePlainTextExampleGraphPath)) {
+    await fs.copy(sourcePlainTextExampleGraphPath, distExampleGraphPath);
+    return;
   }
+
+  /* @todo - Should this be optional? */
```
Alfred-Mountfield, line 64: I guess if a block doesn't need an entity then it shouldn't have to pass an `example-graph`? My concern is that if a user accidentally moves the script, or changes something in a way our complicated webpack and local tooling setup doesn't pick up, it would just silently ignore it rather than complaining.

It might be better to default to users having to explicitly provide a file, but just leaving it empty of entities / elements

libs/block-scripts/shared/webpack-config.js
```diff
@@ -48,11 +48,17 @@ const generateBaseWebpackConfig = async (mode) => {
         ],
         "block entry point",
       ),
+      ...((await fs.pathExists("./src/example-graph.ts"))
+        ? {
+            "example-graph": "./src/example-graph.ts",
+          }
+        : {}),
```
Alfred-Mountfield, line 55: See https://github.com/blockprotocol/blockprotocol/pull/1223/files#r1149557972

If we want to require it we can simplify these lines

apps/site/src/components/pages/hub/block-data-container.tsx
Alfred-Mountfield: Well I tried re-enabling the snackbar and the frontend crashed. And I noticed the current hub doesn't cause the snackbar to actually appear.

apps/site/src/components/pages/hub/block-data-container.tsx
Alfred-Mountfield, line 1: There's strange behaviour I'm seeing (which is present in the video) whereby when I try and use the Heading block, the contents don't update until I adjust the `level` property. I've thrown in some `console.log`s and the actual `Entity` object is completely changed I'm pretty sure, and I'm fairly certain we actually do pass it into the block, so I would have expected it to rerender, but it doesn't seem to when I adjust `textual-content`..

apps/site/src/_pages/docs/1_blocks/02_develop.mdx
```diff
@@ -403,7 +403,10 @@ Once you’ve finished writing your block, run `yarn build` or `npm run build`.
 
 This will produce a compiled version of your code in the `dist` folder, along with a metadata file describing your block (`block-metadata.json`).
 
-It is worth updating the `blockprotocol` object in `package.json` to include your own `icon`, `image`, and `examples` for your block. These will automatically be included in the `block-metadata.json` produced after running `yarn build` or `npm run build`.
+It is worth updating the `blockprotocol` object in `package.json` to include your own `icon`, and `image` for your block.
+These will automatically be included in the `block-metadata.json` produced after running `yarn build` or `npm run build`.
+
+In addition to the above, updating the `example-graph` (either `./src/example-graph.ts` or `./example-graph.json` depending on approach) is a great way to help users understand how to use your block, and how it functions. Many environments (such as the [Þ Hub]) will insert this data when users are viewing blocks.
```
CiaranMn, line 409:
```suggestion
In addition to the above, updating the `example-graph` (either `./src/example-graph.ts` or `./example-graph.json` depending on approach) is a great way to help users understand how to use your block, and how it functions. Many environments (such as the [Þ Hub]("https://blockprotocol.org/hub")) will insert this data when users are viewing blocks.
```

apps/site/src/_pages/docs/5_spec/02_graph.mdx
```diff
@@ -163,9 +163,13 @@ Blocks may also add an `examples` key to `block-metadata.json` to define example
...
-- `examples`: an array of objects conforming to the block's entity type which provide examples of the data that can be passed to it
 - `variants`: an array of objects, each with a `name` and `properties`, and optionally `description`, `icon`, and `examples`. These objects represents different variants of the block that the user can create, where `properties` sets the starting properties. As a simple example, a ‘header’ block might have variants with the `name` ‘Heading 1’ and ‘Heading 2’, which start with `{ "https://blockprotocol.org/@alice/types/property-type/level/": 1 }` and `{ "https://blockprotocol.org/@alice/types/property-type/level/": 2 }` as `properties`, respectively
-  If more complex example data is required, a block’s package may also include a file at the root named `example-graph.json`, which MUST be a JSON object, and MUST contain an `entities` key where each element is a JSON representation of an entity which MUST conform to the structure specified in [entity definition](#structure-specification).
+
+A block using the graph module MUST include an `example-graph.json` file alongside its metadata. The file's contents
```
CiaranMn, line 168: What's the thinking for moving the provision of example data from `MAY` to `MUST`? I'm not totally against it, but it is imposing another requirement on block authors and we should explicitly consider any new requirements.

apps/site/src/components/pages/hub/block-data-container.tsx
CiaranMn: It will be something to do with the use of the hook module. I can look into it when I test this out (just doing a read through of the code at the moment), or in follow up if you merge it before then.

apps/site/src/components/pages/hub/block-data-container.tsx
```diff
@@ -30,6 +31,22 @@ import { BlockVariantsTabs } from "./block-data-container/block-variants-tabs";
 import { SandboxedBlockDemo } from "./block-data-container/sandboxed-block-demo";
 import { BlockExampleGraph, BlockSchema } from "./hub-utils";
 
+const intervalForAllTime =
```
CiaranMn, line 34: Any further thoughts on sharing this somewhere? `@blockprotocol/graph/internal`?

libs/block-scripts/shared/generate-dist-example-graph.js
CiaranMn: We could instead have a warning in the console

libs/block-template-custom-element/src/dev.tsx
```diff
@@ -2,26 +2,11 @@ import { MockBlockDock } from "mock-block-dock";
 import { createRoot } from "react-dom/client";
 
 import packageJson from "../package.json";
+import { blockEntity, exampleGraph } from "./example-graph";
```
CiaranMn, line 5: I'm wondering if this should instead be importing from the generated file, which would catch more problems than importing directly from the generation script. For example, at the moment we do not use `blockEntityRecordId` in `exampleGraph` at all – it could become disconnected from `blockEntity` and nothing would break here. We should probably take `blockEntityRecordId` from `exampleGraph` at least, rather than `blockEntity.metadata.recordId`, if not directly consuming the generated file and just looking for the block entity in the entities.

CiaranMn (commented):
Thanks, looking good! A few comments on a read through. I haven't tried to test it

libs/block-scripts/shared/generate-dist-example-graph.js
Alfred-Mountfield: People basically never see warnings in consoles, especially if `codegen` is just called as part of `yarn build`

apps/site/src/_pages/docs/1_blocks/02_develop.mdx
Alfred-Mountfield: For some reason I thought this was injected and globally defined..

I could have sworn I saw it in some of the other `mdx` files but after searching I can't see that. Will fix, thanks!

apps/site/src/_pages/docs/1_blocks/02_develop.mdx
```diff
@@ -439,11 +442,11 @@ field in your `package.json` points to a public repository containing your code,
 
 Read more about verification in [the FAQ](https://blockprotocol.org/docs/faq#how-does-block-verification-work).
 
-### Improving your block's appearance on the Þ Hub
+### Improving your block's appearance on the [Þ Hub]
```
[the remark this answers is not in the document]
Alfred-Mountfield: > not sure how links work in headers btw

Generally they should work fine but I haven't checked with our rendering strategy, I'll look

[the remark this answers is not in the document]
Alfred-Mountfield: image

Looks fine to me

apps/site/src/_pages/docs/1_blocks/02_develop.mdx
```diff
@@ -439,11 +442,11 @@ field in your `package.json` points to a public repository containing your code,
...
 
 - **To update the preview image**: take a screenshot of your block in action and replace `public/block-preview.png`
 - **To modify the description:** update the `description` in package.json
 - **To change the icon:** add an icon to the `public` folder and update `blockprotocol.icon` in `package.json`
 - add a preview `image` to the `public` folder and update `blockprotocol.image` in `package.json`
 - update your block’s `README.md` – it will appear below your block on its hub page if you change it from the default
-- update the example properties object in `blockprotocol.examples` in `package.json`
+- update the `example-graph` (either `./src/example-graph.ts` or `./example-graph.json` depending on approach) to update the data the block has available to it when a user loads it on the hub
```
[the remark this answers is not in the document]
Alfred-Mountfield: [Capitalize Hub](https://github.com/blockprotocol/blockprotocol/pull/1223/commits/4b418d04269e8a1de811b9127dcc6a09055afc2c)

apps/site/src/_pages/docs/5_spec/02_graph.mdx
Alfred-Mountfield: Oops this is currently inconsistent. [Previous thread for discussion](https://github.com/blockprotocol/blockprotocol/pull/1223#discussion_r1149557972)

apps/site/src/components/pages/hub/block-data-container.tsx
Alfred-Mountfield: It's only used in one place currently so I'm not sure that's necessary 
image

libs/block-template-custom-element/src/dev.tsx
Alfred-Mountfield: > should probably take `blockEntityRecordId` from `exampleGraph` at least, rather than `blockEntity.metadata.recordId`

We 100% should, I only added `blockEntityRecordId` as a thing after writing this and this was an oversight:
[Use exampleGraph.blockEntityRecordId](https://github.com/blockprotocol/blockprotocol/pull/1223/commits/1b0bd218c6865c512cf844653b7c798cb56d4e54)

> and nothing would break here 

With this change I think MBD should break, as it won't be able to find the entity you're identifying with the recordId

libs/block-scripts/shared/block-assets-plugin.js
Alfred-Mountfield: So.. I removed this and everything worked, I am not sure why as I don't believe I've changed anything about `"type": "module"` or any of that. 

Regardless, avoiding the file-endings seems nicer so I've committed it. cc: @CiaranMn

CiaranMn (commented):
A couple of small comments – I haven't been able to test as the block hub isn't loading, but it isn't in prod either. I assume it's something to do with esm.sh

libs/block-scripts/shared/generate-dist-example-graph.js
Alfred-Mountfield: (Not a leading question), would you like it if we make it optional in the spec but required in `block-scripts`?

CiaranMn: I think that's fine. The template is opinionated about things that aren't required in the spec. Really I don't mind what we do except that if we are going to add a new requirement to the spec, I would like a justification for it being required.

Alfred-Mountfield: I've come full circle and I think it might be problematic if this is required and a block doesn't require an entity, or if it expresses its examples through `Variants`.

Going to make it optional for now and throw a warning as you recommended

apps/site/src/_pages/docs/5_spec/02_graph.mdx
Alfred-Mountfield: [Make the example-graph file optional](https://github.com/blockprotocol/blockprotocol/pull/1223/commits/858ae83f0f31420fad60ad094d8b0cf49b9d006a)

libs/block-scripts/shared/generate-dist-example-graph.js
Alfred-Mountfield: [Make the example-graph file optional](https://github.com/blockprotocol/blockprotocol/pull/1223/commits/858ae83f0f31420fad60ad094d8b0cf49b9d006a)

libs/block-scripts/shared/generate-dist-example-graph.js
```diff
@@ -4,17 +4,65 @@ import fs from "fs-extra";
...
+    } = await import(sourceScriptExampleGraphPath);
 
-  if (await fs.pathExists(exampleGraphFilePath)) {
-    await fs.copy(exampleGraphFilePath, exampleGraphInDistFilePath);
+    if (exampleGraph === undefined || !exampleGraph.entities) {
+      throw new Error(
+        "The default export of the example-graph script must be a valid example graph, containing top-level 'entities' and 'blockEntityRecordId' properties.",
+      );
+    }
+    await fs.writeJson(distExampleGraphPath, exampleGraph);
```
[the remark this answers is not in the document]
Alfred-Mountfield: [Use writeFormattedJson](https://github.com/blockprotocol/blockprotocol/pull/1223/commits/74a6a04b86bd8bfb5503b1c26cb99b7cd16499f5)

CiaranMn (approved):

PR #1225, opened 2023-03-28
Consolidate API key pages, update routes and add missing font-family

yusufkinatas:

This PR 
- Removes duplicated code (for API page) from database
- Fixes wrong font-family usage on settings
- Updates `/settings/*` routes to `/account/*`

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204086167143889/1204260759309476/f) _(internal)_

## ❓ How to test this?

1.  Check "My Account" pages and confirm they look like in the design (not the page content, the menu titles), and routes are updated from `/settings/*` to `/account/*`


apps/site/src/components/page-sidebar.tsx
```diff
@@ -61,7 +62,7 @@ type SidebarPageSectionProps = {
 const highlightSection = (isSectionSelected: boolean) => ({
   borderLeft: `3px solid ${
     isSectionSelected ? themeImport.palette.purple[700] : "transparent"
-  }`,
+  } !important`,
```
benwerner01, line 65: Any way we can easily achieve this using specificity where this is used? I think it's best practice to reserve `!important` to a last resort, so just wanted to check.

apps/site/src/pages/account/billing-settings-panel/billing-settings-panel.tsx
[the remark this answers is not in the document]
yusufkinatas: Makes sense, updated!

apps/site/src/components/page-sidebar.tsx
yusufkinatas: I agree, that `!important` is used to override MUI applying `border:none` when it's `:focus-visible`. So I guess it's fair to use here to prevent MUI removing the indicator border.

benwerner01: But can't we use the `:focus-visible` selector to prevent having to use `!important` here then? If it's not worth the trouble, then I'm okay with us leaving it. Just think it's worth trying to avoid `!important` in the codebase where we can.

yusufkinatas: Hmm, yes that makes sense. Thanks for reminding me this angle 😊 

Updated it here https://github.com/blockprotocol/blockprotocol/pull/1225/commits/34265d03cc9a491e6813f4b75982c1339e079538

PR #1228, opened 2023-03-28
Fix mobile styling, icon in edit bar

CiaranMn:

https://github.com/blockprotocol/blockprotocol/pull/1210 introduced some fixes for the type editor in mobile, but involved setting a `min-width` on the `.MuiPaper-root` class that broke mobile styling in various parts of the app (e.g. the homepage, the API key settings page, and the 'confirm code' part of the login modal).

I'm not sure what the `min-width` was required for so I don't know what alternative to introduce in its place, but it needs to be removed as it affects more than the type editor.

In the process I noticed that the edit bar had a weird icon and the spacing in the text could be improved, so I've done that here also.

## ⚠️ Known issues

- This might introduce some issue in the type editor on mobile, but it seems fine for displaying types, which is good enough (editing types on mobile is going to be extremely fiddly anyway).

## ❓ How to test this?

1.  Check that e.g. the homepage displays on mobile, versus https://blockprotocol.org

## 📹 Demo

Revised type bar

Screenshot 2023-03-28 at 16 08 00

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 [discord]: https://blockprotocol.org/discord?utm_medium=organic&utm_source=github_readme_blockprotocol-repo_root
 [graph module]: https://blockprotocol.org/docs/spec/graph-module?utm_medium=organic&utm_source=github_readme_blockprotocol-repo_root
 [hook module]: https://blockprotocol.org/docs/spec/hook-module?utm_medium=organic&utm_source=github_readme_blockprotocol-repo_root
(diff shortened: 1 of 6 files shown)


nonparibus:
@CiaranMn Can we make the icon in the edit bar 'regular' and not 'solid' so it style-matches the rest of the site? thanks

CiaranMn:
> Can we make the icon in the edit bar 'regular' and not 'solid' so it style-matches the rest of the site? thanks

@nonparibus I've created a follow up [task](https://app.asana.com/0/1203179076056186/1204283213882472/f) (internal) for this as I can't find a regular/outline equivalent to this – I'm not even sure it came from FontAwesome as it's not in their list on the web.

nathggns (approved):

PR #1231, opened 2023-03-29
New API for creating users/logging in, marking the user as linked to wordpress

nathggns:

This is the first part of a multi-PR task which will allow you to sign up for BP/generate an API key using the Wordpress plugin.

This PR introduces a new `/api/link-wordpress` API which creates special verification codes which will link an existing user (or create a new one) to a specific wordpress instance, and mark them as being referred (if the user wasn't already verified) by wordpress. 

It also improves consistency in verification code handling, and adds some error case checking. 

**This does not implement the API key generation once a new or existing user has verified their email through this API. This will come in a follow up.** See https://app.asana.com/0/1204086167143889/1204260760205512/f

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204086167143889/1204270515784149/f) _(internal)_

## 🚫 Blocked by

N/A

## 🔍 What does this change?

- New `referrer` field on users, which tracks how their account was created. Will either be empty or `"blockprotocol"` if created on-site. Will be `wordpress` if the account was created via the WP plugin API, or if the user hadn't been verified yet when the email was entered into the WP plugin API.
- New `wordpressInstanceUrls` field on users, which will store an array of wordpress instances the user has claimed to be linked to via the WP plugin API. These are not verified.
- New `addWordpressInstanceUrlAndVerify` function on users used to update a user with wordpressInstanceUrl and referrer when a new special kind of verification code is used
- `User#createVerificationCode` can now take a `wordpressInstanceUrl` to attach to the verification code
- `getVerificationCode` can now take an array of variants. This is so the verify login / verify signup APIs can also look for linkWordpress codes
- `User#hasExceededEmailVerificationRateLimit` now checks all kinds of verification codes, not just login ones. This is a drive-by bug fix – see https://hashintel.slack.com/archives/C02LG39FJAU/p1680088012676019
- New `User#sendLinkWordpressCode` which can create a new special kind of verification code with the wordpressInstanceUrl attached to it, and send a new "Activate your Block Protocol Wordpress plugin" email which will work for both existing users and new users, and will send you to the correct place. 
- New `/api/link-wordpress` API which will either create a user and send an email verification email or send a login email, both with special verification codes which will new users as having been referred by wordpress, and attach the wordpressInstanceUrl to their user when used. 
- The login API now errors if a user hasn't been verified yet, and will now add wordpress info to a user when a linkWordpress verification code is used
- The verify email API now errors if a user has been verified, and will attach wordpress info to a user when used with the new verification codes. We also now mark the code as used, which we didn't before.

N/A

## ⚠️ Known issues

N/A

## 🐾 Next steps

**This does not implement the API key generation once a new or existing user has verified their email through this API. This will come in a follow up.** See https://app.asana.com/0/1204086167143889/1204260760205512/f

This also does not implement the wordpress plugin UI. See https://app.asana.com/0/1204086167143889/1204260760205619/f

## 🛡 What tests cover this?

N/A

## ❓ How to test this?

POST request to `/api/link-wordpress`. Check the following cases:

- Completely new user (wordpress referrer and instance are immediately attached to user)
- Existing user which hasn't been verified yet (wordpress referrer and instance are  attached to user once the verification link has been click on)
- Existing user which has been verified (instance URL is  attached to user once the verification link has been click on – wordpress referrer is never set)


apps/site/src/pages/api/login-with-login-code.api.ts
```diff
@@ -74,7 +84,28 @@ export default createBaseHandler<
...
+            }),
+          );
+        }
+
+        await loginCode.setToUsed(db);
+
+        await user.addWordpressInstanceUrlAndVerify(
+          db,
+          wordpressInstanceUrl,
+          !user.hasVerifiedEmail,
```
thehabbos007, line 104: Shouldn't this always be `updateReferrer = false` for these `login` request?
This is currently updating the referrer on every `linkWordpress` login as far as I can tell (which I suppose should have happened in the sign up step?)

Do we have any calls to `addWordpressInstanceUrlAndVerify` that are ever `false`?

apps/site/src/pages/api/verify-email.api.ts
```diff
@@ -78,7 +88,28 @@ export default createBaseHandler<
...
+              value: code,
+            }),
+          );
+        }
+
+        await emailVerificationCode.setToUsed(db);
+        await user.addWordpressInstanceUrlAndVerify(
+          db,
+          wordpressInstanceUrl,
+          !user.hasVerifiedEmail,
```
thehabbos007, line 107: Similarly to the comment above, because of the exception thrown `if (user.hasVerifiedEmail)`, can this referrer value ever be `false`?

apps/site/src/lib/api/model/verification-code.model.ts
```diff
@@ -5,17 +5,24 @@ import { rword } from "rword";
...
-  variant: VerificationCodeVariant;
   code: string;
   numberOfAttempts: number;
   used: boolean;
   createdAt: Date;
+  variant: VerificationCodeVariant;
+  wordpressInstanceUrl?: string;
 };
 
+// @note not using a discriminated union to match class properties, where that isn't possible
```
thehabbos007, line 20: This double negation is confusing me a little, what does this imply? :sweat_smile:

thehabbos007 (commented):
Thank you, Nate! Works locally for me, just a have some comments :)

apps/site/src/pages/api/login-with-login-code.api.ts
nathggns: Yes, you're right. I did this before putting in the error check, as a hedge in case the decision was not to error. Will fix. 

`addWordpressInstanceUrlAndVerify` will be false for all login requests now. We don't want to set referrer to wordpress for existing accounts.

apps/site/src/lib/api/model/verification-code.model.ts
nathggns: It's trying to explain what I explained at https://github.com/blockprotocol/blockprotocol/pull/1231#discussion_r1153208059 

We can pair on this if you'd like?

apps/site/src/pages/api/verify-email.api.ts
```diff
@@ -78,7 +88,28 @@ export default createBaseHandler<
         );
       }
 
-      await user.update(db, { hasVerifiedEmail: true });
+      if (emailVerificationCode.variant === "linkWordpress") {
+        const { wordpressInstanceUrl } = emailVerificationCode;
+        if (!wordpressInstanceUrl) {
```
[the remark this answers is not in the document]
thehabbos007: That's fair. I don't think we should do it in this PR to keep the scope fairly small. I'm not sure if we have a task for something like this, but it would be ideal if it's some established pattern that we colocate things that are related instead of relying on runtime checks for optional values.

apps/site/src/lib/api/model/verification-code.model.ts
thehabbos007: Nate and I chatted offline about this, see 5051b717c7b4e1f595b9868c4cc4340c6604e9c1

apps/site/src/pages/api/login-with-login-code.api.ts
thehabbos007: Perfect, thank you. Also great you chose the object param so we dont just have stray `false` and `true`s!

thehabbos007 (approved):

PR #1232, opened 2023-03-29
Fix `queryEntities` for MariaDB in WordPress plugin

thehabbos007:

This PR fixes `queryEntities` to support MariaDB (10.2.7+) alongside MySQL (5.7.8+).
The previous setup was mostly functioning, the issues we saw were due to having `\` in front of every `/` in our JSON blobs because of PHP's default `json_encode` behavior. This has been fixed in this PR.

To get MariaDB to work with arbitrary JSON spacing, I had to add conditional casting to JSON (using `JSON_COMPACT` in MariaDB).

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056186/1204069269252814/f) _(internal)_

## 🔍 What does this change?

- Drive-by fixes lack of equality on minimum version check
- Add shared `JSON_DECODE` function to use in the codebase (doesn't escape slashes or unicode characters)
- Adjust query compilation to support MariaDB using `JSON_COMPACT` and changing string quoting

- I've removed README notices about lack of `queryEntities` support in MariaDB

## ⚠️ Known issues

- The JSON keys order has to match on the object being queried and the filter value. This is because the JSON values are still plain text in MariaDB, so minimal processing is done in the DB to check for matches. The `JSON_COMPACT` does make whitespace irrelevant for the JSON object, though
- Some filter operations are case sensitive while others are not. Maybe a difference between the queries that use `LIKE`  and those that use `JSON_SEARCH`

## 🛡 What tests cover this?

- Manual tests :rock: 
- I've ensured that the queries still support providing URLs instead of simple keys

## ❓ How to test this?

- (optional) change DB version of MySQL in the docker-compose to 5.7.8 for minimum version testing.
- Checkout PR, set up WP & BP Plugin
- Create blocks or manual DB entities with custom JSON
- Use the following cURL request (can be pasted into Postman/Insomnia) (doesn't require Auth) to test the various filters:
```curl
curl --request POST \
  --url 'http://localhost:8000/index.php?rest_route=%2Fblockprotocol%2Fentities%2Fquery' \
  --header 'Content-Type: application/json' \
  --data '{
	"operation": {
		"multiFilter": {
			"filters": [
				{
					"operator": "CONTAINS_SEGMENT",
					"field": [
						"https://blockprotocol.org/@blockprotocol/types/property-type/textual-content/"
					],
					"value": "code block"
				}
			],
			"operator": "AND"
		}
	}
}'
```

## 📹 Demo

**MariaDB 10.2.7**:

https://user-images.githubusercontent.com/6988668/228564801-9d6b2274-7282-4ccd-8a74-0c9770888ba4.mp4

**MySQL 5.7.8**:

https://user-images.githubusercontent.com/6988668/228564830-0187d05b-e4a7-43ea-bf7d-3ad1e9492c3a.mp4

Example with URL key:

image


libs/wordpress-plugin/plugin/trunk/server/block-api-endpoints.php
```diff
@@ -409,6 +409,10 @@ function get_all_block_protocol_entities()
   return $entities;
 }
 
+function block_protocol_json_encode(array $value)
+{
+  return json_encode($value, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
```
thehabbos007, line 414: I added `JSON_UNESCAPED_UNICODE` for good measure. Not sure if it would cause issues as we're using the WP charset collation for the table which could be non-utf8?

libs/wordpress-plugin/plugin/trunk/server/block-db-table.php
```diff
@@ -13,10 +16,10 @@ function block_protocol_database_at_version(string $mysql_version, string $maria
 
   if (strpos($db_server_info, 'MariaDB') != false) {
     // site is using MariaDB
-    return strnatcmp($db_version, $mariadb_version) > 0;
+    return strnatcmp($db_version, $mariadb_version) >= 0;
   } else {
     // site is using MySQL
-    return strnatcmp($db_version, $mysql_version) > 0;
+    return strnatcmp($db_version, $mysql_version) >= 0;
```
thehabbos007, line 22: I forgot to add `=` in the previous PR :sweat_smile:

libs/wordpress-plugin/plugin/trunk/server/query.php
```diff
@@ -51,9 +51,9 @@ function block_protocol_encode_filter_value($value)
   if (is_int($value)) {
     return "{$value}";
   } elseif (is_string($value)) {
-    return "'{$value}'";
+    return "\"{$value}\"";
```
thehabbos007, line 54: This made a big difference for some reason :shrug: 
Without it MariaDB would fail on strings.

libs/wordpress-plugin/plugin/trunk/server/query.php
```diff
@@ -71,42 +71,59 @@ function block_protocol_extract_filter_value($filter, $raw = false)
...
 
-function block_protocol_build_string_filter($filter_value, $json_path, $operator, &$sql_fragment, &$values)
+/**
+ * This function will return a SQL string that casts text to JSON. 
+ * Note that this is function is "unsafe" and shouldn't have any direct user data passed into it.
+ */
+function block_protocol_db_json_unsafe_cast(string $value, bool $is_maria_db)
+{
+  if($is_maria_db) {
+    return "JSON_COMPACT({$value})";
```
thehabbos007, line 81: From what I can tell `JSON_COMPACT` even has a binary/non-string representation in the DB in version 10.2.7, so maybe the checks are more efficient when we do this casting? (although there's overhead in the calls..)

libs/wordpress-plugin/plugin/trunk/server/query.php
```diff
@@ -256,9 +287,10 @@ function block_protocol_query_entities($multi_filter, $multi_sort)
   $where_clause = empty($filter_placeholders) ? "" : "WHERE " . implode(" {$operator} ", $filter_placeholders);
   $sort_clause = empty($sort_placeholders) ? "" : "ORDER BY " . implode(", ", $sort_placeholders);
 
+  $selection = block_protocol_entity_selection();
   $table = get_block_protocol_table_name();
   return array(
-    'sql' => "SELECT * FROM {$table} {$where_clause} {$sort_clause}",
+    'sql' => "SELECT {$selection} FROM {$table} {$where_clause} {$sort_clause}",
```
thehabbos007, line 293: I add the same selection here we use for the `subgraph` query for good measure

libs/wordpress-plugin/plugin/trunk/readme.txt
```diff
@@ -36,9 +36,9 @@ You can also build your own block for the Block Protocol without any PHP, and pu
 
 **5.** If you want to add your own block to the plugin, visit https://blockprotocol.org/docs/developing-blocks
 
-Note that you must be using at least MySQL 5.7.8. To check please navigate to `Admin -> Tools -> Site Health -> Info -> Database`.
+Note that you must be using at least MySQL 5.7.8 or MariaDB 10.2.7. To check please navigate to `Admin -> Tools -> Site Health -> Info -> Database`.
 MySQL 5.6 and 5.7 respectively reach(ed) end of life in February 2021 and October 2023.
```
nonparibus, line 40:
```suggestion
**MySQL 5.6** and **5.7** reach(ed) end of life in February 2021 and October 2023 respectively. **MariaDB 10.2** and **10.3** reach(ed) end of life in May 2022, and May 2023.
```

libs/wordpress-plugin/plugin/trunk/server/block-api-endpoints.php
CiaranMn: I tried adding some multibyte unicode characters with this off, and they didn't end up escaped in the db anyway. I wonder if they get unescaped again if stored in a db with a unicode-encoding charset? Could you try putting in some data and querying that contains multibyte unicode characters? If it works anyway it might be worth not adding this in case it messes things for dbs with non-utf8 charsets. But I don't know.

CiaranMn:
@thehabbos007 can you also bump the 'tested up to' from `6.1` to `6.2` please? I've confirmed this works now

libs/wordpress-plugin/plugin/trunk/server/block-api-endpoints.php
thehabbos007: I will try with some of the complicated emojis

thehabbos007: Seems we want this there. The demo here shows with/without the unescape option. TL;DW the escaped (default) causes the emoji to split up and become unqueryable (tested on mariadb 10.2.7)

With

https://user-images.githubusercontent.com/6988668/228780158-26b1b1b0-ea14-44f2-bffa-c861bc870a31.mp4



Without

https://user-images.githubusercontent.com/6988668/228780169-9636e2e8-398c-46b9-b8e4-5f23128a2299.mp4

CiaranMn: Good to have confirmed, thank you for testing this out!

CiaranMn (approved):
Excellent, thank you!

PR #1233, opened 2023-03-29
Fix E2E test CI to handle multiple deployments in repo. Split out workflows

CiaranMn:

Since the `blockprotocol-design-system` project began deploying successfully, we have had deployment events triggers for two separate applications.

Our E2E test workflow for the `blockprotocol` project didn't account for multiple deployment events, meaning that it was triggered by design system deployments, against which the E2E tests fail (because they're testing the wrong app), blocking PR merging.

This PR updates the check for triggering the E2E to ensure that it only works when the environment ends with `blockprotocol` - see example deployment_statuses: [blockprotocol](https://api.github.com/repos/blockprotocol/blockprotocol/deployments/849055133/statuses/1816361288), [blockprotocol-design-system](https://api.github.com/repos/blockprotocol/blockprotocol/deployments/849072047/statuses/1816379609)

It also splits out the E2E test and 'Scheduled every 10 minute Lambda warm up' jobs into their own workflows. They were previously grouped under 'Site deployment' which is confusing and made debugging this harder.

## ❓ How to test this?

1.  Check that workflows run as expected and that the E2E job succeeds.


.github/actions/warm-up-vercel/action.yml
```diff
@@ -0,0 +1,32 @@
...
+      shell: bash
+      run: |
+        cat << EOF > curl-format.txt
+          %{url_effective}\n
+            Status code: %{response_code}\n
+            Time: %{time_total}\n
+          \n
+        EOF
+
+        for PATHNAME in "/" "/hub" "/@hash" "/@hash/blocks/divider/sandboxed-demo" "/api/me" "/api/blocks" "/@blockprotocol/types/entity-type/thing/v/2"; do
```
CiaranMn, line 27: This line has changed since the previous version to add more paths and correct the type path

.github/workflows/e2e-tests.yml
```diff
@@ -0,0 +1,40 @@
...
+  deployment_status:
+
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  e2e_tests:
+    name: E2E tests
+    if: github.event.deployment_status.state == 'success' && endsWith(github.event.deployment_status.environment, 'blockprotocol')
```
CiaranMn, line 13: this line has changed since the previous version to add a check for the environment. See example status events in PR description – it will be something like "Preview – blockprotocol" or "Preview – blockprotocol-design-system"

yusufkinatas (approved):

PR #1238, opened 2023-03-30
Add example of block requirements proposal

CiaranMn:

We want a way of blocks expressing which parts of the protocol they require (or can use), in order that:
1. blocks can be filtered by which features they use, for the purpose of finding blocks of interest
2. an embedding application can determine whether it supports a particular block or not
3. users can give a block permission to use certain modules / features

This PR adds a basic version of the draft proposal to the React block template, and will be expanded with more complex examples and rejection of requests from blocks that they do not specify a requirement for.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056186/1202636018097183/f) _(internal)_

## 🚫 Blocked by

- [ ] RFC to detail the proposal
- [ ] Further examples
- [ ] Handle block requirement declaration in Mock Block Dock

- The proposal will require amendments to the 'developing a block' docs.

## 🐾 Next steps

- Add requirement declarations to our existing blocks.

### DIFF
--- a/libs/block-template-react/package.json
+++ b/libs/block-template-react/package.json
@@ -66,6 +66,22 @@
     "image": "public/block-preview.png",
     "name": "block-template-react",
     "protocol": "0.3",
+    "requirements": {
+      "graph": {
+        "version": "0.3",
+        "messages": {
+          "updateEntity": {
+            "data": {
+              "entityTypeId": [
+                {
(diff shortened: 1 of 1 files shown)


nonparibus:
@CiaranMn Any reason not to merge this in?

PR #1239, opened 2023-03-30
Enable creating multiple API keys

thehabbos007:

This PR enables us to create multiple API keys for the BP backend. 
The API keys are attached to a single user and can be independently revoked and updated.
To keep the site functioning as before, I've changed the frontend to revoke & create temporarily to prevent strange UI states.
I did have a reverted commit that changed the UI to allow multiple creation/revokes but I decided I shouldn't be spending more time on that because of the upcoming UI work.

There's also a drive-by to use time-safe equality checks for API key checking.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1204080201525363/f) _(internal)_

## 🔍 What does this change?

- Remove `revokeAll` call from the `generate-api-key` endpoint
- Add new `revoke-api-key` endpoint
- Add new `update-api-key` endpoint for updating the `displayName`
- Drive-by fix timing-safe equality
- Temporary fix for frontend to make the UI work as before

- The API route documentation has been updated in the README

## 🐾 Next steps

- UI changes [Asana task](https://app.asana.com/0/1204086167143892/1204270888519624/f)

## 🛡 What tests cover this?

- Manual tests

## ❓ How to test this?

1.  Sign up/in and find your `blockprotocol-session-id` cookie value. use that for the following CURL requests (replace URL if using staging environment):
```curl
curl --request POST \
  --url http://localhost:3000/api/me/generate-api-key \
  --header 'Content-Type: application/json' \
  --header 'Cookie: blockprotocol-session-id=COOKIE HERE' \
  --cookie blockprotocol-session-id=*CHANGE* \
  --data '{
	"displayName": "test"
}'
```

```curl
curl --request POST \
  --url http://localhost:3000/api/me/update-api-key \
  --header 'Content-Type: application/json' \
  --header 'Cookie: blockprotocol-session-id=COOKIE HERE' \
  --cookie blockprotocol-session-id=*CHANGE* \
  --data '{
	"publicId": "*CHANGE*",
	"displayName": "Updated!"
}'
```

```curl
curl --request POST \
  --url http://localhost:3000/api/me/revoke-api-key \
  --header 'Content-Type: application/json' \
  --header 'Cookie: blockprotocol-session-id=COOKIE HERE' \
  --cookie blockprotocol-session-id=*CHANGE* \
  --data '{
	"publicId": "*CHANGE*"
}'
```
1. during the requests, check to see if your API keys appear on the site/in the DB

## 📹 Demo

This demo is a little long, but shows the above curl commands running in insomnia, and some edge cases etc. 

https://user-images.githubusercontent.com/6988668/228919453-4a4e4395-d252-4537-a38f-32bd11ac1ef3.mp4

### DIFF
--- a/apps/site/README.md
+++ b/apps/site/README.md
@@ -286,7 +286,7 @@ Revokes a given API key for the authenticated user.
 
   - `publicId`: the public ID portion of the API key to revoke.
 
-- Request Response: Empty
+- Request Response: `SUCCESS`
 
 #### `POST /api/me/update-api-key` [requires cookie authentication]
 
@@ -297,7 +297,7 @@ Update a given API key for the authenticated user.
   - `publicId`: the public ID portion of the API key to update.
   - `displayName`: the new `displayName` that should be present on the API key.
 
-- Request Response: Empty
+- Request Response: `SUCCESS`
 
 #### `POST /api/logout` [requires cookie authentication]
 
--- a/apps/site/src/pages/api/me/revoke-api-key.api.ts
+++ b/apps/site/src/pages/api/me/revoke-api-key.api.ts
@@ -9,7 +9,7 @@ export type ApiRevokeApiKeyBody = {
   publicId: string;
 };
 
-export type ApiRevokeApiKeyResponse = {};
+export type ApiRevokeApiKeyResponse = "SUCCESS";
 
 export default createAuthenticatedHandler<
   ApiRevokeApiKeyBody,
@@ -40,5 +40,5 @@ export default createAuthenticatedHandler<
       );
     }
 
-    res.status(200).json({});
+    res.status(200).json("SUCCESS");
   });
--- a/apps/site/src/pages/api/me/update-api-key.api.ts
+++ b/apps/site/src/pages/api/me/update-api-key.api.ts
@@ -10,7 +10,7 @@ export type ApiUpdateApiKeyBody = {
   displayName: string;
 };
 
-export type ApiUpdateApiKeyResponse = {};
+export type ApiUpdateApiKeyResponse = "SUCCESS";
 
 export default createAuthenticatedHandler<
   ApiUpdateApiKeyBody,
@@ -44,5 +44,5 @@ export default createAuthenticatedHandler<
     }
     // If it was found but not updated, the operation was a no-op.
 
-    res.status(200).json({});
+    res.status(200).json("SUCCESS");
   });


apps/site/README.md
```diff
@@ -265,19 +265,40 @@ Retrieves the user object of the currently logged in user.
...
 
+#### `POST /api/me/revoke-api-key` [requires cookie authentication]
+
+Revokes a given API key for the authenticated user.
+
+- Request Body:
+
+  - `publicId`: the public ID portion of the API key to revoke.
+
+- Request Response: Empty
```
CiaranMn, line 289: It's not really an empty response, it's `"{}"`. We could instead make it `true` (i.e. `res.status(200).send(true)`) or `"SUCCESS"` (we do this for logout) or something. I think it's fine to be an empty object too, since basically if the request doesn't error it has succeeded, but we should describe it as such here.

apps/site/src/lib/api/model/api-key.model.ts
```diff
@@ -206,13 +211,42 @@ export class ApiKey {
...
+    const response = await db
       .collection<ApiKeyDocument>(ApiKey.COLLECTION_NAME)
-      .updateMany(
-        { user: params.user.toRef(), revokedAt: { $eq: null } },
-        { $set: { revokedAt: new Date() } },
+      .updateOne(
+        {
+          user: params.user.toRef(),
+          publicId: params.publicId,
+          revokedAt: { $eq: null },
```
CiaranMn, line 229: This means that users cannot rename a key once it has been revoked – I'm not sure we need to impose that restriction (users might want to revoke an old 'Production' key and then rename it to 'Production – old' or something). I agree that we don't want to revoke keys that have already been revoked because it makes the timestamp misleading.

apps/site/src/pages/api/me/revoke-api-key.api.ts
```diff
@@ -0,0 +1,44 @@
...
+export type ApiRevokeApiKeyBody = {
+  publicId: string;
+};
+
+export type ApiRevokeApiKeyResponse = {};
+
+export default createAuthenticatedHandler<
+  ApiRevokeApiKeyBody,
+  ApiRevokeApiKeyResponse
+>(false)
```
CiaranMn, line 17: `false` means that users who have not completed signup can use this route. 

I see that I put this on the original API key handler which explains why it's here, but I was wrong. We can just omit it as it will default to `true` (check if the user has completed sign up).

```suggestion
>()
```

thehabbos007: Addressed in 6755f5529c5ed899bc23590559c28f50d5b3e573, also changed it for Create/List

apps/site/src/pages/api/me/update-api-key.api.ts
```diff
@@ -0,0 +1,47 @@
...
+  publicId: string;
+  displayName: string;
+};
+
+export type ApiUpdateApiKeyResponse = {};
+
+export default createAuthenticatedHandler<
+  ApiUpdateApiKeyBody,
+  ApiUpdateApiKeyResponse
+>(false)
```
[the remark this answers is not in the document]
thehabbos007: Addressed in 6755f5529c5ed899bc23590559c28f50d5b3e573,

apps/site/src/lib/api/model/api-key.model.ts
thehabbos007: I see, I thought we didn't return the revoked keys in the `List` endpoint, but that's just because we filter them away in the UI. I'll adjust this

thehabbos007: I've fixed this in 1d0b2c7713dd297c38d9d106f04dd275ae670db6
also now treat idempotent updates as successful

apps/site/README.md
thehabbos007: Changed to "SUCCESS" in a30e7f5eaf912f1d03ce4f541f4b04b6c2ea34e3 for consistency with logout. I think `true` would be fine too, but both are strongly typed.

I didn't go with `{}` because it's catch-all type and we don't get errors in the `.send` if we provide another value.

CiaranMn (approved):

PR #1245, opened 2023-03-31
Reapply first WordPress migration for users of older version

thehabbos007:

This PR adds a new migration that simply re-runs the first (through the `IF NOT EXISTS` clause) to ensure that old users have a database. We previously didn't bail if the DB wasn't applied correctly, which means some users are stuck with unmigrated DBs at `migration_version` 2

I've also changed the migration code to be automated rather than manual and error prone.

Please do test this locally. I may have missed some edge cases. Also feel free to push commits if this doesn't get approved/merged today :)

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204086167143889/1204298348614666/f) _(internal)_

## 🔍 What does this change?

- Add new migration code for automatic migration management
- Drive-by add migration version to the sentry errors/telemetry data
- Add `migration_2` that reapplies `migration_1`

## ⚠️ Known issues

This isn't much of a direct issue, but we have to account for the previous version numbering in the new migration code, so we are depending on incrementing the version numbers somewhat unintuitively. This is to have it fit the former migration code.

## 🛡 What tests cover this?

- Manual testing

## ❓ How to test this?

1.  Set up the WordPress instance locally on **`main`**. Install + activate the plugin. Go to the setting page of the plugin
1.  Check DB for `block_protocol_db_migration_version` in the `wp_options` table. It should be `2`
1.  Checkout the branch
1.  Reload your setting page
1.  Confirm that the `block_protocol_db_migration_version` is now `3`


libs/wordpress-plugin/plugin/trunk/server/block-db-table.php
```diff
@@ -75,21 +83,35 @@ function block_protocol_set_migration_version_to(int $migration_version)
...
+  // Apply new migrations sequentially, skipping already-applied ones.
+  foreach ($migrations as $index => $migration) {
+    // Fetch the saved version at each iteration as it should be changing
+    $saved_version = (int) get_site_option('block_protocol_db_migration_version');
+    $expected_migration_version = $index + 1;
+    if ($saved_version <= $expected_migration_version) {
+      ($migration)();
+      block_protocol_set_migration_version_to($expected_migration_version + 1);
+    }
+  }
```
thehabbos007, line 114: The main change is that we previously checked for `block_protocol_db_migration_version`, which would be `null` in the beginning (cast to 0 I believe) and the set the migration level to `2` after applying migration `1`.

Now we use the same numbering scheme but calculate the current/next migration versions automatically from the array indices.
We have
`[ 0 => "block_protocol_migration_1", 1 => "block_protocol_migration_1"]`
which is translated to 
`[ 1 => "block_protocol_migration_1", 2 => "block_protocol_migration_1"]`
in the loop. We then apply the migration if the current migration is less than or equal to the current loop iteration.
So we'd be applying `1` on a fresh install of the plugin, and on existing installs that have migration set to `2` in their DB, we'd apply `2` because of the equality check. `block_protocol_db_migration_version` is always set to `current migration index + 1 + 1` which would mean we set 
`"block_protocol_migration_1" to 2, "block_protocol_migration_1" to 3`

CiaranMn (approved):

PR #1246, opened 2023-04-03
Show card to generate API key for wordpress when using link-wordpress API 

nathggns:

New UI in dashboard for generating API key when signed up / logged in via the link wordpress API. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204086167143889/1204260760205512/f) _(internal)_

## 🚫 Blocked by

- [x] https://github.com/blockprotocol/blockprotocol/pull/1231

## 🔍 What does this change?

**Outdated… see PR comments**

- New `signOut` global function to ensure other side effects (Sentry, removing the `wordpressInstanceUrl` from session storage) are triggered. It's a global function impacting context though, so you need to pass the `setUser` function. This is now used for existing ways you can sign out too.
- New `DashboardWordpressSection` component which displays the UI for generating/copying an API key, with the name of "Wordpress". It also links you back to a passed `wordpressInstanceUrl`. In a future PR, this will likely be updated to append a settings path to it, so this will be just the root URL for the wordpress instance
- The verification code screen (used by login and signup) now settings the wordpress instance URL in session storage if provided by the API when checking a verification code, and also passes up a redirectPath if one is provided by the API. This allows the check verification code APIs for login/signup to take you to the dashboard showing the wordpress card. The APIs now also provide these values.
- New `setWordpressInstanceUrlSession`/`getWordpressInstanceUrlSessionOnce` APIs wrapping sessionStorage. This allows the login/signup page to communicate with the dashboard screen, without using a query param, which [could be unsafe](https://hashintel.slack.com/archives/C02TWBTT3ED/p1680513055932689)
- The dashboard page now calls `getWordpressInstanceUrlSessionOnce` on mount, and then conditionally shows `DashboardWordpressSection` if there is a wordpress URL in session storage
- The login/signup pages no longer immediately redirect you to `/dashboard` if you're already logged in, provided a verification code is present in the URL. This is to allow you get to the dashboard with wordpress card using the link-wordpress API, even if you're already logged in

N/A

## ⚠️ Known issues

N/A

## 🐾 Next steps

- Wordpress plugin UI https://app.asana.com/0/1204086167143889/1204260760205619/f – 

## 🛡 What tests cover this?

N/A

## ❓ How to test this?

Use the API created in https://github.com/blockprotocol/blockprotocol/pull/1231 both for a new and existing user, and ensure you see the wordpress card as in the video below when you open the URL you're given by the API. 

## 📹 Demo

The design has updated slightly since this. See https://hashintel.slack.com/archives/C02NW18QGUR/p1680517515214059

https://user-images.githubusercontent.com/719814/229494872-24808703-1944-484c-961e-a19b8909a20f.mov


apps/site/src/components/pages/dashboard/dashboard-wordpress-section.tsx
```diff
@@ -0,0 +1,213 @@
...
+import { FontAwesomeIcon } from "../../icons";
+import { Link } from "../../link";
+
+const defaultApiKey =
+  "b10ck5.9faa5da6664f7229999439d5433d4ac2.c549e923-c3b6-451f-baba-20b391e6033d";
+
+export const DashboardWordpressSection = ({
+  wordpressInstanceUrl,
+}: {
+  wordpressInstanceUrl: string;
```
CiaranMn, line 17: I suspect that we are going to store both
(a) the WP instance URL (e.g. https://example.com) and
(b) the WP BP plugin settings page to send them back to (https://example.com/?????)

One potential approach would be to send (b) when signing up, store the origin as `wordpressInstanceUrl` as we do now in the db, and store the full settings page URL in the db (but only need one value for it).

I don't think we can rely on the path to the settings page being the same across instances because people can change the admin path (defaults to `/wp-admin`).

I think you could look at this as part of doing the WP side of it though, as it'll be easier to check / iterate.

apps/site/src/components/pages/dashboard/dashboard-wordpress-section.tsx
```diff
@@ -0,0 +1,213 @@
...
+  };
+
+  const generateApiKey = () => {
+    if (generating) {
+      return;
+    }
+
+    setGenerating(true);
+    void apiClient
+      .generateApiKey({ displayName: "Wordpress" })
```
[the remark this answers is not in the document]
nathggns: I'll fix it in other parts of the codebase in another PR

apps/site/src/components/pages/dashboard/dashboard-wordpress-section.tsx
nathggns: I've done it as part of this PR.

apps/site/src/pages/api/link-wordpress.api.ts
```diff
@@ -7,6 +7,7 @@ import { formatErrors } from "../../util/api";
 export type ApiSignupRequestBody = {
```
CiaranMn, line 7: Just noticed that these types have the same name as in `signup.api.ts`. I think it's worth renaming these, but it could be done in the follow up PR.

apps/site/src/pages/api/link-wordpress.api.ts
```diff
@@ -20,18 +21,28 @@ export default createBaseHandler<ApiSignupRequestBody, ApiSignupResponse>()
...
       allow_fragments: false,
       allow_query_components: false,
     }),
   )
+  .use(
+    bodyValidator("wordpressSettingsUrl").isURL({
+      protocols: ["http", "https"],
+      require_protocol: true,
+      require_tld: process.env.NODE_ENV === "production",
+      allow_fragments: false,
```
CiaranMn, line 34: I assume we will remove fragments in the WordPress side before submitting to this API. If not I'd allow them and remove them here.

CiaranMn (approved):
Looks good – a couple of comments that don't block merging.

apps/site/src/pages/api/link-wordpress.api.ts
nathggns: Discussed offline. No use case for including fragments (hashes/anchors) in these URLs. They won't be sent by the wordpress plugin as they're never sent to the server anyway

CiaranMn (approved):

PR #1249, opened 2023-04-04
Rename `/@user/all-types` to `/@user/types`

yusufkinatas:

Changes profile tab slug for convenience. 

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203358502199087/1204015967491453/f) _(internal)_

## ❓ How to test this?

1.  Open `/@user/types` or manually open an user's types tab on their public profile.
2. Confirm that it's working as expected, and `@user/all-types` throws an 404 error.

## 📹 Demo

Screenshot 2023-04-04 at 12 18 43
Screenshot 2023-04-04 at 12 19 08


benwerner01:
Should we add a redirect in the Next config from `/@[user]/all-types` to `/@[user]/types`?

yusufkinatas:
> Should we add a redirect in the Next config from `/@[user]/all-types` to `/@[user]/types`?

There is no reference to `all-types` left in the codebase. Checked HASH and there is no ref there too.

benwerner01:
> > Should we add a redirect in the Next config from `/@[user]/all-types` to `/@[user]/types`?
> 
> There is no reference to `all-types` left in the codebase. Checked HASH and there is no ref there too.

I didn't mean for internal references, more for external references (such as blog posts, links from other websites, etc). It's also possible that the browser cache of a user auto-completes this URL, and it should be given a 301 redirect instead of a 404. I'm not confident any of these are going to happen, but I think it'd be good practice to add a redirect just in case.

yusufkinatas:
Hmm, that makes sense. Added redirection here https://github.com/blockprotocol/blockprotocol/pull/1249/commits/90ffaa6962b02bf72bd1bdf82187eae741210c6c

@benwerner01

benwerner01 (approved):
Thanks Yusuf, looks good to me 👍

nonparibus:
For future reference, what @benwerner01 suggests is best-practice here. In the future we can set up monitoring on redirects (which doesn't exist today) and deprecate any that really aren't being used at all, but accounting for the possibility of external inbound references (from across our multiple repos: internal, Þ, and #, as well as external sources) seems prudent.

PR #1251, opened 2023-04-05
BP design system

luisbettencourt:
## 🔗 Related links

- [Asana task](link) _(internal)_

## 🚫 Blocked by

- [ ] ...

## 🔍 What does this change?

- ...

- The docs for x need updating to explain that y

## 🛡 What tests cover this?

-

## ❓ How to test this?

1.  Try X
1.  Confirm that Y


nonparibus:
@luisbettencourt With today's internal decision, I think we _could_ move the `block-design-system` into the HASH repo. If you agree, please feel free to open this PR there instead. This should improve the DX for us internally of working with blocks, as it means the package we're relying on is in the same repo as the blocks.

I personally still think it would be nice to keep this separate from the existing `hash-design-system` which contains HASH brand assets, but @CiaranMn may disagree or think there is good reason not to bother having it separated out.

luisbettencourt:
Can't say I haven't thought about this already, but I'll wait for @CiaranMn's opinion before going ahead with that. A nice advantage of moving it there would be not having to duplicate the prepublish and postpublish scripts (and all the others which they are dependant on) to the blockprotocol repo.

CiaranMn:
Let's put it in the `hash` repo. We can try it being its own separate thing.

nonparibus:
At https://github.com/hashintel/hash/tree/main/libs/%40hashintel/block-design-system

PR #1254, opened 2023-04-07
Bump `wasm-pack` to version `0.11.0`

TimDiekmann:

We currently use `0.10.3` and the `0.11` comes with new features we likely want to use such as `--weak-refs` and a newer version of `wasm-opt`


Alfred-Mountfield (approved):
Tested locally, cheers!

Ran into an issue where it couldn't find `wasm-opt`, I needed to manually call 
- `cargo uninstall wasm-pack`
- `cargo uninstall wasm-opt`

But it worked on a fresh install so this just seems like migration pains and is likely only going to happen once

Issue #1185, opened 2023-03-16
Display RFCs on the Þ site alongside `spec` and `docs`


nonparibus:
We currently store a number of RFCs in the root of our [`blockprotocol/blockprotocol` repository](https://github.com/blockprotocol/blockprotocol) on GitHub, under a top-level [`rfcs` directory](https://github.com/blockprotocol/blockprotocol/tree/main/rfcs). We want to leave this where it currently resides. However, we want to ensure that the accepted RFCs within it (all of those RFC documents [available here](https://github.com/blockprotocol/blockprotocol/tree/main/rfcs/text)) are directly browsable on the Þ website, linked to from under the "RFCs" section of the [roadmap page](https://blockprotocol.org/docs/spec/roadmap#rfcs), and hosted at URLs such as `https://blockprotocol.org/docs/spec/roadmap/0352-graph-type-system` (in the case of the RFC located at `https://github.com/blockprotocol/blockprotocol/blob/main/rfcs/text/0352-graph-type-system.md`).

This work builds on the changes introduced in https://github.com/blockprotocol/blockprotocol/pull/1183

PR #1255, opened 2023-04-07
[BP-8] Display RFCs on the Þ site alongside `spec` and `docs`

gitstart-blockprotocol:
🌟 What is the purpose of this PR?
---
Display RFCs on the Þ site alongside spec and docs

🔗  Related links
---
- [Github](https://github.com/blockprotocol/blockprotocol/issues/1185)

❓How to test this?
---
https://www.loom.com/share/9c6a81c0c05c418c812e120d146ab55f

---
This code was written and reviewed by GitStart Community. Growing future engineers, one PR at a time.


nonparibus (changes requested):
The approach taken in this PR needs revising. We don't want to duplicate the contents of the [RFCs folder](https://github.com/blockprotocol/blockprotocol/tree/main/rfcs) within the `roadmap` directory (as seen here), but should instead ensure that it remains in-place (as a single source of truth which can never become "out of sync" with the live site).

rfcs/text/0_0352-graph-type-system.md
```diff
@@ -1841,7 +1841,7 @@ This would accept Entity instances with the following shape
 #### Specifying there is a List of Links
 
 The Entity Type can also express that it can have multiple outgoing links of the same type.
-Furthermore list of links can be constrainted with `minItems` and `maxItems`.
+Furthermore list of links can be constrained with `minItems` and `maxItems`.
```
nonparibus, line 1844: Nice spot. Thanks.

apps/site/src/util/mdx-utils.tsx
```diff
@@ -272,3 +281,172 @@ export const getAllPages = (params: {
     });
   });
 };
+
+const getRfcsPages = (): SiteMapPage[] => {
```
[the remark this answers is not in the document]
gitstart-blockprotocol: Okay, we'd get working on that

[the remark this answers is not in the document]
gitstart-blockprotocol: @CiaranMn, So we refactored but couldn't find a way to abstract away the functionality without introducing a lot of complexity. So we opted to abstract the setup and keep the object construction separate for the `getRoadmapSubPages`

rfcs/text/0352-graph-type-system.md
```diff
@@ -1,10 +1,3 @@
-- **Feature Name:** `graph-type-system`
-- **Start Date:** 2022-06-14
-- **RFC PR:** [blockprotocol/blockprotocol#352](https://github.com/blockprotocol/blockprotocol/pull/352)
-- **RFC Discussion:** [blockprotocol/blockprotocol#418](https://github.com/blockprotocol/blockprotocol/discussions/418)
-
-<!-- markdownlint-disable  MD036 -->
-
```
CiaranMn, line 7: This should not be deleted. It is information we want to be present as part of the RFC. Please ask if you unsure before deleting information from a document, it is probably there for a reason.

How does it look with the information present?

rfcs/text/0352-graph-type-system.md
```diff
@@ -116,12 +109,8 @@ Given the motivation above and some other implications of the current design, th
 
 As well as addressing these existing shortcomings, this RFC seeks to:
 
-<!-- markdownlint-disable  MD029 -->
-
-3.  define a constrained type system that is able to describe any JSON object structure
-4.  refine the current way of defining links between entities, encapsulating the approach within the type system
-
-<!-- markdownlint-enable  MD029 -->
```
CiaranMn, line 124: What prompted this change? These numbers were meant to be a continuation of the numbers above.

CiaranMn (changes requested):
The refactor to re-use code looks good, however the RFC MD files being moved to their original location has made it easier to spot that information has been deleted from them. We don't want content deleted from files in order to display them. 

There are also some linting errors related to the format of the MD files which should disappear when the deleted content is restored, although worth checking before requesting review again. Thanks!

rfcs/text/0352-graph-type-system.md
gitstart-blockprotocol: Hello @CiaranMn, Note has been taken. The reason it was deleted though was that it wasn't being rendered, an error was continuously thrown and prevented the application from launching

rfcs/text/0352-graph-type-system.md
gitstart-blockprotocol: Screenshot 2023-04-20 at 12 30 10

CiaranMn: The problem is the HTML comment lines disabling markdownlint rules. As the error message says, comments in MDX are expected to be in a different syntax.

I can think of two options here:
1. Rename the files to `.mdx` – let's try and avoid that for now as it will probably have wider implications
2. Deal with the `markdownlint` disable rules some other how, so we don't have to include HTML comments in the file. I suggest trying this first as it will hopefully be simpler.

You can disable `markdownlint` rules via config files as well as comments in the doc. We have one such file, `.markdownlint.json`, in the root of the repo. I suggest that we don't disable the rules across the whole repo, though, as they are still useful to have. Let's try adding a new `.markdownlint.json` file inside `rfcs` that just disables the rules we are currently disabling via HTML comments. 

You should be able to extend the root config file, and add the two new disable options – see [docs here](https://github.com/DavidAnson/markdownlint#optionsconfig)

Please let me know if you have any issues.

gitstart-blockprotocol: Hello @CiaranMn, the second option has been implemented but it doesn't serve the function, the command `lint:markdownlint` completely disregards the `.markdownlint.json` config file in the `rfcs` folder

CiaranMn: I've pushed a change [here](https://github.com/blockprotocol/blockprotocol/pull/1255/files/02d4868b9b206d6e85114a8a49ee4f1e0864ca1e..a7a08585c9b5d7765567966ad8b1f7932cb98904) to replace the CLI used to run `markdownlint`. The previous one wasn't checking nested config files, the new one does.

Please restore the content deleted previously and then it should be good to review again.

CiaranMn: I pushed https://github.com/blockprotocol/blockprotocol/pull/1255/commits/edd7efd81d09c872c956073d08c2def103ec668b to fix a linting issue

CiaranMn (approved):
Apologies for the delay in finalising this one – looks good, thank you! I'll get it merged

PR #1259, opened 2023-04-12
Add files for self-hosting esm.sh, make URL configurable

CiaranMn:

We use esm.sh to load dependencies for blocks in the demo sandbox, e.g. the [Code block](https://blockprotocol.org/@hash/blocks/code).

esm.sh is a great project, but it has some stability issues. Updates occasionally mean that some imports fail, which means that the block preview fails to load, breaking both user-facing pages and our CI (which test the block pages).

This PR introduces a configurable URL for the esm.sh server and files for deploying our own, pinned to a known-working version. This is already deployed to and updated to default to https://esm.blocksandbox.org – the environment variable allows for testing other/local deployments of esm.sh, including reverting to esm.sh itself if there's an issue with our server.

This _might_ fix the fact that block previews don't work under certain security settings (due to mysterious sessionStorage access attempts), but as this only happens in production we won't know for sure until it's merged.

**Note**
This PR intentionally does not change links in the HTML template which point to esm.sh. I'm clear at this point what the future of our esm.sh server is (we may rewrite the sandbox altogether to not load ESM modules in this way), and don't want to introduce external dependencies on it.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203193968491715/1204086165607422/f) _(internal)_

## 🐾 Next steps

- Once deployed to prod, see if it has fixed the incognito issue.

## 🛡 What tests cover this?

- Existing block page integration tests.

## ❓ How to test this?

1.  View the deployment
2. Check that hub pages work


CiaranMn:
Switched the origin of the server, need to update the SSL certificate

CiaranMn:
Closing this for now – to be resurrected if issues with esm.sh arise again

vilkinsons:
We've now archived https://github.com/hashdeps/esm.sh

PR #1261, opened 2023-04-13
Block template react scss cleanup

luisbettencourt:

This PR strips down the contents of the scss file of `block-template-react` to include only the styles useful for the Hello World example.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1203179076056186/1204069980884590/f) _(internal)_

## 🔍 What does this change?

- Removed variables;
- Made styles apply to classes instead of html elements;
- Removed styles not useful for the Hello World example.

This PR:

- [ ] I am unsure / need advice


libs/block-template-react/src/base.module.scss
```diff
@@ -1,251 +1,45 @@
...
 
+.block {
   /* –––––––––– Helper styles –––––––––– */
 
-  font-family: var(--font-family);
-  color: var(--primary-text-base);
-
-  /* Typography */
+  font-family: system-ui, "Inter", "Open Sans", "Helvetica Neue", "Arial",
+    sans-serif;
```
CiaranMn, line 9: We don't load these fonts so I'd just get rid of them
```suggestion
  font-family: system-ui, "Helvetica Neue", "Arial", sans-serif;
```
```

luisbettencourt: Done in 37b0a92d5d1fe8df92ccd12b4509ec0d28d6a8b3

CiaranMn: See suggestion above

luisbettencourt: I misunderstood, thank you. 7b2f14a38eb4968cb6b76d2560a15656eea3a637

CiaranMn (approved):

PR #1264, opened 2023-04-14
Fix determinism of block codegen

Alfred-Mountfield:

The codegen utility for blocks has some logic for ensuring that generated type names are unique. It first differentiates by title, then kind (entity type, property type etc.), and then starts an incrementing counter. The incrementing counter lead to some non-determinism in codegen outputs (which is problematic for CI checks) as depending on the order the types were encountered, one type with a given title might come before another type of the same kind with the same title, whereas sometimes their orders might be reversed.

This is partially demonstrated with two clashing `Name` types in a commit in this PR (although testing and proving it is unfortunately not straight forward as you just need to rerun `codegen` until it happens).

This PR fixes this by sorting by base URL when establishing the incrementing counter.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1202805690238892/1204258875913354/f) _(internal)_

## 🚫 Blocked by

- N/A

## 🔍 What does this change?

See commits

This PR:

No

## ⚠️ Known issues

N/A

## 🐾 Next steps

More codegen improvements

## 🛡 What tests cover this?

Both block templates call codegen, so this should at least ensure they aren't broken

## ❓ How to test this?

1.  Checkout [Re-order deterministically](https://github.com/blockprotocol/blockprotocol/commit/3ed2bdc2e284680934c267ed1d4644bbb3886537) and play around with it

## 📹 Demo

This is difficult to demo, but see [Re-order deterministically](https://github.com/blockprotocol/blockprotocol/commit/3ed2bdc2e284680934c267ed1d4644bbb3886537)

### DIFF
--- a/.changeset/old-clocks-sneeze.md
+++ b/.changeset/old-clocks-sneeze.md
@@ -1,0 +1,5 @@
+---
+"@blockprotocol/graph": minor
+---
+
+Fix determinism in codegen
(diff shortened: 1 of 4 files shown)


thehabbos007 (dismissed):
LGTM :rock: :rocket:

thehabbos007 (approved):

PR #1257, opened 2023-04-10
Update API key page to allow users to create multiple keys

yusufkinatas:
## 🌟 What is the purpose of this PR?

Update the API keys panel on profile to let users create / edit / revoke multiple API keys instead of just having one API key.

## 🔗 Related links

- [Asana task 1](https://app.asana.com/0/1204086167143892/1204270888519633/f) _(internal)_
- [Asana task 2](https://app.asana.com/0/1204086167143892/1204270888519624/f) _(internal)_
- [Asana task minor bugfix](https://app.asana.com/0/1204086167143892/1204345535567014/f) _(internal)_
- [Design](https://www.figma.com/file/Fpmde3E9nJMb4VLkPTYgrM/blockprotocol.org?node-id=4619%3A31258&t=J5773M1n52ElnUv3-1) _(internal)_

## 🔍 What does this change?

- Create new API keys panel
- Fix `Button` component changing width-height while loading
- Create `CopyToClipboardButton` component
- Remove unused components used in `api-keys-panel`
- Adjust vertical padding of `size="small"` button on MUI theme 

## ❓ How to test this?

1.  Checkout the branch / view the deployment
1.  Open API Keys panel
1.  Compare it with design, and confirm it's working as expected.

## 📹 Demo

https://user-images.githubusercontent.com/39553853/231387355-0d49afcb-08e8-4ef4-aa0e-7f04877d2975.mov

### Before the button component `loading` bugfix
https://user-images.githubusercontent.com/39553853/231388229-8911ff64-99a3-41db-9d43-409776d7140f.mov

### After
https://user-images.githubusercontent.com/39553853/231388221-afc180a3-d626-4f9d-9d25-fc742b3b9253.mov

PR #1267, opened 2023-04-18
Improve performance on API keys page

nathggns:
There's still lots to do here (which I've tried to leave comments for), but this indicates sort of what you need to do – storing everything in a single context means every component (and all its children) need to re-render whenever you do any interaction. On my computer, where I have lots of API keys, that's regularly taking 100ms-200ms to do.


nathggns:
This is related to https://github.com/blockprotocol/blockprotocol/pull/1257

nathggns:
Leaving as draft as probably not to be merged directly – just an indication of the approach

nathggns:
This has been superseded by https://github.com/blockprotocol/blockprotocol/pull/1271

PR #1286, opened 2023-05-15
Manually install `just` instead of using `just-install`

TimDiekmann:

It's likely to be rate-limited in CI when running `just-install`. This manually installs just instead.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204274556967336/1204486732268406/f) _(internal)_

This PR:

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "check-dependency-version-consistency": "3.0.3",
     "dotenv-flow": "3.2.0",
     "husky": "8.0.3",
-    "just-install": "^1.0.11",
     "knip": "^1.6.1",
     "lint-staged": "13.1.0",
     "lockfile-lint": "4.10.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -50,7 +50,6 @@
     "check-dependency-version-consistency": "3.0.3",
     "dotenv-flow": "3.2.0",
     "husky": "8.0.3",
-    "just-install": "^1.0.11",
```
thehabbos007, line 53: I suppose people will need to install it manually now that it isn't installed with `yarn`

thehabbos007 (approved):
LGTM. We should just quickly consider the comment below

package.json
TimDiekmann: Yes, this is sadly required 😞 (we considered this in the linked slack thread)

PR #1288, opened 2023-05-15
Provide a vercel-install script

TimDiekmann:

Vercel limits the length of the install script to 256 characters. To install multiple dependencies, we easily hit the limit. This adds a script, which does the install step

## 🔗 Related links

- #1286 

This PR:

---
  - https://app.asana.com/0/0/1204596705932049

### DIFF
--- a/.github/vercel.install.sh
+++ b/.github/vercel.install.sh
@@ -3,7 +3,7 @@
 set -euxo pipefail
 
 # Change to the root directory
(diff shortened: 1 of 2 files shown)


thehabbos007 (approved):
Thank you. Builds are passing, we should look into splitting the build configs out like we do in the HASH repo, e.g. https://github.com/hashintel/hash/blob/main/apps/hashdotdev/vercel.json

PR #1289, opened 2023-05-15
Provide a `vercel.json` per project

TimDiekmann:

Similar to what we have at HASH this adds `vercel.json` per project to persist commands in VCS and being able to run different commands per branch. This is a direct follow-up of #1288 and as a result is blocked on it.

## 🔗 Related links

- [Asana Task](https://app.asana.com/0/1204315348455207/1204596705932063/f) (_internal_)

## 🚫 Blocked by

- #1288

### DIFF
--- a/apps/site/vercel-install.sh
+++ b/apps/site/vercel-install.sh
@@ -3,7 +3,7 @@
 set -euxo pipefail
 
 # Change to the root directory
-pushd "$(dirname "$0")/.."
+pushd "$(dirname "$0")/../.."
 
 # Install dependencies
 curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | sh -s -- --to /usr/local/bin
--- a/apps/site/vercel-install.sh
+++ b/apps/site/vercel-install.sh
@@ -3,7 +3,7 @@
 set -euxo pipefail
 
 # Change to the root directory
-pushd "$(dirname "$0")/../.."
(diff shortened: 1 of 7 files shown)


apps/site/vercel-install.sh
```diff
@@ -3,7 +3,7 @@
 set -euxo pipefail
 
 # Change to the root directory
-pushd "$(dirname "$0")/.."
+pushd "$(dirname "$0")/../.."
```
thehabbos007, line 6: Wild idea, but would `git rev-parse --show-toplevel` work instead of us having to put the path? :smile:

libs/@blockprotocol/design-system/vercel-install.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -euxo pipefail
+
+# Change to the root directory
+pushd "$(dirname "$0")/../../.."
```
thehabbos007, line 6: Same here if above can be done :+1:

apps/site/vercel-install.sh
TimDiekmann: Great idea, I wish it was mine 😄 
[Use git rev-parse](https://github.com/blockprotocol/blockprotocol/pull/1289/commits/7d86e43d2823837bbe10105f23ceb708397b5e2a)

thehabbos007 (approved):

PR #1291, opened 2023-05-17
Provide a `repr::EntityType::merge_parent` method

TimDiekmann:

For a closed entity type it's required to merge an entity type with its parent.

## 🔗 Related links

- [Asana task](https://app.asana.com/0/1204315348455207/1204596705932094/f) _(internal)_
- [Slack thread](https://hashintel.slack.com/archives/C03TWBSRT8B/p1684314549509999) _(internal)_

## 🔍 What does this change?

- Add a `merge_parent` method, the function is documented 

This PR:

- [x] does not modify any publishable blocks or libraries
- [ ] I am unsure / need advice

A documentation for the method was added

## 🛡 What tests cover this?

- A simple test case has been added

### DIFF
--- a/libs/@blockprotocol/type-system/crate/src/ontology/entity_type/repr.rs
+++ b/libs/@blockprotocol/type-system/crate/src/ontology/entity_type/repr.rs
@@ -1,10 +1,14 @@
-use std::{collections::HashMap, str::FromStr};
+use std::{
+    collections::{HashMap, HashSet},
+    str::FromStr,
+};
 
 use serde::{Deserialize, Serialize};
 #[cfg(target_arch = "wasm32")]
 use tsify::Tsify;
 
 use crate::{
+    ontology::entity_type::error::MergeEntityTypeError,
     repr,
     url::{BaseUrl, ParseVersionedUrlError, VersionedUrl},
     ParseEntityTypeError,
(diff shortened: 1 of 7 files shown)


libs/@blockprotocol/type-system/crate/src/ontology/entity_type/repr.rs
```diff
@@ -52,6 +56,54 @@ pub struct EntityType {
...
+    ///
+    /// This will:
+    ///   - remove the other entity type from the `allOf`
+    ///   - merge the `properties` and `required` fields
+    ///   - merge the `links` field
+    ///
+    /// # Notes
+    ///
+    /// - This does not validate the resulting entity type.
+    /// - The `required` field may have a different order after merging.
```
thehabbos007, line 70: Follow-up from above. I like this wording better :)

thehabbos007 (commented):
Looks mostly good to me. Just have a comment about how we talk about this operation :)

libs/@blockprotocol/type-system/crate/src/ontology/entity_type/repr.rs
TimDiekmann: The "This will:" section was mainly written by Copilot, it's better than me 🙃

thehabbos007 (approved):

PR #1298, opened 2023-06-19
Fix schema links

TimDiekmann:

Targets #428 as a separated PR to allow easier reviews

### DIFF
--- a/rfcs/text/0428-extending-and-duplicating-types.md
+++ b/rfcs/text/0428-extending-and-duplicating-types.md
@@ -986,8 +986,6 @@ While various implementations likely will want to implement checking for the sim
 
 - We haven't specified how projecting/selecting properties of a supertype from a subtype instance is possible. It is an open question how we actually pick out the exact properties of a subtype to provide a valid supertype instance in embedding applications.
   - This may be especially difficult if a property type is defined as being an array of `oneOf` a given set of property type objects. Detecting which sub-schema applies to a given array element when accounting for dropped properties, **may** be a very difficult task.
-- Type extension might prove to have strange side-effects when considering link types which extend from one another.
-  - Say `Person` defines `Knows` and `Has Friend` links. If `Has Friend` extends `Knows`, then technically any `Has Friend` instance is also a `Knows` instance, and would need to satisfy the link constraints set on both. This has implications for `minItems` and `maxItems`, as well as for link destinations. If `Person` says it only `Knows` other `Person` entities, but `Has Friend` links to both `Person` and `Pet` entities, then a `Has Friend` link to a `Pet` entity would break the constraint set for `Knows`.
 
 # Future possibilities
 
--- a/rfcs/text/0428-extending-and-duplicating-types.md
+++ b/rfcs/text/0428-extending-and-duplicating-types.md
@@ -206,6 +206,14 @@ Furthermore, resolving the intentions of the existing definition for the message
 
 We propose that we should treat type updates as complete replacements, so implementation is much more straightforward for embedding applications. Partial schema updates also add some level of indirection, and that may obfuscate error sources and error reasons. Therefore, updates to types must be complete replacements rather than partial schema updates.
 
+### Changes to the link type system
+
+The discussed extension to the type system has implications on link types which extend from each other. Given the above `Person` and saying it defines `Knows` and `Has Friend` links, and that those links are **ordered**. Currently, the order of a link is defined in a given direction, e.g. _left-to-right_, and has a single value. With the inheritance of link types this would presently imply, that the order of the `Knows` link _within the `Knows` links list_ would have to be the same as the order of the `Has Friend` link _within the `Has Friend` links list_.
+
+**To mitigate this problem, we decided to remove link ordering**. This means, that links will no longer have an inherent order and no assumption is being made of the order.
+
+This change will affect any code that relies on the order property of link types. If applications have such code, they will need to update it to use a different mechanism for ordering links. A possibility is to add an ordering property to the link type itself. This makes the type-system itself easier to reason about whereas it will also allow application to use other types for their ordering instead of being forced to use integral values.
+
 # Reference-level explanation
 
 [reference-level-explanation]: #reference-level-explanation
@@ -980,7 +988,6 @@ While various implementations likely will want to implement checking for the sim
   - This may be especially difficult if a property type is defined as being an array of `oneOf` a given set of property type objects. Detecting which sub-schema applies to a given array element when accounting for dropped properties, **may** be a very difficult task.
 - Type extension might prove to have strange side-effects when considering link types which extend from one another.
   - Say `Person` defines `Knows` and `Has Friend` links. If `Has Friend` extends `Knows`, then technically any `Has Friend` instance is also a `Knows` instance, and would need to satisfy the link constraints set on both. This has implications for `minItems` and `maxItems`, as well as for link destinations. If `Person` says it only `Knows` other `Person` entities, but `Has Friend` links to both `Person` and `Pet` entities, then a `Has Friend` link to a `Pet` entity would break the constraint set for `Knows`.
-  - Say `Person` defines `Knows` and `Has Friend` links, and that those links are **ordered**. At the moment, the order of a link is defined in a given direction (e.g. left to right) and has a single value. That would presently imply that the order of the `Knows` link _within the `Knows` links list_ would have to be the same as the order of the `Has Friend` link _within the `Has Friend` links list_.
 
 # Future possibilities
 
--- a/rfcs/text/0428-extending-and-duplicating-types.md
+++ b/rfcs/text/0428-extending-and-duplicating-types.md
@@ -218,7 +218,7 @@ We'll add the following fields to the existing Entity Type meta schema definitio
 
 ```json
 {
-  "$id": "https://blockprotocol.org/types/modules/graph/0.3.1/schema/entity-type",
+  "$id": "https://blockprotocol.org/types/modules/graph/0.3/schema/entity-type",
   "type": "object",
   ...,
   "properties": {
@@ -230,7 +230,7 @@ We'll add the following fields to the existing Entity Type meta schema definitio
         "properties": {
           "$ref": {
             "$comment": "Valid reference to an existing Entity Type version",
-            "$ref": "https://blockprotocol.org/types/modules/graph/0.3.1/schema/versioned-url"
+            "$ref": "https://blockprotocol.org/types/modules/graph/0.3/schema/versioned-url"
           },
           "additionalProperties": false
         }
@@ -248,7 +248,7 @@ Given a _supertype_ `Person`:
 
 ```json
 {
-  "$schema": "https://blockprotocol.org/types/modules/graph/0.3.1/schema/entity-type",
+  "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/entity-type",
   "kind": "entityType",
   "$id": "https://blockprotocol.org/@alice/entity-type/person/v/1",
   "type": "object",
@@ -272,7 +272,7 @@ and a _subtype_ `Employee`:
 
 ```json
 {
-  "$schema": "https://blockprotocol.org/types/modules/graph/0.3.1/schema/entity-type",
+  "$schema": "https://blockprotocol.org/types/modules/graph/0.3/schema/entity-type",
   "kind": "entityType",
   "$id": "https://blockprotocol.org/@alice/entity-type/employee/v/1",
   "type": "object",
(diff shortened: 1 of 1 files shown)


rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -206,19 +206,27 @@ Furthermore, resolving the intentions of the existing definition for the message
...
 
-In the Block Protocol, we will allow type extension through the `allOf` JSON Schema keyword which applies all constraints from the sub-schemas within the `allOf` array..
+In the Block Protocol, we will allow type extension through the `allOf` JSON Schema keyword which applies all constraints from the sub-schemas within the `allOf` array.
 
 We'll add the following fields to the existing Entity Type meta schema definition:
 
 ```json
 {
-  "$id": "https://blockprotocol.org/types/modules/graph/0.3.1/schema/entity-type",
+  "$id": "https://blockprotocol.org/types/modules/graph/0.3/schema/entity-type",
```
Alfred-Mountfield, line 229: The patch version was on purpose. The RFC would result in a new version of the spec. Although it was previously backwards compatible, but now with the removal of link ordering it will not be, so we might want to change this to `0.4`

Alfred-Mountfield (changes requested):

rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -206,19 +206,27 @@ Furthermore, resolving the intentions of the existing definition for the message
 
 We propose that we should treat type updates as complete replacements, so implementation is much more straightforward for embedding applications. Partial schema updates also add some level of indirection, and that may obfuscate error sources and error reasons. Therefore, updates to types must be complete replacements rather than partial schema updates.
 
+### Changes to the link type system
+
+The discussed extension to the type system has implications on link types which extend from each other. Given the above `Person` and saying it defines `Knows` and `Has Friend` links, and that those links are **ordered**. Currently, the order of a link is defined in a given direction, e.g. _left-to-right_, and has a single value. With the inheritance of link types this would presently imply, that the order of the `Knows` link _within the `Knows` links list_ would have to be the same as the order of the `Has Friend` link _within the `Has Friend` links list_.
```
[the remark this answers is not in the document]
TimDiekmann: Thanks for extending the section. I applied your changes (even though, GH did not really display them correctly 😆) and added a section on the other issues we encountered: [`7748d48` (#1298)](https://github.com/blockprotocol/blockprotocol/pull/1298/commits/7748d48580e597b5f2edd7e436a8911394734de0)

rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -206,19 +206,95 @@ Furthermore, resolving the intentions of the existing definition for the message
 
 We propose that we should treat type updates as complete replacements, so implementation is much more straightforward for embedding applications. Partial schema updates also add some level of indirection, and that may obfuscate error sources and error reasons. Therefore, updates to types must be complete replacements rather than partial schema updates.
 
+### Removing link ordering
+
+Adding a type extension mechanism causes issues with the behaviour of links in situations where an entity type refers to multiple link types where one extends the other.
+
```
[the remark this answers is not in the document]
TimDiekmann: I added this to the section as well: [`fac523f` (#1298)](https://github.com/blockprotocol/blockprotocol/pull/1298/commits/fac523fb72686c304db478a3fb407cfe26e324eb)

rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -206,19 +206,95 @@ Furthermore, resolving the intentions of the existing definition for the message
...
+```ts
+HasFriendCharlie.leftToRightOrder = 1;
+HasFriendDave.leftToRightOrder = 0;
+```
+
+Giving us:
+
+```ts
+getHasFriendLinks(alice); // [ HasFriendDave, HasFriendCharlie ]
+```
```
thehabbos007, line 272: A bit of a nit, but we dont have other examples like these in the RFC where we usually just write plain JSON. Any reason why we opted to use pseudeo-code blocks for this instead of writing out explicitly how the data looks?

rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -978,9 +1053,6 @@ While various implementations likely will want to implement checking for the sim
 
 - We haven't specified how projecting/selecting properties of a supertype from a subtype instance is possible. It is an open question how we actually pick out the exact properties of a subtype to provide a valid supertype instance in embedding applications.
```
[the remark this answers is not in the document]
TimDiekmann: I added the intention as a bullet point below: [`d5d8639` (#1298)](https://github.com/blockprotocol/blockprotocol/pull/1298/commits/d5d8639f63f7ea3e80d5a53e15b370296df0b2ed)

rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -206,19 +206,95 @@ Furthermore, resolving the intentions of the existing definition for the message
...
+- Furthermore, there is a data modelling question as to if links can be ordered by a single property. There are many cases where the ordering might be use case dependent (as opposed to something intrinsic in the data), which is already partially supported by the `orderBy` field on the `query` methods. Various data models may also wish to have multi-tiered ordering, e.g. ordering first by one property and then secondly by another. As the current implementation does not account for all such use cases, and fractional indexing would not either, we opt to remove it for now and allow users to order links themselves at the application level. Should this decision cause problems, ordering can be added back in the future, with a design that takes issues into consideration including those posed by inheritance.
```
thehabbos007, line 283:
```suggestion
- Furthermore, there is a data modeling question regarding whether links can be ordered based on a single property. There are many cases where the ordering could be dependent on the specific use case (as opposed to something intrinsic in the data), which is already partially supported by the `orderBy` field on the `query` methods. Various data models may also wish to have multi-tiered ordering, e.g. ordering first by one property and then secondly by another. As the current implementation does not account for all such use cases, and fractional indexing would not either, we opt to remove it for now and allow users to order links themselves at the application level. Should this decision cause problems, ordering can be added back in the future, with a design that takes issues into consideration including those posed by inheritance.
```

rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -206,19 +206,95 @@ Furthermore, resolving the intentions of the existing definition for the message
...
+
+- integer-based indexing is not very flexible. For example, if you want to add a link to the middle of the list, you'd have to renumber all the indices after the insertion point. To mitigate this issue, it's possible to use fractional indexing. However, some links may should not be ordered by the same index but rely on a different ordering mechanism. For example, a `HasFriend` link may be ordered by the date the friendship was established, while a `Knows` link may be ordered by the date the acquaintance was established. Even ordering by multiple properties or other factors may be desirable.
+- assuming an application, which is implementing the type system, supports collaborative editing, it's possible that two users will add links to the same entity at the same time. This will result in a conflict, which will be difficult to resolve.
+
+These issues (and there are probably more) are the reason why we're going to remove link ordering from the Block Protocol type system.
+
 # Reference-level explanation
```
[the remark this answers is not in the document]
TimDiekmann: I added it to the section: [`ea82692` (#1298)](https://github.com/blockprotocol/blockprotocol/pull/1298/commits/ea8269204ee52ad39dd4b460d8dec1aa9c7dc66b)

thehabbos007 (commented):
I have a couple comments, will just drop them here while I look further :)

rfcs/text/0428-extending-and-duplicating-types.md
TimDiekmann: I can imagine that using JSON objects here everywhere will get pretty verbose. cc @Alfred-Mountfield

thehabbos007: I don't mean that we should use the full JSON objects, but just that these operations are sort of an unknown/new syntax at this point and that we'd have to be a little more explicit about what it means

rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -206,19 +206,95 @@ Furthermore, resolving the intentions of the existing definition for the message
...
+
+```ts
+getHasFriendLinks(alice); // [ HasFriendDave, HasFriendCharlie ]
+```
+
+But what about the `getKnowsLinks`? There is only one `leftToRightOrder` on each link entity, so the order will apply to _both_ lists, which could result in conflicting indices. Changing the order (index) of one of the elements in one collection will affect its position in the other.
+
+#### Other issues with link ordering
+
+The issue above caused us to revisit the thinking around link ordering, and is the instigator for its removal being included in this RFC. However, there are a number of other shortcomings in the current implementation, bolstering the decision to remove it (as opposed to iterating on it):
```
thehabbos007, line 278:
```suggestion
The issue above caused us to revisit the thinking around link ordering, and is the motivator for removing link ordering in this RFC. However, there are a number of other shortcomings in the current implementation, bolstering the decision to remove it (as opposed to iterating on it):
```
I think `instigator` is usually a person? Perhaps we should change this

rfcs/text/0428-extending-and-duplicating-types.md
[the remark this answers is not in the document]
Alfred-Mountfield: Can we make it part of the same bullet point, or a sub-bullet? At the moment it reads like a _distinct_ 'unresolved question'

rfcs/text/0428-extending-and-duplicating-types.md
Alfred-Mountfield: I had written it in the comment in pseudo code as part of an explanation of what I think we needed to add. I didn't necessarily mean we should copy paste it into the RFC 😅 

I agree that having big JSON objects would be far too verbose, so I'm not sure how we go about representing this in a concise manner. Perhaps there is a more intuitive syntax we can use, or perhaps there's a more explainable one as you say @thehabbos007

rfcs/text/0428-extending-and-duplicating-types.md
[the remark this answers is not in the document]
TimDiekmann: I updated the part for the `links` section as well: [`ffeee06` (#1298)](https://github.com/blockprotocol/blockprotocol/pull/1298/commits/ffeee062342bc5c25fde70b76f1febfe77338dd2) I also mentioned the `graph-module` as `createEntity` and `updateEntity` has to be changed slightly: [`039ef50` (#1298)](https://github.com/blockprotocol/blockprotocol/pull/1298/commits/039ef5056b9e3ff68488070a163b1ec1a887c865)

rfcs/text/0428-extending-and-duplicating-types.md
TimDiekmann: What about plain text? 🙃

rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -776,6 +851,123 @@ The above change applies to these Block Protocol operations (`updateEntityType`
 
 which must all make use of complete schemas (absent of `$id`) in place of partial ones. These examples originate from the [Type System RFC](./0352-graph-type-system.md#interfacing-with-types-1).
 
+## Removing link ordering
+
+The [current meta schema for `Entity Type`](https://blockprotocol.org/types/modules/graph/0.4/schema/entity-type) allows a `ordered` property inside of `links` and [current meta schema for `Link Data`](https://blockprotocol.org/types/modules/graph/0.3/schema/link-data) includes the [link ordering schema](https://blockprotocol.org/types/modules/graph/0.3/schema/link-orders) as properties, which consists of `leftToRightOrder` and `rightToLeftOrder`. With the removal of link ordering, these will be removed in the next version of the meta schema, so the next version of the `Entity Type` meta schema will look like this (the changes to the `allOf` property from above were omitted):
```
Alfred-Mountfield, line 856: Just a few nits, the link data and link orders are only schemas, they're not metaschemas (as they don't describe a schema):

```suggestion
The [current metaschema for `Entity Type`](https://blockprotocol.org/types/modules/graph/0.4/schema/entity-type) allows an `ordered` property inside of `links` and [current schema for `Link Data`](https://blockprotocol.org/types/modules/graph/0.3/schema/link-data) includes the [link ordering schema](https://blockprotocol.org/types/modules/graph/0.3/schema/link-orders) as properties, which consists of `leftToRightOrder` and `rightToLeftOrder`. With the removal of link ordering, these will be removed in the next version of the metaschema, so the next version of the `Entity Type` metaschema will look like this (the changes to the `allOf` property from above were omitted):
```

rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -776,6 +851,123 @@ The above change applies to these Block Protocol operations (`updateEntityType`
...
+      }
+      // ...
+    }
+    // ...
+  }
+  // ...
+}
+```
+
+and the next version of the `Link Data` meta schema will become this:
```
Alfred-Mountfield, line 895:
```suggestion
and the next version of the `Link Data` metaschema will become this:
```

rfcs/text/0428-extending-and-duplicating-types.md
```diff
@@ -776,6 +851,123 @@ The above change applies to these Block Protocol operations (`updateEntityType`
...
+      }
+      // ...
+    }
+    // ...
+  }
+  // ...
+}
+```
+
+and the next version of the `Link Data` metaschema will become this:
```
TimDiekmann, line 895:
```suggestion
and the next version of the `Link Data` schema will become this:
```

Alfred-Mountfield (approved):

PR #1311, opened 2023-08-03
Add suffix to name of auto-generated data types

CiaranMn:

When generating TypeScript types from ontology type schemas, we want to avoid giving identical names to different types.

Because an entity type and a data type did not have any suffix on their name in the existing code, entity types and data types with identical titles (e.g. `Text`) could end up with identical type names.

This PR fixes that by giving a `DataType` suffix to data types, e.g. `TextDataType`.

This PR:


benwerner01 (approved):
Thanks Ciaran, looks good to me 👍

PR #1330, opened 2023-08-31
Remove debug logs, fix type RegExp check

CiaranMn:

Remove debug logs added in #1329 – the issue was a redirect on the domain rather than anything in the application logic. Block pages are working again.

Also fixes a RegExp check which was updated in the same PR to avoid false positives on NextJS 'JSON for page' fetches, but missed a check.


CiaranMn:
Force merged as CI was failling due to the bug in production fixed by this PR

PR #1349, opened 2023-11-20
H-1345: Refactor non-temporal BP methods to always have non-temporal returns

benwerner01:

This PR refactors several non-temporal BP methods to always have non-temporal return types, even when they are passed temporal subgraphs. This fixes blocks that are using non-temporal methods when they receive temporal subgraph from their EA.

This PR also adds previous HASH app patches to the `@blockprotocol/graph` package.

## 🔗 Related links

- H-1345

## 🔍 What does this change?

- adds a `forceNonTemporal` optional parameter to the 5 functions in `link-entity.ts`, and sets these to true in `non-temporal/stdlib.ts`
- adds a `rewriteTypeId` parameter to the codegen script (previously patched in the HASH App)
- handles edges with no temporal intersection gracefully in `getIncomingLinksForEntity`/`getOutgoingLinksForEntity` (previously patched for `getOutgoingLinksForEntity` in the HASH app)

This PR:

- [x] modifies an **npm**-publishable library and **I have added a changeset file(s)**=

The changes in this PR:

The changes in this PR:

## 🛡 What tests cover this?

None

## ❓ How to test this?

One way of testing this would be to:
1. Use `yarn workspace @blockprotocol/graph link` to make the package available locally
2. Use `yarn workspace @blocks/address link @blockprotocol/graph` to link the package to the Address block in the HASH app, which requires the changes to work
3. Use the address block and refresh the page, which will pass a temporal subgraph to `useEntitySubgraph` and break if a temporal return type is provided instead instead of the non-temporal return, and should work if the non-temporal return is returned.

## 📹 Demo

image

### DIFF
--- a/libs/@blockprotocol/graph/src/codegen/parameters.ts
+++ b/libs/@blockprotocol/graph/src/codegen/parameters.ts
@@ -23,6 +23,7 @@ export type CodegenParameters = {
   typeNameOverrides?: {
     [sourceTypeId: string]: string;
   };
+  rewriteTypeId?: (typeId: VersionedUrl) => VersionedUrl;
   /** Generate look-up maps with aliases for all type URLs */
   typeIdAliases?:
     | { enabled: false }
@@ -174,8 +175,9 @@ export const validateCodegenParameters = (
 
 export type ProcessedCodegenParameters = Omit<
   Required<CodegenParameters>,
-  "targets"
+  "targets" | "rewriteTypeId"
 > & {
+  rewriteTypeId: CodegenParameters["rewriteTypeId"];
   targets: {
     [fileName: string]: {
       sourceTypeIds: VersionedUrl[];
@@ -226,6 +228,7 @@ export const processCodegenParameters = (
 
   return {
     ...parameters,
+    rewriteTypeId: parameters.rewriteTypeId,
     targets,
     typeIdAliases: parameters.typeIdAliases ?? { enabled: true },
     typeNameOverrides: parameters.typeNameOverrides ?? {},
--- a/libs/@blockprotocol/graph/src/codegen/parameters.ts
+++ b/libs/@blockprotocol/graph/src/codegen/parameters.ts
@@ -23,7 +23,12 @@ export type CodegenParameters = {
   typeNameOverrides?: {
     [sourceTypeId: string]: string;
   };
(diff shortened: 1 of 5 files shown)


libs/@blockprotocol/graph/src/shared/stdlib/subgraph/edge/link-entity.ts
```diff
@@ -187,12 +184,7 @@ export const getIncomingLinksForEntity = <Temporal extends boolean>(
...
 
             if (intersection === null) {
-              throw new Error(
-                `No entity revision was found which overlapped the given edge, subgraph was likely malformed.\n` +
-                  `EntityId: ${linkEntityId}\n` +
-                  `Search Interval: ${JSON.stringify(searchInterval)}\n` +
-                  `Edge Valid Interval: ${JSON.stringify(edgeInterval)}`,
-              );
+              continue;
             }
```
benwerner01, line 188: @CiaranMn we only patched this for `getOutgoingLinksForEntity`, but don't see why we shouldn't do this for `getIncomingLinksForEntity`

libs/@blockprotocol/graph/src/codegen/parameters.ts
```diff
@@ -23,6 +23,7 @@ export type CodegenParameters = {
   typeNameOverrides?: {
     [sourceTypeId: string]: string;
   };
+  rewriteTypeId?: (typeId: VersionedUrl) => VersionedUrl;
```
CiaranMn, line 26: Suggest we rename this to make clear that it is only rewriting for fetching, it doesn't rewrite the id in the generated schema. I've suggested a name but anything that makes the distinction would work
```suggestion
  /**
   * Enables fetching the schema from a different URL to the one in its schema.
   * The return of this function will be used as the URL to fetch the type from.
   * Does NOT affect the id in the generated schema.
   */
  getFetchUrlFromTypeId?: (typeId: VersionedUrl) => VersionedUrl;
```

benwerner01: That's a fair point, have addressed in https://github.com/blockprotocol/blockprotocol/pull/1349/commits/31ab2d8d08f7d22f51496d7ee8790b4f91e5d30a

CiaranMn (dismissed):

CiaranMn (approved):

PR #1350, opened 2023-11-25
BP-70, BP-59: Implement (de)serialize directly on ontology types

TimDiekmann:

As a quick solution to distinguish between format validation and content validation, the raw module should be made private and the serde implementation should use that internally. This greatly improves the DX around the types.

## 🔍 What does this change?

- Make the `raw` module internal
- move `merge_parent` from `raw::EntityType` to `EntityType`
- Implement `Serialize` and `Deserialize` on `DataType`, `PropertyType`, and `EntityType` via the `raw::` representation
- Make the inheritance field public as this is required downstream. I expect to also make the other fields public but not in this PR as the downstream PR for this PR is already big enough.
- Remove the bound on `ValidateUrl` to read the properties and required field on `Object`

This PR:

- [x] does not modify any publishable blocks or libraries, or modifications do not need publishing

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/@blockprotocol/type-system/crate/src/ontology/entity_type/error.rs
+++ b/libs/@blockprotocol/type-system/crate/src/ontology/entity_type/error.rs
@@ -32,6 +32,9 @@ pub enum ParseEntityTypeError {
 
 #[derive(Debug, PartialEq, Eq, Error)]
 pub enum MergeEntityTypeError {
-    #[error("`{0}` is not contained in the `allOf` property of `{1}`")]
-    NotInAllOf(VersionedUrl, VersionedUrl),
+    #[error("`{parent}` is not contained in the `allOf` property of `{child}`")]
+    DoesNotInheritFrom {
+        child: VersionedUrl,
+        parent: VersionedUrl,
+    },
 }
--- a/libs/@blockprotocol/type-system/crate/src/ontology/entity_type/error.rs
+++ b/libs/@blockprotocol/type-system/crate/src/ontology/entity_type/error.rs
@@ -4,7 +4,7 @@ use thiserror::Error;
 use tsify::Tsify;
 
 use crate::{
-    url::{ParseBaseUrlError, ParseVersionedUrlError},
+    url::{ParseBaseUrlError, ParseVersionedUrlError, VersionedUrl},
     ParseAllOfError, ParseLinksError, ParsePropertyTypeObjectError,
 };
 
@@ -33,5 +33,5 @@ pub enum ParseEntityTypeError {
 #[derive(Debug, PartialEq, Eq, Error)]
 pub enum MergeEntityTypeError {
     #[error("`{0}` is not contained in the `allOf` property of `{1}`")]
-    NotInAllOf(String, String),
+    NotInAllOf(VersionedUrl, VersionedUrl),
 }
--- a/libs/@blockprotocol/type-system/crate/src/ontology/entity_type/links/mod.rs
+++ b/libs/@blockprotocol/type-system/crate/src/ontology/entity_type/links/mod.rs
@@ -13,7 +13,9 @@ use crate::{
 };
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Links(HashMap<VersionedUrl, MaybeOrderedArray<Option<OneOf<EntityTypeReference>>>>);
+pub struct Links(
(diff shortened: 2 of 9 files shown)


libs/@blockprotocol/type-system/crate/src/ontology/entity_type/error.rs
```diff
@@ -33,5 +33,5 @@ pub enum ParseEntityTypeError {
 #[derive(Debug, PartialEq, Eq, Error)]
 pub enum MergeEntityTypeError {
     #[error("`{0}` is not contained in the `allOf` property of `{1}`")]
-    NotInAllOf(String, String),
+    NotInAllOf(VersionedUrl, VersionedUrl),
```
indietyp, line 36: do we maybe want to name the fields instead of using a tuple?

libs/@blockprotocol/type-system/crate/src/ontology/entity_type/links/mod.rs
```diff
@@ -13,7 +13,9 @@ use crate::{
 };
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Links(HashMap<VersionedUrl, MaybeOrderedArray<Option<OneOf<EntityTypeReference>>>>);
+pub struct Links(
+    pub(crate) HashMap<VersionedUrl, MaybeOrderedArray<Option<OneOf<EntityTypeReference>>>>,
```
indietyp, line 17: quite a complex type, this looks like something clippy would usually complain about. 🤔

libs/@blockprotocol/type-system/crate/src/ontology/entity_type/mod.rs
```diff
@@ -7,20 +7,23 @@ mod wasm;
...
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(try_from = "raw::EntityType", into = "raw::EntityType")]
 pub struct EntityType {
     id: VersionedUrl,
     title: String,
     description: Option<String>,
     property_object: Object<ValueOrArray<PropertyTypeReference>>,
-    inherits_from: AllOf<EntityTypeReference>,
+    pub inherits_from: AllOf<EntityTypeReference>,
```
indietyp, line 26: specific reason why only `inherits_from` is pub?

libs/@blockprotocol/type-system/crate/src/ontology/mod.rs
```diff
@@ -37,7 +37,8 @@ pub use shared::{
 // Re-export the repr contents so they're nicely grouped and so that they're easier to import in
 // a non-ambiguous way where they don't get confused with their non repr counterparts.
 // For example, `import crate::raw` lets you then use `raw::DataType`
-pub mod raw {
+#[allow(unreachable_pub, unused_imports)]
```
indietyp, line 40: why allow instead of fixing the lints?

indietyp (commented):
Sadly cannot request changes because of missing privileges :/

libs/@blockprotocol/type-system/crate/src/ontology/entity_type/mod.rs
TimDiekmann: I mentioned this in the description: it’s the only field which is required to be pub for our downstream usage. I'm planning on making them all public.

libs/@blockprotocol/type-system/crate/src/ontology/mod.rs
TimDiekmann: Because depending on the target (host/wasm) or build type (build/test) a different set is required.

libs/@blockprotocol/type-system/crate/src/ontology/entity_type/links/mod.rs
TimDiekmann: I agree but I won’t change this here as it’s internal only.

libs/@blockprotocol/type-system/crate/src/ontology/entity_type/error.rs
TimDiekmann: Good point, made it a struct `{ child: _, parent: _}`

indietyp (commented):
LGTM (cannot approve because of privileges)

TimDiekmann:
Thank you @indietyp!

nonparibus (approved):

PR #1358, opened 2024-01-11
H-1868: Remove `completeText` from the wordpress plugin

benwerner01:

This PR removes `completeText` from the wordpress plugin service module, by upgrading `@blockprotocol/service` to the latest version.

## 🔗 Related links

- H-1868

This PR:

The changes in this PR:

The changes in this PR:

### DIFF
--- a/libs/wordpress-plugin/plugin/trunk/block/edit-or-preview/edit/service-callbacks.tsx
+++ b/libs/wordpress-plugin/plugin/trunk/block/edit-or-preview/edit/service-callbacks.tsx
@@ -126,17 +126,6 @@ export const constructServiceModuleCallbacks = (
         displayToast,
       ),
 
-    /** @todo: remove this when `@blockprotocol/service` is updated */
-    openaiCompleteText: async ({ data }) =>
-      callService(
-        {
-          providerName: "openai",
-          methodName: "completeText",
-          data,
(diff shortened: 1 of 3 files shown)


libs/wordpress-plugin/plugin/trunk/block/edit-or-preview/edit/service-callbacks.tsx
```diff
@@ -126,17 +126,6 @@ export const constructServiceModuleCallbacks = (
...
-    openaiCompleteText: async ({ data }) =>
-      callService(
-        {
-          providerName: "openai",
-          methodName: "completeText",
-          data,
-        },
-        displayToast,
-      ),
-
```
benwerner01, line 139: @CiaranMn I suppose we could re-publish the wordpress plugin to reflect this change? But might not be worth the effort if it's not straightforward, as it doesn't add any functionality and won't improve the experience in any way (just changes where blocks would get an error if they try to call `completeText`)

CiaranMn: No need... once we publish the latest blocks any instance that adds them won't have it as an option

CiaranMn (approved):

PR #1374, opened 2024-03-13
H-2387: Fix `@blockprotocol/graph` to account for draftIds

CiaranMn:

Account for an entity potentially being keyed by an entityId that has a suffix indicating it's a draft – part of ongoing evolution of the spec to be reflected back into docs once stable. 

## 🔗 Related links

- Used in https://github.com/hashintel/hash/pull/4166

This PR:

The changes in this PR:

- [x] are in a state where docs changes are not _yet_ required but will be

The changes in this PR:

### DIFF
--- a/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/element/entity.ts
+++ b/libs/@blockprotocol/graph/src/shared/stdlib/subgraph/element/entity.ts
@@ -1,7 +1,7 @@
 import { Entity, EntityId, EntityRevisionId } from "../../../types/entity.js";
 import { Subgraph } from "../../../types/subgraph.js";
-import { isEntityVertex } from "../../../types/subgraph/vertices.js";
-import { TimeInterval, Timestamp } from "../../../types/temporal-versioning.js";
+import { isEntityVertex, Vertices } from "../../../types/subgraph/vertices.js";
+import { TimeInterval } from "../../../types/temporal-versioning.js";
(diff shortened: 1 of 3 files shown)


libs/@blockprotocol/graph/src/shared/stdlib/subgraph/element/entity.ts
```diff
@@ -36,6 +36,37 @@ export const getEntities = <Temporal extends boolean>(
...
+  const entityRevisions = subgraph.vertices[entityId];
+
+  if (entityRevisions) {
+    return entityRevisions;
+  }
+
+  // check for the presence of draft versions of the entity in the subgraph
+  const draftEntityIds = Object.keys(subgraph.vertices).filter((id) =>
+    id.startsWith(`${entityId}~`),
+  );
```
TimDiekmann, line 52: If `entityId` already contains two `~` we can return `undefined` here as an early return. Small optimization which I don't know worth implementing it.

TimDiekmann (approved):

libs/@blockprotocol/graph/src/shared/stdlib/subgraph/element/entity.ts
CiaranMn: Might as well! Done in https://github.com/blockprotocol/blockprotocol/pull/1374/commits/7c12c757a1967935a7b5c5244fc7053a8e0e758f

